PR #24, opened 2018-01-11
[ElmSharp] Add exception logic to Widget.SetFocus()

darkleem:
### Description of Change ###
Focus Chain gets lost due to the malfunction of calling SetFocus(false).
Eventually, the focus is missing forever on the app. 
This change is about adding an exeption logic to make focus chain normalized.

### DIFF
--- a/src/ElmSharp/ElmSharp/Widget.cs
+++ b/src/ElmSharp/ElmSharp/Widget.cs
@@ -294,6 +294,13 @@ namespace ElmSharp
         public void SetFocus(bool isFocus)
         {
             Interop.Elementary.elm_object_focus_set(RealHandle, isFocus);
+
+            // Temporary code due to focus issue (will be removed at Tizen 5.0)
+            if (this is Window)
+            {
+                if (isFocus && !Interop.Evas.evas_object_focus_get(RealHandle))
+                    Interop.Evas.evas_object_focus_set(RealHandle, true);
+            }
         }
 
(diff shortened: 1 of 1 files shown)


src/ElmSharp/ElmSharp/Widget.cs
```diff
@@ -294,6 +294,13 @@ public bool IsAutoMirroredMode
         public void SetFocus(bool isFocus)
         {
             Interop.Elementary.elm_object_focus_set(RealHandle, isFocus);
+
+            // Temporary code due to focus issue (will be removed at Tizen 5.0)
+            if (this is Window window)
```
JoonghyunCho: 'window' is never used on the following block

darkleem: done

src/ElmSharp/ElmSharp/Widget.cs
```diff
@@ -463,8 +470,8 @@ public int GetPartOpacity(string part)
             return a;
         }
 
-	/// <summary>
-        /// Sends a signal to the edje object of the widget.
+        /// <summary>
+        /// Send a signal to the edje object of the widget.
         /// </summary>
```
WonyoungChoi: "Sends" is correct.

darkleem: thanks~ done

PR #30, opened 2018-01-12
[Information] Restore converted key before calling a callback

kichankwon:
### Description of Change ###

Restore converted key before calling a callback.

### Bugs Fixed ###

Callback provides invalid string key if it is converted.


kichankwon:
Please merge this patch.

PR #32, opened 2018-01-15
[Recorder] Change recorder feature

hsgwon:
### Description of Change ###

Change recorder feature

### API Changes ###

 - ACR: http://suprem.sec.samsung.net/jira/browse/TCSACR-129

Change feature of recorder API
: "http://tizen.org/feature/microphone" -> "http://tizen.org/feature/media.audio_recording"
  "http://tizen.org/feature/camera" -> "http://tizen.org/feature/media.video_recording"

### DIFF
--- a/src/Tizen.Multimedia.Recorder/Recorder/VideoRecorder.Capabilities.cs
+++ b/src/Tizen.Multimedia.Recorder/Recorder/VideoRecorder.Capabilities.cs
@@ -47,6 +47,11 @@ namespace Tizen.Multimedia
 
         private static IEnumerable<Size> LoadVideoResolutions(CameraDevice device)
         {
+            if (!Features.IsSupported(RecorderFeatures.VideoRecorder))
+            {
+                throw new NotSupportedException("Video Recorder is not supported.");
+            }
+
             using (var camera = new Camera(device))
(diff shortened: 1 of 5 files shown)


hsgwon:
I updated missing feature tag that you mentioned.

src/Tizen.Multimedia.Recorder/Recorder/VideoRecorder.Capabilities.cs
```diff
@@ -49,6 +49,11 @@ private static IEnumerable<Size> LoadVideoResolutions(CameraDevice device)
         {
             using (var camera = new Camera(device))
             {
+                if (!Features.IsSupported(RecorderFeatures.VideoRecorder))
```
coderhyme: I think if the feature is not supported, the camera doesn't need to be created which means checking needs to be first followed by camera creation.
And I wonder it is possilbe that camera can be created when the video recorder feature is false, if so, checking video recorder feature after camera creation is meaningless.

hsgwon: I'll update it to check video_recording feature first.

PR #35, opened 2018-01-16
[ElmSharp] Changed Widget.SetFocus() internal routine

darkleem:
### Description of Change ###
Focus Chain gets lost due to the malfunction of calling SetFocus(false).
Eventually, the focus is missing forever on the app.

We created an internal API to return the initial focus state to the same state as when the App was first launched. This change is about adding an exeption logic to make focus chain normalized.


arosis78:
I think below code is more clear.

if (win != IntPtr.Zero && Eo.eo_class_name_get(Eo.eo_class_get(win)) == "Elm_Win")
{
	Evas.evas_object_focus_set(win, true);
}

PR #38, opened 2018-01-17
[ElmSharp] Add explicit operator for native handle

idkiller:
Add global dictionary for map for every EvasObject and native handle
Add explicit converter can be used to get managed EvasObject from native handle

Change-Id: I253ffeaa3599b885862bb10054627f500854e2d1

### Description of Change ###

Add global dictionary for map for every EvasObject and native handle
Add explicit converter can be used to get managed EvasObject from native handle

### Bugs Fixed ###

--

### API Changes ###

Add Explicit converter to EvasObject from IntPtr

Added
public static explicit operator EvasObject(IntPtr handle)

### Behavioral Changes ###

--

### DIFF
--- a/src/ElmSharp/ElmSharp/EvasObject.cs
+++ b/src/ElmSharp/ElmSharp/EvasObject.cs
@@ -125,6 +125,8 @@ namespace ElmSharp
         private Interop.Eext.EextEventCallback _backButtonHandler;
         private Interop.Eext.EextEventCallback _moreButtonHandler;
 
+        private static Dictionary<IntPtr, EvasObject> s_handleTable = new Dictionary<IntPtr, EvasObject>();
+
         /// <summary>
         /// Sets or gets the handle for EvasObject.
         /// </summary>
@@ -884,6 +886,13 @@ namespace ElmSharp
             return obj.Handle;
         }
 
+	/// <summary>
+        /// Define cast to EvasObject operator from IntPtr
+        /// </summary>
+        /// <param name="handle">Native handle to EvasObject</param>
+        /// <since_tizen> preview </since_tizen>
+        public static explicit operator EvasObject(IntPtr handle) => EvasObject.s_handleTable.TryGetValue(handle, out EvasObject obj) ? obj : null;
+
         /// <summary>
         /// Requests the keyname key events to be directed to the current object.
         /// </summary>
@@ -1063,6 +1072,8 @@ namespace ElmSharp
                 Handle = CreateHandle(parent);
                 Debug.Assert(Handle != IntPtr.Zero);
 
+                s_handleTable[Handle] = this;
+
                 (parent as Window)?.AddChild(this);
 
                 OnRealized();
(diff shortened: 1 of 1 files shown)


src/ElmSharp/ElmSharp/EvasObject.cs
```diff
@@ -125,6 +125,8 @@ public abstract class EvasObject
         private Interop.Eext.EextEventCallback _backButtonHandler;
         private Interop.Eext.EextEventCallback _moreButtonHandler;
 
+        private static Dictionary<IntPtr, EvasObject> _objects = new Dictionary<IntPtr, EvasObject>();
```
WonyoungChoi: There is no adding item block in this code.

idkiller: Oh... I miss some piece of code in merging.

idkiller: I add missed code in line 1083

idkiller:
Okay I add code for remove EvasObject at deleted callback is called.

src/ElmSharp/ElmSharp/EvasObject.cs
darkleem: The static variable must be prefixed with s_ ...
and I guess the name is not good.
Would not it be nice to have an intuitive name like s_HandleToItemTable in an ItemObject?

src/ElmSharp/ElmSharp/EvasObject.cs
```diff
@@ -1100,20 +1119,24 @@ public void Unrealize()
 
                 Interop.Evas.evas_object_del(toBeDeleted);
                 Parent = null;
+                _objects.Remove(toBeDeleted);
```
darkleem: I think I should check the Key. Otherwise, it seems that Handle is already Intptr.Zero when it is the second remove.

idkiller: It's not IntPtr.Zero on handle is removed on native side.
although Dictionary.Remove method is does not fail if toBeDeleted is IntPtr.Zero.

idkiller:
Okay, I rename _objects to s_handleTable it has already EvasObject type in this generic parameter.
"handle" is means usage of IntPtr type as dictionary key.

src/ElmSharp/ElmSharp/EvasObject.cs
```diff
@@ -884,6 +886,21 @@ public void Lower()
...
 
+	/// <summary>
+        /// Define cast to EvasObject operator from IntPtr
+        /// </summary>
+        /// <param name="handle">Native handle to EvasObject</param>
+        /// <since_tizen> preview </since_tizen>
+        public static explicit operator EvasObject(IntPtr handle)
+        {
+            EvasObject obj;
+            if (EvasObject.s_handleTable.TryGetValue(handle, out obj))
```
rookiejava: Can use it as single line.
` if (EvasObject.s_handleTable.TryGetValue(handle, out EvasObject obj))`

idkiller: Okay I change it to one line.

PR #50, opened 2018-01-18
[StreamRecorder] fix after as derefence issue

wantfire:
### Description of Change ###
as operator can have null.
It is needed to check for null value.

Checker: DEREF_AFTER_AS.INSTANT
StreamRecorder.cs
Line:  329

### DIFF
--- a/src/Tizen.Multimedia.StreamRecorder/StreamRecorder/StreamRecorder.cs
+++ b/src/Tizen.Multimedia.StreamRecorder/StreamRecorder/StreamRecorder.cs
@@ -326,7 +326,8 @@ namespace Tizen.Multimedia
                         throw new InvalidOperationException("Video option is not set.");
                     }
 
-                    if (AreVideoTypesMatched(_sourceFormat, (packet.Format as VideoMediaFormat).MimeType) == false)
+                    if ((packet.Format is VideoMediaFormat format) &&
+                        (AreVideoTypesMatched(_sourceFormat, format.MimeType) == false))
                     {
                         throw new InvalidOperationException("Video format does not match.");
                     }


src/Tizen.Multimedia.StreamRecorder/StreamRecorder/StreamRecorder.cs
```diff
@@ -326,7 +326,8 @@ public void PushBuffer(MediaPacket packet)
                         throw new InvalidOperationException("Video option is not set.");
                     }
 
-                    if (AreVideoTypesMatched(_sourceFormat, (packet.Format as VideoMediaFormat).MimeType) == false)
+                    if (!(packet.Format is VideoMediaFormat format) ||
```
coderhyme: It is not required actually, because only VideoMediaFormat can have Video as its type.

I don't think it needs to be fixed but if we have to modify it to check format is the VideoMediaFormat, then it should be checked in a separate if statement and throw exception with a clearer message(it is meaningless in my opinion).

Or you can replace whole switch statement with "Switch statements with patterns" which is introduced in c# 7.0.
(see https://blogs.msdn.microsoft.com/dotnet/2017/03/09/new-features-in-c-7-0/)

wantfire:
I agree with you.
I will close  this PR.

But I have one question about "pattern match"
Do you recommand code as below?

`
            switch (packet.Format)
            {
                case AudioMediaFormat audioFormat:
                    if (_audioEnabled == false)
                    {
                        throw new InvalidOperationException("Audio option is not set.");
                    }

                    if (audioFormat.Type != MediaFormatType.Audio)
                    {
                        throw new InvalidOperationException("Audio format does not match.");
                    }
                    break;

                case VideoMediaFormat videoFormat:
                    if (_videoEnabled == false)
                    {
                        throw new InvalidOperationException("Video option is not set.");
                    }

                    if (videoFormat.Type != MediaFormatType.Video ||
                        (AreVideoTypesMatched(_sourceFormat, videoFormat.MimeType) == false))
                    {
                        throw new InvalidOperationException("Video format does not match.");
                    }
                    break;

                default:
                    throw new ArgumentException("Packet is not valid.");
            }

            Native.PushStreamBuffer(Handle, MediaPacket.Lock.Get(packet).GetHandle())
                .ThrowIfError("Failed to push buffer.");
`

Anyway thanks for your comment.

wantfire:
Code is corrupted.

``            switch (packet.Format)
            {
                case AudioMediaFormat audioFormat:
                    if (_audioEnabled == false)
                    {
                        throw new InvalidOperationException("Audio option is not set.");
                    }

                    if (audioFormat.Type != MediaFormatType.Audio)
                    {
                        throw new InvalidOperationException("Audio format does not match.");
                    }
                    break;

                case VideoMediaFormat videoFormat:
                    if (_videoEnabled == false)
                    {
                        throw new InvalidOperationException("Video option is not set.");
                    }

                    if (videoFormat.Type != MediaFormatType.Video ||
                        (AreVideoTypesMatched(_sourceFormat, videoFormat.MimeType) == false))
                    {
                        throw new InvalidOperationException("Video format does not match.");
                    }
                    break;

                default:
                    throw new ArgumentException("Packet is not valid.");
            }

            Native.PushStreamBuffer(Handle, MediaPacket.Lock.Get(packet).GetHandle())
                .ThrowIfError("Failed to push buffer.");
``

src/Tizen.Multimedia.StreamRecorder/StreamRecorder/StreamRecorder.cs
wantfire: I agree with you.
I will close this PR.

But I have one question about "pattern match"
Do you recommand code as below?

```
 	switch (packet.Format)
 	{
            case AudioMediaFormat audioFormat:
                if (_audioEnabled == false)
                {
                    throw new InvalidOperationException("Audio option is not set.");
                }

                if (audioFormat.Type != MediaFormatType.Audio)
                {
                    throw new InvalidOperationException("Audio format does not match.");
                }

                break;

            case VideoMediaFormat videoFormat:
                if (_videoEnabled == false)
                {
                    throw new InvalidOperationException("Video option is not set.");
                }

                if (videoFormat.Type != MediaFormatType.Video ||
                    (AreVideoTypesMatched(_sourceFormat, videoFormat.MimeType) == false))
                {
                    throw new InvalidOperationException("Video format does not match.");
                }
                break;

            default:
                throw new ArgumentException("Packet is not valid.");
        }

        Native.PushStreamBuffer(Handle, MediaPacket.Lock.Get(packet).GetHandle())
            .ThrowIfError("Failed to push buffer.");


```

Anyway thanks for your comment.

coderhyme: That can be an option if it needs to be fixed.
If you use switch with patterns, no need to check Type. (videoFormat.Type != MediaFormatType.Video)
Type is a helper property indicating the type of format so that it can be cast to concrete type.

wantfire:
I will close  without any modification as maintainer's opinion.

PR #51, opened 2018-01-18
[ElmSharp] Add ContainerBase for Custom Container.

idkiller:
### Description of Change ###

Add ContainerBase for User-defined Containers or Layouts.

ContainerBase is basically Elementary Box.
but it does not have options properties of box or methods of the box.
it has the only method for adding or removing a child and the user must define the custom function for layout the container.

A custom container that inherits a Box class can't block or modify the method for child adding or removing. this class should solve this problem.

### Bugs Fixed ###

--

### API Changes ###

Added:
 - protected virtual void Container.OnChildAdded(EvasObject child)
 - protected virtual void Container.OnChildRemoved(EvasObject child)
 - public abstract class ContainerBase : Container

### Behavioral Changes ###

Container class will invoke the OnChildAdded when Container.AddChild is called.
Container class will invoke the OnChildRemoved when Container.RemoveChild is called.


idkiller:
ContainerBase has own Background and Clipper.
It can be User can create own container or layout with own background and layout.

src/ElmSharp/ElmSharp/ContainerBase.cs
```diff
@@ -0,0 +1,204 @@
...
+            Interop.Evas.evas_object_clip_unset(child);
+        }
+
+        /// <summary>
+        /// Creates a widget handle.
+        /// </summary>
+        /// <param name="parent">Parent EvasObject</param>
+        /// <returns>Handle IntPtr</returns>
+        /// <since_tizen> preview </since_tizen>
+        protected override IntPtr CreateHandle(EvasObject parent)
```
darkleem: When an inherited subobject overrides CreateHandle (), all functions associated with _layoutCallback will be expanded, but will be in an inactive state. Is this the intended action? I do not think it is a good structure.

PR #58, opened 2018-01-19
[ElmSharp] Changed Background widget Color property internal implementation

arosis78:
[ElmSharp] Changed Background widget Color property internal implementation to Widget BackgroundColor property

Signed-off-by: Jeonghyun Yun <jh0506.yun@samsung.com>

### Description of Change ###

There are two ways for setting background color to Background widget. (Color & BackgroundColor property)
Two properties are almost same for Background widget but internal implementation is different.
I changed Background widget Color property internal implementation to Widget BackgroundColor property.

### DIFF
--- a/src/ElmSharp/ElmSharp/Background.cs
+++ b/src/ElmSharp/ElmSharp/Background.cs
@@ -43,26 +43,11 @@ namespace ElmSharp
         {
             get
             {
-                int r = 0;
-                int g = 0;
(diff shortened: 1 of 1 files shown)


myroot:
How about change PR title? 
`Background => Background widget`

arosis78:
PR title is changed. (Background -> Background widget)

PR #64, opened 2018-01-19
[TEEC] Guard session close in destructor, apply 'Dispose Pattern'

krzydyn:
### Description of Change ###

Prevent multiple destruction of inernal objects.

### Bugs Fixed ###

- N/A

### API Changes ###

None

None

### Behavioral Changes ###

No info in log file about context not fount or session_imp is null

### DIFF
--- a/src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
+++ b/src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
@@ -19,6 +19,8 @@ using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Runtime.CompilerServices;
+using System.IO;
 
 namespace Tizen.Security.TEEC
 {
@@ -326,8 +328,10 @@ namespace Tizen.Security.TEEC
     /// This type denotes a TEE Session, the logical container linking a client application with a particular trusted application.
     /// </summary>
     /// <since_tizen> 3 </since_tizen>
-    public sealed class Session
+    public sealed class Session : IDisposable
     {
+        private bool disposed = false;
+        private bool opened = false;
         private Context context;
         private IntPtr session_imp;
         private IntPtr opptr;
@@ -340,14 +344,37 @@ namespace Tizen.Security.TEEC
             for (int i=0; i < 4; ++i) shm[i] = null;
         }
 
+        // internal since class is sealed
+        internal void Dispose(bool disposing)
+        {
+            if (disposed) {
+                return ;
+            }
+            if (opened) {
+                Interop.Libteec.CloseSession(session_imp);
+                opened = false;
+            }
+            Marshal.FreeHGlobal(this.opptr);
+            Marshal.FreeHGlobal(this.session_imp);
+            this.opptr = IntPtr.Zero;
+            this.session_imp = IntPtr.Zero;
+            disposed = true;
+        }
+
         /// <summary>
         /// Destructor of the class.
         /// </summary>
         ~Session()
         {
+            Dispose(false);
+        }
+
+        /// <summary>
+        /// Disposable interface implememtation.
+        /// </summary>
+        public void Dispose()
+        {
             Close();
-            Marshal.FreeHGlobal(this.opptr);
-            Marshal.FreeHGlobal(this.session_imp);
         }
 
         internal UInt32 InitParam(ref Interop.TEEC_Parameter32[] dst, int i, Parameter src)
@@ -550,6 +577,7 @@ namespace Tizen.Security.TEEC
 
             //MAYBE map origin of return code to specyfic Exception
             Interop.CheckNThrowException(ret, string.Format("OpenSession('{0}')", destination));
+            opened = true;
         }
         internal void Open64(Guid destination, uint loginMethod, byte[] connectionData, Parameter[] paramlist)
         {
@@ -582,6 +610,7 @@ namespace Tizen.Security.TEEC
 
             //MAYBE map origin of return code to specyfic Exception
             Interop.CheckNThrowException(ret, string.Format("OpenSession('{0}')", destination));
+            opened = true;
         }
 
         /// <summary>
@@ -596,7 +625,8 @@ namespace Tizen.Security.TEEC
         /// <exception cref="NotSupportedException">The required feature is not supported.</exception>
         /// <exception cref="InvalidOperationException">The operation is invalid.</exception>
         public void Close() {
-            Interop.Libteec.CloseSession(session_imp);
+            Dispose(true);
+            GC.SuppressFinalize(this);
         }
 
         /// <summary>
@@ -689,6 +719,8 @@ namespace Tizen.Security.TEEC
     /// <since_tizen> 3 </since_tizen>
     public sealed class Context : IDisposable
     {
+        private bool disposed = false;
+        private bool initialized = false;
         internal IntPtr context_imp;
 
         /// <summary>
@@ -711,6 +743,7 @@ namespace Tizen.Security.TEEC
             try {
                 int ret = Interop.Libteec.InitializeContext(name, context_imp);
                 Interop.CheckNThrowException(ret, string.Format("InititalizeContext('{0}')", name));
+				initialized = true;
             }
             catch (global::System.DllNotFoundException e)
             {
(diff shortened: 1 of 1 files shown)


src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
```diff
@@ -597,6 +600,7 @@ internal void Open64(Guid destination, uint loginMethod, byte[] connectionData,
         /// <exception cref="InvalidOperationException">The operation is invalid.</exception>
         public void Close() {
```
myroot: How about check closed before CloseSession?

JoonghyunCho:
Please update the PR description

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
krzydyn: I think the implementation should behave correctly. This is C# wrapper only.
I have put a call to Close() in destructor, to prevent resource leakage and this special call have to be guarded but when programmer calls it double it should be forwarded to implementation.

Documentation says:
*This function closes a Session which has been opened with a Trusted Application.
All Commands within the Session MUST have completed before calling this function.
The Implementation MUST do nothing if the session parameter is NULL.
The implementation of this function MUST NOT be able to fail: after this function returns the Client
Application must be able to consider that the Session has been closed.*

myroot: To prevent leakage of unmanaged resource, 
Implement IDisposable interface

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
```diff
@@ -328,6 +328,7 @@ public sealed class Value : BaseParameter<TEFValueType>
     /// <since_tizen> 3 </since_tizen>
     public sealed class Session
     {
+        private bool closed = false;
```
myroot: Close only valid after call of Open32/Open64, 

So, `closed` value initially should be true, 
and It changed to false in Open32/Open64

krzydyn: Done

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
krzydyn: Done

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
```diff
@@ -345,7 +346,14 @@ public sealed class Session
         /// </summary>
         ~Session()
         {
-            Close();
+            Dispose();
```
myroot: Please apply Dispose pattern 

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern

myroot: It cause double free, When Dispose was called by user

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
```diff
@@ -340,14 +344,37 @@ public sealed class Session
             for (int i=0; i < 4; ++i) shm[i] = null;
         }
 
+        // internal since class is sealed
+        internal void Dispose(bool disposing)
```
myroot: Why not private?

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
```diff
@@ -340,14 +344,37 @@ public sealed class Session
...
+        {
+            Dispose(false);
+        }
+
+        /// <summary>
+        /// Disposable interface implememtation.
+        /// </summary>
+        public void Dispose()
         {
             Close();
```
myroot: Please add 
`GC.SuppressFinalize(this);`
It make disable the finalizer

myroot:
Please apply Dispose pattern
```c#
public class DisposableResourceHolder : IDisposable {  
    public DisposableResourceHolder(){  
    }  

    // Do not change it
    public void Dispose(){  
        Dispose(true);  
        GC.SuppressFinalize(this);  
    }  

    protected virtual void Dispose(bool disposing){  
        if (disposing){  
            // release Managed resource
        }  
        // release Unmanaged resource
    }  
}  
```

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
```diff
@@ -596,7 +625,8 @@ internal void Open64(Guid destination, uint loginMethod, byte[] connectionData,
         /// <exception cref="NotSupportedException">The required feature is not supported.</exception>
         /// <exception cref="InvalidOperationException">The operation is invalid.</exception>
         public void Close() {
-            Interop.Libteec.CloseSession(session_imp);
+            Dispose(true);
```
myroot: Do not call Dispose in Close()

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
krzydyn: Can be private as well. I will do.

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
krzydyn: Close have to be identical to Dispose
* providing method Close(), in addition to the Dispose(), if close is standard terminology in the area.  it is important that you make the Close implementation identical to Dispose *

To make so, I put a call to Close() from Dispose(), and implementation of Close() calls Dispose(true); and GC.SuppressFinalize(this);
So I don;t understand your point.

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
krzydyn: What about this patern?:
```
public class Stream : IDisposable {  
    IDisposable.Dispose(){  
        Close();  
    }  
    public void Close(){  
        Dispose(true);  
        GC.SuppressFinalize(this);  
    }  
}  
```

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
myroot: This object has two state.
 1. Created : Resource allocated.
 2. Opened : Session was established

OpenXX method transit to Opened state from Created

In my opinion, Close is an API that corresponds to OpenXX

So, If Open was transit to Opened from Created
Close should transit to Created from Opened

It is more resonable interface and more flexible to use.

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
myroot: Stream dose not provide open method

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
myroot: It's my point
http://dotnetguts.blogspot.kr/2007/06/difference-between-close-and-dispose.html?m=1

src/Tizen.Security.TEEC/Tizen.Security.TEEC/Libteec.cs
krzydyn: Note that Open32, Open64 are internal.
The only way to create Session is to call context.OpenSession(...) and after
call session.Close() session is not usable any more (one can't open it again).
So making Dispose() and Close() work identical makes sense.

myroot: Ok, I understand your point

But please, consider to refactoring
Session should be created from factory method instead of context.OpenSession.
context.OpenSession just use Factory method to meet high cohesion
it is a case of Expert pattern
https://en.m.wikipedia.org/wiki/GRASP_(object-oriented_design)#Information_expert

PR #73, opened 2018-01-24
[Multimedia] Moved enums of AudioManager into separate files.

coderhyme:
### Description of Change ###

Structuring code.


src/Tizen.Multimedia/AudioManager/AudioDeviceIoDirection.cs
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (c) 2016 Samsung Electronics Co., Ltd All Rights Reserved
```
hsgwon: How about change year : 2016 -> 2018

PR #88, opened 2018-01-26
[TEEC] Guard session close in destructor, apply 'Dispose Pattern'

krzydyn:
### Description of Change ###

Prevent multiple destruction of inernal objects.

### Bugs Fixed ###

- N/A

### API Changes ###

None

None

### Behavioral Changes ###

No info in log file about context not fount or session_imp is null


terry2000s:
Resolve conflict first.

PR #102, opened 2018-01-30
[SystemSettings] Change remove event handler for all of keys.

jinwangan:
Signed-off-by: jinwang.an <jinwang.an@samsung.com>

### Description of Change ###

When Tizen.System.SystemSettings caller try to remove event handler event if there is no event handler
in SystemSettings Key,

There will be log "There is no event handler" with "Tizen.System.SystemSettings" tag.


kionias:
event handler check null logic.

PR #119, opened 2018-02-09
[Tizen.Network.WiFi] Use lock on checking if instance is null

chleun-moon:
### Description of Change ###

WiFiManagerImple class has only single instance, but in multi-thread environment, race condition occurs when checking if the instance is null

### Bugs Fixed ###

- Add lock before checkint if the instance is null

### DIFF
--- a/src/Tizen.Network.WiFi/Tizen.Network.WiFi/WiFiManagerImpl.cs
+++ b/src/Tizen.Network.WiFi/Tizen.Network.WiFi/WiFiManagerImpl.cs
@@ -46,7 +46,7 @@ namespace Tizen.Network.WiFi
 
     internal partial class WiFiManagerImpl
     {
-        private static WiFiManagerImpl _instance = null;
+        private static Lazy<WiFiManagerImpl> _instance = null;
         private Dictionary<IntPtr, Interop.WiFi.VoidCallback> _callback_map = new Dictionary<IntPtr, Interop.WiFi.VoidCallback>();
         private int _requestId = 0;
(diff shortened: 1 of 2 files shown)


src/Tizen.Network.WiFi/Tizen.Network.WiFi/WiFiManagerImpl.cs
```diff
@@ -121,13 +122,16 @@ internal static WiFiManagerImpl Instance
         {
             get
             {
-                if (_instance == null)
+                lock (_instanceLock)
                 {
-                    Log.Debug(Globals.LogTag, "Instance is null");
```
myroot: It is better to use a Lazy class

src/Tizen.Network.WiFi/Tizen.Network.WiFi/WiFiManagerImpl.cs
```diff
@@ -124,10 +124,13 @@ internal static WiFiManagerImpl Instance
                 if (_instance == null)
                 {
                     Log.Debug(Globals.LogTag, "Instance is null");
-                    _instance = new WiFiManagerImpl();
+                    _instance = new Lazy<WiFiManagerImpl>(() =>
+                            {
```
myroot: Wrong use Lazy

myroot: It should be initialized in constructor

PR #136, opened 2018-03-02
[Connection][WiFi]Apply TizenSynchronizationContext for event handling

chleun-moon:
### Description of Change ###
Registering events should be executed in main thread.

### Behavioral Changes ###
All events are registered in main thread.
As a result, they are delivered to main thread.

### DIFF
--- a/src/Tizen.Network.Connection/Tizen.Network.Connection/ConnectionInternalManager.cs
+++ b/src/Tizen.Network.Connection/Tizen.Network.Connection/ConnectionInternalManager.cs
@@ -116,19 +116,25 @@ namespace Tizen.Network.Connection
         {
             add
             {
-                if (_ConnectionTypeChanged == null)
-                {
-                    context.Post((x) => { ConnectionTypeChangedStart(); }, null);
-                }
-                _ConnectionTypeChanged += value;
+                context.Post((x) =>
+                        {
+                            if (_ConnectionTypeChanged == null)
+                            {
+                                ConnectionTypeChangedStart();
+                            }
+                            _ConnectionTypeChanged += value;
+                        }, null);
             }
             remove
             {
-                _ConnectionTypeChanged -= value;
-                if (_ConnectionTypeChanged == null)
-                {
-                    context.Post((x) => { ConnectionTypeChangedStop(); }, null);
-                }
+                context.Post((x) =>
+                        {
+                            _ConnectionTypeChanged -= value;
+                            if (_ConnectionTypeChanged == null)
+                            {
+                                ConnectionTypeChangedStop();
+                            }
+                        }, null);
             }
         }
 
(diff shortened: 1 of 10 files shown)


src/Tizen.Network.Connection/Tizen.Network.Connection/ConnectionInternalManager.cs
```diff
@@ -120,29 +105,6 @@ private ConnectionInternalManager()
...
-            Log.Debug(Globals.LogTag, ">>> Handle: " + GetHandle());
-            if (disposed)
-                return;
-
-            if (disposing)
-            {
-                // Free managed objects.
-            }
-
-            UnregisterEvents();
```
myroot: If anymore not use it (`UnregisterEvents()`), please remove it's implementation

src/Tizen.Network.Connection/Tizen.Network.Connection/ConnectionProfile.cs
```diff
@@ -36,6 +37,8 @@ public class ConnectionProfile : IDisposable
 
         private Interop.ConnectionProfile.ProfileStateChangedCallback _profileChangedCallback;
 
+        private static TizenSynchronizationContext context = new TizenSynchronizationContext();
+
```
myroot: IDisposable interface of ConnectionProfile was wrong implemented,

unmanaged resource (in this case ProfileHandle) should always released in Dispose(bool) method. 
managed resource is only released when disposing was true

src/Tizen.Network.WiFi/Tizen.Network.WiFi/WiFiManager.cs
```diff
@@ -51,10 +53,19 @@ public override bool IsInvalid
         /// </summary>
         protected override bool ReleaseHandle()
         {
-            Interop.WiFi.Deinitialize(this.handle);
+            Interop.WiFi.Deinitialize(tid, this.handle);
             this.SetHandle(IntPtr.Zero);
             return true;
         }
+
+        internal int TID
```
myroot: do not make set only property

in this case use method

src/Tizen.Network.Connection/Tizen.Network.Connection/ConnectionInternalManager.cs
```diff
@@ -81,6 +83,8 @@ internal class ConnectionInternalManager
         private Interop.Connection.ConnectionAddressChangedCallback _proxyAddressChangedCallback;
         private Interop.Connection.EthernetCableStateChangedCallback _ethernetCableStateChangedCallback;
 
+        private static TizenSynchronizationContext context = new TizenSynchronizationContext();
```
myroot: How about add `s_`  prefix for priviate static variable

PR #141, opened 2018-03-07
[NUI] Fix parent get issue

huiyueun:
Signed-off-by: huiyu <huiyu@huiyu-linux.tn.corp.samsungelectronics.net>


dongsug-song:
Dear SRCN members,
Hello,
could you review this code?
we got VD 100% crash issue on CSFS => Source => USB memory stick icon => up key => infomation steps.
so we recommended CSFS to change this Parent as GetParent() to solve this.
and this patch is for resolving crash issue and memory leak issue both.
but this is kind of critical so I want you to review this patch carefully.
thank you.
best regards,
ssong.

src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -4332,7 +4332,8 @@ public new View Parent
 
                 if (basehandle is Layer)
```
darkleem: Using 'Is-expressions with patterns' gives you better readability and better performance.

e.g.  if( basehandle is Layer layer )

PR #142, opened 2018-03-07
[NUI] Fix parent get issue

huiyueun:
Signed-off-by: huiyu <huiyu@huiyu-linux.tn.corp.samsungelectronics.net>


dongsug-song:
Dear SRCN members,
Hello,
could you review this code?
we got VD 100% crash issue on CSFS => Source => USB memory stick icon => up key => infomation steps.
so we recommended CSFS to change this Parent as GetParent() to solve this.
and this patch is for resolving crash issue and memory leak issue both.
but this is kind of critical so I want you to review this patch carefully.
thank you.
best regards,
ssong.

huiyueun:
Parent is deprecated.  deleted

PR #245, opened 2018-05-10
[ElmSharp] Added some APIs for EvasGL and EvasImage

arosis78:
### Description of Change ###

Added some APIs for EvasGL and EvasImage


arosis78:
abandoned

PR #275, opened 2018-05-30
[NUI] Fix property issue

huiyueun:


adeelkazmi:
Connecting to a property notification will reduce performance quite substantially so I would advise against pushing something like this.  As mentioned earlier, this is a value, i.e. struct in c#, type and GetProperty does not return an object, all it does is return a copy of the current value.

WonyoungChoi:
@huiyueun Please close this PR if you no longer need it.

dongsug-song:
This is fixed by the patch of https://github.com/Samsung/TizenFX/pull/338 (View.cs)
so I close this.

PR #327, opened 2018-07-09
[NUI] Modify Resources to XamlResources

huiyueun:


dongsug-song:
this will be ACR later =>

/// This will be public opened in tizen_5.0 after ACR done. Before ACR, need to be hidden as inhouse API.  
public ResourceDictionary Resources

PR #337, opened 2018-07-18
[MediaPlayer] add API about playback time to set/get in nonoseconds

kqjy777:
### API Changes ###

 - ACR: http://suprem.sec.samsung.net/jira/browse/TCSACR-164

Added:
 - public async Task SetPlayPositionAsyncNanos(long position, bool accurate)
 - public long GetPlayPositionNanos()
 - public long GetDurationNanos()

### DIFF
--- a/src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
@@ -551,16 +551,17 @@ namespace Tizen.Multimedia
             return playPosition;
         }
 
-        private void SetPlayPosition(long seconds, bool accurate, bool nano,
+        private void SetPlayPosition(long position, bool accurate, bool nanoseconds,
             NativePlayer.SeekCompletedCallback cb)
         {
-            var ret = !nano ? NativePlayer.SetPlayPosition(Handle, Convert.ToInt32(seconds), accurate, cb, IntPtr.Zero) :
-                NativePlayer.SetPlayPositionNanos(Handle, seconds, accurate, cb, IntPtr.Zero);
+            //Check if it is nanoseconds or milliseconds.
+            var ret = !nanoseconds ? NativePlayer.SetPlayPosition(Handle, Convert.ToInt32(position), accurate, cb, IntPtr.Zero) :
+                NativePlayer.SetPlayPositionNanos(Handle, position, accurate, cb, IntPtr.Zero);
 
             //Note that we assume invalid param error is returned only when the position value is invalid.
             if (ret == PlayerErrorCode.InvalidArgument)
             {
-                throw new ArgumentOutOfRangeException(nameof(seconds), seconds,
+                throw new ArgumentOutOfRangeException(nameof(position), position,
                     "The position is not valid.");
             }
             if (ret != PlayerErrorCode.None)
--- a/src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
@@ -549,15 +549,16 @@ namespace Tizen.Multimedia
             return playPosition;
         }
 
-        private void SetPlayPosition(int milliseconds, bool accurate,
+        private void SetPlayPosition(long seconds, bool accurate, bool nano,
             NativePlayer.SeekCompletedCallback cb)
         {
-            var ret = NativePlayer.SetPlayPosition(Handle, milliseconds, accurate, cb, IntPtr.Zero);
+            var ret = !nano ? NativePlayer.SetPlayPosition(Handle, (int)seconds, accurate, cb, IntPtr.Zero) :
+                NativePlayer.SetPlayPositionNanos(Handle, seconds, accurate, cb, IntPtr.Zero);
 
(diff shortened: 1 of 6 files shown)


src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -549,15 +551,17 @@ public int GetPlayPosition()
             return playPosition;
         }
 
-        private void SetPlayPosition(int milliseconds, bool accurate,
+        private void SetPlayPosition(long position, bool accurate, bool nanoseconds,
             NativePlayer.SeekCompletedCallback cb)
         {
-            var ret = NativePlayer.SetPlayPosition(Handle, milliseconds, accurate, cb, IntPtr.Zero);
+            //Check if it is nanoseconds or milliseconds.
+            var ret = !nanoseconds ? NativePlayer.SetPlayPosition(Handle, Convert.ToInt32(position), accurate, cb, IntPtr.Zero) :
```
hsgwon: Does ```Convert.ToInt32(position)``` really needed? 
Isn't it enough just explicit casting ```(int)position```?

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -549,15 +551,17 @@ public int GetPlayPosition()
...
+                NativePlayer.SetPlayPositionNanos(Handle, position, accurate, cb, IntPtr.Zero);
```
hsgwon: How about ```Nanosecond(s)``` instead of ```SetPlayPositionNonos```?

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -595,7 +603,74 @@ public int GetPlayPosition()
...
+        /// <exception cref="InvalidOperationException">The player is not in the valid state.</exception>
+        /// <seealso cref="SetPlayPositionAsync(int, bool)"/>
+        /// <seealso cref="SetPlayPositionAsyncNanos(long, bool)"/>
+        /// <seealso cref="GetPlayPosition"/>
+        /// <since_tizen> 5 </since_tizen>
+        public long GetPlayPositionNanos()
+        {
+            ValidatePlayerState(PlayerState.Ready, PlayerState.Paused, PlayerState.Playing);
+
+            long playPosition = 0;
```
hsgwon: How about below?
```
NativePlayer.GetPlayPositionNanos(Handle, out long playPosition).
```

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -549,15 +551,17 @@ public int GetPlayPosition()
             return playPosition;
         }
 
-        private void SetPlayPosition(int milliseconds, bool accurate,
+        private void SetPlayPosition(long position, bool accurate, bool nanoseconds,
```
hsgwon: Don't we need to consider extensibility on new unit?
(Is there any possibility to add new unit in the future? )

kqjy777: i think not. but if then, i'll make new enum.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
kqjy777: i think it is not needed. i'll change it to cast only.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
kqjy777: that looks better

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
kqjy777: ok

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -595,7 +603,74 @@ public int GetPlayPosition()
...
+        /// <exception cref="ObjectDisposedException">The player has already been disposed of.</exception>
+        /// <exception cref="InvalidOperationException">The player is not in the valid state.<br/>
+        ///     -or-<br/>
+        ///     In case of non-seekable content, the player will return error and keep playing without changing the play position.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">The specified position is not valid.</exception>
+        /// <seealso cref="SetPlayPositionAsync(int, bool)"/>
+        /// <seealso cref="GetPlayPosition"/>
+        /// <seealso cref="GetPlayPositionNanos"/>
+        /// <since_tizen> 5 </since_tizen>
+        public async Task SetPlayPositionAsyncNanos(long position, bool accurate)
```
hsgwon: It seems that the difference between ```SetPlayPositionAsyncNanos``` and ```SetPlayPositionAsync``` is only the unit for the position. 
How about extracting common part to another private method?

kqjy777: ok i'll try it

PR #367, opened 2018-08-03
[MediaContent] Added exception handling to catch p/invoke error return

hsgwon:
### Bugs Fixed ###

- Native FW was changed to return error when there's no matched record with playlistId.
  So, C# layer should catch the exception to handle it correctly.


jiyong2:
It looks good to me.

PR #373, opened 2018-08-07
[MediaVision] Fix an issue that callback was made on a garbage collected delegate

tyoungroy:
Signed-off-by: Tae-Young Chung <ty83.chung@samsung.com>

### Bugs Fixed ###
- TC is sometimes block because of issue that callback was made on a garbage collected delegate.
- To fix it, declare the member variable for callback.

### DIFF
--- a/src/Tizen.Multimedia.Vision/MediaVision/SurveillanceEngine.cs
+++ b/src/Tizen.Multimedia.Vision/MediaVision/SurveillanceEngine.cs
@@ -35,6 +35,7 @@ namespace Tizen.Multimedia.Vision
         private IntPtr _handle = IntPtr.Zero;
         private bool _disposed = false;
 
+        internal EventOccurredCallback _eventDetectedCallback;
         /// <summary>
         /// Initializes a new instance of the <see cref="SurveillanceEngine"/> class.
         /// </summary>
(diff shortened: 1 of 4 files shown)


src/Tizen.Multimedia.Vision/MediaVision/SurveillanceEngine.cs
```diff
@@ -35,6 +35,7 @@ public abstract class SurveillanceEngine : IDisposable
         private IntPtr _handle = IntPtr.Zero;
         private bool _disposed = false;
 
+        internal EventOccurredCallback OnEventDetected;
```
hsgwon: How about use the word ```_eventDetectedCallback```? It seems like eventHandler to me.

hsgwon: It seems that Recognized and Detected events in several classes are using this sindle field together. Is it OK?

hsgwon: I missed that all those classes are not just seperate instance but base and derived class relation.  It's OK.

tyoungroy:
I change "OnEventDetected" to "_eventDetectedCallback" and update commit.

PR #377, opened 2018-08-08
[Tizen.Multimedia] add logs to check GCHandle

kqjy777:
### Description of Change ###

[Tizen.Multimedia] add logs to check GCHandle

### DIFF
--- a/src/Tizen.Multimedia/Common.Internal/ObjectKeeper.cs
+++ b/src/Tizen.Multimedia/Common.Internal/ObjectKeeper.cs
@@ -40,6 +40,7 @@ namespace Tizen.Multimedia
             {
                 Target = obj;
                 _handle = GCHandle.Alloc(obj);
+                Tizen.Log.Info("Tizen.Multimedia", "GCHandle is allocated.");
             }
 
             ~ObjectKeeperImpl()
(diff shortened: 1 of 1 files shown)


src/Tizen.Multimedia/Common.Internal/ObjectKeeper.cs
```diff
@@ -40,6 +40,7 @@ internal ObjectKeeperImpl(T obj)
             {
                 Target = obj;
                 _handle = GCHandle.Alloc(obj);
+                Tizen.Log.Info("Tizen.Multimedia", "GCHandle is allocated.");
```
hsgwon: You don't need to use 'Tizen.'.

src/Tizen.Multimedia/Common.Internal/ObjectKeeper.cs
```diff
@@ -63,6 +64,7 @@ public override void Dispose()
             {
                 Dispose(true);
                 GC.SuppressFinalize(this);
+                Tizen.Log.Warn("Tizen.Multimedia", "GCHandle is disposed.");
```
hsgwon: It's not warning situation. After lifecycle of ObjectKeeper is done, ```using``` statement will call Dispose() or user have to call it explicitly to free GCHandle.
Please change Warn to Info.

PR #398, opened 2018-08-17
[Multimedia.MediaTool] Add APIs for rotation and flip

hsgwon:
### Description of Change ###

Add APIs for rotation and flip. 
Rotation and flip are in same enum at native fw. But, Rotation and Flips enums are already used in Multimedia so we use it.

### API Changes ###

 - TCSACR-174
   ( It's related with Native ACR-1214.)

Added:
 - Rotation { get; set; } // Property
 - Flip { get; set; } // Property

### DIFF
--- a/src/Tizen.Multimedia/MediaTool/MediaPacket.cs
+++ b/src/Tizen.Multimedia/MediaTool/MediaPacket.cs
@@ -216,7 +216,6 @@ namespace Tizen.Multimedia
                 ValidateNotDisposed();
 
                 int ret = Interop.MediaPacket.GetDts(_handle, out var value);
-
                 MultimediaDebug.AssertNoError(ret);
 
                 return value;
@@ -227,7 +226,6 @@ namespace Tizen.Multimedia
                 ValidateNotLocked();
 
                 int ret = Interop.MediaPacket.SetDts(_handle, value);
-
                 MultimediaDebug.AssertNoError(ret);
             }
         }
@@ -245,7 +243,6 @@ namespace Tizen.Multimedia
                 ValidateNotDisposed();
 
                 int ret = Interop.MediaPacket.IsEncoded(_handle, out var value);
-
                 MultimediaDebug.AssertNoError(ret);
 
                 return value;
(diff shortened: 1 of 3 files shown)


src/Tizen.Multimedia/MediaTool/MediaPacket.cs
```diff
@@ -252,6 +252,79 @@ public bool IsEncoded
             }
         }
 
+        /// <summary>
+        /// Gets or sets the rotation value of the current packet.
+        /// </summary>
+        /// <exception cref="ArgumentException">The specified value to set is invalid.</exception>
+        /// <exception cref="ObjectDisposedException">The MediaPacket has already been disposed of.</exception>
+        /// <since_tizen> 5 </since_tizen>
```
kqjy777: it needs to add InvalidOperationException for ValidateNotLocked.

src/Tizen.Multimedia/MediaTool/MediaPacket.cs
```diff
@@ -252,6 +252,79 @@ public bool IsEncoded
...
+
+        /// <summary>
+        /// Gets or sets the flip value of the current packet.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="Flips.None"/> will be ignored in set case. It's not supported in Native FW.
+        /// </remarks>
+        /// <exception cref="ArgumentException">The specified value to set is invalid.</exception>
+        /// <exception cref="ObjectDisposedException">The MediaPacket has already been disposed of.</exception>
+        /// <since_tizen> 5 </since_tizen>
```
kqjy777: it needs to add InvalidOperationException for ValidateNotLocked.

src/Tizen.Multimedia/MediaTool/MediaPacket.cs
hsgwon: Done

src/Tizen.Multimedia/MediaTool/MediaPacket.cs
hsgwon: Done

PR #421, opened 2018-08-28
English review for RPCPort

Reni-90:
### Description of Change ###
Minor grammar changes done.
**Exception.cs:**
Question:
Line 46:
The exception occurs, if a socket is not connected or when a non connected socket is used.

**Parcel.cs**
Minor grammar changes done according to Native API reference.
Question:
1.  Can we use <c></c> tags to emphasize data types instead of ' '?
For example,
 /// Writes a 'byte' value into parcel object.

**StubBase.cs**
Question:
Line 89:
Whether stub allows only trusted proxy or not.
Can we rephrase as:
If stub allows only trusted proxy then <c>true</c>, <c>false</c> otherwise .
Or do we mention 'true' or 'false' only for return values.

### Bugs Fixed ###

-NA

### API Changes ###

-NA

### Behavioral Changes ###

-NA

### DIFF
--- a/src/Tizen.Applications.Common/Tizen.Applications.RPCPort/Exceptions.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications.RPCPort/Exceptions.cs
@@ -43,7 +43,7 @@ namespace Tizen.Applications.RPCPort
     public class InvalidProtocolException : InvalidOperationException { }
 
     /// <summary>
-    /// Exception class which will be thrown when not connected socket is used.
+    /// Exception class, which is thrown when a non connected socket is used.
     /// </summary>
     /// <since_tizen> 5 </since_tizen>
     public class NotConnectedSocketException : InvalidOperationException { }
--- a/src/Tizen.Applications.Common/Tizen.Applications.RPCPort/Exceptions.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications.RPCPort/Exceptions.cs
@@ -19,37 +19,37 @@ using System;
 namespace Tizen.Applications.RPCPort
 {
     /// <summary>
-    /// Exception class for invalid IO
+    /// Exception class for invalid IO.
     /// </summary>
     /// <since_tizen> 5 </since_tizen>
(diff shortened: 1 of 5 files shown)


pjh9216:
Question:
Line 46:
The exception occurs, if a socket is not connected or when a non connected socket is used.
==> Fine. Please change it

Question:
Line 89:
Whether stub allows only trusted proxy or not.
Can we rephrase as:
If stub allows only trusted proxy then true, false otherwise .
Or do we mention 'true' or 'false' only for return values.
==> Fine. Please change it

Question:
    Can we use tags to emphasize data types instead of ' '?
    For example,
    /// Writes a 'byte' value into parcel object.
==> I have no idea to emphasize data types with XML elements. 
If you know the way, please use it.

Reni-90:
Question:
Line 46:
The exception occurs, if a socket is not connected or when a non connected socket is used.
==> Fine. Please change it

Could you please elaborate. 
The exception occurs because the socket is not connected.
OR
The exception occurs because a non connected socket is used.
Please correct me if I am wrong.


Question:
Can we use tags to emphasize data types instead of ' '?
For example,
/// Writes a 'byte' value into parcel object.
==> I have no idea to emphasize data types with XML elements.
If you know the way, please use it.

According to Native APIs( https://developer.tizen.org/development/api-references/native-application?redirect=https://developer.tizen.org/dev-guide/4.0.0/org.tizen.native.mobile.apireference/group__CAPI__RPC__PORT__MODULE.html), The datatypes are not emphasized.
Therefore, to maintain consistency I will follow the same style used in Native APIs.
Need your opinion.

pjh9216:
Line 46.
==> The exception occurs because a non connected socket is used.

The datatypes are not emphasized.
Therefore, to maintain consistency I will follow the same style used in Native APIs.
Need your opinion.
==> Fine. Please do it.

Reni-90:
@pjh9216 
Thank you for explaining.

Reni-90:
@pjh9216 
Changes are updated. Please verify.

PR #424, opened 2018-08-29
[ABIChecker] Add feature to check internal api changes

JoonghyunCho:
### Description of Change ###

Add a new feature to ABIChecker to test internl API changes.

### Bugs Fixed ###

N/A

### API Changes ###

N/A

### Behavioral Changes ###

ABIChecker used to add only `ACR Required` tag when there is public api including not browsable changes.
Now, the checker can  distinguish the internal api(not browsable) changes so it can add `Internal API changes` tag.


tools/src/ABIChecker/ABITester.cs
```diff
@@ -72,7 +82,75 @@ bool InternalDirectoryCheck()
                     }
                 }
             }
-            return errCount > 0 ? false : true;
+            return directoryResult;
+        }
+
+        ABITestResult CheckChagedModule(FileInfo[] baseDllFiles, FileInfo[] targetDllFiles)
```
WonyoungChoi: mis-spell, CheckChangedModule

tools/src/ABIChecker/ABIChecker.cs
```diff
@@ -0,0 +1,156 @@
...
+                {
+                    if (baseFile.Name == targetFile.Name)
+                    {
+                        try
+                        {
+                            var fileResult = CheckFile(baseFile.FullName, targetFile.FullName);
+                            Console.WriteLine($"{count++}. {baseFile.ToString()} : {((fileResult & ABITestResult.ACRRequired) == ABITestResult.ACRRequired ? "FAIL" : "PASS")}");
+                            directoryResult |= fileResult;
+                        }
+                        catch (Exception)
```
WonyoungChoi: Unnecessary try/catch block. The method CheckFile does not throw any exception.

tools/src/ABIChecker/ABIChecker.cs
```diff
@@ -0,0 +1,156 @@
...
+                if (targetFile == string.Empty)
+                {
+                    changedAPIList = _checker.CheckAssemblyFile(Assembly.LoadFrom(baseFile));
+                }
+                else
+                {
+                    changedAPIList = _checker.CheckAssemblyFile(Assembly.LoadFrom(baseFile), Assembly.LoadFile(targetFile));
+                }
+                var internalAPIList = new List<MemberInfo>();
+                for (int i = changedAPIList.Count - 1; i >= 0; i--)
```
WonyoungChoi: why not use 'foreach'?

tools/src/ABIChecker/AssemblyChecker.cs
```diff
@@ -0,0 +1,99 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Checker_ABI
+{
+    internal class AssemblyChecker
+    {
+        private const BindingFlags s_PublicOnlyFlags = BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly | BindingFlags.Static;
```
WonyoungChoi: Do not use s_* prefix for private const variable. It should be just "PublicOnlyFlags".

tools/src/ABIChecker/ABIChecker.cs
JoonghyunCho: Removing the item of List inside the loop ruins the List.

PR #425, opened 2018-08-29
[AudioManager] Fix to use sound_manager_is_device_running_by_id() for…

sangchul1011:
… Interop.AudioDevice.IsDeviceRunning

### Description of Change ###
Since device state is mutable variable, it needs to get the value from server on the fly.

### Bugs Fixed ###
- AudioDevice does not include the running state. So it should be obtained by checking with id. (Use the id instead of the ptr of new copied AudioDevice object)


sangchul1011:
1. it does not work properly. When the callback is invoked and if checking the "IsRunning" member, an invalid value can be obtained.
2. Okay, I'll remove it soon.

PR #429, opened 2018-08-30
[Tizen.Applications] Add finalizer in TizenSyncronizationContext

pjh9216:
### Bugs Fixed ###

- The native source should be removed before GC

### DIFF
--- a/src/Tizen.Applications.Common/Tizen.Applications/TizenSynchronizationContext.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications/TizenSynchronizationContext.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


src/Tizen.Applications.Common/Tizen.Applications/TizenSynchronizationContext.cs
```diff
@@ -115,7 +125,7 @@ private void Post(Action action)
                 id = _transactionId++;
             }
             _handlerMap.TryAdd(id, action);
-            Interop.Glib.IdleAdd(_wrapperHandler, (IntPtr)id);
+            _source = Interop.Glib.IdleAdd(_wrapperHandler, (IntPtr)id);
```
semun-lee: I think Post() could be called multiple times, and _source could be overwritten. Could you check it?

PR #447, opened 2018-09-04
[AudioManager] Add new APIs for USB Audio output device

sangchul1011:
1. Add new APIs regarding sample format and rate for USB audio output device.
2.  'AvoidResampling' and 'MediaStreamOnly' properties are added in AudioDevice class.

 - ACR: TCSACR-182


src/Tizen.Multimedia/AudioManager/AudioDevice.cs
```diff
@@ -104,6 +107,185 @@ public bool IsRunning
...
+        /// <remarks>
+        /// This device should be <see cref="AudioDeviceType.UsbAudio"/> type and <see cref="AudioDeviceIoDirection.Output"/> direction.
+        /// </remarks>
+        /// <exception cref="ArgumentException">This device is not valid or is disconnected.</exception>
+        /// <exception cref="InvalidOperationException">This device is not <see cref="AudioDeviceType.UsbAudio"/> type or is not <see cref="AudioDeviceIoDirection.Output"/> direction.</exception>
+        /// <since_tizen> 5 </since_tizen>
+        public IEnumerable<AudioSampleFormat> GetSupportedSampleFormats()
+        {
+            IntPtr formats;
+            uint numOfElems;
```
hsgwon: We don't use abbreviation in C#.

src/Tizen.Multimedia/AudioManager/AudioSampleFormat.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        /// <summary>
+        /// Unsigned 8 bit samples.
+        /// </summary>
+        U8,
+
+        /// <summary>
+        /// Signed 16 bit samples.
+        /// </summary>
+        S16_LE,
```
hsgwon: Don't use underline. please refer ```AudioSampleType``` enum newly updated.

src/Tizen.Multimedia/AudioManager/AudioSampleFormat.cs
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Tizen.Multimedia
+{
+    public enum AudioSampleFormat
```
hsgwon: put this enum into AudioIOEnums.cs or seperate them all like this.

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
```diff
@@ -104,6 +107,185 @@ public bool IsRunning
...
+
+            Interop.AudioDevice.GetSupportedSampleFormats(_id, out formats, out numOfElems).
```
hsgwon: ```(_id, out formats, out numOfElems)``` -> ``` (_id, out IntPtr formats, out uint numOfElements)```

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
```diff
@@ -104,6 +107,185 @@ public bool IsRunning
...
+        {
+            IntPtr rates;
+            uint numOfElems;
+
+            Interop.AudioDevice.GetSupportedSampleRates(_id, out rates, out numOfElems).
+                ThrowIfError("Failed to get supported sample formats");
+
+            return RetrieveRates();
+
+            IEnumerable<uint> RetrieveRates()
```
hsgwon: How about extracting mehod to use commonly with GetSupportedSampleFormats()?

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
```diff
@@ -104,6 +107,185 @@ public bool IsRunning
...
+        /// <exception cref="ArgumentException">This device is not valid or is disconnected.</exception>
+        /// <exception cref="InvalidOperationException">This device is not <see cref="AudioDeviceType.UsbAudio"/> type or is not <see cref="AudioDeviceIoDirection.Output"/> direction.</exception>
+        /// </remarks>
+        /// <since_tizen> 5 </since_tizen>
+        public AudioSampleFormat SampleFormat
+        {
+            get
+            {
+                Interop.AudioDevice.GetSampleFormat(_id, out AudioSampleFormat format).
+                    ThrowIfError("Failed to get sample format of the device");
```
WonyoungChoi: X AVOID throwing exceptions from property getters. 
see https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
```diff
@@ -104,6 +107,185 @@ public bool IsRunning
...
+        /// </remarks>
+        /// <exception cref="ArgumentException">This device is not valid or is disconnected.</exception>
+        /// <exception cref="InvalidOperationException">This device is not <see cref="AudioDeviceType.UsbAudio"/> type or is not <see cref="AudioDeviceIoDirection.Output"/> direction.</exception>
+        /// <since_tizen> 5 </since_tizen>
+        public uint SampleRate
+        {
+            get
+            {
+                Interop.AudioDevice.GetSampleRate(_id, out uint rate).
+                    ThrowIfError("Failed to get sample rate of the device");
```
WonyoungChoi: X AVOID throwing exceptions from property getters. 
see https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
```diff
@@ -104,6 +107,185 @@ public bool IsRunning
...
+        /// </remarks>
+        /// <exception cref="ArgumentException">This device is not valid or is disconnected.</exception>
+        /// <exception cref="InvalidOperationException">This device is not <see cref="AudioDeviceType.UsbAudio"/> type or is not <see cref="AudioDeviceIoDirection.Output"/> direction.</exception>
+        /// <since_tizen> 5 </since_tizen>
+        public bool AvoidResampling
+        {
+            get
+            {
+                Interop.AudioDevice.GetAvoidResampling(_id, out bool enabled).
+                    ThrowIfError("Failed to get avoid-resampling property of the device");
```
WonyoungChoi: X AVOID throwing exceptions from property getters. 
see https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
```diff
@@ -104,6 +107,185 @@ public bool IsRunning
...
+        /// </remarks>
+        /// <exception cref="ArgumentException">This device is not valid or is disconnected.</exception>
+        /// <exception cref="InvalidOperationException">This device is not <see cref="AudioDeviceType.UsbAudio"/> type or is not <see cref="AudioDeviceIoDirection.Output"/> direction.</exception>
+        /// <since_tizen> 5 </since_tizen>
+        public bool MediaStreamOnly
+        {
+            get
+            {
+                Interop.AudioDevice.GetMediaStreamOnly(_id, out bool enabled).
+                    ThrowIfError("Failed to get media-stream-only property of the device");
```
WonyoungChoi: X AVOID throwing exceptions from property getters. 
see https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
```diff
@@ -104,6 +107,185 @@ public bool IsRunning
...
+        /// <exception cref="ArgumentException">This device is not valid or is disconnected.</exception>
+        /// <exception cref="InvalidOperationException">This device is not <see cref="AudioDeviceType.UsbAudio"/> type or is not <see cref="AudioDeviceIoDirection.Output"/> direction.</exception>
+        /// </remarks>
+        /// <since_tizen> 5 </since_tizen>
+        public IEnumerable<uint> GetSupportedSampleRates()
+        {
+            IntPtr rates;
+            uint numOfElems;
+
+            Interop.AudioDevice.GetSupportedSampleRates(_id, out rates, out numOfElems).
```
hsgwon: ```out IntPtr rates, out uint numberOfElements```

src/Tizen.Multimedia/AudioManager/AudioSampleFormat.cs
sangchul1011: I think that no need to put this enum to audio io's.
There's a chance to add more enum values in audio manager regardless of audio io's, I think currently it is fine to divide each file as is.

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
sangchul1011: done.

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
sangchul1011: done.

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
sangchul1011: I'll revise it with set/get methods instead of it.

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
sangchul1011: I'll revise it with set/get methods instead of it.

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
sangchul1011: I'll revise it with set/get methods instead of it.

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
sangchul1011: I'll revise it with set/get methods instead of it.

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
sangchul1011: The latest patch, as a result, implementation of RetrieveXXX() differs from each other because of adding convert rate value in GetSupportedSampleRates().

WonyoungChoi:
Run Build Checker

WonyoungChoi:
Run Build Checker

JoonghyunCho:
Run Build Checker

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
```diff
@@ -104,9 +107,252 @@ public bool IsRunning
...
+        /// </remarks>
+        /// <exception cref="ArgumentException">This device is not valid or is disconnected.</exception>
+        /// <exception cref="InvalidOperationException">This device is not <see cref="AudioDeviceType.UsbAudio"/> type or is not <see cref="AudioDeviceIoDirection.Output"/> direction.</exception>
+        /// <since_tizen> 5 </since_tizen>
+        public IEnumerable<AudioSampleFormat> GetSupportedSampleFormats()
+        {
+            Interop.AudioDevice.GetSupportedSampleFormats(_id, out IntPtr formats, out uint numberOfElements).
+                ThrowIfError("Failed to get supported sample formats");
+
+            return RetrieveFormats();
```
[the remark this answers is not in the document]
sangchul1011: I think it's good to comply with the same style GetConnectedDevices() in AudioManager.cs.
We can take advantage of 'local' function as well.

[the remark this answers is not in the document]
myroot: I ask again, why do you use 'local' function in here and GetConnectedDevices()?

What is 'local' function adevantage?

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
```diff
@@ -104,9 +107,252 @@ public bool IsRunning
...
+                case 5: return 88200;
+                case 6: return 96000;
+                case 7: return 192000;
+                default:
+                    Log.Error(Tag, $"unknown value from core:{value}");
+                    return 0;
+            }
+        }
+
+        private uint ConvertRateToCoreValue(uint rate)
```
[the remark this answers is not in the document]
hsgwon: It causes some performance advantage. Please refer below full deaciption of CA1822.

Members that do not access instance data or call instance methods can be marked as static (Shared in Visual Basic). After you mark the methods as static, the compiler will emit nonvirtual call sites to these members. Emitting nonvirtual call sites will prevent a check at runtime for each call that makes sure that the current object pointer is non-null. This can achieve a measurable performance gain for performance-sensitive code. In some cases, the failure to access the current object instance represents a correctness issue.

src/Tizen.Multimedia/Interop/Interop.Device.cs
```diff
@@ -60,6 +60,36 @@ internal static partial class AudioDevice
...
+            [DllImport(Libraries.SoundManager, EntryPoint = "sound_manager_set_sample_format_by_id")]
+            internal static extern AudioManagerError SetSampleFormat(int deviceId, AudioSampleFormat format);
+
+            [DllImport(Libraries.SoundManager, EntryPoint = "sound_manager_get_sample_format_by_id")]
+            internal static extern AudioManagerError GetSampleFormat(int deviceId, out AudioSampleFormat format);
+
+            [DllImport(Libraries.SoundManager, EntryPoint = "sound_manager_get_supported_sample_rates_by_id")]
+            internal static extern AudioManagerError GetSupportedSampleRates(int deviceId, out IntPtr rates, out uint numberOfElements);
+
+            [DllImport(Libraries.SoundManager, EntryPoint = "sound_manager_set_sample_rate_by_id")]
```
hsgwon: Why don't you use ENUM like Native FW? ```rate``` parameter is ```sound_sample_rate_e``` enum.
Is there any reason for this? 
I think, if you introduce it at C# layer, you don't need to covert it at all.

```
typedef enum {
	SOUND_SAMPLE_RATE_8000,  
	...
	SOUND_SAMPLE_RATE_192000,
} sound_sample_rate_e;
```

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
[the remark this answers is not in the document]
hsgwon: As I said before, I think it's better to extract local functions to method. Because, the simillar logic is in GetSupportedSampleRates() method as local function.

How about extract it as generic method? (after remove converting logic. please refer below comment.)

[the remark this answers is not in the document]
sangchul1011: the advantage is as below.
"Local functions make the intent of your code clear. Anyone reading you code can see that the method is not callable except by the containing method. For team projects, they also make it impossible for another developer to mistakenly call the method directly from elsewhere in the class or struct."

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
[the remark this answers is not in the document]
sangchul1011: Ok, I'll apply it as static considering the performance advantage.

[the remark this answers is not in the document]
hsgwon: Please check below comment before you change this.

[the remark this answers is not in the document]
sangchul1011: Sure, I'm considering your comments including below.
Seems to think which is more useful for user.

src/Tizen.Multimedia/Interop/Interop.Device.cs
sangchul1011: We need to think that the change like Native APIs is really useful for C# user from the C# API point of view.

hsgwon: I think enum value is more useful and help to prevent misuse(intellisense help to choice it).  Actually AudioIO API use sample rate as int, but there's no resctition for it. But your API doesn't allowed another sample rate expect you predefined.

sangchul1011: It differs from audio-io situation. In case of audio-io, s/w resampler is in charge of resampling, so it's not that strict. In contrast, in case of use device, these are from supported values of h/w device(of course via ALSA). So if user set any value without quiring supported rates from API, it will occurs an error. This situation is also same even if we make it enum(intellisense does not know the supported sample rates anyway).

src/Tizen.Multimedia/AudioManager/AudioDevice.cs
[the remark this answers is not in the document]
myroot: I don't feel to extract method as local function, because `GetSupportedSampleFormats` method is obviously simple.
But, I respect your opinion

PR #457, opened 2018-09-06
[PrivacyPrivilegeManager] Add new API for checking & requesting multiple privileges

xerrni:
### Description of Change ###

C# bindings for checking & requesting multiple privileges using single API call

### API Changes ###
 - TCSACR-184
(It's related Native ACR-1253.)

#### Class PrivacyPrivilegeManager: ####
Method:
```csharp
public static IEnumerable<CheckResult> CheckPermissions(IEnumerable<string> privileges);
public static Task<RequestMultipleResponseEventArgs> RequestPermissions(IEnumerable<string> privileges);
```
#### Class PermissionRequestResponse: ####
Event:
```csharp
public RequestResult Result { get; internal set; }
public string Privilege { get; internal set; }
```
#### Class RequestResponseEventArgs : EventArgs ####
Property:
```csharp
public PermissionRequestResponse Response { get; }
```
#### Class RequestMultipleResponseEventArgs : EventArgs ####
Property:
```csharp
public CallCause Cause { get; internal set; }
public IEnumerable<PermissionRequestResponse> Responses { get; internal set; }
```

### DIFF
--- a/src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
+++ b/src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
@@ -1,5 +1,5 @@
-﻿/*
- * Copyright (c) 2017 Samsung Electronics Co., Ltd All Rights Reserved
+/*
+ * Copyright (c) 2017 - 2018 Samsung Electronics Co., Ltd All Rights Reserved
  *
  * Licensed under the Apache License, Version 2.0 (the License);
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
 using Tizen.Internals.Errors;
 
 namespace Tizen.Security
@@ -27,22 +29,51 @@ namespace Tizen.Security
     public static class PrivacyPrivilegeManager
     {
         private const string LogTag = "Tizen.Privilege";
-        private static Interop.PrivacyPrivilegeManager.RequestResponseCallback s_requestResponseCb;
-        private static IDictionary<string, ResponseContext> s_responseMap = new Dictionary<string, ResponseContext>();
+        private static Interop.PrivacyPrivilegeManager.RequestResponseCallback s_requestResponseCb =
+                       (Interop.PrivacyPrivilegeManager.CallCause cause, Interop.PrivacyPrivilegeManager.RequestResult result,
+                           string privilege, IntPtr userData) =>
+                        {
+                            try
+                            {
+                                if (s_responseMap.TryGetValue(privilege, out ResponseContext ctx) && ctx != null)
+                                {
+                                    ctx.FireEvent((CallCause)cause, (RequestResult)result);
 
-        static PrivacyPrivilegeManager()
+                                }
+                                else
+                                {
+                                    s_responseMap.Remove(privilege);
+                                    Log.Error(LogTag, "No listener for: " + privilege);
+                                }
+                            }
+                            catch (Exception e)
+                            {
+                                Log.Error(LogTag, "Exception in callback : " + e.Message);
+                            }
+                            s_PrivilegesInProgress.Remove(privilege);
+                        };
+
+        private static IDictionary<string, ResponseContext> s_responseMap = new Dictionary<string, ResponseContext>();
+        private static HashSet<string> s_PrivilegesInProgress = new HashSet<string>();
+
+        private static string[] CheckPrivilegesArgument(IEnumerable<string> privileges, string methodName)
         {
-            s_requestResponseCb = (Interop.PrivacyPrivilegeManager.CallCause cause, Interop.PrivacyPrivilegeManager.RequestResult result, string privilege, IntPtr userData) =>
+            if (privileges == null || !privileges.Any())
             {
-                try
+                Log.Error(LogTag, "privileges for " + methodName + " are null or empty.");
+                throw new ArgumentException("privileges for " + methodName + " are null or empty.");
+            }
+
+            foreach (var privilege in privileges)
+            {
+                if (string.IsNullOrEmpty(privilege))
                 {
-                    s_responseMap[privilege].FireEvent((CallCause)cause, (RequestResult) result);
+                    Log.Error(LogTag, " At least one privilege for " + methodName + " is null or empty.");
+                    throw new ArgumentException(" At least one privilege for " + methodName + " is null or empty.");
                 }
-                catch (Exception e)
-                {
-                    Log.Error(LogTag, "Exception in callback : " + e.Message);
-                }
-            };
+            }
+
+            return privileges as string[] ?? privileges.ToArray();
         }
 
         /// <summary>
@@ -85,6 +116,61 @@ namespace Tizen.Security
         }
 
         /// <summary>
+        /// Gets the status of a privacy privileges permission.
+        /// </summary>
+        /// <param name="privileges">The privacy privileges to be checked.</param>
+        /// <returns>The permission setting for a respective privileges.</returns>
+        /// <exception cref="ArgumentException">Thrown when an invalid parameter is passed.</exception>
+        /// <exception cref="OutOfMemoryException">Thrown when a memory error occurred.</exception>
+        /// <exception cref="System.IO.IOException">Thrown when the method failed due to an internal I/O error.</exception>
+        /// <example>
+        /// <code>
+        ///     string[] privileges = new [] {"http://tizen.org/privilege/account.read",
+        ///                                   "http://tizen.org/privilege/alarm"};
+        ///     CheckResult[] results = PrivacyPrivilegeManager.CheckPermissions(privileges).ToArray();
+        ///     List&lt;string&gt; privilegesWithAskStatus = new List&lt;string&gt;();
+        ///     for (int iterator = 0; iterator &lt; results.Length; ++iterator)
+        ///     {
+        ///         switch (results[iterator])
+        ///         {
+        ///             case CheckResult.Allow:
+        ///                 // Privilege can be used
+        ///                 break;
+        ///             case CheckResult.Deny:
+        ///                 // Privilege can't be used
+        ///                 break;
+        ///             case CheckResult.Ask:
+        ///                 // User permission request required
+        ///                 privilegesWithAskStatus.Add(privileges[iterator]);
+        ///                 break;
+        ///         }
+        ///     }
+        ///     PrivacyPrivilegeManager.RequestPermissions(privilegesWithAskStatus);
+        /// </code>
+        /// </example>
+        /// <since_tizen> 5 </since_tizen>
+        public static IEnumerable<CheckResult> CheckPermissions(IEnumerable<string> privileges)
+        {
+            string[] privilegesArray = CheckPrivilegesArgument(privileges, "CheckPermissions");
+
+            Interop.PrivacyPrivilegeManager.CheckResult[] results = new Interop.PrivacyPrivilegeManager.CheckResult[privilegesArray.Length];
+            int ret = (int)Interop.PrivacyPrivilegeManager.CheckPermissions(privilegesArray, (uint)privilegesArray.Length, results);
+            if (ret != (int)Interop.PrivacyPrivilegeManager.ErrorCode.None)
+            {
+                Log.Error(LogTag, "Failed to check permission");
+                throw PrivacyPrivilegeManagerErrorFactory.GetException(ret);
+            }
+
+            CheckResult[] checkResults = new CheckResult[results.Length];
+            for (int iterator = 0; iterator < results.Length; ++iterator)
+            {
+                checkResults[iterator] = (CheckResult)results[iterator];
+            }
+            return checkResults;
+        }
+
+
+        /// <summary>
         /// Triggers the permission request for a user.
         /// </summary>
         /// <param name="privilege">The privacy privilege to be requested.</param>
@@ -112,11 +198,117 @@ namespace Tizen.Security
         /// <since_tizen> 4 </since_tizen>
         public static void RequestPermission(string privilege)
         {
+            if (!s_PrivilegesInProgress.Add(privilege))
+            {
+                Log.Error(LogTag, "Request for this privilege: " + privilege + " is already in progress.");
+                throw new ArgumentException("Request for this privilege: " + privilege + " is already in progress.");
+            }
+
             int ret = (int)Interop.PrivacyPrivilegeManager.RequestPermission(privilege, s_requestResponseCb, IntPtr.Zero);
             if (ret != (int)Interop.PrivacyPrivilegeManager.ErrorCode.None)
             {
                 Log.Error(LogTag, "Failed to request permission");
+                s_PrivilegesInProgress.Remove(privilege);
                 throw PrivacyPrivilegeManagerErrorFactory.GetException(ret);
+            }
+        }
+
+        /// <summary>
+        /// Triggers the permissions request for a user.
+        /// </summary>
+        /// <param name="privileges">The privacy privileges to be requested.</param>
+        /// <exception cref="ArgumentException">Thrown when an invalid parameter is passed.</exception>
+        /// <exception cref="OutOfMemoryException">Thrown when a memory error occurred.</exception>
+        /// <exception cref="System.IO.IOException">Thrown when the method failed due to an internal I/O error.</exception>
+        /// <returns>Permission request Task</returns>
+        /// <example>
+        /// <code>
+        ///     string[] privileges = new [] {"http://tizen.org/privilege/account.read",
+        ///                                   "http://tizen.org/privilege/alarm"};
+        ///     CheckResult[] results = PrivacyPrivilegeManager.CheckPermissions(privileges).ToArray();
+        ///     List&lt;string&gt; privilegesWithAskStatus = new List&lt;string&gt;();
+        ///     for (int iterator = 0; iterator &lt; results.Length; ++iterator)
+        ///     {
+        ///         switch (results[iterator])
+        ///         {
+        ///             case CheckResult.Allow:
+        ///                 // Privilege can be used
+        ///                 break;
+        ///             case CheckResult.Deny:
+        ///                 // Privilege can't be used
+        ///                 break;
+        ///             case CheckResult.Ask:
+        ///                 // User permission request required
+        ///                 privilegesWithAskStatus.Add(privileges[iterator]);
+        ///                 break;
+        ///         }
+        ///     }
+        ///     IEnumerable&lt;PermissionRequestResponse&gt; responses = PrivacyPrivilegeManager.RequestPermissions(privilegesWithAskStatus).Result;
+        ///     //handle responses
+        /// </code>
+        /// </example>
+        /// <since_tizen> 5 </since_tizen>
+        public static Task<RequestMultipleResponseEventArgs> RequestPermissions(IEnumerable<string> privileges)
+        {
+            string[] privilegesArray = CheckPrivilegesArgument(privileges, "RequestPermissions");
+
+            for (int iterator = 0; iterator < privilegesArray.Length; ++iterator)
+            {
+                if (!s_PrivilegesInProgress.Add(privilegesArray[iterator]))
+                {
+                    Log.Error(LogTag, "Request for this privilege: " + privilegesArray[iterator] + " is already in progress.");
+
+                    for (int removeIterator = iterator - 1; removeIterator >= 0; --removeIterator)
+                    {
+                        s_PrivilegesInProgress.Remove(privilegesArray[removeIterator]);
+                    }
+                    Log.Error(LogTag, "Request for this privilege: " + privilegesArray[iterator] + " is already in progress.");
+                    throw new ArgumentException("Request for this privilege: " + privilegesArray[iterator] + " is already in progress.");
+                }
+            }
+
+            Log.Info(LogTag, "Sending request for permissions: " + string.Join(" ", privilegesArray));
+
+            TaskCompletionSource<RequestMultipleResponseEventArgs> permissionResponsesTask = new TaskCompletionSource<RequestMultipleResponseEventArgs>();
+            int ret = (int)Interop.PrivacyPrivilegeManager.RequestPermissions(privilegesArray, (uint)privilegesArray.Length,
+                        (Interop.PrivacyPrivilegeManager.CallCause cause, Interop.PrivacyPrivilegeManager.RequestResult[] results,
+                        string[] requestedPrivileges, uint privilegesCount, IntPtr userData) =>
+                        {
+                            Log.Info(LogTag, "Sending request for permissions: ");
+                            RequestMultipleResponseEventArgs requestResponse = new RequestMultipleResponseEventArgs();
+                            PermissionRequestResponse[] permissionResponses = new PermissionRequestResponse[privilegesCount];
+
+                            for (int iterator = 0; iterator < privilegesCount; ++iterator)
+                            {
+                                permissionResponses[iterator] = new PermissionRequestResponse
+                                {
+                                    Privilege = requestedPrivileges[iterator],
+                                    Result = (RequestResult)results[iterator]
+                                };
+                            }
+                            requestResponse.Cause = (CallCause)cause;
+                            requestResponse.Responses = permissionResponses;
+
+                            foreach (string privilege in requestedPrivileges)
+                            {
+                                s_PrivilegesInProgress.Remove(privilege);
+                            }
+                            permissionResponsesTask.SetResult(requestResponse);
+                        }, IntPtr.Zero);
+
+            if (ret != (int)Interop.PrivacyPrivilegeManager.ErrorCode.None)
+            {
+                Log.Error(LogTag, "Failed to request permissions.");
+                foreach (string privilege in privileges)
+                {
+                    s_PrivilegesInProgress.Remove(privilege);
+                }
+                throw PrivacyPrivilegeManagerErrorFactory.GetException(ret);
+            }
+            else
+            {
+                Log.Info(LogTag, "Requesting permissions successfull.");
+                return permissionResponsesTask.Task;
             }
         }
 
@@ -173,9 +365,10 @@ namespace Tizen.Security
         /// <since_tizen> 4 </since_tizen>
         public static WeakReference<ResponseContext> GetResponseContext(string privilege)
         {
-            if (!s_responseMap.ContainsKey(privilege))
+            if (!(s_responseMap.TryGetValue(privilege, out ResponseContext ctx) && ctx != null))
             {
-                s_responseMap[privilege] = new ResponseContext(privilege);
+                ctx = new ResponseContext(privilege);
+                s_responseMap[privilege] = ctx;
             }
             return new WeakReference<ResponseContext>(s_responseMap[privilege]);
         }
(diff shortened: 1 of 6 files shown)


src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -120,6 +211,83 @@ public static void RequestPermission(string privilege)
...
+        ///                 // User permission request required
+        ///                 privilegesWithAskStatus.Add(privileges[iterator]);
+        ///                 break;
+        ///         }
+        ///     }
+        ///     PrivacyPrivilegeManager.RequestPermissions(privilegesWithAskStatus.ToArray());
+        /// </code>
+        /// </example>
+        /// <since_tizen> 5 </since_tizen>
+        public static void RequestPermissions(string[] privileges)
```
myroot: Why don't you provide async method?

public static Task<RequestResponse> RequestPermissions(string[] privileges)

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -180,6 +348,91 @@ public static WeakReference<ResponseContext> GetResponseContext(string privilege
...
+        ///
+        ///     PrivacyPrivilegeManager.GetResponseContext(privilegesWithAskStatus.ToArray()).TryGetTarget(out context);
+        ///     if (context != null)
+        ///     {
+        ///         context.ResponseFetched -= PPM_RequestMultipleResponses;
+        ///     }
+        /// </code>
+        /// </example>
+        /// <since_tizen> 5 </since_tizen>
+        public static WeakReference<MultipleResponseContext> GetResponseContext(string[] privileges)
```
myroot: Why do you use WeakReference?

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
myroot: And do not use array in API
Use `IEnumerable<string>` instead of `string[]`

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
xerrni: 1. To prevent memory leaks when handlers that are attached to event sources will not be destroyed at the same time.

Reference:
https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/aa970850(v=vs.100)
2. To be consistent with
```csharp
public static WeakReference<ResponseContext> GetResponseContext(string privilege)
```

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
myroot: 2. Why do you think so?
Please run below code
``` c#
            if (typeof(string[]).IsAssignableFrom(typeof(IEnumerable<string>))) {
                System.Console.WriteLine("true");
            }            
            string[] result = new string[10];
            IEnumerable<string> mylist = result;
```

myroot: Please see .net framework design guide

X DO NOT use ArrayList or List<T> in public APIs.
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
myroot: Why do you think so?
It is not meaory leaks, because it released by GC,
and why do you thinks it should be destroyed at the same time?

WeakReference should not used in API

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
myroot: And 
`public static void RequestPermission(string privilege)`
It should be deprecated because it is not good usability

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
myroot: I know your intent but implemention was wrong

It should be like below changes
https://github.com/myroot/TizenFX/commit/9579961b3931dd52d92cb6e03f8f2a7f3b5a8cb1

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/RequestResponseMultipleEventArgs.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// </summary>
+        /// <since_tizen> 5 </since_tizen>
+        public IEnumerable<RequestResult> Results { get; internal set; }
+
+        /// <summary>
+        /// The privileges for which a permission was requested for.
+        /// </summary>
+        /// <since_tizen> 5 </since_tizen>
+        public IEnumerable<string> Privileges { get; internal set; }
+    }
```
myroot: It is uncomfortable to use. 
Let's see below use case
``` c#
void Handler(RequestMultipleResponseEventArgs arg) {
   var privileges = arg.Privileges.ToList();
   int index = 0;
   foreach (var result in arg.Results) {
       Console.WriteLine($"Result {privileges[index]} : {result}");
       index++;
   }
}
```
Results and Privileges should be a pair becuase it is a related data. 

Consider change to IList from IEnumerable
or Make a some class to combine `Result` and `Privilege`
``` c#
class Response {
    public RequestResult result { get; }
    public string privilege { get; }
}

public class RequestMultipleResponseEventArgs : EventArgs
{
     public CallCause cause { get; }
     public IEnumerable<Response> Responses {get; }
}
public class RequestResponseEventArgs : EventArgs
{
     public CallCause cause { get; }
     public Response Response {get; }
}

```

WonyoungChoi:
Run Build Checker

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -120,6 +237,100 @@ public static void RequestPermission(string privilege)
...
+        ///                 // User permission request required
+        ///                 privilegesWithAskStatus.Add(privileges[iterator]);
+        ///                 break;
+        ///         }
+        ///     }
+        ///     PrivacyPrivilegeManager.RequestPermissions(privilegesWithAskStatus);
+        /// </code>
+        /// </example>
+        /// <since_tizen> 5 </since_tizen>
+        public static void RequestPermissions(IEnumerable<string> privileges)
```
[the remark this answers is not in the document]
xerrni: developer may not be interested in responseContext.
Subscribing to response is and should be optional.

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -180,6 +391,106 @@ public static WeakReference<ResponseContext> GetResponseContext(string privilege
...
+            else
+            {
+                privilegesArray = privileges.ToArray();
+            }
+
+            if (privilegesArray == null || privilegesArray.Length == 0)
+            {
+                throw new ArgumentException("Unable to get response context for null argument or 0 privileges.");
+            }
+            string joinedPrivileges = string.Join("", privilegesArray);
```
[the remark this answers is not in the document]
xerrni: I need to check privileges count see 473 line: privilegesArray.Length == 0

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/RequestResponseMultipleEventArgs.cs
xerrni: I can change is for `RequestMultipleResponseEventArgs` class.
`RequestResponseEventArgs` was introduced in tizen 4. We don`t want to break backwards compatibility so I could extend it.
```csharp
public class RequestResponseEventArgs : EventArgs
    {
        public CallCause cause { get; internal set; }

        public RequestResult result { get; internal set; }

        public string privilege { get; internal set; }
        public Response Response {get; }
    }
```

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -120,6 +237,100 @@ public static void RequestPermission(string privilege)
...
+            }
+
+            if (privilegesArray == null || privilegesArray.Length == 0)
+            {
+                Log.Error(LogTag, "privileges for RequestPermissions are null or privileges.Length == 0");
+                throw new ArgumentException("privileges for RequestPermissions are null or privileges.Length == 0");
+            }
+
+            int iterator = 0;
+            for (iterator = 0; iterator < privilegesArray.Length; ++iterator)
```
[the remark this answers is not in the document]
xerrni: I still have to check if each Add operation will be successful (307 line).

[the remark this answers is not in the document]
xerrni: I moved checking privileges argument to separate function as it`s used in 3 places.

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -180,6 +391,106 @@ public static WeakReference<ResponseContext> GetResponseContext(string privilege
...
+        ///     catch (Exception)
+        ///     {
+        ///         //handle exception
+        ///     }
+        /// </code>
+        /// </example>
+        /// <since_tizen> 5 </since_tizen>
+        public static MultipleResponseContext GetResponseContext(IEnumerable<string> privileges)
+        {
+            string[] privilegesArray;
```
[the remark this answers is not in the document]
xerrni: done

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -180,6 +391,106 @@ public static WeakReference<ResponseContext> GetResponseContext(string privilege
...
+        ///         PrivacyPrivilegeManager.GetResponseContext(privilegesWithAskStatus).ResponseFetched -= PPM_RequestMultipleResponses;
+        ///     }
+        ///     catch (Exception)
+        ///     {
+        ///         //handle exception
+        ///     }
+        /// </code>
+        /// </example>
+        /// <since_tizen> 5 </since_tizen>
+        public static MultipleResponseContext GetResponseContext(IEnumerable<string> privileges)
```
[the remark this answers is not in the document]
xerrni: And it should not rise when developer set context to null.

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -219,7 +531,50 @@ internal ResponseContext(string privilege)
...
-                _ResponseFetched?.Invoke(null, new RequestResponseEventArgs() { cause = _cause, result = _result, privilege = _privilege });
+                _ResponseFetched?.Invoke(null, new RequestResponseEventArgs { cause = _cause, result = _result, privilege = _privilege });
+            }
+        }
+        /// <summary>
+        /// This class manages event handlers of the privileges permission requests.
+        /// This class enables having event handlers for multiple privileges.
+        /// </summary>
+        /// <since_tizen> 5 </since_tizen>
+        public class MultipleResponseContext
```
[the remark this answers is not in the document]
xerrni: `MultipleResponseContext EventHandler` has to be accessed from outside.

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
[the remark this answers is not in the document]
myroot: Why do you think so? Subscribing is mandatory, if not, developer got a security exception

For example, to access media storage, must grant "http://tizen.org/privilege/mediastorage" privilege
So, developer write a code like below
``` c#
  TaskCompletionSource<bool> tcs = new TaskCompletionSource<bool>();
  var response = PrivacyPrivilegeManager.GetResponseContext("http://tizen.org/privilege/mediastorage");
  PrivacyPrivilegeManager.ResponseContext target;
  response.TryGetTarget(out target);
  target.ResponseFetched += (s, e) =>
  {
      tcs.SetResult(e.result == RequestResult.AllowForever);
   };
   PrivacyPrivilegeManager.RequestPermission("http://tizen.org/privilege/mediastorage");
   bool result = await tcs.Task;
   if (result) {
       // access to media storage
    } else {
        //show error message
   }
```

If you provide as a Task<RequestResult>, developer can make more simple code
``` c#
 public static Task<RequestResult> RequestPermission(string privilege)
```
``` c#
    var result = await PrivacyPrivilegeManager.RequestPermission("http://tizen.org/privilege/mediastorage");
    if (result == RequestResult.AllowForever) {
           // access to media storage
    } else {
           // show error message
    }
```
What do you think about the difference in usability between these two codes?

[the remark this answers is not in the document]
myroot: And I hope to you distinguish `event` vs `asynchronous result`.
`ResponseFetched` is not event, 
It is a respone of `RequestPermission` so, it is a asynchronus result

`event` can raise more than one. 
`asynchronous result` does not raise if not do request.

`event` has long term lifecycle. it is same with manager object, 
`asynchronous result` has short term lifecycle, it's lifetime was end when request was completed.

And, I think, ResponseFetched should not be implement as event

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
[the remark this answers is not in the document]
myroot: What about this scenario?
``` c#
GetResponseContext(...).ResponseFetched += ....;
GC.Collect();
GC.WaitForPendingFinalizers();
RequestPermissions(...);
//ResponseFetched never raise
```

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/Response.cs
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Tizen.Security
+{
+    /// <summary>
+    /// Holds information about privilege request result.
+    /// </summary>
+    public class Response
```
myroot: `Response` class name is too common

How about `PermissionRequestResponse`?

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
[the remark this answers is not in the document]
myroot: Why not use Overlaps method?
If Overlaps method return true, Add operation never be failed.

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -120,6 +203,101 @@ public static void RequestPermission(string privilege)
...
+                    throw new ArgumentException("Request for this privilege: " + privilegesArray[iterator] + " is already in progress.");
+                }
+            }
+
+            Log.Info(LogTag, "Sending request for permissions: " + string.Join(" ", privilegesArray));
+
+            TaskCompletionSource<RequestMultipleResponseEventArgs> permissionResponsesTask = new TaskCompletionSource<RequestMultipleResponseEventArgs>();
+            int ret = (int)Interop.PrivacyPrivilegeManager.RequestPermissions(privilegesArray, (uint)privilegesArray.Length,
+                        (Interop.PrivacyPrivilegeManager.CallCause cause, Interop.PrivacyPrivilegeManager.RequestResult[] results,
+                        string[] _privileges, uint privilegesCount, IntPtr userData) =>
```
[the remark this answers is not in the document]
xerrni: because `privileges` is an argument to upper method.

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -120,6 +203,101 @@ public static void RequestPermission(string privilege)
...
+                            {
+                                s_PrivilegesInProgress.Remove(privilege);
+                            }
+                            permissionResponsesTask.SetResult(requestResponse);
+                        }, IntPtr.Zero);
+
+            if (ret != (int)Interop.PrivacyPrivilegeManager.ErrorCode.None)
+            {
+                Log.Error(LogTag, "Failed to request permissions.");
+                throw PrivacyPrivilegeManagerErrorFactory.GetException(ret);
```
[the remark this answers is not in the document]
xerrni: done

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -112,6 +189,12 @@ public static CheckResult CheckPermission(string privilege)
         /// <since_tizen> 4 </since_tizen>
         public static void RequestPermission(string privilege)
         {
+            if (!s_PrivilegesInProgress.Add(privilege))
+            {
+                Log.Error(LogTag, "Request for this privilege: " + privilege + " is already in progress.");
+                throw new ArgumentException("Request for this privilege: " + privilege + " is already in progress.");
+            }
+
```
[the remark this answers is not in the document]
xerrni: I added it in 290 line

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -120,6 +203,101 @@ public static void RequestPermission(string privilege)
...
+            Log.Info(LogTag, "Sending request for permissions: " + string.Join(" ", privilegesArray));
+
+            TaskCompletionSource<RequestMultipleResponseEventArgs> permissionResponsesTask = new TaskCompletionSource<RequestMultipleResponseEventArgs>();
+            int ret = (int)Interop.PrivacyPrivilegeManager.RequestPermissions(privilegesArray, (uint)privilegesArray.Length,
+                        (Interop.PrivacyPrivilegeManager.CallCause cause, Interop.PrivacyPrivilegeManager.RequestResult[] results,
+                        string[] _privileges, uint privilegesCount, IntPtr userData) =>
+                        {
+                            Log.Info(LogTag, "Sending request for permissions: ");
+                            RequestMultipleResponseEventArgs requestResponse = new RequestMultipleResponseEventArgs();
+                            PermissionRequestResponse[] permissionResponses = new PermissionRequestResponse[privilegesCount];
```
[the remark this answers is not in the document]
xerrni: I used `Array[]` instead of `List` because I know the responses count and they doesn`t change during execution.

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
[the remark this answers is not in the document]
xerrni: I have to throw Exception because delegate `RequestResponseCallback` will not be called.

When `ppm_request_permissions` returns error it just end execution and `RequestMultipleResponseCallback` delegate will not be called.

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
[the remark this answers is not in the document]
myroot: Do not use `_` prefix. because it is for field variable.

How about use `string[] requestedPrivileges`

myroot:
Please fix this issue
https://github.com/Samsung/TizenFX/issues/466

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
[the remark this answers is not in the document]
xerrni: done

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -203,23 +386,27 @@ internal ResponseContext(string privilege)
...
             private event EventHandler<RequestResponseEventArgs> _ResponseFetched;
 
             internal void FireEvent(CallCause _cause, RequestResult _result)
             {
-                _ResponseFetched?.Invoke(null, new RequestResponseEventArgs() { cause = _cause, result = _result, privilege = _privilege });
+                _ResponseFetched?.Invoke(null, new RequestResponseEventArgs { cause = _cause, result = _result, privilege = _privilege });
+            }
+            ~ResponseContext()
+            {
+                s_responseMap.Remove(_privilege);
```
myroot: It has race condition issue, destructor was executed on GC Thread.

and it seem to be dead code, 
because, this object was strongly referenced by s_responseMaps

If _privilege was not existed in s_responseMaps, you do not need to remove from s_responseMap
If _privilege was existed in s_responseMaps, This object never destroyed

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -203,23 +386,27 @@ internal ResponseContext(string privilege)
                 add
                 {
                     _ResponseFetched += value;
+                    if (!s_responseMap.ContainsKey(_privilege))
+                    {
+                        s_responseMap[_privilege] = new ResponseContext(_privilege);
```
myroot: Why do you create a new ResponseContext?
The result of request will send to new created ResponseContext instead this object, 
You intend it?

myroot:
Here is **Updated** solution for you
https://github.com/myroot/TizenFX/commit/f26192773abb76aa190a841cad33b46fb94a890c

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
xerrni: done

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
xerrni: corrected to 
```csharp
this
```

xerrni:
> Here is **Updated** solution for you
> [myroot@f261927](https://github.com/myroot/TizenFX/commit/f26192773abb76aa190a841cad33b46fb94a890c)

Please read my comment under your solution.

myroot:
Your solution fail below case
and My solution does  not change API
Did you check my solution?

``` c#
static ResponseContext context = null; PrivacyPrivilegeManager.GetResponseContext("some privilege").TryGetTarget(out context); context.ResponseFetched += some_handler; context.ResponseFetched -= some_handler; // context was removed from s_responseMap PrivacyPrivilegeManager.GetResponseContext("some privilege"); // a new context was created and added in s_responseMap context.ResponseFetched += some_handler; // It will not fired when Response was fetched
```

xerrni:
> Your solution fail below case
> and My solution does not change API
> Did you check my solution?
> 
> ```cs
> static ResponseContext context = null; 
> static ResponseContext context2 = null; 
> PrivacyPrivilegeManager.GetResponseContext("some privilege").TryGetTarget(out context); context.ResponseFetched += some_handler; context.ResponseFetched -= some_handler; // context was removed from s_responseMap
> PrivacyPrivilegeManager.GetResponseContext("some privilege").TryGetTarget(out context2);  // a new context was created and added in s_responseMap 
> context.ResponseFetched += some_handler; // It will not fired when Response was fetched
> context2.ResponseFetched += .... ;
> ```

I fixed this.
Can you please no block this Pull Request. It has to be merged before Tizen 5.0 is released.
I will be happy to make another commit with other fixes not related to this Pull request.

myroot:
It was not fixed. If you want to merge to master, 
PLEASE, SUBMIT CORRECT CODE.

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
myroot: ??

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -27,22 +29,51 @@ namespace Tizen.Security
...
+                            try
+                            {
+                                if (s_responseMap.TryGetValue(privilege, out ResponseContext ctx) && ctx != null)
+                                {
+                                    ctx.FireEvent((CallCause)cause, (RequestResult)result);
+
+                                }
+                                else
+                                {
+                                    s_responseMap.Remove(privilege);
```
myroot: You does not need call Remove, 
becuase, TryGetValue was truning false, it means privilege was not contained into dictionary

myroot:
If you want separate a PR from bug fix PR.
DO NOT CHANGE CODE that is not related to Added API

src/Tizen.Security.PrivacyPrivilegeManager/Tizen.Security/PrivacyPrivilegeManager.cs
```diff
@@ -27,22 +29,51 @@ namespace Tizen.Security
...
     {
         private const string LogTag = "Tizen.Privilege";
-        private static Interop.PrivacyPrivilegeManager.RequestResponseCallback s_requestResponseCb;
+        private static Interop.PrivacyPrivilegeManager.RequestResponseCallback s_requestResponseCb =
+                       (Interop.PrivacyPrivilegeManager.CallCause cause, Interop.PrivacyPrivilegeManager.RequestResult result,
+                           string privilege, IntPtr userData) =>
+                        {
+                            try
+                            {
+                                if (s_responseMap.TryGetValue(privilege, out ResponseContext ctx) && ctx != null)
```
myroot: `ctx != null` is dead code, 
because, if does not existed ResponseContext, TryGetValue returing false, so `ctx != null` was not executed.
and if existed ResponseContext, ctx can not be null unless you insert null.

myroot:
https://github.com/Samsung/TizenFX/issues/466 issue was fixed, 
but code is not good

myroot:
Why don't you apply this(https://github.com/myroot/TizenFX/commit/f26192773abb76aa190a841cad33b46fb94a890c) solution? Do you have any problem to apply?

myroot:
Please reopen when ready to update

PR #461, opened 2018-09-10
[ElmSharp] Add efl_util input generator APIs in ElmSharp

arosis78:
### Description of Change ###

Add new APIs for efl-util input generator APIs in ElmSharp

### Bugs Fixed ###

None

### API Changes ###

Added:
 - void InputGenerator.GenerateKeyEvent(string key, int pressed);
 - void InputGenerator.GenerateMouseEvent(int buttons, InputPointerType type, int x, int y)
 - void InputGenerator.GenerateTouchEvent(int index, InputTouchType type, int x, int y)

### Behavioral Changes ###

None

### DIFF
--- a/src/ElmSharp/ElmSharp/InputGenerator.cs
+++ b/src/ElmSharp/ElmSharp/InputGenerator.cs
@@ -1,0 +1,242 @@
+﻿/*
+ * Copyright (c) 2016 Samsung Electronics Co., Ltd All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+
+namespace ElmSharp
+{
+    /// <summary>
+    /// Enumeration of device type generated events.
+    /// </summary>
+    /// <since_tizen> preview </since_tizen>
+    public enum InputDeviceType
+    {
+        /// <summary>
+        /// None.
+        /// </summary>
+        /// <remarks>
+        /// Deprecated since 3.0.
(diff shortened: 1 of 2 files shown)


src/ElmSharp/ElmSharp/InputGenerator.cs
```diff
@@ -0,0 +1,242 @@
...
+        /// </remarks>
+        All = (TouchScreen | Keyboard),
+
+        /// <summary>
+        /// Max.
+        /// </summary>
+        /// <remarks>
+        /// Deprecated since 3.0.
+        /// </remarks>
+        Max = (1 << 10),
```
myroot: I think, we don't need a Max value.

src/ElmSharp/ElmSharp/InputGenerator.cs
```diff
@@ -0,0 +1,242 @@
...
+        /// </summary>
+        End,
+
+        /// <summary>
+        /// Max.
+        /// </summary>
+        /// <remarks>
+        /// Deprecated since 3.0.
+        /// </remarks>
+        Max = 10,
```
myroot: I think, we don't need a Max value.

src/ElmSharp/ElmSharp/InputGenerator.cs
```diff
@@ -0,0 +1,242 @@
...
+        protected virtual void Dispose(bool disposing)
+        {
+            if (_isDisposed)
+                return;
+
+            if (disposing)
+            {
+                Interop.Eutil.efl_util_input_deinitialize_generator(_handle);
+            }
+
```
myroot: Usually unmanaged object always released ,whether disposing is true or not
Is there some issue when execute efl_util_input_deinitialize_generator API in GC thread?

src/ElmSharp/ElmSharp/InputGenerator.cs
arosis78: Done

src/ElmSharp/ElmSharp/InputGenerator.cs
arosis78: Done

src/ElmSharp/ElmSharp/InputGenerator.cs
arosis78: Right. This API don't have thread safety, so there is issue when execute in GC thread.
But resource and device will be managed on window system when process is killed.

PR #489, opened 2018-10-01
[Shortcut] Update add, delete func

mk5004lee:
- use sync api

Signed-off-by: mk5004.lee <mk5004.lee@samsung.com>

Description of Change

Change the native API to sync.
 shortcut_add_to_home -> shortcut_add_to_home_sync
 shortcut_add_to_home_widget -> shortcut_add_to_home_widget_sync
 shortcut_remove_from_home -> shortcut_remove_from_home_sync

Bugs Fixed
•Crash occured when native code is abnormal.
 There was a problem with DotNet_TCT.


mk5004lee:
Yes.

WonyoungChoi:
@mk5004lee I will try to push this patch to MCD for wearable products. But in case of VD, you should discuss with your VD counter part if you really want to apply to TV products because you should consider the OCN schedules.

PR #515, opened 2018-10-16
Push

sameerppradhan:
### Description of Change ###

Changing the class to static class to prevent delegate from being garbage collected if its the only cause.

### Bugs Fixed ###

P181005-00070

### API Changes ###

 - ACR: 

Added:
 - SafeBundleHandle Bundle.SafeBundleHandle { get; } // Property
 - void Bundle.AddItem(string key, string value);

Changed:
 - object Bundle.ReceiveItem(string key) => object Bundle.GetItem(string key);

### Behavioral Changes ###

Describe any non-bug related behavioral changes that may change how users app behaves when upgrading to this version of the codebase.

### DIFF
--- a/src/Tizen.Messaging.Push/Tizen.Messaging.Push/PushClient.cs
+++ b/src/Tizen.Messaging.Push/Tizen.Messaging.Push/PushClient.cs
@@ -84,7 +84,7 @@ namespace Tizen.Messaging.Push
         /// <param name="pushAppId"> The Push Application ID Registered with the server.</param>
         public static void PushServiceConnect(string pushAppId)
         {
-            PushImpl.Instance.PushServiceConnect(pushAppId);
+            PushImpl.PushServiceConnect(pushAppId);
         }
 
         /// <summary>
@@ -93,7 +93,7 @@ namespace Tizen.Messaging.Push
         /// <since_tizen> 3 </since_tizen>
         public static void PushServiceDisconnect()
         {
-            PushImpl.Instance.PushServiceDisconnect();
+            PushImpl.PushServiceDisconnect();
             //PushImpl.Reset();
         }
 
@@ -107,7 +107,7 @@ namespace Tizen.Messaging.Push
         /// </returns>
         public static Task<ServerResponse> PushServerRegister()
         {
-            return PushImpl.Instance.PushServerRegister();
+            return PushImpl.PushServerRegister();
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


myroot:
Value type does not garbage collected. IntPtr is vaule type

sameerppradhan:
http://168.219.243.246:8090/pages/viewpage.action?pageId=28300363
Was asked to follow the above link and about the non-static member.

sameerppradhan:
AFAIK, static classes are used to create for utility classes where all the member functions are static.
And we are creating instances inside the class too. 
When you try to create an instance to the static class, it again generates a compile time error, because the static members can be accessed directly with its class name.

WonyoungChoi:
@sameerppradhan I still can't understand why the _connection varaible should be changed to a static variable. The PushImpl has a singleton pattern and the variable is a member of the singleton instance. So that variable is not removed by GC.

myroot:
> 
> 
> AFAIK, static classes are used to create for utility classes where all the member functions are static.
> And we are creating instances inside the class too.
> When you try to create an instance to the static class, it again generates a compile time error, because the static members can be accessed directly with its class name.

If you want make a static class, no more create a instance

myroot:
It is way to change static class https://github.com/myroot/TizenFX/commit/65ce6c4106a2b816ba30c8ffaf354feb851aa9a1

And Do you know about IntPtr? it is a umnaged object, It does not released by GC.

sameerppradhan:
> @sameerppradhan I still can't understand why the _connection varaible should be changed to a static variable. The PushImpl has a singleton pattern and the variable is a member of the singleton instance. So that variable is not removed by GC.

@WonyoungChoi We are suspecting in the long run this is getting removed by GC and thats why the crash is occurring. After merging this, the automated test will be run on this.

sameerppradhan:
> It is way to change static class [myroot@65ce6c4](https://github.com/myroot/TizenFX/commit/65ce6c4106a2b816ba30c8ffaf354feb851aa9a1)
> 
> And Do you know about IntPtr? it is a umnaged object, It does not released by GC.

@myroot I have changed the static class as suggested. Thanks

sameerppradhan:
> > It is way to change static class [myroot@65ce6c4](https://github.com/myroot/TizenFX/commit/65ce6c4106a2b816ba30c8ffaf354feb851aa9a1)
> > And Do you know about IntPtr? it is a umnaged object, It does not released by GC.
> 
> @myroot I have changed the static class as suggested. Thanks

PR #526, opened 2018-10-31
[DataControl] Add current caller client appid and current provider id property

gotoextreme:
### Description of Change ###
Add current caller client appid and current provider id property

### API Changes ###
 - ACR: TCSACR-201

Added:
   - string CurrentClientAppId{ get; } // Property
   - string CurrentProviderId{ get; } // Property


src/Tizen.Applications.DataControl/Tizen.Applications.DataControl/Provider.cs
```diff
@@ -53,6 +54,22 @@ public string DataID
             private set;
         }
 
+        /// <summary>
+        /// Gets the latest client appid.
+        /// </summary>
+        /// <since_tizen> 5 </since_tizen>
```
WonyoungChoi: since_tizen : 6

PR #537, opened 2018-11-06
[NUI] Fix to make null assigning to ResourceUrl possible

dongsug-song:
### Description of Change ###
[NUI] Fix to make null assigning to ResourceUrl possible

### API Changes ###
N/A


dongsug-song:
this is the fix to allow null assigning on ResourceUrl property

myroot:
I think this PR make not allow null assign. 
becuase, Can't store a null value into ResourceUrl
``` c#
obj.ResourceUrl = null;
var shuldbeTrue = obj.ResourceUrl == null;
//shuldebeTrue is false
```

dongsug-song:
thank you for reviewing, but this ResourceUrl null assigning has been allowed form tizen_4.0 by requesting of VD. this is kind of action to reset ResourceUrl then ImageView discards previous picture shown to show nothing finally.

myroot:
How about update `propertyChanged` callback of `ResourceUrl` and `UpdateImage` method to handle null

PR #545, opened 2018-11-09
[Tizen.Applications.Common] Add SendLaunchRequestAsync Method

hjhun:

### Description of Change ###

Add a new method to support launch request API asynchronoulsy

### API Changes ###

 - ACR: TCSACR-198

Added:
 - Task<AppControlResult> AppControl.SendLaunchRequestAsync(AppControl launchRequest, AppControlReplyCallback replyAfterLaunching);
 - AppControlResult // enumeration

### DIFF
--- a/src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
@@ -660,6 +660,7 @@ namespace Tizen.Applications
         /// </remarks>
         /// <param name="launchRequest">The AppControl.</param>
         /// <param name="replyAfterLaunching">The callback function to be called when the reply is delivered.</param>
+        /// <returns>A task with the result of the launch request.</returns>
         /// <exception cref="ArgumentException">Thrown when failed because of the argument is invalid.</exception>
         /// <exception cref="Exceptions.AppNotFoundException">Thrown when the application to run is not found.</exception>
         /// <exception cref="Exceptions.LaunchRejectedException">Thrown when the launch request is rejected.</exception>
--- a/src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
@@ -18,6 +18,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Threading.Tasks;
 
 namespace Tizen.Applications
 {
@@ -45,8 +46,9 @@ namespace Tizen.Applications
     {
         private const string LogTag = "Tizen.Applications";
 
-        private static Dictionary<int, Interop.AppControl.ReplyCallback> s_replyNativeCallbackMaps = new Dictionary<int, Interop.AppControl.ReplyCallback>();
-        private static int s_replyNativeCallbackId = 0;
+        private static Dictionary<int, Interop.AppControl.ResultCallback> s_resultNativeCallbackMaps = new Dictionary<int, Interop.AppControl.ResultCallback>();
+        private static Dictionary<int, AppControlReplyCallback> s_replyCallbackMaps = new Dictionary<int, AppControlReplyCallback>();
+        private static int s_reaustId = 0;
 
         private readonly SafeAppControlHandle _handle;
 
@@ -123,7 +125,26 @@ namespace Tizen.Applications
             }
         }
 
-        #region Public Properties
+        private static Interop.AppControl.ReplyCallback s_replyNativeCallback = (launchHandle, replyHandle, result, userData) =>
+        {
+            int requestId = (int)userData;
+            lock (s_replyCallbackMaps)
+            {
+                if (s_replyCallbackMaps.ContainsKey(requestId))
+                {
+                    s_replyCallbackMaps[requestId](new AppControl(launchHandle), new AppControl(replyHandle), (AppControlReplyResult)result);
+                    if (result != Interop.AppControl.AppStartedStatus)
+                    {
+                        lock (s_replyCallbackMaps)
+                        {
+                            s_replyCallbackMaps.Remove(requestId);
+                        }
+                    }
+                }
+            }
+        };
+
+#region Public Properties
 
         /// <summary>
         /// Gets the SafeAppControlHandle instance.
@@ -423,7 +444,7 @@ namespace Tizen.Applications
             }
         }
 
-        #endregion // Public Properties
+#endregion // Public Properties
 
         /// <summary>
         /// Retrieves all applications that can be launched to handle the given app_control request.
@@ -544,26 +565,12 @@ namespace Tizen.Applications
             if (replyAfterLaunching != null)
             {
                 int id = 0;
-                lock (s_replyNativeCallbackMaps)
+                lock (s_replyCallbackMaps)
                 {
-                    id = s_replyNativeCallbackId++;
-                    s_replyNativeCallbackMaps[id] = (launchRequestHandle, replyRequestHandle, result, userData) =>
-                    {
-                        if (replyAfterLaunching != null)
-                        {
-                            Log.Debug(LogTag, "Reply Callback is launched");
-                            replyAfterLaunching(new AppControl(launchRequestHandle), new AppControl(replyRequestHandle), (AppControlReplyResult)result);
-                            if (result != Interop.AppControl.AppStartedStatus)
-                            {
-                                lock (s_replyNativeCallbackMaps)
-                                {
-                                    s_replyNativeCallbackMaps.Remove(id);
-                                }
-                            }
-                        }
-                    };
+                    id = s_reaustId++;
+                    s_replyCallbackMaps[id] = replyAfterLaunching;
                 }
-                err = Interop.AppControl.SendLaunchRequest(launchRequest._handle, s_replyNativeCallbackMaps[id], IntPtr.Zero);
+                err = Interop.AppControl.SendLaunchRequest(launchRequest._handle, s_replyNativeCallback, (IntPtr)id);
             }
             else
             {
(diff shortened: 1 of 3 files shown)


src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
```diff
@@ -640,6 +642,113 @@ public static void SendTerminateRequest(AppControl terminateRequest)
...
+                };
+            }
+
+            if (replyCallback != null)
+            {
+                int replyId = 0;
+                lock (_replyNativeCallbackMaps)
+                {
+                    replyId = _replyNativeCallbackId++;
+                    _replyNativeCallbackMaps[replyId] = (launchRequestHandle, replyRequestHandle, result, userData) =>
```
myroot: How about use a same Interop.AppControl.ReplyCallback and getting a user calllback by userData?
``` c#
 // a static constructor
 s_replyNativeCallback = (launchHandle, replyHandle, result, userData) => {
     int key = (int)userData;
    if (_replyCallbackMaps.Contains(key))
    {
           _resultCallbackMaps[key](new AppControl(launchRequestHandle), new AppControl(replyRequestHandle), (AppControlReplyResult)result);
           _resultCallbackMaps.Remove(key);
     }
 }

....

 err = Interop.AppControl.SendLaunchRequestAsync(request._handle, s_resultNativeCallback, s_replyNativeCallback , (IntPtr)key);
```

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
```diff
@@ -45,8 +45,10 @@ public class AppControl
     {
         private const string LogTag = "Tizen.Applications";
 
-        private static Dictionary<int, Interop.AppControl.ReplyCallback> s_replyNativeCallbackMaps = new Dictionary<int, Interop.AppControl.ReplyCallback>();
-        private static int s_replyNativeCallbackId = 0;
+        private static Dictionary<int, Interop.AppControl.ReplyCallback> _replyNativeCallbackMaps = new Dictionary<int, Interop.AppControl.ReplyCallback>();
```
myroot: Why remove `s_` prefix? it means a static variable

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
```diff
@@ -640,6 +642,113 @@ public static void SendTerminateRequest(AppControl terminateRequest)
...
+        /// <exception cref="ArgumentException">Thrown when failed because of the argument is invalid.</exception>
+        /// <exception cref="Exceptions.AppNotFoundException">Thrown when the application to run is not found.</exception>
+        /// <exception cref="Exceptions.LaunchFailedException">Thrown when the request failed to launch the application.</exception>
+        /// <exception cref="Exceptions.LaunchRejectedException">Thrown when the launch request is rejected.</exception>
+        /// <exception cref="Exceptions.OutOfMemoryException">Thrown when the memory is insufficient.</exception>
+        /// <exception cref="Exceptions.PermissionDeniedException">Thrown when the permission is denied.</exception>
+        /// <exception cref="TimeoutException">Thrown when failed because of timeout.</exception>
+        /// <privilege>http://tizen.org/privilege/appmanager.launch</privilege>
+        /// <since_tizen> 6 </since_tizen>
+        public static void SendLaunchRequestAsync(AppControl request, AppControlResultCallback resultCallback, AppControlReplyCallback replyCallback)
```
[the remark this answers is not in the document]
hjhun: I'll update it.

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
hjhun: I'll consider it.

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
hjhun: I'll modify it.

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
hjhun: Done

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
hjhun: Done

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
```diff
@@ -640,6 +638,99 @@ public static void SendTerminateRequest(AppControl terminateRequest)
...
+        /// <privilege>http://tizen.org/privilege/appmanager.launch</privilege>
+        /// <since_tizen> 6 </since_tizen>
+        public static async Task<AppControlResult> SendLaunchRequestAsync(AppControl launchRequest, AppControlReplyCallback replyAfterLaunching)
+        {
+            if (launchRequest == null)
+            {
+                throw new ArgumentNullException(nameof(launchRequest));
+            }
+
+            return await Task.Run(() =>
```
myroot: Is it safe for multithreaded?
Task.Run is working on worker thread, as I know ApplicationFW API dose not safe for multithread.

I recommend to use TaskCompletionSource
Here is guide for you
http://suprem.sec.samsung.net/confluence/display/TNET/How+to+Implement+Native+Async+API+to+C%23+async+API

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
```diff
@@ -640,6 +638,99 @@ public static void SendTerminateRequest(AppControl terminateRequest)
...
+
+                if (replyAfterLaunching != null)
+                {
+                    int requestId = 0;
+                    lock (s_replyNativeCallbackMaps)
+                    {
+                        requestId = s_replyNativeCallbackId++;
+                        s_replyNativeCallbackMaps[requestId] = replyAfterLaunching;
+
+                        s_replyNativeCallback = (launchRequestHandle, replyRequestHandle, result, userData) =>
```
myroot: Why do you re-create a callback for each request?
I recommand to create a s_replayNativeCallback in static constructor.

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
```diff
@@ -640,6 +638,99 @@ public static void SendTerminateRequest(AppControl terminateRequest)
...
+                            throw new Exceptions.LaunchFailedException("Launch failed");
+                        case Interop.AppControl.ErrorCode.PermissionDenied:
+                            throw new Exceptions.PermissionDeniedException("Permission denied");
+
+                        default:
+                            throw new Exceptions.LaunchRejectedException("Launch rejected");
+                    }
+                }
+
+                return appControlResult;
```
myroot: It will always returing AppControlResult.None

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
hjhun: I'll prepare the patchset to fix it. Thanks.

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
hjhun: It's not. The method can return None or AppNotFoundException or LaunchFailed. I'll adjust return values.

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
myroot: You something wrong understand. no need to use lock and Task.Run
It already imlemented as asynchroniously. just handling result callback with TaskCompletionSource

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
myroot: As i know ResultCallback is working as asynchronously
and you do not wait ResultCallback

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
hjhun: I already got it. I already checked the URL.

hjhun:
I know that. You already told me about that.
Hmm... I think I should add some comments to Native API doxygen. The result is not always None. If the caller doesn't have a permission, the result is permission denied error. And, if the timeout error is occurred(as recv() timeout), it is launch failed error.

myroot:
the result that you said, is come from result callback. and it is working as async.
But you do not wait that call of result callback.
some error case, result callback could be working as sync. but normal case it will working as async.
and appControlResult only update in result callback

myroot:
Ok, I check your native code. result callback is not async.
It should be remove '_async' suffix

hjhun:
> the result that you said, is come from result callback. and it is working as async.
> But you do not wait that call of result callback.
> some error case, result callback could be working as sync. but normal case it will working as async.
> and appControlResult only update in result callback
> 
> As an other word,
> if native result callback was working as asynchronous, your code is wrong
> 
> if native result callback was working as synchronous, your code is right. but native api naming is wrong

hjhun:
> the result that you said, is come from result callback. and it is working as async.
> But you do not wait that call of result callback.
> some error case, result callback could be working as sync. but normal case it will working as async.
> and appControlResult only update in result callback
> 
> As an other word,
> if native result callback was working as asynchronous, your code is wrong
> 
> if native result callback was working as synchronous, your code is right. but native api naming is wrong

My code is wrong. The native result callback is working as asynchronous.

src/Tizen.Applications.Common/Tizen.Applications/AppControlResult.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        /// <summary>
+        /// Invalid parameter.
+        /// </summary>
+        InvalidParameter = Interop.AppControl.ErrorCode.InvalidParameter,
+
+        /// <summary>
+        /// Out of memory.
+        /// </summary>
+        OutOfMemory = Interop.AppControl.ErrorCode.OutOfMemory,
```
myroot: Is this error code(out of memory) valid for this function?

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
```diff
@@ -640,6 +652,86 @@ public static void SendTerminateRequest(AppControl terminateRequest)
...
+            Interop.AppControl.ErrorCode err;
+
+            int resultId = 0;
+            lock (s_resultNativeCallbackMaps)
+            {
+                resultId = s_resultNativeCallbackId++;
+                s_resultNativeCallbackMaps[resultId] = (IntPtr handle, int result, IntPtr userData) =>
+                {
+                    task.SetResult((AppControlResult)result);
+                    s_resultNativeCallbackMaps.Remove((int)userData);
```
myroot: you pass a replyId  into user_data
(see 701)
and you pass IntPtr.Zero into user_data
(see 705)

Did you test this code?

myroot: in this case use local delegate
```c#
var resultCb = (IntPtr handle, int result, IntPtr userData) => {
task.SetResult((AppControlResult)result);
};
```

src/Tizen.Applications.Common/Tizen.Applications/AppControlResult.cs
hjhun: It's my mistake.  So, I've deleted it.

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
hjhun: I made a sample app and tested it separately. There was a mistake in moving the code.
When I checked with your suggested code, a crash occurred. The local delegate has been deleted by the garbage collector. Hmm.. According to the guide you wrote in confluence, the delegate is managed as a dictionary object. I think it's right.

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
```diff
@@ -640,6 +647,82 @@ public static void SendTerminateRequest(AppControl terminateRequest)
...
+            lock (s_resultNativeCallbackMaps)
+            {
+                requestId = s_reaustId++;
+                s_resultNativeCallbackMaps[requestId] = (handle, result, userData) =>
+                {
+                    task.SetResult((AppControlResult)result);
+                    lock (s_resultNativeCallbackMaps)
+                    {
+                        s_resultNativeCallbackMaps.Remove((int)userData);
+                    }
```
hjhun: lock (s_resultNativeCallbackMaps)
{
    if (s_resultNativeCallbackMaps.Contains((int)userData)) 
    {
        s_resultNativeCallbackMaps.Remove((int)userData);
    }
}

PR #547, opened 2018-11-13
[Applications.Common] Add missing application APIs

gotoextreme:
### Description of Change ###
Add missing application APIs

- app_get_device_orientation
- app_get_name
- app_get_version

### API Changes ###
 - ACR: TCSACR-197

 - DeviceOrientation CurrentDeviceOrientation {get;} // Property
 - string Name {get;} // Property
 - string Version {get;} // Property


src/Tizen.Applications.Common/Tizen.Applications/EventSystem.cs
```diff
@@ -0,0 +1,172 @@
...
+using System.Text;
+using static Interop.EventSystem;
+
+namespace Tizen.Applications
+{
+    /// <summary>
+    /// The EventSystem API provides functions to receive events.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class EventSystem : IDisposable
```
semun-lee: I think the name "EventSystem" is too specific. Actually, this class does not about "system" it is just about "Event". How about renaming to Event?

src/Tizen.Applications.Common/Tizen.Applications/EventSystem.cs
```diff
@@ -0,0 +1,172 @@
...
+        /// <exception cref="System.OutOfMemoryException">Thrown when out of memory.</exception>
+        /// <example>
+        /// <code>
+        /// EventSystem ev = new EventSystem("tizen.system.event.screen_autorotate_state");
+        /// ev.EventReceived += EventReceivedCallback;
+        /// ev.Listen();
+        /// </code>
+        /// </example>
+        /// <since_tizen> 6 </since_tizen>
+        public void Listen()
```
semun-lee: Is this API really needed? As an app developer, calling Listen() after adding event handler is duplicated job, I think.

src/Tizen.Applications.Common/Tizen.Applications/EventSystem.cs
```diff
@@ -0,0 +1,172 @@
...
+        /// <example>
+        /// <code>
+        /// EventSystem ev = new EventSystem("tizen.system.event.screen_autorotate_state");
+        /// ev.EventReceived += EventReceivedCallback;
+        /// ev.Listen();
+        /// ev.StopListening();
+        /// </code>
+        /// </example>
+        /// <since_tizen> 6 </since_tizen>
+        public void StopListening()
```
semun-lee: I think this API may be removed also. We can remove the native handler when all the event handlers are removed.

src/Tizen.Applications.Common/Tizen.Applications/ApplicationRunningContext.cs
```diff
@@ -255,6 +255,32 @@ public void Terminate()
             }
         }
 
+        /// <summary>
+        /// Terminates the background application.
+        /// </summary>
+        /// <exception cref="ArgumentException">Thrown when failed of invalid argument.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when failed because of permission denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when failed because of system error.</exception>
+        /// <privilege>http://tizen.org/privilege/appmanager.kill</privilege>
```
semun-lee: app_manager_request_terminate_bg_app() is a public API with http://tizen.org/privilege/appmanager.kill.bgapp privilege. I think this API can be public also.

src/Tizen.Applications.Common/Tizen.Applications/Event.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Text;
+using static Interop.Event;
+
+namespace Tizen.Applications
+{
+    /// <summary>
+    /// The Event API provides functions to receive events.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class Event : IDisposable
```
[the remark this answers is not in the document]
myroot: Please do not use reserved keyword on class name

src/Tizen.Applications.Common/Tizen.Applications/ApplicationEventArgs.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// An extended EventArgs class, which contains event data.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class ApplicationEventArgs : EventArgs
+    {
+        /// <summary>
+        /// The message passed from the event system.
+        /// </summary>
+        /// <since_tizen> 3 </since_tizen>
+        public Bundle EventData { get; internal set; }
```
myroot: Why not `Data`? It is already named as `EventArgs`
So, `Event` is duplication of word. 
```
eventArgs.EventData vs eventArgs.Data
```

src/Tizen.Applications.Common/Tizen.Applications/Event.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// Event ev = new Event("tizen.system.event.usb_status");
+        /// ev += EventReceivedCallback;
+        /// </code>
+        /// </example>
+        /// <since_tizen> 6 </since_tizen>
+        public Event(string eventName)
+        {
+            if (String.IsNullOrEmpty(eventName))
+            {
+                throw new InvalidOperationException("null event name");
```
myroot: Why not InvalidArgumentException?

src/Tizen.Applications.Common/Tizen.Applications/Event.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+            _eventName = eventName;
+            _eventCallBack = (string name, IntPtr eventData, IntPtr userData) =>
+            {
+                ApplicationEventArgs args = new ApplicationEventArgs()
+                {
+                    EventData = new Bundle(new SafeBundleHandle(eventData, false))
```
myroot: Why not has ownership of Bundle?
It has a issue lifecycle of `EventData`

Q1. eventData was valid on outside of this callback?

Q1-1. if valid outside of this callback, 
   Who is release evetData and when is released? If anyone dose not release it, 
    new created Bundle object shoud has owner ship of this eventData
Q1-2. if invalid outside of this callback.
  You should copy a eventData to make sure valid on outside of this callback.

src/Tizen.Applications.Common/Tizen.Applications/Event.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <param name="eventName">The name of the event.</param>
+        /// <exception cref="System.ArgumentException">Thrown when eventName is null or empty.</exception>
+        /// <example>
+        /// <code>
+        /// Event ev = new Event("tizen.system.event.usb_status");
+        /// ev += EventReceivedCallback;
+        /// </code>
+        /// </example>
+        /// <since_tizen> 6 </since_tizen>
+        public Event(string eventName)
```
myroot: It is too poor usability to use. Developer hard to know which eventName is available on this parameter.
Please check AppControlOperations class

gotoextreme: I'll make another PR for Event.
For now, i need to discuss design for Event with my coworkers.
Thank you for your code review.

src/Tizen.Applications.Common/Tizen.Applications/Application.cs
```diff
@@ -82,6 +82,47 @@ public ApplicationInfo ApplicationInfo
             }
         }
 
+        /// <summary>
+        /// Gets the current device orientation.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public DeviceOrientation CurrentDeviceOrientation
```
myroot: I think It seem to valid only for UIApplication

src/Tizen.Applications.Common/Tizen.Applications/ApplicationRunningContext.cs
```diff
@@ -255,6 +255,31 @@ public void Terminate()
...
 
+        /// <summary>
+        /// Terminates the background application.
+        /// </summary>
+        /// <exception cref="ArgumentException">Thrown when failed of invalid argument.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when failed because of permission denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when failed because of system error.</exception>
+        /// <privilege>http://tizen.org/privilege/appmanager.kill.bgapp</privilege>
+        /// <since_tizen> 6 </since_tizen>
+        public void TerminateBgApp()
```
myroot: Why it added in ApplicationRunningContext class instead of ApplicationManager?

myroot: What is different from `Terminate()`

myroot:
You change file mode to 755 from 644. I think source code does not need execute permission.

src/Tizen.Applications.Common/Tizen.Applications/ApplicationRunningContext.cs
gotoextreme: We intend to use ApplicationRunningContext as managing application's running context, so it contains Resume, Terminate methods.

And TerminateBgApp is terminate only background application as summary described

src/Tizen.Applications.Common/Tizen.Applications/Application.cs
gotoextreme: I removed this attribute. I'll move it to CoreUIApplication

src/Tizen.Applications.Common/Tizen.Applications/ApplicationRunningContext.cs
myroot: As your comment
How about adding overloading method of Terminate()?

```suggestion
        public void Terminate(bool onlyForBackground);
```

myroot: And need to update description. 

`Terminates the application if it is running on background`

gotoextreme: Terminate(false); case is internal API we do not have plan to open it public

myroot: How about add in ApplicationManager class as static?

public static class ApplicationManager {
   public static void TerminateBackgroundApplication(ApplicationRunningContext app);
}

gotoextreme: Our design intend to managing app running context using ApplicationRunningContext.
So, Resume() and internal Terminate() methods also included in ApplicationRunningContext.
static TerminateBackgroundApplication method in ApplicationManager class is not consistent with our design.

myroot: Terminate is acceptable for method of ApplicationRunningContext.

But, TerminateBackgroundApplication is not acceptable for ApplicationRunningContext

myroot: as you say, Terminate is internal API and if we can change api signature, How about change behavior of Terminate() as terminate background app?
and orignal api is changed to Terminate(bool includeForground)

src/Tizen.Applications.Common/Tizen.Applications/ApplicationRunningContext.cs
```diff
@@ -156,6 +159,34 @@ public string ApplicationId
             }
         }
 
+        /// <summary>
+        /// Gets whether the application is terminated.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public bool IsTerminated
```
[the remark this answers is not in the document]
semun-lee: This change is for supporting legacy native API feature in C# API. I think changing State property to support live state is a kind of API semantic break. I'd like to keep IsTerminated property seperately.

However, I think it would be better if there is a 'remark' explaining this can be used to check 'current state' of RunningContext in contrast with State property.

src/Tizen.Applications.Common/Tizen.Applications/ApplicationManager.cs
```diff
@@ -170,6 +171,31 @@ public static async Task<IEnumerable<ApplicationInfo>> GetInstalledApplicationsA
...
 
+        /// <summary>
+        /// Terminates the application if it is running on background.
+        /// </summary>
+        /// <exception cref="ArgumentException">Thrown when failed of invalid argument.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when failed because of permission denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when failed because of system error.</exception>
+        /// <privilege>http://tizen.org/privilege/appmanager.kill.bgapp</privilege>
+        /// <since_tizen> 6 </since_tizen>
+        public static void TerminateBackgroundApplication(ApplicationRunningContext app)
```
semun-lee: Let's add remark.

> This function returns after it just sends a request for terminating a background application.
Platform will decide if the target application could be terminated or not according to the state of the target application.

src/Tizen.Applications.Common/Interop/Interop.AppCommon.cs
```diff
@@ -71,6 +71,9 @@ internal enum ResourceCategory : int
         [DllImport(Libraries.AppCommon, EntryPoint = "app_get_version")]
         internal static extern ErrorCode AppGetVersion(out string version);
 
+        [DllImport(Libraries.AppCommon, EntryPoint = "app_get_device_orientation")]
+        internal static extern DeviceOrientation AppGetDeviceOrientation();
```
myroot: This API not used

PR #551, opened 2018-11-14
[AudioManager] Apply dispose pattern with the finalizer

sangchul1011:

### Description of Change ###
Apply dispose pattern with the finalizer

### DIFF
--- a/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
+++ b/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
@@ -59,6 +59,14 @@ namespace Tizen.Multimedia
         }
 
         /// <summary>
+        /// Finalizes an instance of the AudioStreamPolicy class.
+        /// </summary>
+        ~AudioStreamPolicy()
+        {
+            Dispose(false);
+        }
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
```diff
@@ -320,12 +329,14 @@ public void Dispose()
...
+            if (_disposed)
             {
-                if (_handle != null)
-                {
-                    _handle.Dispose();
-                }
+                return;
+            }
+
+            if (_handle != null)
```
myroot: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern
Dispose() only called when disposing was true

sangchul1011: I missed the _handle inherit safeHandle. I'll remove the finalizer again which is not necessary in this class and make some modifications in Dispose(bool disposing) again.

sangchul1011: It is now replaced with https://github.com/Samsung/TizenFX/pull/553

PR #596, opened 2018-12-04
[STT/TTS] Fix desctructor to call dispose

syeonholic:
Signed-off-by: sooyeon.kim <sooyeon.kim@samsung.com>

### Description of Change ###

Fix desctructor to call dispose in STT/TTS

### DIFF
--- a/src/Tizen.Uix.Stt/Tizen.Uix.Stt/SttClient.cs
+++ b/src/Tizen.Uix.Stt/Tizen.Uix.Stt/SttClient.cs
@@ -364,6 +364,14 @@ namespace Tizen.Uix.Stt
         }
 
         /// <summary>
+        /// Destructor to destroy a STT instance.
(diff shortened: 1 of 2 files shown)


myroot:
now only if disposing was true, _handle(unmanaged object) was released.
if it called from destructor, _handle does not released.

myroot:
Unmanaged object is alway released in Dispose(boo) method

``` c#
        {
            if (!disposedValue)
            {
                lock (thisLock)
                {
                    if (_handle != IntPtr.Zero)
                    {
                        SttError error = SttDestroy(_handle);
                        if (error != SttError.None)
                        {
                            Log.Error(LogTag, "Destroy Failed with error " + error);
                        }
                        _handle = IntPtr.Zero;
                   }
            }

                disposedValue = true;
            }
        }
```

PR #600, opened 2018-12-06
[NUI] Disconnect all native signals of Window class when application …

dongsug-song:
### Description of Change ###
[NUI] Disconnect all native signals of Window class when application is terminated
  - defensive code for timing critical issue (application is terminating and CLR is disposing)

### API Changes ###
None

### DIFF
--- a/src/Tizen.NUI/src/public/Window.cs
+++ b/src/Tizen.NUI/src/public/Window.cs
@@ -1789,5 +1789,63 @@ namespace Tizen.NUI
             }
         }
 
+        /// <summary>
+        /// Disconnect all native signals
+        /// </summary>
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Window.cs
```diff
@@ -1789,5 +1789,63 @@ protected override bool ReleaseHandle()
             }
         }
 
+        /// <summary>
+        /// Disconnect all native signals
+        /// </summary>
+        /// <since_tizen> 5 </since_tizen>
```
WonyoungChoi: internal scope apis does not require the since_tizen tag

PR #616, opened 2018-12-17
[Applications.Common] Fix Metadata Getter

hjhun:
- Using static function

### Description of Change ###

To avoid handling callback function(relocating memory position) by GC, this patch changes to use static function.

### API Changes ###

 - ACR: N/A

### DIFF
--- a/src/Tizen.Applications.Common/Tizen.Applications/ApplicationInfo.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications/ApplicationInfo.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Runtime.InteropServices;
 
 namespace Tizen.Applications
 {
@@ -193,6 +194,21 @@ namespace Tizen.Applications
             }
         }
 
+        private static Interop.ApplicationManager.AppInfoMetadataCallback s_appInfoMetadataCallback = (string key, string value, IntPtr userData) =>
+        {
+            GCHandle handle = (GCHandle)userData;
+            IDictionary<string, string> metadata = (handle.Target as IDictionary<string, string>);
+
+            if (key.Length != 0)
+            {
+                if (!metadata.ContainsKey(key))
+                {
+                    metadata.Add(key, value);
+                }
+            }
+            return true;
+        };
+
         /// <summary>
         /// Gets the application's metadata.
         /// </summary>
@@ -201,27 +217,20 @@ namespace Tizen.Applications
         {
             get
             {
-                IDictionary<string, string> metadata = new Dictionary<String, String>();
-
-                Interop.ApplicationManager.AppInfoMetadataCallback cb = (string key, string value, IntPtr userData) =>
-                {
-                    if (key.Length != 0)
-                    {
-                        if (!metadata.ContainsKey(key))
-                            metadata.Add(key, value);
-                    }
-                    return true;
-                };
+                IDictionary<string, string> metadata = new Dictionary<string, string>();
+                GCHandle handle = GCHandle.Alloc(metadata);
 
                 IntPtr infoHandle = GetInfoHandle();
                 if (infoHandle != IntPtr.Zero)
                 {
-                    err = Interop.ApplicationManager.AppInfoForeachMetadata(infoHandle, cb, IntPtr.Zero);
+                    err = Interop.ApplicationManager.AppInfoForeachMetadata(infoHandle, s_appInfoMetadataCallback, (IntPtr)handle);
                     if (err != Interop.ApplicationManager.ErrorCode.None)
                     {
                         Log.Warn(LogTag, "Failed to get application metadata of " + _applicationId + ". err = " + err);
                     }
                 }
+
+                handle.Free();
                 return metadata;
             }
         }


hjhun:
My comments was wrong. This patch is not for preventing the delegate function from deleting.
After GC, the memory position is relocated. The native code cannot handle it(relocated memory) directly. 

I got the crash issue:
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:58
[Current thread is 1 (LWP 14488)]


(gdb)bt
#0  __GI_raise (sig=sig@entry=6) at /usr/src/debug/glibc-2.24/signal/../sysdeps/unix/sysv/linux/raise.c:58
#1  0xb654d47c in __GI_abort () at /usr/src/debug/glibc-2.24/stdlib/abort.c:89
#2  0xb6360186 in __gnu_cxx::__verbose_terminate_handler () at /usr/src/debug/gcc-6.2.1/obj/armv7l-tizen-linux-gnueabi/libstdc++-v3/libsupc++/../../../../libstdc++-v3/libsupc++/vterminate.cc:95
#3  0xb635ebf4 in __cxxabiv1::__terminate (handler=<optimized out>) at /usr/src/debug/gcc-6.2.1/obj/armv7l-tizen-linux-gnueabi/libstdc++-v3/libsupc++/../../../../libstdc++-v3/libsupc++/eh_terminate.cc:47
#4  0xb635ec42 in std::terminate () at /usr/src/debug/gcc-6.2.1/obj/armv7l-tizen-linux-gnueabi/libstdc++-v3/libsupc++/../../../../libstdc++-v3/libsupc++/eh_terminate.cc:57
#5  0xb635ee4e in __cxxabiv1::__cxa_throw (obj=0xada28540, tinfo=0x1, dest=0xb3ccb405 <PAL_SEHException::~PAL_SEHException()>) at /usr/src/debug/gcc-6.2.1/obj/armv7l-tizen-linux-gnueabi/libstdc++-v3/libsupc++/../../../../libstdc++-v3/libsupc++/eh_throw.cc:87
#6  0xb3e7ffea in RtlpRaiseException (ExceptionRecord=0xada26438, ContextRecord=0xada26298) at /usr/src/debug/coreclr-2.1.4/src/pal/src/exception/seh-unwind.cpp:452
#7  0xb3e7ffa6 in RaiseException (dwExceptionCode=<optimized out>, dwExceptionFlags=<optimized out>, nNumberOfArguments=15, lpArguments=<optimized out>) at /usr/src/debug/coreclr-2.1.4/src/pal/src/exception/seh-unwind.cpp:519
#8  0xb3bbddfe in COMPlusThrowBoot (hr=<optimized out>) at /usr/src/debug/coreclr-2.1.4/src/vm/excep.cpp:167
#9  0xb3c4e08a in TheUMEntryPrestubWorker (pUMEntryThunk=0xb3749274) at /usr/src/debug/coreclr-2.1.4/src/vm/dllimportcallback.cpp:1064
#10 0xb3cd3a92 in TheUMEntryPrestub () from ./symbols/usr/share/dotnet/shared/Microsoft.NETCore.App/2.1.4/libcoreclr.so
#11 0xb45712a0 in app_info_foreach_app_metadata_cb (metadata_key=<optimized out>, metadata_value=<optimized out>, user_data=<optimized out>) at /usr/src/debug/capi-appfw-app-manager-0.7.1/src/app_info.c:189
#12 0xb62cb14e in pkgmgrinfo_appinfo_foreach_metadata (handle=<optimized out>, metadata_func=0xb4571291 <app_info_foreach_app_metadata_cb>, user_data=user_data@entry=0xadcfe594) at /usr/src/debug/pkgmgr-info-0.13.5/src/pkgmgrinfo_appinfo.c:2701

src/Tizen.Applications.Common/Tizen.Applications/ApplicationInfo.cs
```diff
@@ -201,27 +217,20 @@ public string ApplicationType
...
-                {
-                    if (key.Length != 0)
-                    {
-                        if (!metadata.ContainsKey(key))
-                            metadata.Add(key, value);
-                    }
-                    return true;
-                };
+                IDictionary<string, string> metadata = new Dictionary<string, string>();
+                GCHandle handle = GCHandle.Alloc(metadata);
```
myroot: How about hold only cb object on CGHandle?
```c#
            get
            {
                IDictionary<string, string> metadata = new Dictionary<String, String>();

                Interop.ApplicationManager.AppInfoMetadataCallback cb = (string key, string value, IntPtr userData) =>
                {
                    if (key.Length != 0)
                    {
                        if (!metadata.ContainsKey(key))
                            metadata.Add(key, value);
                    }
                    return true;
                };
                var handle = GCHandle.Alloc(cb, GCHandleType.Pinned);

                IntPtr infoHandle = GetInfoHandle();
                if (infoHandle != IntPtr.Zero)
                {
                    err = Interop.ApplicationManager.AppInfoForeachMetadata(infoHandle, cb, IntPtr.Zero);
                    if (err != Interop.ApplicationManager.ErrorCode.None)
                    {
                        Log.Warn(LogTag, "Failed to get application metadata of " + _applicationId + ". err = " + err);
                    }
                }
                handle.Free();

                return metadata;
            }
```

hjhun: I've tested your suggestion with GCHandleType(Pinned).;
ArgumentException is occurred.

01-01 10:41:54.178+0900 I/DOTNET_LAUNCHER(P12042, T12048): System.ArgumentException: Object contains non-primitive or non-blittable data.
01-01 10:41:54.178+0900 I/DOTNET_LAUNCHER(P12042, T12048):    at System.Runtime.InteropServices.GCHandle.InternalAlloc(Object value, GCHandleType type)
01-01 10:41:54.178+0900 I/DOTNET_LAUNCHER(P12042, T12048):    at System.Runtime.InteropServices.GCHandle.Alloc(Object value, GCHandleType type)

myroot: Please refer this article
https://blogs.msdn.microsoft.com/cbrumme/2003/05/06/asynchronous-operations-pinning/

>Along the same lines, managed Delegates can be marshaled to unmanaged code, where they are exposed as unmanaged function pointers. Calls on those pointers will perform an unmanaged to managed transition; a change in calling convention; entry into the correct AppDomain; and any necessary argument marshaling. Clearly the unmanaged function pointer must refer to a fixed address. It would be a disaster if the GC were relocating that! This leads many applications to create a pinning handle for the delegate. This is completely unnecessary. The unmanaged function pointer actually refers to a native code stub that we dynamically generate to perform the transition & marshaling. This stub exists in fixed memory outside of the GC heap.

일단, delegate는 pinning할 필요가 없다고 합니다.
Crash가 발생한 원인이 다른곳에 있는지 확인이 필요합니다.

semun-lee: >     // Check for ShutDown scenario.  This happens only when we have initiated shutdown 
>     // and someone is trying to call in after the CLR is suspended.  In that case, we
>     // must either raise an unmanaged exception or return an HRESULT, depending on the
>     // expectations of our caller.
>     if (!CanRunManagedCode())
> 

this comment might be helpful.

https://github.com/dotnet/coreclr/blob/master/src/vm/dllimportcallback.cpp#L851

WonyoungChoi: @hjhun Could you register this app crash issue to https://github.sec.samsung.net/dotnet/home/issues? With the callstack, It seems that the VM can't execute the managed code in special case.

hjhun: https://github.sec.samsung.net/dotnet/home/issues/252

myroot: crash callstack과 delegate memory reallocated 이슈와는 관계가 없는것으로 보입니다.

semun-lee: Let's abandon this patch and trace issue with https://github.sec.samsung.net/dotnet/home/issues/252

hjhun:
This is unnecessary. We should check the issue again.

PR #623, opened 2018-12-19
[NUI] Remove VK/GL switching

dongsug-song:
### Description of Change ###
[NUI] Remove VK/GL switching
- Add plugin-parser in dali-extension, this makes symbolic link of dali-csharp-binder.so in Application's local lib folder

### API Changes ###
None


elishateng:
Hello, Mr. SSong,
After the patch (b9a8cfc4eab8c5d436440381da7c1b7972f9f0d6 nui-vulkan backend support (#353) ) merged, there are still some patches related Pinvoke merged, I think these changes should be also included into this patch.

PR #625, opened 2018-12-19
[Metadata] Add new API and deprecate old one

hsgwon:
### Description of Change ###
Add `Composer` property and deprecate `Author` property

### API Changes ###
 - TCSACR-209

### DIFF
--- a/src/Tizen.Multimedia.Metadata/MetadataExtractor/Metadata.cs
+++ b/src/Tizen.Multimedia.Metadata/MetadataExtractor/Metadata.cs
@@ -196,7 +196,8 @@ namespace Tizen.Multimedia
             Album = extractor.GetMetadata(MetadataExtractorAttr.Album);
             AlbumArtist = extractor.GetMetadata(MetadataExtractorAttr.AlbumArtist);
             Genre = extractor.GetMetadata(MetadataExtractorAttr.Genre);
-            Author = extractor.GetMetadata(MetadataExtractorAttr.Author);
+            Author = extractor.GetMetadata(MetadataExtractorAttr.Composer);
+            Composer = extractor.GetMetadata(MetadataExtractorAttr.Composer);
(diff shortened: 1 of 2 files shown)


src/Tizen.Multimedia.Metadata/MetadataExtractor/Metadata.cs
```diff
@@ -196,7 +196,8 @@ internal Metadata(MetadataExtractor extractor)
             Album = extractor.GetMetadata(MetadataExtractorAttr.Album);
             AlbumArtist = extractor.GetMetadata(MetadataExtractorAttr.AlbumArtist);
             Genre = extractor.GetMetadata(MetadataExtractorAttr.Genre);
-            Author = extractor.GetMetadata(MetadataExtractorAttr.Author);
+            Author = extractor.GetMetadata(MetadataExtractorAttr.Composer);
```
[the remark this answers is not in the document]
hsgwon: It cannot be removed in this version.

src/Tizen.Multimedia.Metadata/MetadataExtractor/MetadataExtractorAttr.cs
```diff
@@ -34,7 +34,8 @@ internal enum MetadataExtractorAttr
         Album,
         AlbumArtist,
         Genre,
-        Author,
+        Author, // This will be deprecated. Please use `Composer` instead.
```
rookiejava: This is for internal use, so you can remove it. Or, if you want to keep it as obsolete, 
```cs
[Obsolete("This will be deprecated. Please use `Composer` instead.")]
Author,
``` 
would be better.

hsgwon: OK I'll remove it.

PR #632, opened 2018-12-26
[DevicePolicyManager] Add new DevicePolicyManager API

y01k:
### Description of Change ###

Add new DevicePolicyManager API

### API Changes ###

 - ACR: TCSACR-211

### DIFF
--- a/src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/BluetoothPolicy.cs
+++ b/src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/BluetoothPolicy.cs
@@ -19,7 +19,7 @@ using System;
 namespace Tizen.Security.DevicePolicyManager
 {
     /// <summary>
-    /// The BluetoothPolicy provides methods to control bluetooth policies.
+    /// The BluetoothPolicy provides methods to manage Bluetooth policies.
     /// </summary>
     /// <since_tizen> 6 </since_tizen>
     public class BluetoothPolicy
@@ -45,7 +45,7 @@ namespace Tizen.Security.DevicePolicyManager
         }
 
         /// <summary>
-        /// The BluetoothPolicyChanged event is raised when the bluetooth policy is changed.
+        /// The BluetoothPolicyChanged event is raised when the Bluetooth policy is changed.
         /// </summary>
         /// <since_tizen> 6 </since_tizen>
         public event EventHandler<PolicyChangedEventArgs> BluetoothPolicyChanged
@@ -97,7 +97,7 @@ namespace Tizen.Security.DevicePolicyManager
         }
 
         /// <summary>
-        /// The BluetoothTetheringPolicyChanged event is raised when the bluetooth tethering policy is changed.
+        /// The BluetoothTetheringPolicyChanged event is raised when the Bluetooth tethering policy is changed.
         /// </summary>
         /// <since_tizen> 6 </since_tizen>
         public event EventHandler<PolicyChangedEventArgs> BluetoothTetheringPolicyChanged
--- a/src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/BluetoothPolicy.cs
+++ b/src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/BluetoothPolicy.cs
@@ -56,8 +56,10 @@ namespace Tizen.Security.DevicePolicyManager
                 {
                     AddBluetoothPolicyChangedCallback();
                 }
+
                 _bluetoothPolicyChanged += value;
             }
+
             remove
             {
                 _bluetoothPolicyChanged -= value;
@@ -94,6 +96,7 @@ namespace Tizen.Security.DevicePolicyManager
                 Log.Error(Globals.LogTag, "Failed to remove policy changed callback, name " + _bluetoothPolicyName + ", ret : " + ret);
                 throw DevicePolicyManagerErrorFactory.GetException(ret);
             }
+
             _bluetoothPolicyChangedCallback = null;
             _bluetoothCallbackId = 0;
         }
@@ -110,8 +113,10 @@ namespace Tizen.Security.DevicePolicyManager
                 {
                     AddBluetoothTetheringPolicyChangedCallback();
                 }
+
                 _bluetoothTetheringPolicyChanged += value;
             }
+
             remove
             {
                 _bluetoothTetheringPolicyChanged -= value;
@@ -148,6 +153,7 @@ namespace Tizen.Security.DevicePolicyManager
                 Log.Error(Globals.LogTag, "Failed to remove policy changed callback, name " + _bluetoothTetheringPolicyName + ", ret : " + ret);
                 throw DevicePolicyManagerErrorFactory.GetException(ret);
             }
+
             _bluetoothTetheringPolicyChangedCallback = null;
             _bluetoothTetheringCallbackId = 0;
         }
--- a/src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/BluetoothPolicy.cs
+++ b/src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/BluetoothPolicy.cs
@@ -81,6 +81,7 @@ namespace Tizen.Security.DevicePolicyManager
             int ret = Interop.DevicePolicyManager.AddPolicyChangedCallback(_dpm.GetHandle(), _bluetoothPolicyName, _bluetoothPolicyChangedCallback, IntPtr.Zero, out _bluetoothCallbackId);
             if (ret != (int)Interop.DevicePolicyManager.DpmError.None)
             {
+                Log.Error(Globals.LogTag, "Failed to add policy changed callback, name " + _bluetoothPolicyName + ", ret : " + ret);
                 throw DevicePolicyManagerErrorFactory.GetException(ret);
             }
         }
@@ -90,6 +91,7 @@ namespace Tizen.Security.DevicePolicyManager
             int ret = Interop.DevicePolicyManager.RemovePolicyChangedCallback(_dpm.GetHandle(), _bluetoothCallbackId);
             if (ret != (int)Interop.DevicePolicyManager.DpmError.None)
             {
+                Log.Error(Globals.LogTag, "Failed to remove policy changed callback, name " + _bluetoothPolicyName + ", ret : " + ret);
                 throw DevicePolicyManagerErrorFactory.GetException(ret);
             }
             _bluetoothPolicyChangedCallback = null;
@@ -133,6 +135,7 @@ namespace Tizen.Security.DevicePolicyManager
             int ret = Interop.DevicePolicyManager.AddPolicyChangedCallback(_dpm.GetHandle(), _bluetoothTetheringPolicyName, _bluetoothTetheringPolicyChangedCallback, IntPtr.Zero, out _bluetoothTetheringCallbackId);
             if (ret != (int)Interop.DevicePolicyManager.DpmError.None)
             {
+                Log.Error(Globals.LogTag, "Failed to add policy changed callback, name " + _bluetoothTetheringPolicyName + ", ret : " + ret);
                 throw DevicePolicyManagerErrorFactory.GetException(ret);
             }
         }
@@ -142,6 +145,7 @@ namespace Tizen.Security.DevicePolicyManager
             int ret = Interop.DevicePolicyManager.RemovePolicyChangedCallback(_dpm.GetHandle(), _bluetoothTetheringCallbackId);
             if (ret != (int)Interop.DevicePolicyManager.DpmError.None)
             {
+                Log.Error(Globals.LogTag, "Failed to remove policy changed callback, name " + _bluetoothTetheringPolicyName + ", ret : " + ret);
                 throw DevicePolicyManagerErrorFactory.GetException(ret);
             }
             _bluetoothTetheringPolicyChangedCallback = null;
--- a/src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/BluetoothPolicy.cs
+++ b/src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/BluetoothPolicy.cs
@@ -26,7 +26,7 @@ namespace Tizen.Security.DevicePolicyManager
     {
         private readonly string _bluetoothPolicyName = "bluetooth";
         private readonly string _bluetoothTetheringPolicyName = "bluetooth_tethering";
-        private readonly DevicePolicyManager handle;
+        private readonly DevicePolicyManager _dpm;
         private int _bluetoothCallbackId;
         private int _bluetoothTetheringCallbackId;
 
@@ -41,7 +41,7 @@ namespace Tizen.Security.DevicePolicyManager
 
         internal BluetoothPolicy(DevicePolicyManager dpm)
         {
-            handle = dpm;
+            _dpm = dpm;
         }
 
         /// <summary>
@@ -78,7 +78,7 @@ namespace Tizen.Security.DevicePolicyManager
                 };
             }
 
-            int ret = Interop.DevicePolicyManager.AddPolicyChangedCallback(handle.GetHandle(), _bluetoothPolicyName, _bluetoothPolicyChangedCallback, IntPtr.Zero, out _bluetoothCallbackId);
+            int ret = Interop.DevicePolicyManager.AddPolicyChangedCallback(_dpm.GetHandle(), _bluetoothPolicyName, _bluetoothPolicyChangedCallback, IntPtr.Zero, out _bluetoothCallbackId);
             if (ret != (int)Interop.DevicePolicyManager.DpmError.None)
             {
                 throw DevicePolicyManagerErrorFactory.GetException(ret);
@@ -87,7 +87,7 @@ namespace Tizen.Security.DevicePolicyManager
 
         private void RemoveBluetoothPolicyChangedCallback()
         {
-            int ret = Interop.DevicePolicyManager.RemovePolicyChangedCallback(handle.GetHandle(), _bluetoothCallbackId);
+            int ret = Interop.DevicePolicyManager.RemovePolicyChangedCallback(_dpm.GetHandle(), _bluetoothCallbackId);
             if (ret != (int)Interop.DevicePolicyManager.DpmError.None)
             {
                 throw DevicePolicyManagerErrorFactory.GetException(ret);
@@ -130,7 +130,7 @@ namespace Tizen.Security.DevicePolicyManager
                 };
             }
 
-            int ret = Interop.DevicePolicyManager.AddPolicyChangedCallback(handle.GetHandle(), _bluetoothTetheringPolicyName, _bluetoothTetheringPolicyChangedCallback, IntPtr.Zero, out _bluetoothTetheringCallbackId);
+            int ret = Interop.DevicePolicyManager.AddPolicyChangedCallback(_dpm.GetHandle(), _bluetoothTetheringPolicyName, _bluetoothTetheringPolicyChangedCallback, IntPtr.Zero, out _bluetoothTetheringCallbackId);
             if (ret != (int)Interop.DevicePolicyManager.DpmError.None)
             {
                 throw DevicePolicyManagerErrorFactory.GetException(ret);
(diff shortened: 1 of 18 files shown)


src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/PasswordPolicy.cs
```diff
@@ -0,0 +1,202 @@
...
+        /// Complex characters are all non-alphabetic characters; that is, numbers and symbols.
+        /// </summary>
+        /// <returns>Number of minimum complex char in password.</returns>
+        /// <since_tizen> 6 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/dpm.password</privilege>
+        /// <privlevel>partner</privlevel>
+        /// <exception cref="ArgumentException">Thrown when failed because of invalid handle of DevicePolicyManager.</exception>
+        /// <exception cref="TimeoutException">Thrown when failed because of timeout.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when application does not have privilege to access this method.</exception>
+        public int GetMinComplexChars()
```
myroot: How about rename to MinimumRequiredComplexChars
```suggestion
        public int MinimumRequiredComplexChars { get { ....  }; }
```

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/TelephonyPolicy.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        /// <summary>
+        /// Checks whether the text messaging is allowed or not.
+        /// </summary>
+        /// <param name="simId">SIM identifier</param>
+        /// <returns>true if the messaging is allowed, false otherwise.</returns>
+        /// <since_tizen> 6 </since_tizen>
+        /// <exception cref="ArgumentException">Thrown when failed because of invalid parameter.</exception>
+        /// <exception cref="TimeoutException">Thrown when failed because of timeout.</exception>
+        public bool GetMessagingState(string simId)
```
myroot: How about rename to  `IsMessagingAllowed`
```suggestion
        public bool IsMessagingAllowed(string simId)
```

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/DevicePolicyManager.cs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    static internal class Globals
+    {
+        internal const string LogTag = "Tizen.Security.DPM";
+    }
+
+    internal static class DevicePolicyManagerErrorFactory
+    {
+        static internal Exception GetException(int error)
```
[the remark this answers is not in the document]
y01k: How about this? 
```
static internal void ThrowException(int error) {
    .... 
    switch (errCode)
    {
        case Interop.DevicePolicyManager.DpmError.InvalidParameter:
               throw new ArgumentException("Invalid parameter");
        ....
    }
} 
```

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/EmailPolicy.cs
```diff
@@ -0,0 +1,121 @@
...
+
+using System;
+
+namespace Tizen.Security.DevicePolicyManager
+{
+    /// <summary>
+    /// The EmailPolicy provides methods to manage email policies.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class EmailPolicy
```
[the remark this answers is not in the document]
y01k: I'll change like below. Is it okay? 

```
class EmailPolicy : IDisposable
{

        ~EmailPolicy()
        {
            this.Dispose(false);
        }

        {
            if (!_disposed)
            {
                if (disposing)
                {
                    // to be used if there are any other disposable objects
                }

                if (_popImapCallbackId != 0)
                {
                    RemovePopImapPolicyChangedCallback();
                }

                _disposed = true;
            }
       }
        public void Dispose()
        {
            Dispose(true);
            GC.SuppressFinalize(this);
        }
}
```

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/EmailPolicy.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        /// <summary>
+        /// Checks whether the access to POP or IMAP email is allowed or not.
+        /// </summary>
+        /// <returns>true if the POP or IMAP email is allowed, false otherwise.</returns>
+        /// <since_tizen> 6 </since_tizen>
+        /// <exception cref="ArgumentException">Thrown when failed because of invalid handle of DevicePolicyManager.</exception>
+        /// <exception cref="TimeoutException">Thrown when failed because of timeout.</exception>
+        public bool GetPopImapState()
```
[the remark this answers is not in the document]
y01k: I considered using the property all policies. but the GetMessagingState in TelephonyPolicy is needed argument. and I wanted to apply the same format to all policies. so I use method all policies. but Is that better using property policies except the GetMessagingState in TelephonyPolicy?

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/DevicePolicyManager.cs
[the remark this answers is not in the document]
myroot: I don't prefer it, 
because it always provide a same call stack ends with `DevicePolicyManagerErrorFactory.ThrowException` to app developer when exception was raised

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/EmailPolicy.cs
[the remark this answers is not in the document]
myroot: It will cause second issue.
```
var policy = new DevicePolicyManager().GetPolicy<EmailPolicy>();
policy.PopImapPolicyChanged += (s, e) => { // it will be not raised }
policy = null;
System.GC.Collect();
```
when native callback was riased C# event is not raised

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/EmailPolicy.cs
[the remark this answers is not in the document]
myroot: In C# prefer property, if it is a value of object.
Method only use to represent a action on about object

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/EmailPolicy.cs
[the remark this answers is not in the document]
myroot: But, Your solution also acceptable, with remarks about lifetime

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/EmailPolicy.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            _dpm = dpm;
+        }
+
+        /// <summary>
+        /// Checks whether the access to POP or IMAP email is allowed or not.
+        /// </summary>
+        /// <returns>true if the POP or IMAP email is allowed, false otherwise.</returns>
+        /// <since_tizen> 6 </since_tizen>
+        /// <exception cref="ArgumentException">Thrown when failed because of invalid handle of DevicePolicyManager.</exception>
```
myroot: Wrong exception type, because there is no argument in this method

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/EmailPolicy.cs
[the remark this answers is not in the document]
myroot: We guide to avoid throwing exception from property getter. 

>X AVOID throwing exceptions from property getters. 
    Property getters should be simple operations and should not have any preconditions. If a getter can throw an exception, it should probably be redesigned to be a method. Notice that this rule does not apply to indexers, where we do expect exceptions as a result of validating the arguments.

If you must report exception to app developer, use method. 
but, if you can replace exception with default value like `false` and it is a natural behavior to app developer. we recommand to use property getter.
and this case seem to be better to use property getter

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/WifiPolicy.cs
```diff
@@ -0,0 +1,244 @@
...
+            this.Dispose();
+        }
+
+        /// <summary>
+        /// Releases any unmanaged resources used by this object.
+        /// </summary>
+        public void Dispose()
+        {
+            this.Dispose(true);
+            GC.SuppressFinalize(this);
```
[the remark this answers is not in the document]
y01k: Done

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/PolicyState.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        DISALLOWED = 0,
+        /// <summary>
+        /// The policy status is allowed.
+        /// </summary>
+        ALLOWED,
+        /// <summary>
+        /// The policy status is unknown.
+        /// </summary>
+        UNKNOWN
```
myroot: Is it really need? why not simply use Disallowed?
`Unknow` can be returned in normal state?

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/StoragePolicy.cs
```diff
@@ -0,0 +1,146 @@
...
+                    {
+                        Log.Error(Globals.LogTag, e.ToString());
+                    }
+                }
+
+                _disposed = true;
+            }
+        }
+
+        /// <summary>
```
myroot: Why not provide API to query current policy?

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/UsbPolicy.cs
```diff
@@ -0,0 +1,146 @@
...
+            }
+        }
+
+        private void AddUsbTetheringPolicyChangedCallback()
+        {
+            if (_usbTetheringPolicyChangedCallback == null)
+            {
+                _usbTetheringPolicyChangedCallback = (string name, string state, IntPtr userData) =>
+                {
+                    _usbTetheringPolicyChanged?.Invoke(null, new PolicyChangedEventArgs(name, state));
```
myroot: Why pass `null` instead of `this`? `sender` can be null when event is a static
```suggestion
                    _usbTetheringPolicyChanged?.Invoke(this, new PolicyChangedEventArgs(name, state));
```

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/PolicyChangedEventArgs.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// Get name of the policy.
+        /// </summary>
+        public string Name { get; }
+
+        /// <summary>
+        /// Get the current state of the policy.
+        /// </summary>
+        public string State { get; }
```
myroot: Why state is string instead of PolicyStat?

string not useful for app developer, can't expect which value can be used

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/DevicePolicyManager.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <returns>An instance of policy.</returns>
+        /// <since_tizen> 6 </since_tizen>
+        /// <exception cref="InvalidOperationException">Thrown when failed to create instance of the policy.</exception>
+        public T GetPolicy<T>() where T : DevicePolicy
+        {
+            T policy = Activator.CreateInstance(typeof(T),
+                System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance,
+                null, new object[] { this }, null) as T;
+
+            if (policy == null)
```
myroot: policy can't be `null`, if can't create a object CreateInstace will throw exception.

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/DevicePolicy.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    /// <summary>
+    /// The abstract class for Policy instances.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public abstract class DevicePolicy
+    {
+        internal readonly DevicePolicyManager _dpm;
+
+        internal DevicePolicy()
```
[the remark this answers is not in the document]
y01k: I removed unnecessary default constructors.

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/DevicePolicy.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Tizen.Security.DevicePolicyManager
+{
+    /// <summary>
+    /// The abstract class for Policy instances.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public abstract class DevicePolicy
+    {
+        internal readonly DevicePolicyManager _dpm;
```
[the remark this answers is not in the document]
y01k: Done

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/PolicyState.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// <summary>
+    /// Enumeration for the device policy state.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public enum PolicyState
+    {
+        /// <summary>
+        /// The policy status is disallowed.
+        /// </summary>
+        DISALLOWED = 0,
```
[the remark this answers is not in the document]
y01k: Done

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/UsbPolicy.cs
```diff
@@ -0,0 +1,146 @@
...
+                _disposed = true;
+            }
+        }
+
+        /// <summary>
+        /// The UsbTetheringPolicyChanged event is raised when the usb tethering policy is changed.
+        /// </summary>
+        /// <remarks>This event will be removed automatically when UsbPolicy is destroyed.</remarks>
+        /// <since_tizen> 6 </since_tizen>
+        public event EventHandler<PolicyChangedEventArgs> UsbTetheringPolicyChanged
```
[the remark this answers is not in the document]
y01k: I Added properties.

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/WifiPolicy.cs
```diff
@@ -0,0 +1,244 @@
...
+            _wifiPolicyChangedCallback = null;
+            _wifiCallbackId = 0;
+        }
+
+        /// <summary>
+        /// The WifiHotspotPolicyChanged event is raised when the wifi hotspot policy is changed.
+        /// </summary>
+        /// <remarks>This event will be removed automatically when WifiPolicy is destroyed.</remarks>
+        /// <since_tizen> 6 </since_tizen>
+        public event EventHandler<PolicyChangedEventArgs> WifiHotspotPolicyChanged
```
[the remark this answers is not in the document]
y01k: I added properties.

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/PolicyState.cs
y01k: I remove "Unknown". and I added comment default value of properties.

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/StoragePolicy.cs
y01k: I added properties.

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/UsbPolicy.cs
y01k: I changed null to this.

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/PolicyChangedEventArgs.cs
y01k: I changed string to PolicyState.

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/DevicePolicyManager.cs
y01k: I remove null checking. I added try-catch and rethrow exception. but I can't be sure it is the right solution. I considered below options.
1. remove exception and just return policy
2. add try-catch and rethrow InvalidOperationException
3. add try-catch and return default(T)

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/BluetoothPolicy.cs
```diff
@@ -0,0 +1,262 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Gets whether the bluetooth tethering state change is allowed or not.
+        /// </summary>
+        /// <value>Allowed if the change is allowed, Disallowed otherwise. The default value is Allowed.</value>
+        /// <seealso cref="PolicyState"/>
+        /// <since_tizen> 6 </since_tizen>
+        public PolicyState IsBluetoothTetheringAllowed
```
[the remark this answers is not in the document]
y01k: I removed PolicyState type and replaced to bool.

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/PolicyChangedEventArgs.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// Get name of the policy.
+        /// </summary>
+        public string Name { get; }
+
+        /// <summary>
+        /// Get the current state of the policy.
+        /// </summary>
+        public PolicyState State { get; }
```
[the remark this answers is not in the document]
y01k: Done

y01k:
I added comment to each policy name.

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/PasswordPolicy.cs
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        /// <summary>
+        /// Gets the number of days password expires.
+        /// </summary>
+        /// <value>Number of days after which the password expires. The default value is maximum of int.</value>
+        /// <since_tizen> 6 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/dpm.password</privilege>
+        /// <privlevel>partner</privlevel>
+        public int ExpiresDay
```
js3576: Please consider changing the name to "DaysToExpiration".

y01k: I changed the name to "DaysToExpiration".

src/Tizen.Security.DevicePolicyManager/Tizen.Security.DevicePolicyManager/WifiPolicy.cs
```diff
@@ -0,0 +1,274 @@
...
+        /// <remarks>This is used in <see cref="PolicyChangedEventArgs.PolicyName"/>.</remarks>
+        /// <since_tizen> 6 </since_tizen>
+        public static string WifiPolicyName => "Wifi";
+
+        /// <summary>
+        /// The Wifi hotspot policy name. This represents <see cref="WifiPolicy.IsWifiHotspotAllowed"/>.
+        /// </summary>
+        /// <remarks>This is used in <see cref="PolicyChangedEventArgs.PolicyName"/>.</remarks>
+        /// <since_tizen> 6 </since_tizen>
+        public static string WifiHotspotPolicyName => "WifiHotspot";
```
myroot: it should be readonly

y01k: Done

PR #644, opened 2019-01-08
removed account_destroy() function call

abhishekvijay:
### Description of Change ###
fix submitted for resolving account-manager dll crash issue coming with DotNet test suite. The crash was coming due to calling of native function - account_destroy(). On analysis it is found that this destroy call is not required as account-manager service itself destroy's the instance when its lifecycle gets complete.


src/Tizen.Account.AccountManager/Tizen.Account.AccountManager/SafeAccountHandle.cs
```diff
@@ -60,7 +60,6 @@ public override bool IsInvalid
         /// <returns>true if the handle is released successfully</returns>
         protected override bool ReleaseHandle()
         {
-            Interop.Account.Destroy(this.handle);
             this.SetHandle(IntPtr.Zero);
```
myroot: If it does not has owner ship of handle 
please use `public SafeAccountHandle(IntPtr existingHandle, bool ownsHandle) ` with ownsHandle = false

myroot:
Please rebase to master

src/Tizen.Account.AccountManager/Tizen.Account.AccountManager/SafeAccountHandle.cs
myroot: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.safehandle.-ctor?view=netframework-4.7.2

>If the ownsHandle parameter is false, ReleaseHandle is never called; thus, it is not recommended to use this parameter value as your code may leak resources.

abhishekvijay: Yes false is used with this API.

myroot: If you remove Destroy method call in ReleaseHandle, 

It will cause memory leak when use `CreateAccount`
``` c#
        public static Account CreateAccount()
        {
            SafeAccountHandle handle;
            AccountError err = (AccountError)Interop.Account.Create(out handle);
            if (err != AccountError.None)
            {
                throw AccountErrorFactory.CreateException(err, "Failed to create new error.");
            }

            return new Account(handle);
        }
```

myroot: Why do you think the reason for the crash is Interop.Account.Destroy?

abhishekvijay: As per the test reports of native test suite, we found that crash occurs immediately after account_destroy() call. JIRA issues - TFDF-272 & TFDF-270 are raised for the same.

myroot: Do you know reproduce step?
When a crash occurred?

myroot: Please check this solution (https://github.com/myroot/TizenFX/commit/8d60061e591196a830f7dace1c3516b0c294ff57) and please test it

``` c#
internal static extern int QueryAccountById(int accountId, ref SafeAccountHandle handle);
```
It will free passed handle. but ownership does not passed by C# layer. so it cause double free

myroot:
Please update a new PR with my recommended solution.

abhishekvijay:
With these suggested changes along with previous commit changes, we found 12 TC's failing for DotNet Test suite. After removing these suggested changes and also removing the call to function - Interop.Account.Destroy(this.handle); from ReleaseHandle() function, there is no problem. Please suggest if its OK to remove call to "Interop.Account.Destroy(this.handle);" .

myroot:
Nop, do not remove "Interop.Account.Destroy(this.handle);" .
It is memory leak.

please find reason of crach

PR #656, opened 2019-01-16
[Applications] Fix static analysis issue

hjhun:
To avoid ArgumentNullException, we have to check the args before calling
base.Run().

### Description of Change ###

This patch is for checking args parameter before calling base.run().
Currently, if args is null, ArgumentNullException can be occurred.

### API Changes ###

 - ACR: Not Required

### DIFF
--- a/src/Tizen.Applications.Common/Tizen.Applications/CoreApplication.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications/CoreApplication.cs
@@ -101,17 +101,6 @@ namespace Tizen.Applications
         /// <since_tizen> 3 </since_tizen>
         public override void Run(string[] args)
         {
-            base.Run(args);
-
-            _backend.AddEventHandler(EventType.Created, OnCreate);
-            _backend.AddEventHandler(EventType.Terminated, OnTerminate);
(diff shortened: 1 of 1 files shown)


src/Tizen.Applications.Common/Tizen.Applications/CoreApplication.cs
```diff
@@ -124,6 +113,18 @@ public override void Run(string[] args)
                 args.CopyTo(argsClone, 1);
             }
             argsClone[0] = string.Empty;
+
+            base.Run(argsClone);
```
myroot: base.Run(args);

Why not allow calling with null?

myroot:
What is problem to throwing ArgumentNullException on null argument?
It need to throw ArgumentNullException to notify to App developer that argument was passing a null reference and we do not allow null

myroot:
If you allow null reference on args, please fix it do not throwing exception on null reference
https://github.com/Samsung/TizenFX/blob/e3ef2c83139d7f3985809aa2ac5be8a740b520dd/src/Tizen.Applications.Common/Tizen.Applications/Application.cs#L121-L124

hjhun:
I think I were a little impatient.
This request is not needed.

PR #661, opened 2019-01-16
Revert "[Applications] Fix static analysis issue"

myroot:
Reverts Samsung/TizenFX#659

It cause behavior change
and It is **NOT** issue throwing exception when args was null
If you want to allow null reference on args, 
please fix below code
https://github.com/Samsung/TizenFX/blob/e3ef2c83139d7f3985809aa2ac5be8a740b520dd/src/Tizen.Applications.Common/Tizen.Applications/Application.cs#L121-L124


hjhun:
I think I was a little impatient.
And, I think the pull request 656 and 659 are not needed.
Thanks for reviewing.

PR #663, opened 2019-01-17
[Bluetooth] Fix API document

Wootak:
[Bluetooth] Fix API document during ACR


src/Tizen.Network.Bluetooth/Tizen.Network.Bluetooth/BluetoothAdapter.cs
```diff
@@ -401,7 +401,7 @@ public static void Disable()
         }
 
         /// <summary>
-        /// Enables the discoverable mode.
+        /// Enables the discoverable mode for the duration.
```
myroot: This API is no duration parameter.

journey2w: myroot is correct. You have to change the summary on line 418.

PR #668, opened 2019-01-17
[Calendar] Fix svace defect  WGID#14567

sameerppradhan:
### Description of Change ###
Just added a null check before using the variable.

### API Changes ###

 - ACR:


src/Tizen.Pims.Calendar/Tizen.Pims.Calendar/CalendarStructs.cs
```diff
@@ -126,7 +126,7 @@ public int CompareTo(CalendarTime other)
         public override bool Equals(object obj)
         {
             var other = obj as CalendarTime;
-            if (_type != other._type)
+            if (other != null && _type != other._type)
```
myroot: if (other == null || _type != other._type)

sameerppradhan:
Updated as per the review comment.

Thanks

PR #669, opened 2019-01-18
[InputMethod] Remove lock from InputMethodEditor

Inhong:
IME does not need to consider multi-threading.


Inhong:
Dear @imfine98.
Please check the new comment.

PR #691, opened 2019-01-29
[STC] Add C# API wrapper for Stc CAPIs

lkasana:
Signed-off-by: Lokesh <l.kasana@samsung.com>

### Description of Change ###
Add C# API wrapper for Smart Traffic Control (Stc) CAPIs

### API Changes ###
 - ACR: TCSACR-217 (http://suprem.sec.samsung.net/jira/browse/TCSACR-217)

### DIFF
--- a/src/Tizen.Network.Stc/Interop/Interop.Libraries.cs
+++ b/src/Tizen.Network.Stc/Interop/Interop.Libraries.cs
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Samsung Electronics Co., Ltd All Rights Reserved
+ * Copyright (c) 2019 Samsung Electronics Co., Ltd All Rights Reserved
  *
  * Licensed under the Apache License, Version 2.0 (the License);
  * you may not use this file except in compliance with the License.
@@ -16,7 +16,7 @@
 
 internal static partial class Interop
 {
-    internal static partial class Libraries
+    internal static class Libraries
     {
         public const string Stc = "libcapi-network-stc.so.0";
         public const string Glib = "libglib-2.0.so.0";
--- a/src/Tizen.Network.Stc/Interop/Interop.Libraries.cs
+++ b/src/Tizen.Network.Stc/Interop/Interop.Libraries.cs
@@ -1,0 +1,25 @@
+/*
+ * Copyright (c) 2016 Samsung Electronics Co., Ltd All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+internal static partial class Interop
+{
+    internal static partial class Libraries
+    {
+        public const string Stc = "libcapi-network-stc.so.0";
+        public const string Glib = "libglib-2.0.so.0";
+        public const string Libc = "libc.so.6";
+    }
+}
--- a/src/Tizen.Network.Stc/Interop/Interop.Libraries.cs
+++ b/src/Tizen.Network.Stc/Interop/Interop.Libraries.cs
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Samsung Electronics Co., Ltd All Rights Reserved
+ * Copyright (c) 2018 Samsung Electronics Co., Ltd All Rights Reserved
  *
  * Licensed under the Apache License, Version 2.0 (the License);
  * you may not use this file except in compliance with the License.
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc.csproj
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc.csproj
@@ -1,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Version>1.0.0</Version>
+    <Authors>Samsung Electronics</Authors>
+    <Copyright>© Samsung Electronics Co., Ltd All Rights Reserved</Copyright>
+    <Description>Provides the Stc APIs for Tizen .NET</Description>
+    <PackageProjectUrl>https://www.tizen.org/</PackageProjectUrl>
+    <PackageLicenseUrl>https://www.apache.org/licenses/LICENSE-2.0</PackageLicenseUrl>
+    <PackageIconUrl>https://developer.tizen.org/sites/default/files/images/tizen-pinwheel-on-light-rgb_64_64.png</PackageIconUrl>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TargetFramework>netstandard1.3</TargetFramework>
+    <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
+    <SignAssembly>True</SignAssembly>
+    <AssemblyOriginatorKeyFile>Tizen.Network.Stc.snk</AssemblyOriginatorKeyFile>
+    <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
+    <DelaySign>False</DelaySign>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Tizen" Version="1.0.5" />
+    <PackageReference Include="Tizen.Network.Connection" Version="1.0.18" />
+  </ItemGroup>
+
+</Project>
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc.csproj
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc.csproj
@@ -1,27 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <Version>1.0.0</Version>
-    <Authors>Samsung Electronics</Authors>
-    <Copyright>© Samsung Electronics Co., Ltd All Rights Reserved</Copyright>
-    <Description>Provides the Stc APIs for Tizen .NET</Description>
-    <PackageProjectUrl>https://www.tizen.org/</PackageProjectUrl>
-    <PackageLicenseUrl>https://www.apache.org/licenses/LICENSE-2.0</PackageLicenseUrl>
-    <PackageIconUrl>https://developer.tizen.org/sites/default/files/images/tizen-pinwheel-on-light-rgb_64_64.png</PackageIconUrl>
-  </PropertyGroup>
-
-  <PropertyGroup>
-    <TargetFramework>netstandard1.3</TargetFramework>
-    <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
-    <SignAssembly>True</SignAssembly>
-    <AssemblyOriginatorKeyFile>Tizen.Network.Stc.snk</AssemblyOriginatorKeyFile>
-    <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
-    <DelaySign>False</DelaySign>
+    <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Tizen" Version="1.0.5" />
-    <PackageReference Include="Tizen.Network.Connection" Version="1.0.18" />
+    <PackageReference Include="..\Tizen\Tizen.csproj" />
+    <PackageReference Include="..\Tizen.Log\Tizen.Log.csproj" />
   </ItemGroup>
 
 </Project>
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc.csproj
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc.csproj
@@ -5,8 +5,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="..\Tizen\Tizen.csproj" />
-    <PackageReference Include="..\Tizen.Log\Tizen.Log.csproj" />
+    <ProjectReference Include="..\Tizen\Tizen.csproj" />
+    <ProjectReference Include="..\Tizen.Log\Tizen.Log.csproj" />
   </ItemGroup>
 
 </Project>
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc/NamespaceDoc.cs
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc/NamespaceDoc.cs
@@ -1,0 +1,6 @@
+/**
+<summary>
+The Tizen.Network.Stc namespace provides classes to manage and use Smart Traffic Control (STC).
+</summary>
+*/
+namespace Tizen.Network.Stc {}
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc/NamespaceDoc.cs
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc/NamespaceDoc.cs
@@ -1,6 +1,0 @@
-/**
-<summary>
-The Tizen.Network.Stc namespace provides classes to manage and use Smart Traffic Control (STC).
-</summary>
-*/
-namespace Tizen.Network.Stc {}
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
@@ -72,7 +72,7 @@ namespace Tizen.Network.Stc
     /// <summary>
     /// Internal Enumeration for network interface type.
     /// </summary>
-    internal enum NativeInterfaceTypeEnum
+    internal enum NativeNetworkInterface
     {
         /// <summary>
         /// Network interface type is Unknown.
@@ -103,7 +103,7 @@ namespace Tizen.Network.Stc
     /// <summary>
     /// Enumeration for network interface type.
     /// </summary>
-    public enum InterfaceTypeEnum
+    public enum NetworkInterface
     {
         /// <summary>
         /// Network interface type is datacall. 
@@ -254,7 +254,7 @@ namespace Tizen.Network.Stc
     /// <summary>
     /// Enumeration for return type of the callback.
     /// </summary>
-    public enum CallbackRet
+    internal enum CallbackRet
     {
         /// <summary>
         /// Cancel
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
@@ -1,0 +1,258 @@
+﻿/*
+ * Copyright (c) 2016 Samsung Electronics Co., Ltd All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+using Tizen.Internals.Errors;
+
+namespace Tizen.Network.Stc
+{
+    /// <summary>
+    /// Enumeration for Stc time period.
+    /// </summary>
+    public enum StcTimePeriod
+    {
+        /// <summary>
+        /// Unknown
+        /// </summary>
+        Unknown = 0,
+        /// <summary>
+        /// Hour
+        /// </summary>
+        Hour = 3600,
+        /// <summary>
+        /// Day
+        /// </summary>
+        Day = 86400,
+        /// <summary>
+        /// Week
+        /// </summary>
+        Week = 604800,
+        /// <summary>
+        /// Month
+        /// </summary>
+        Month = 2419200
+    }
+
+    /// <summary>
+    /// Enumeration for network interface type.
+    /// </summary>
+    public enum StcIfaceType
+    {
+        /// <summary>
+        /// Network interface type is Unknown.
+        /// </summary>
+        Unknown = 0,
+        /// <summary>
+        /// Network interface type is datacall. 
+        /// </summary>
+        Datacall,
+        /// <summary>
+        /// Network interface type is  wifi.
+        /// </summary>
+        Wifi,
+        /// <summary>
+        /// Network interface type is  wired.
+        /// </summary>
+        Wired,
+        /// <summary>
+        /// Network interface type is bluetooth.
+        /// </summary>
+        Bluetooth,
+        /// <summary>
+        /// Network interface type is all.
+        /// </summary>
+        All
+    }
+
+    /// <summary>
+    /// Enumeration for network roaming type
+    /// </summary>
+    public enum StcRoamingType
+    {
+        /// <summary>
+        /// Not in roaming.
+        /// </summary>
+        Disabled,
+        /// <summary>
+        /// In roaming.
+        /// </summary>
+        Enabled
+    }
+
+    /// <summary>
+    /// Enumeration for hardware network protocol type. 
+    /// </summary>
+    public enum StcProtocolType
+    {
+        /// <summary>
+        /// Network unknown.
+        /// </summary>
+        Unknown,
+        /// <summary>
+        /// Network no service.
+        /// </summary>
+        DatacallNoService,
+        /// <summary>
+        /// Network emergency.
+        /// </summary>
+        DatacallEmergency,
+        /// <summary>
+        /// Network search 1900.
+        /// </summary>
+        DatacallSearch,
+        /// <summary>
+        /// Network 2G.
+        /// </summary>
+        Datacall2G,
+        /// <summary>
+        /// Network 2.5G.
+        /// </summary>
+        Datacall2_5G,
+        /// <summary>
+        /// Network EDGE.
+        /// </summary>
+        Datacall2_5G_Edge,
+        /// <summary>
+        /// Network UMTS.
+        /// </summary>
+        Datacall3G,
+        /// <summary>
+        /// Network HSDPA.
+        /// </summary>
+        DatacallHsdpa,
+        /// <summary>
+        /// Network LTE.
+        /// </summary>
+        DatacallLte
+    }
+
+    /// <summary>
+    /// Enumeration for monitored process state.
+    /// </summary>
+    public enum StcProcessState
+    {
+        /// <summary>
+        /// Unknown state.
+        /// </summary>
+        Unknown,
+        /// <summary>
+        /// Foreground state.
+        /// </summary>
+        Foreground,
+        /// <summary>
+        /// Background state.
+        /// </summary>
+        Background
+   }
+
+    /// <summary>
+    /// Enumeration for network connection period type.
+    /// @details Last received/sent mean counting data from the first connection of each interface.
+    /// </summary>
+    public enum StcConnPeriodType
+    {
+        /// <summary>
+        /// Unknown period.
+        /// </summary>
+        Unknown,
+        /// <summary>
+        /// Last received data.
+        /// </summary>
+        LastReceivedData,
+        /// <summary>
+        /// Last sent data.
+        /// </summary>
+        LastSentData,
+        /// <summary>
+        /// Total received data.
+        /// </summary>
+        TotalReceivedData,
+        /// <summary>
+        /// Total sent data.
+        /// </summary>
+        TotalSentData
+    }
+
+    /// <summary>
+    /// Enumeration for return type of the callback.
+    /// </summary>
+    public enum StcCallbackRet
+    {
+        /// <summary>
+        /// Cancel
+        /// </summary>
+        Cancel,
+        /// <summary>
+        /// Continue
+        /// </summary>
+        Continue
+    }
+
+    /// <summary>
+    /// Enumeration for the Stc error code.
+    /// </summary>
+    public enum StcError
+    {
+        /// <summary>
+        /// Successful.
+        /// </summary>
+        None = ErrorCode.None,
+        /// <summary>
+        /// Operation not permitted.
+        /// </summary>
+        NotPermitted = ErrorCode.NotPermitted,
+        /// <summary>
+        /// Out of memory.
+        /// </summary>
+        OutOfMemory = ErrorCode.OutOfMemory,
+        /// <summary>
+        /// Permission denied.
+        /// </summary>
+        PermissionDenied = ErrorCode.PermissionDenied,
+        /// <summary>
+        /// Device or the resource is busy.
+        /// </summary>
+        ResourceBusy = ErrorCode.ResourceBusy,
+        /// <summary>
+        /// Invalid operation.
+        /// </summary>
+        InvalidOperation = ErrorCode.InvalidOperation, 
+        /// <summary>
+        /// Invalid function parameter.
+        /// </summary>
+        InvalidParameter = ErrorCode.InvalidParameter,
+        /// <summary>
+        /// Not supported.
+        /// </summary>
+        NotSupported = ErrorCode.NotSupported,
+        /// <summary>
+        /// Operation failed.
+        /// </summary>
+        OperationFailed = -0x02F80000 | 0x01,
+        /// <summary>
+        /// Not initialized.
+        /// </summary>
+        NotInitialized = -0x02F80000 | 0x02,
+        /// <summary>
+        /// Already initialized the client.
+        /// </summary>
+        AlreadyInitialized = -0x02F80000 | 0x03,
+        /// <summary>
+        /// In progress.
+        /// </summary>
+        InProgress = -0x02F80000 | 0x04
+    }
+}
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
@@ -1,5 +1,5 @@
 ﻿/*
- * Copyright (c) 2016 Samsung Electronics Co., Ltd All Rights Reserved
+ * Copyright (c) 2018 Samsung Electronics Co., Ltd All Rights Reserved
  *
  * Licensed under the Apache License, Version 2.0 (the License);
  * you may not use this file except in compliance with the License.
@@ -49,7 +49,7 @@ namespace Tizen.Network.Stc
     /// <summary>
     /// Enumeration for network interface type.
     /// </summary>
-    public enum StcIfaceType
+    public enum StcInterfaceType
     {
         /// <summary>
         /// Network interface type is Unknown.
@@ -204,7 +204,7 @@ namespace Tizen.Network.Stc
     /// <summary>
     /// Enumeration for the Stc error code.
     /// </summary>
-    public enum StcError
+    internal enum StcError
     {
         /// <summary>
         /// Successful.
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
@@ -20,9 +20,9 @@ using Tizen.Internals.Errors;
 namespace Tizen.Network.Stc
 {
     /// <summary>
-    /// Enumeration for Stc time period.
+    /// Internal Enumeration for Stc time period.
     /// </summary>
-    public enum TimePeriodType
+    internal enum NativeTimePeriodType
     {
         /// <summary>
         /// Unknown
@@ -47,14 +47,64 @@ namespace Tizen.Network.Stc
     }
 
     /// <summary>
-    /// Enumeration for network interface type.
+    /// Enumeration for Stc time period.
     /// </summary>
-    public enum Interface
+    public enum TimePeriodType
+    {
+        /// <summary>
+        /// Hour
+        /// </summary>
+        Hour = 3600,
+        /// <summary>
+        /// Day
+        /// </summary>
+        Day = 86400,
+        /// <summary>
+        /// Week
+        /// </summary>
+        Week = 604800,
+        /// <summary>
+        /// Month
+        /// </summary>
+        Month = 2419200
+    }
+
+    /// <summary>
+    /// Internal Enumeration for network interface type.
+    /// </summary>
+    internal enum NativeInterfaceTypeEnum
     {
         /// <summary>
         /// Network interface type is Unknown.
         /// </summary>
         Unknown = 0,
+        /// <summary>
+        /// Network interface type is datacall.
+        /// </summary>
+        Datacall,
+        /// <summary>
+        /// Network interface type is  wifi.
+        /// </summary>
+        Wifi,
+        /// <summary>
+        /// Network interface type is  wired.
+        /// </summary>
+        Wired,
+        /// <summary>
+        /// Network interface type is bluetooth.
+        /// </summary>
+        Bluetooth,
+        /// <summary>
+        /// Network interface type is all.
+        /// </summary>
+        All
+    }
+
+    /// <summary>
+    /// Enumeration for network interface type.
+    /// </summary>
+    public enum InterfaceTypeEnum
+    {
         /// <summary>
         /// Network interface type is datacall. 
         /// </summary>
@@ -78,7 +128,7 @@ namespace Tizen.Network.Stc
     }
 
     /// <summary>
-    /// Enumeration for network roaming type
+    /// Internal Enumeration for network roaming type
     /// </summary>
     internal enum RoamingType
     {
@@ -93,9 +143,9 @@ namespace Tizen.Network.Stc
     }
 
     /// <summary>
-    /// Enumeration for hardware network protocol type. 
+    /// Internal Enumeration for hardware network protocol type.
     /// </summary>
-    public enum NetworkProtocol
+    internal enum NativeNetworkProtocol
     {
         /// <summary>
         /// Network unknown.
@@ -140,7 +190,50 @@ namespace Tizen.Network.Stc
     }
 
     /// <summary>
-    /// Enumeration for monitored process state.
+    /// Enumeration for hardware network protocol type.
+    /// </summary>
+    public enum NetworkProtocol
+    {
+        /// <summary>
+        /// Network no service.
+        /// </summary>
+        NoService,
+        /// <summary>
+        /// Network emergency.
+        /// </summary>
+        Emergency,
+        /// <summary>
+        /// Network search 1900.
+        /// </summary>
+        Search,
+        /// <summary>
+        /// Network 2G.
+        /// </summary>
+        Datacall_2G,
+        /// <summary>
+        /// Network 2.5G.
+        /// </summary>
+        Datacall_2_5G,
+        /// <summary>
+        /// Network EDGE.
+        /// </summary>
+        Datacall_2_5G_Edge,
+        /// <summary>
+        /// Network UMTS.
+        /// </summary>
+        Datacall_3G,
+        /// <summary>
+        /// Network HSDPA.
+        /// </summary>
+        Hsdpa,
+        /// <summary>
+        /// Network LTE.
+        /// </summary>
+        Lte
+    }
+
+    /// <summary>
+    /// Internal Enumeration for monitored process state.
     /// </summary>
     internal enum ProcessState
     {
@@ -156,7 +249,7 @@ namespace Tizen.Network.Stc
         /// Background state.
         /// </summary>
         Background
-   }
+    }
 
     /// <summary>
     /// Enumeration for return type of the callback.
@@ -174,7 +267,7 @@ namespace Tizen.Network.Stc
     }
 
     /// <summary>
-    /// Enumeration for the Stc error code.
+    /// Internal Enumeration for the Stc error code.
     /// </summary>
     internal enum StcError
     {
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
@@ -22,7 +22,7 @@ namespace Tizen.Network.Stc
     /// <summary>
     /// Enumeration for Stc time period.
     /// </summary>
-    public enum StcTimePeriod
+    public enum TimePeriodType
     {
         /// <summary>
         /// Unknown
@@ -49,7 +49,7 @@ namespace Tizen.Network.Stc
     /// <summary>
     /// Enumeration for network interface type.
     /// </summary>
-    public enum StcInterfaceType
+    public enum Interface
     {
         /// <summary>
         /// Network interface type is Unknown.
@@ -80,7 +80,7 @@ namespace Tizen.Network.Stc
     /// <summary>
     /// Enumeration for network roaming type
     /// </summary>
-    public enum StcRoamingType
+    internal enum RoamingType
     {
         /// <summary>
         /// Not in roaming.
@@ -95,7 +95,7 @@ namespace Tizen.Network.Stc
     /// <summary>
     /// Enumeration for hardware network protocol type. 
     /// </summary>
-    public enum StcProtocolType
+    public enum NetworkProtocol
     {
         /// <summary>
         /// Network unknown.
@@ -104,45 +104,45 @@ namespace Tizen.Network.Stc
         /// <summary>
         /// Network no service.
         /// </summary>
-        DatacallNoService,
+        NoService,
         /// <summary>
         /// Network emergency.
         /// </summary>
-        DatacallEmergency,
+        Emergency,
         /// <summary>
         /// Network search 1900.
         /// </summary>
-        DatacallSearch,
+        Search,
         /// <summary>
         /// Network 2G.
         /// </summary>
-        Datacall2G,
+        Datacall_2G,
         /// <summary>
         /// Network 2.5G.
         /// </summary>
-        Datacall2_5G,
+        Datacall_2_5G,
         /// <summary>
         /// Network EDGE.
         /// </summary>
-        Datacall2_5G_Edge,
+        Datacall_2_5G_Edge,
         /// <summary>
         /// Network UMTS.
         /// </summary>
-        Datacall3G,
+        Datacall_3G,
         /// <summary>
         /// Network HSDPA.
         /// </summary>
-        DatacallHsdpa,
+        Hsdpa,
         /// <summary>
         /// Network LTE.
         /// </summary>
-        DatacallLte
+        Lte
     }
 
     /// <summary>
     /// Enumeration for monitored process state.
     /// </summary>
-    public enum StcProcessState
+    internal enum ProcessState
     {
         /// <summary>
         /// Unknown state.
@@ -159,37 +159,9 @@ namespace Tizen.Network.Stc
    }
 
     /// <summary>
-    /// Enumeration for network connection period type.
-    /// @details Last received/sent mean counting data from the first connection of each interface.
-    /// </summary>
-    public enum StcConnPeriodType
-    {
-        /// <summary>
-        /// Unknown period.
-        /// </summary>
-        Unknown,
-        /// <summary>
-        /// Last received data.
-        /// </summary>
-        LastReceivedData,
-        /// <summary>
-        /// Last sent data.
-        /// </summary>
-        LastSentData,
-        /// <summary>
-        /// Total received data.
-        /// </summary>
-        TotalReceivedData,
-        /// <summary>
-        /// Total sent data.
-        /// </summary>
-        TotalSentData
-    }
-
-    /// <summary>
     /// Enumeration for return type of the callback.
     /// </summary>
-    public enum StcCallbackRet
+    public enum CallbackRet
     {
         /// <summary>
         /// Cancel
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
@@ -1,5 +1,5 @@
 ﻿/*
- * Copyright (c) 2018 Samsung Electronics Co., Ltd All Rights Reserved
+ * Copyright (c) 2019 Samsung Electronics Co., Ltd All Rights Reserved
  *
  * Licensed under the Apache License, Version 2.0 (the License);
  * you may not use this file except in compliance with the License.
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
@@ -28,7 +28,7 @@ namespace Tizen.Network.Stc
     public class NetworkStatistics : IDisposable
     {
         private IntPtr _infoHandle = IntPtr.Zero;
-        private bool _disposed = false;
+        private bool _disposed;
 
         internal NetworkStatistics(IntPtr handle)
         {
@@ -54,12 +54,16 @@ namespace Tizen.Network.Stc
         private void Dispose(bool disposing)
         {
             if (_disposed)
+            {
                 return;
+            }
 
             if (disposing)
             {
+                Interop.Stc.Info.StatsDestroy(_infoHandle);
                 _infoHandle = IntPtr.Zero;
             }
+
             _disposed = true;
         }
 
@@ -163,17 +167,24 @@ namespace Tizen.Network.Stc
         /// <privilege>http://tizen.org/privilege/network.get</privilege>
         /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
         /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
-        public Interface InterfaceType
+        public InterfaceTypeEnum InterfaceType
         {
             get
             {
-                Interface iface;
-                int ret = Interop.Stc.Info.GetInterfaceType(_infoHandle, out iface);
+                NativeInterfaceTypeEnum ifaceType;
+                int ret = Interop.Stc.Info.GetInterfaceType(_infoHandle, out ifaceType);
                 if (ret != (int)StcError.None)
                 {
                     Log.Error(Globals.LogTag, "Failed to get Interface type from info, Error - " + (StcError)ret);
                 }
-                return iface;
+                if (ifaceType == NativeInterfaceTypeEnum.Unknown)
+                {
+                    throw new InvalidOperationException("Interface Type is Unknown.");
+                }
+                else
+                {
+                    return (InterfaceTypeEnum)ifaceType;
+                }
             }
         }
 
@@ -243,7 +254,9 @@ namespace Tizen.Network.Stc
                 }
 
                 if(roaming == RoamingType.Disabled)
+                {
                     return false;
+                }
                 return true;
             }
         }
@@ -260,13 +273,20 @@ namespace Tizen.Network.Stc
         {
             get
             {
-                NetworkProtocol protocol;
+                NativeNetworkProtocol protocol;
                 int ret = Interop.Stc.Info.GetProtocol(_infoHandle, out protocol);
                 if (ret != (int)StcError.None)
                 {
                     Log.Error(Globals.LogTag, "Failed to get Protocol type from info, Error - " + (StcError)ret);
                 }
-                return protocol;
+                if (protocol == NativeNetworkProtocol.Unknown)
+                {
+                    throw new InvalidOperationException("Network protocol is Unknown.");
+                }
+                else
+                {
+                    return (NetworkProtocol)protocol;
+                }
             }
         }
 
@@ -291,8 +311,7 @@ namespace Tizen.Network.Stc
 
                 if(state == ProcessState.Unknown)
                 {
-                    Log.Error(Globals.LogTag, "Getting 'Unknown' as Process state from info");
-                    throw new InvalidOperationException("Operation failed");
+                    throw new InvalidOperationException("Process state is Unknown.");
                 }
                 else if(state == ProcessState.Foreground)
                 {
--- a/src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
+++ b/src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
@@ -25,24 +25,24 @@ namespace Tizen.Network.Stc
     /// </summary>
     /// <since_tizen> 6 </since_tizen>
 
-    public class StcInfo : IDisposable
+    public class NetworkStatistics : IDisposable
     {
         private IntPtr _infoHandle = IntPtr.Zero;
         private bool _disposed = false;
 
-        internal StcInfo(IntPtr handle)
+        internal NetworkStatistics(IntPtr handle)
         {
             Log.Debug(Globals.LogTag, "New Info. Handle: " + handle);
             _infoHandle = handle;
         }
 
-        ~StcInfo()
+        ~NetworkStatistics()
         {
             Dispose(false);
         }
 
         /// <summary>
-        /// A method to destroy the managed StcInfo objects.
+        /// A method to destroy the managed NetworkStatistics objects.
         /// </summary>
         /// <since_tizen> 6 </since_tizen>
         public void Dispose()
@@ -163,11 +163,11 @@ namespace Tizen.Network.Stc
         /// <privilege>http://tizen.org/privilege/network.get</privilege>
         /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
         /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
-        public StcInterfaceType InterfaceType
+        public Interface InterfaceType
         {
             get
             {
-                StcInterfaceType iface;
+                Interface iface;
                 int ret = Interop.Stc.Info.GetInterfaceType(_infoHandle, out iface);
                 if (ret != (int)StcError.None)
                 {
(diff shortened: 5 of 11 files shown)


src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
```diff
@@ -0,0 +1,258 @@
...
+{
+    /// <summary>
+    /// Enumeration for Stc time period.
+    /// </summary>
+    public enum StcTimePeriod
+    {
+        /// <summary>
+        /// Unknown
+        /// </summary>
+        Unknown = 0,
```
myroot: Consider removing `Unknown` value. 

Why need `Unknown` value? 
it seem to a error case if get `Unknown` value. 
If it is really error case, you should throw exception instead of `Unknown` value

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// <summary>
+        /// Continue
+        /// </summary>
+        Continue
+    }
+
+    /// <summary>
+    /// Enumeration for the Stc error code.
+    /// </summary>
+    public enum StcError
```
myroot:
```suggestion
    internal enum StcError
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,352 @@
...
+        /// Gets the statistics information an application matched rule asynchronously.
+        /// </summary>
+        /// <since_tizen> tizen_5.5 </since_tizen>
+        /// <param name="infoCb">The callback is called for each application that used network in between timestamps specified.</param>
+        /// <param name="userData">The user data passed to the callback function.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public void GetStats(StatsInfoObjectCb.InfoCallback infoCb, IntPtr userData)
```
myroot: Why don't you provide async method?
```suggestion
        public Task<StcInfo> GetStats()
```

In C# userData never needed.
because C# provide a closure concept

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,352 @@
...
+        /// Gets the statistics information of each application asynchronously. The callback is called for each application that used network in between timestamps specified.
+        /// </summary>
+        /// <since_tizen> tizen_5.5 </since_tizen>
+        /// <param name="infoCb">The callback is called for each application that used network in between timestamps specified.</param>
+        /// <param name="userData">The user data passed to the callback function.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public void ForeachStats(StatsInfoObjectCb.InfoCallback infoCb, IntPtr userData)
```
myroot:
```suggestion
        public IEnumerable<StcInfo> GetStats()
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,238 @@
...
+
+        /// <summary>
+        /// A property to get interface name from statistics information.
+        /// </summary>
+        /// <since_tizen> tizen_5.5 </since_tizen>
+        /// <value>Interface name.</value>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
+        public string IfaceName
```
myroot: X DO NOT use abbreviations or contractions as part of identifier names.
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,352 @@
...
+            _foreachStatsObjectCb = infoCb;
+
+            _foreachStatsCb = (int result, IntPtr info, IntPtr userDataBack) =>
+            {
+                if(result != (int)StcError.None)
+                {
+                    Log.Error(Globals.LogTag, "ForeachStats failed, Error - " + (StcError)result);
+                    StcErrorFactory.ThrowStcException(result);
+                }
+                return _foreachStatsObjectCb(new StcInfo(info), userDataBack);
```
myroot: Please clarify ownership and lifetime of info

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// <summary>
+        /// Get the time interval from statistics information.
+        /// </summary>
+        /// <since_tizen> tizen_5.5 </since_tizen>
+        /// <param name="from">The beginning of the time interval.</param>
+        /// <param name="to">The end of the time interval.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public void GetTimeInterval(out long from, out long to)
```
myroot: Do not use out param

```suggestion
        public DateTime From { get; }
        public DateTime To { get; }
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,238 @@
...
+
+        /// <summary>
+        /// A property to get the interface type from statistics information.
+        /// </summary>
+        /// <since_tizen> tizen_5.5 </since_tizen>
+        /// <value>Interface type.</value>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
+        public StcIfaceType Iface
```
myroot: X DO NOT use abbreviations or contractions as part of identifier names.
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,352 @@
...
+        /// <summary>
+        /// Set the time interval for statistics rule.
+        /// </summary>
+        /// <since_tizen> tizen_5.5 </since_tizen>
+        /// <param name="from">The beginning of the time interval.</param>
+        /// <param name="to">The end of the time interval.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public void SetTimeInterval(long from, long to)
```
myroot:
```suggestion
        public void SetTimeInterval(DateTime from, DateTime to)
```

src/Tizen.Network.Stc/Tizen.Network.Stc/NamespaceDoc.cs
```diff
@@ -0,0 +1,6 @@
+/**
+<summary>
+The Tizen.Network.Stc namespace provides classes to manage and use Smart Traffic Control (STC).
+</summary>
+*/
+namespace Tizen.Network.Stc {}
```
m-macias: Please remove this file, because the namespace documentation is generated using DocFX and there is no need for the existence of such a file.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
```diff
@@ -0,0 +1,258 @@
+﻿/*
+ * Copyright (c) 2016 Samsung Electronics Co., Ltd All Rights Reserved
```
m-macias: Please change the year in all commited files.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,238 @@
...
+
+        ~StcInfo()
+        {
+            Dispose(false);
+        }
+
+        /// <summary>
+        /// A method to destroy the managed StcInfo objects.
+        /// </summary>
+        /// <since_tizen> tizen_5.5 </since_tizen>
```
m-macias: Current Platform version is 5.5 and API version is 6. Please change it to <since_tizen> 6 </since_tizen>

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,238 @@
...
+            {
+                _infoHandle = IntPtr.Zero;
+            }
+            _disposed = true;
+        }
+
+        /// <summary>
+        /// A property to get the application ID from statistics information.
+        /// </summary>
+        /// <since_tizen> tizen_5.5 </since_tizen>
```
m-macias: Current Platform version is 5.5 and API version is 6. Please change it to <since_tizen> 6 </since_tizen>

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,238 @@
...
+                    return "";
+                }
+                return appId;
+            }
+        }
+
+        /// <summary>
+        /// A property to get interface name from statistics information.
+        /// </summary>
+        /// <since_tizen> tizen_5.5 </since_tizen>
```
m-macias: Current Platform version is 5.5 and API version is 6. Please change it to <since_tizen> 6 </since_tizen>

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,238 @@
...
+                    Log.Error(Globals.LogTag, "Failed to get Interface type from info, Error - " + (StcError)ret);
+                }
+                return iface;
+            }
+        }
+
+        /// <summary>
+        /// Get the counters from statistics information.
+        /// </summary>
+        /// <since_tizen> tizen_5.5 </since_tizen>
```
m-macias: Current Platform version is 5.5 and API version is 6. Please change it to <since_tizen> 6 </since_tizen>

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// <param name="incoming">The incoming counter.</param>
+        /// <param name="outgoing">The outgoing counter.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public void GetCounter(out long incoming, out long outgoing)
```
m-macias: CA1021: Avoid out parameters. Either split it to 2 different methods (maybe even properties) or create a struct/class as a return type for this method, which aggregates the result.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,75 @@
...
+            Interop.Stc.Deinitialize(this.handle);
+            this.SetHandle(IntPtr.Zero);
+            return true;
+        }
+    }
+
+    /// <summary>
+    /// A class which is used to manage Smart Traffic control (Stc).<br/>
+    /// </summary>
+    public static class StcManager
```
m-macias: What is the point of a class which doesn't have any visible methods/properties? Maybe make it invisible as well.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,352 @@
...
+        /// <summary>
+        /// Get the time interval for statistics rule.
+        /// </summary>
+        /// <since_tizen> tizen_5.5 </since_tizen>
+        /// <param name="from">The beginning of the time interval.</param>
+        /// <param name="to">The end of the time interval.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public void GetTimeInterval(out long from, out long to)
```
m-macias: CA1021: Avoid out parameters. Either split it to 2 different methods (maybe even properties) or create a struct/class as a return type for this method, which aggregates the result.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: This is as per CAPI Enums.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: Done as per suggestion.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: 1) userData is removed from function parameter as per your suggestion.
2) The callback is called for each application to provide its information. So i can't  find what we have to await for inside this method. As we don't know how many times this callback will be called.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: The callback is called for each application to provide its information. So i can't  find what we have to await for inside this method. As we don't know how many times this callback will be called.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: done as per suggestion.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: I am not sure about it. But i think as 'new' is used, so there is no matter of scope of variable, and user have to call destroy for each 'info' object.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Changed as per your suggestion.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/NamespaceDoc.cs
lkasana: Done

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: done

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done as per your suggestion.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: done

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: done

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: done

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: It is split into two properties.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
lkasana: There is a method:
"public static SafeStcHandle GetStcHandle()".
What should i do?

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: It is split into two properties.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
myroot: It is not CAPI. it can't be reason
if CAPI returing Unknown you should throw exception

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
myroot:
```suggestion
        public IEnumerable<StcInfo> GetStats();
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
myroot: See my suggestion. I recommend to use IEnumerable

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
myroot: `IntPtr info` is ONLY valid in callback scope. 
If you want to extand the lifetime of `info` you need to use some native API to copy

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
myroot: Please study about C# async / await, Task

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: I am uncertain about how to do it. Because,
1) If "Interop.Stc.Rule.GetTimePeriod(_ruleHandle, out timePeriod)" gave **'Unknown'** value for CAPI enum 'stc_time_period_e', then it needs to be handled by corresponding C# enum 'StcTimePeriod'.

2) If somehow 'StcTimePeriod' enum variable does not get initialized or assigned a value from the getter method, then it has a default value of zero. And that's why we need to declare a zero(Unknown) inside this enum to avoid errors.
Reference: "https://blog.submain.com/c-enum-definition-examples-best-practices-and-pitfalls/" says: "Pitfall #2: Assign a value to zero to avoid business logic errors in switch statements."

3) I think that getters don't throw exceptions.
Reference: "https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property"
says: "X AVOID throwing exceptions from property getters."

myroot: >     1. If "Interop.Stc.Rule.GetTimePeriod(_ruleHandle, out timePeriod)" gave **'Unknown'** value for CAPI enum 'stc_time_period_e', then it needs to be handled by corresponding C# enum 'StcTimePeriod'.
First, CAPI will return `stc_time_period_e` `Unknown` only error cases, 
So, You should report error to app developer. 

>     2\. If somehow 'StcTimePeriod' enum variable does not get initialized or assigned a value from the getter method, then it has a default value of zero. And that's why we need to declare a zero(Unknown) inside this enum to avoid errors.
        Reference: "https://blog.submain.com/c-enum-definition-examples-best-practices-and-pitfalls/" says: "Pitfall #2: Assign a value to zero to avoid business logic errors in switch statements."
Yes, You should not assign a zero value(like Unknown) to avoid error case

>    3\. I think that getters don't throw exceptions.
        Reference: "https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property"
        says: "X AVOID throwing exceptions from property getters."

Yes, It is **AVOID** not **DO NOT**
And, you can chage to method not property

myroot: I think, we can ignore a error when only a default value is working well on error case
for example, 
about On/Off enum values, Off value is working well as default value on error case
But your case not working as default value
``` c#
StcRule stcRule1 = ...
StcRule stcRule2 = ...
var period = stcRule1.TimePeriod;
...
stcRule2.TimePeriod = period; // It will cause exception
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,292 @@
...
+
+        /// <summary>
+        /// A property to get the roaming type from statistics information.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <value>Roaming type.</value>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
+        public StcRoamingType Roaming
```
myroot: How aobut replace `StcRomaingType` to bool
```suggestion
        public bool IsRomaing
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// <summary>
+        /// In roaming.
+        /// </summary>
+        Enabled
+    }
+
+    /// <summary>
+    /// Enumeration for hardware network protocol type. 
+    /// </summary>
+    public enum StcProtocolType
```
myroot: How about remove prefix `Datacall` ?

myroot: How about rename to `NetworkProtocal`

1. Stc - no need, because already including `Stc` in namspace
2. Type - no need it, Everything is type
3. Network - need it becuase it is a category of network protocol

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// Background state.
+        /// </summary>
+        Background
+   }
+
+    /// <summary>
+    /// Enumeration for network connection period type.
+    /// @details Last received/sent mean counting data from the first connection of each interface.
+    /// </summary>
+    public enum StcConnPeriodType
```
myroot:
```suggestion
    public enum ConnectionPeriod
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// </summary>
+        LastSentData,
+        /// <summary>
+        /// Total received data.
+        /// </summary>
+        TotalReceivedData,
+        /// <summary>
+        /// Total sent data.
+        /// </summary>
+        TotalSentData
```
myroot: Why these are a period?

LastDataReceivedData, LastSentData, TotalReceivedData, TotalSentData

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
myroot: No used anywhere

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,376 @@
...
+        /// <summary>
+        /// Gets the statistics information an application matched rule asynchronously.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="infoCb">The callback is called for each application that used network in between timestamps specified.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public void GetStats(StatsInfoObjectCb.InfoCallback infoCb)
```
myroot: Please use Task

myroot: This API should move to `StcManager`

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,292 @@
...
+using System.Collections.Generic;
+
+namespace Tizen.Network.Stc
+{
+    /// <summary>
+    /// A class for managing the Stc Info (statistics information).
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+
+    public class StcInfo : IDisposable
```
myroot:
```suggestion
    public class NetworkStatistics : IDisposable
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,292 @@
...
+
+        /// <summary>
+        /// A property to get the process state from statistics information.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <value>Process state.</value>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
+        public StcProcessState State
```
myroot:
```suggestion
        public boo IsForeground
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
m-macias: Currently this method is invisible in the IDE, because you used EditorBrowsableState.Never attribute. In that case the developer will find an empty static class with no usability. If there are no more known dependencies of this class, please make it also invisible (add EditorBrowsableState.Never). Otherwise, just remove this attribute from the method, because I see no reason why we want to hide to it in the first place.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
myroot:
```suggestion
        public bool IsForegroundProcess
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: Shall i do like this in the getter?
->
                 if(timePeriod == TimePeriodType.Unknown)
                 {
                     Log.Error(Globals.LogTag, "Getting 'Unknown' as timePeriod");
                     throw new InvalidOperationException("Operation failed");
                 }
                 else
                     return timePeriod;

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
lkasana: I made the whole class invisible.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: Done.

lkasana: Applied it for every enum.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: removed this enum as it's not used.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: removed this whole enum as not used.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: Currently we use _ruleHandle of the StcRule object for which this API is called. If this API is moved to StcManager, it will be difficult to get _ruleHandle.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
myroot: No, You should revmoe Unknown value in TimePeriodType enum

``` c#
   if (nativePeriod == NativeTimePeriodType.Unknow)
          throw new InvalidOperationException( ... );
   else
          return (TimePeriodType)nativePeriod;
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,373 @@
...
+        /// <summary>
+        /// Gets the statistics information an application matched rule asynchronously.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="infoCb">The callback is called for each application that used network in between timestamps specified.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public void GetStats(StatisticsCallback infoCb)
```
myroot:
```suggestion
        public void Task<NetworkStatistics> GetStatistics()
```

It should be move to StcManager class

``` c#
 class StcManager {
...
  public static Task<NetworkStatistics> GetStatistics(StcRule rule)
...
```
You can access to _ruleHandle in StcManager class
if you declare as `internal`

``` c#
internal IntPtr _ruleHandle = IntPtr.Zero;
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,373 @@
...
+        /// <summary>
+        /// Gets the statistics information of each application asynchronously.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="infoCb">The callback is called for each application that used network in between timestamps specified.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public void ForeachStats(StatisticsCallback infoCb)
```
myroot:
```suggestion
        public void Task<IEnumerable<NetworkStatistics>> GetAllStatistics()
```

It should be move to StcManager class

``` c#
 class StcManager {
...
  public static Task<NetworkStatistics> GetAllStatistics(StcRule rule)
...
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,373 @@
...
+        /// <summary>
+        /// Gets the total statistics information by interface type asynchronously.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="infoCb">The callback is called for each application that used network in between timestamps specified.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public void GetTotalStats(StatisticsCallback infoCb)
```
myroot: It should move to StcManager class

``` c#
 class StcManager {
...
  public static Task<NetworkStatistics> GetAllStatistics()
...
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,373 @@
...
+        public void GetStats(StatisticsCallback infoCb)
+        {
+            if (_disposed)
+            {
+                throw new ObjectDisposedException("Invalid StcRule instance (Object may have been disposed or released)");
+            }
+
+            _getStatsObjectCb = infoCb;
+
+            _getStatsCb = (int result, IntPtr info, IntPtr userData) =>
```
myroot: info handle only valid in callback,
Handle is invalid on outside of callback

So, you need to pack all data into c# layer

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,305 @@
...
+    /// A class for managing the Stc Info (statistics information).
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+
+    public class NetworkStatistics : IDisposable
+    {
+        private IntPtr _infoHandle = IntPtr.Zero;
+        private bool _disposed = false;
+
+        internal NetworkStatistics(IntPtr handle)
```
myroot: handle is invalid on outside of callback

So, you need to pack all data into c# layer

``` c#
string appId;
Interop.Stc.Info.GetAppId(_infoHandle, out appId);
AppId = appId;

string ifaceName;
int ret = Interop.Stc.Info.GetInterfaceName(_infoHandle, out ifaceName);
InterfaceName = ifaceName;
....
```

src/Tizen.Network.Stc/Tizen.Network.Stc.csproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
WonyoungChoi: This csproj is a obsolete style. Please refer the csproj file in other modules.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// <summary>
+        /// Month
+        /// </summary>
+        Month = 2419200
+    }
+
+    /// <summary>
+    /// Enumeration for network interface type.
+    /// </summary>
+    public enum Interface
```
m-macias: I suggest something more verbose like `InterfaceType`, because just `Interface` can be misleading.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: Done for all similar enums.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: This is changed to 'InterfaceTypeEnum'.

src/Tizen.Network.Stc/Tizen.Network.Stc.csproj
lkasana: file changed.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Now clone() CAPI is available, and i am using it this time.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: Now clone() CAPI is available, and i am using it this time.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: moved.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: moved.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: moved.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: done

src/Tizen.Network.Stc/Tizen.Network.Stc.csproj
```diff
@@ -0,0 +1,12 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="..\Tizen\Tizen.csproj" />
```
WonyoungChoi: Use ProjectReference instead of PackageReference

lkasana: done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,324 @@
...
+        private IntPtr _infoHandle = IntPtr.Zero;
+        private bool _disposed;
+
+        internal NetworkStatistics(IntPtr handle)
+        {
+            Log.Debug(Globals.LogTag, "New Info. Handle: " + handle);
+            _infoHandle = handle;
+        }
+
+        ~NetworkStatistics()
```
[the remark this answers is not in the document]
lkasana: Shall i make it as 'internal' or put XML comments?

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,275 @@
...
+            Log.Debug(Globals.LogTag, "New StcRule.");
+            int ret = Interop.Stc.Rule.Create(StcManagerImpl.Instance.GetSafeHandle(), out _ruleHandle);
+            if(ret != (int)StcError.None)
+            {
+                Log.Error(Globals.LogTag, "Failed to create Rule handle, Error - " + (StcError)ret);
+                StcErrorFactory.ThrowStcException(ret);
+            }
+        }
+
+        ~StcRule()
```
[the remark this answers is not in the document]
lkasana: Shall i make it internal or put comments?

lkasana:
Hi @myroot 
Codacy is showing some warnings, which include:
"Remove the exception throwing from this property getter, or refactor the property into a method."

Please suggest, what should i do?

myroot:
> 
> 
> Hi @myroot
> Codacy is showing some warnings, which include:
> "Remove the exception throwing from this property getter, or refactor the property into a method."
> 
> Please suggest, what should i do?

Yes, you should change it

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
```diff
@@ -0,0 +1,323 @@
...
+        /// <summary>
+        /// Background state.
+        /// </summary>
+        Background
+    }
+
+    /// <summary>
+    /// Enumeration for return type of the callback.
+    /// </summary>
+    public enum CallbackRet
```
myroot: Why public?

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,324 @@
...
+        private void Dispose(bool disposing)
+        {
+            if (_disposed)
+            {
+                return;
+            }
+
+            if (disposing)
+            {
+                Interop.Stc.Info.StatsDestroy(_infoHandle);
```
myroot: You should always release a unmanaged object. 
if not, it is a resource leak when disposed by gc

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,324 @@
...
+        public string AppId
+        {
+            get
+            {
+                string appId;
+                int ret = Interop.Stc.Info.GetAppId(_infoHandle, out appId);
+                if (ret != (int)StcError.None)
+                {
+                    Log.Error(Globals.LogTag, "Failed to get AppId from Info, Error - " + (StcError)ret);
+                    return "";
```
myroot:
```suggestion
                    return string.Empty;
```
or do not return in here
appId will empty string, so you can just returing appId

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,324 @@
...
+        public string InterfaceName
+        {
+            get
+            {
+                string ifaceName;
+                int ret = Interop.Stc.Info.GetInterfaceName(_infoHandle, out ifaceName);
+                if (ret != (int)StcError.None)
+                {
+                    Log.Error(Globals.LogTag, "Failed to get interface name from info, Error - " + (StcError)ret);
+                    return "";
```
myroot:
```suggestion
                    return string.Empty;
```
or do not return in here
ifaceName will empty string, so you can just returing ifaceName

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
```diff
@@ -0,0 +1,323 @@
...
+        /// <summary>
+        /// Network interface type is all.
+        /// </summary>
+        All
+    }
+
+    /// <summary>
+    /// Enumeration for network interface type.
+    /// </summary>
+    public enum InterfaceTypeEnum
```
myroot:
```suggestion
    public enum InterfaceType
```
or
```suggestion
    public enum NetworkInterface
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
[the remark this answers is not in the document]
myroot: put a description

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,324 @@
...
+                }
+                else
+                {
+                    return (NetworkProtocol)protocol;
+                }
+            }
+        }
+
+        /// <summary>
+        /// A property to get the process state from statistics information.
```
myroot: Could you explain more details?
I can't understand your description.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <summary>
+        /// Gets the total statistics information by interface type asynchronously.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <returns>The total statistics information of the applications that used network in between timestamps specified.</returns>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public static Task<NetworkStatistics> GetTotalStatisticsAsync(StcRule rule)
```
myroot: Accoding to your description. it only need a interface type for paramter. 
Why do you pass a StcRule intead of InterfaceType?

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
```diff
@@ -0,0 +1,233 @@
...
+                StcErrorFactory.ThrowStcException(ret);
+            }
+            return handle;
+        }
+
+        internal Task<NetworkStatistics> GetStatisticsAsync(StcRule rule)
+        {
+            if (rule._disposed)
+            {
+                throw new ObjectDisposedException("Invalid StcRule instance (Object may have been disposed or released)");
```
myroot: Use ArgumentException

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
```diff
@@ -0,0 +1,233 @@
...
+            }
+
+            TaskCompletionSource<NetworkStatistics> task = new TaskCompletionSource<NetworkStatistics>();
+
+            Interop.Stc.StatsInfoCallback getStatsCb = (int result, IntPtr info, IntPtr userData) =>
```
myroot: You need to protect a delegate object, if not, it released by GC

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
```diff
@@ -0,0 +1,233 @@
...
+            }
+
+            return task.Task;
+        }
+
+        internal Task<IEnumerable<NetworkStatistics>> GetAllStatisticsAsync(StcRule rule)
+        {
+            if (rule._disposed)
+            {
+                throw new ObjectDisposedException("Invalid StcRule instance (Object may have been disposed or released)");
```
myroot: Use ArgumentException

lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
lkasana: Done. I declared a private delegate object in StcManagerImpl and used it later.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
lkasana: done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
lkasana: Changed the description.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: Changed the enum name to 'NetworkInterface'.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done. I have removed the "if" check.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: Done. Thank you.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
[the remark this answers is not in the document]
lkasana: Added description.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <value>Represents the validity of the handle.</value>
+        public override bool IsInvalid
+        {
+            get
+            {
+                return this.handle == IntPtr.Zero;
+            }
+        }
+
+        protected override bool ReleaseHandle()
```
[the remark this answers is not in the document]
lkasana: Added description.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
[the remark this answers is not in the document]
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// <summary>
+        /// Gets the total statistics information by interface type and according to the given rule, asynchronously.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <returns>The total statistics information of the applications that used network in between timestamps specified.</returns>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public static Task<NetworkStatistics> GetTotalStatisticsAsync(StcRule rule)
```
m-macias: Parameters of methods must have valid documentation. Method has 1 parameter, 0 are documented.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// <summary>
+        /// Gets statistics information of each application that used network in between specified timestamps and matches the given rule, asynchronously.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <returns>A list of the NetworkStatistics objects.</returns>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public static Task<IEnumerable<NetworkStatistics>> GetAllStatisticsAsync(StcRule rule)
```
m-macias: Parameters of methods must have valid documentation. Method has 1 parameter, 0 are documented.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// <summary>
+        /// Gets the statistics information of an application, which matched the rule, asynchronously. The rule must contain an Application ID.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <returns>The Statistics information of the application that used network in between timestamps specified.</returns>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public static Task<NetworkStatistics> GetStatisticsAsync(StcRule rule)
```
m-macias: Parameters of methods must have valid documentation. Method has 1 parameter, 0 are documented.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,120 @@
...
+            get
+            {
+                return (this.handle == IntPtr.Zero);
+            }
+        }
+
+        /// <summary>
+        /// Release the Stc handle
+        /// </summary>
+        protected override bool ReleaseHandle()
```
m-macias: Methods which returns non-void value must have exactly one 'returns' section.

src/Tizen.Network.Stc/Interop/Interop.Libraries.cs
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) 2018 Samsung Electronics Co., Ltd All Rights Reserved
```
m-macias: Please update the year in each file to 2019.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,329 @@
...
+        /// A method to destroy the managed NetworkStatistics objects.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        private void Dispose(bool disposing)
```
myroot:
```suggestion
        protected void Dispose(bool disposing)
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,329 @@
...
+            if (_disposed)
+            {
+                return;
+            }
+
+            int ret = Interop.Stc.Info.StatsDestroy(_infoHandle);
+            if (ret != (int)StcError.None)
+            {
+                Log.Error(Globals.LogTag, "Failed to destroy network statistics handle, Error - " + (StcError)ret);
+                StcErrorFactory.ThrowStcException(ret);
```
myroot: do not throw exception on `Dispose` method

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,329 @@
...
+            get
+            {
+                RoamingType roaming;
+                int ret = Interop.Stc.Info.GetRoaming(_infoHandle, out roaming);
+                if (ret != (int)StcError.None)
+                {
+                    Log.Error(Globals.LogTag, "Failed to get Roaming type from info, Error - " + (StcError)ret);
+                }
+
+                if(roaming == RoamingType.Disabled)
```
myroot:
```suggestion
                return roming == RoamingType.Enabled;
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,329 @@
...
+                if (ret != (int)StcError.None)
+                {
+                    Log.Error(Globals.LogTag, "Failed to get IsForegroundProcess from info, Error - " + (StcError)ret);
+                }
+
+                if(state == ProcessState.Unknown)
+                {
+                    throw new InvalidOperationException("Process state is Unknown.");
+                }
+                else if(state == ProcessState.Foreground)
```
myroot:
```suggestion
                return state == ProcessState.Foreground;
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,275 @@
...
+        /// A method to destroy the managed StcRule objects.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        private void Dispose(bool disposing)
```
myroot:
```suggestion
        protected void Dispose(bool disposing)
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,275 @@
...
+            if (_disposed)
+            {
+                return;
+            }
+
+            int ret = Interop.Stc.Rule.Destroy(_ruleHandle);
+            if (ret != (int)StcError.None)
+            {
+                Log.Error(Globals.LogTag, "Failed to destroy Rule handle, Error - " + (StcError)ret);
+                StcErrorFactory.ThrowStcException(ret);
```
myroot: Do not throw exception on Dispose method

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
```diff
@@ -0,0 +1,236 @@
...
+            int ret = Interop.Stc.Initialize(out handle);
+            if (ret != (int)StcError.None)
+            {
+                Log.Error(Globals.LogTag, "Failed to initialize Stc, Error - " + (StcError)ret);
+                StcErrorFactory.ThrowStcException(ret);
+            }
+            return handle;
+        }
+
+        internal Task<NetworkStatistics> GetStatisticsAsync(StcRule rule)
```
myroot: Your implementation was not support multiple request before completion of previous request

``` c#
 var task1 = obj.GetStatisticsAsync(rule1);
 var task2 = obj.GetStatisticsAsync(rule2);
 // task1 is never completed
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
```diff
@@ -0,0 +1,236 @@
...
+        {
+            if (rule._disposed)
+            {
+                throw new ArgumentException("Invalid StcRule object (Object may have been disposed or released).");
+            }
+
+            TaskCompletionSource<NetworkStatistics> task = new TaskCompletionSource<NetworkStatistics>();
+
+            _getStatsCb = (int result, IntPtr info, IntPtr userData) =>
```
myroot: it is overwrited
and captured variable (task) is also changed. so, previous task lost

you need save each task per request
please refer this implementation https://github.com/Samsung/TizenFX/blob/master/src/ElmSharp/ElmSharp/EcoreMainloop.cs

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
lkasana: Done.

src/Tizen.Network.Stc/Interop/Interop.Libraries.cs
lkasana: Done

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Removed the exception.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
myroot: What is difference from `GetAllStatisticsAsync`?

lkasana:
For difference in GetAllStatisticsAsync() and GetTotalStatisticsAsync():

GetAllStatisticsAsync() is to get network statistics of each application, that used network in between specified timestamps. It further uses ForeachAllStats() to iterate over each app's data. 

While GetTotalStatisticsAsync() gives network statistics by interface type.
I talked to Mr. Hyunuk Tak about it previoulsy. He replied:
"
[Lokesh Kasana]
"stc_get_total_stats" also returns multiple stats info handles in multiple callbacks.

[Lokesh Kasana] - 10:41 
so i will not be able to know how many callbacks i will receive.

[탁현욱 / Hyunuk Tak] - 10:45                 
If there are some interfaces, it will be called each, hmm...

[탁현욱 / Hyunuk Tak] - 10:46                 
I think we can get total data usage using proper rule.

[탁현욱 / Hyunuk Tak]
So i think 'stc_get_total_stats' API seems to be overlapping.
"

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// Gets statistics information of each application that used network in between specified timestamps and matches the given rule, asynchronously.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="rule"> The StcRule object.</param>
+        /// <returns>A list of the NetworkStatistics objects.</returns>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public static Task<IEnumerable<NetworkStatistics>> GetAllStatisticsAsync(StcRule rule)
```
myroot:
```suggestion
        public static Task<IEnumerable<NetworkStatistics>> GetAllStatisticsAsync(string appid, DateTime from, DateTime to)
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// Gets the total statistics information by interface type and according to the given rule, asynchronously.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="rule"> The StcRule object.</param>
+        /// <returns>The total statistics information of the applications that used network in between timestamps specified.</returns>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public static Task<NetworkStatistics> GetTotalStatisticsAsync(StcRule rule)
```
myroot:
```suggestion
        public static Task<NetworkStatistics> GetStatisticsAsync(NetworkInterface interface)
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
myroot:
```suggestion
        public static Task<NetworkStatistics> GetStatisticsAsync(string appid, StcRule rule)
```

myroot:
Please, re-design your APIs, consider purpose of API and constraints.
I recommand do not expose `StcRule` class. it should be reconstructed. 
Do not directly mapping core API into C# layer

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
myroot: Please clarify inputs parameters

lkasana:
STC CAPI has this behaviour that user can maintain multiple Rules. Rule has multiple variables within it, which can be used as per future CAPI requirements. By exposing rule class, user gets the option of maintaining several Rule objects and get statistics according to any rule. Recently, there were some CAPI additions and updations, and therefore i was told to keep the design similar to CAPI.

myroot:
> 
> 
> STC CAPI has this behaviour that user can maintain multiple Rules. Rule has multiple variables within it, which can be used as per future CAPI requirements. By exposing rule class, user gets the option of maintaining several Rule objects and get statistics according to any rule. Recently, there were some CAPI additions and updations, and therefore i was told to keep the design similar to CAPI.

Please give a detail specification of Rule. I will check it and guide a API design

lkasana:
For this query:
https://github.com/Samsung/TizenFX/pull/691#discussion_r275578970

GetTotalStatisticsAsync() is now deprecated in core STC package.

myroot:
@lkasana Could close and Re-open for the efficiency?
Currently too many comments make slow down.

PR #705, opened 2019-02-12
[NUI] Added new ViewAdded event on Window

arosis78:
### Description of Change ###
Added new ViewAdded event on Window

### DIFF
--- a/src/Tizen.NUI/src/public/Window.cs
+++ b/src/Tizen.NUI/src/public/Window.cs
@@ -274,6 +274,17 @@ namespace Tizen.NUI
             }
         }
 
+        /// <summary>
+        /// ViewConnected will be triggered when the view connected on Window
+        /// </summary>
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Window.cs
```diff
@@ -274,6 +274,17 @@ internal Window(Rectangle windowPosition, string name, string className) : this(
             }
         }
 
+        /// <summary>
+        /// ViewConnected will be triggered when the view connected on Window
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public event EventHandler ViewConnected;
```
dongsug-song: sorry for late reply. please change this name as "ViewAdded" 
and, please add this attribute => 

arosis78:
Update done. Please check & review patch again.

PR #598, opened 2018-12-05
[Bluetooth] Add Bluetooth Internal APIs

Wootak:
### Description of Change ###

- Add Adapter Enable/Disable APIs (internal)
- Add Adapter Discoverable mode API (internal)
- Add Socket APIs (internal)
- Add Audio AG APIs (internal)
- Add 'Client' property in SocketConnection class
- Deprecated AcceptStateChangedEventArgs.Server and BluetoothAdapter.DestroyServerSocket()

### API Changes ###

 - ACR: TCSACR-207, TCSACR-208

PR #709, opened 2019-02-13
[Bluetooth] Fix the socket callback issue

dh79pyun:
### Description of Change ###
AcceptStateChanged callback is not invoked in application. Because the static callback is not added by default.

### API Changes ###

 - ACR: N/A

### DIFF
--- a/src/Tizen.Network.Bluetooth/Tizen.Network.Bluetooth/BluetoothServerSocket.cs
+++ b/src/Tizen.Network.Bluetooth/Tizen.Network.Bluetooth/BluetoothServerSocket.cs
@@ -36,12 +36,6 @@ namespace Tizen.Network.Bluetooth
         internal int socketFd;
         private bool disposed = false;
 
-        internal BluetoothServerSocket()
-        {
-            StaticAcceptStateChanged += OnAcceptStateChanged;
-            StaticConnectionRequested += OnConnectionRequested;
-        }
-
         private void OnConnectionRequested(Object s, SocketConnectionRequestedEventArgs e)
         {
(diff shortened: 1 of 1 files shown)


src/Tizen.Network.Bluetooth/Tizen.Network.Bluetooth/BluetoothServerSocket.cs
```diff
@@ -132,6 +132,10 @@ private static void UnregisterAcceptStateChangedEvent()
         internal BluetoothServerSocket(int socketFd)
         {
             Log.Info (Globals.LogTag, "Constructing server socket");
+
```
myroot: How about call default constructor ?
``` c#
        internal BluetoothServerSocket(int socketFd) : this()
        {
            this.socketFd = socketFd;
        }
```

or remove unused default constructor

and Constructor should be placed the top of class definition
(next of field declare)

dh79pyun: I think,, this concept was already discussed in the previous pull request - https://github.com/Samsung/TizenFX/pull/598
* My patchset is just fixing the bug in C# BT dll.

@ Wootak
- Is it possible to answer above question?

Wootak: I removed unused default constructor and move this constructor to top of class definition.

dh79pyun:
BUILD is not triggered. So remain comment for triger

PR #710, opened 2019-02-13
[Tizen.Applications.Common] Add a timer for handling GC collection

pjh9216:
- Once apps receive low memory events, GC collection will happen at the
  same time.
- This patch is for distribution of GC collection event.

### DIFF
--- a/src/Tizen.Applications.Common/Tizen.Applications/CoreApplication.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications/CoreApplication.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Globalization;
 using System.Text;
+using System.Timers;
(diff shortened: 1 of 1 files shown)


src/Tizen.Applications.Common/Tizen.Applications/CoreApplication.cs
```diff
@@ -175,6 +178,14 @@ protected virtual void OnAppControlReceived(AppControlReceivedEventArgs e)
         protected virtual void OnLowMemory(LowMemoryEventArgs e)
         {
             LowMemory?.Invoke(this, e);
+            sTimer = new Timer(new Random().Next() % (10 * 1000));
```
myroot:
```suggestion
            sTimer = new Timer(new Random().Next(10 * 1000));
```

PR #711, opened 2019-02-14
[AudioManager] Add HasStreamOnDevice method to AudioStreamPolicy

sangchul1011:

### Description of Change ###
Add **HasStreamOnDevice** method to AudioStreamPolicy class. This functionality has already exist in NativeAPI since Tizen 3.0.
It checks whether if any stream from the current AudioStreamPolicy is using the device.

### API Changes ###
 - ACR: TCSACR-220

### DIFF
--- a/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
+++ b/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
@@ -314,7 +314,7 @@ namespace Tizen.Multimedia
         /// <exception cref="ObjectDisposedException">The <see cref="AudioStreamPolicy"/> has already been disposed of.</exception>
         /// <seealso cref="AudioManager.GetConnectedDevices()"/>
         /// <since_tizen> 6 </since_tizen>
-        public bool IsStreamOnDevice(AudioDevice device)
+        public bool HasStreamOnDevice(AudioDevice device)
         {
             if (device == null)
             {
--- a/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
+++ b/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
@@ -309,6 +309,14 @@ namespace Tizen.Multimedia
         /// </summary>
         /// <returns>true if any audio stream from the current AudioStreamPolicy is using the device; otherwise, false.</returns>
         /// <param name="device">The device to be checked.</param>
+        /// <remarks>
+        /// The AudioStreamPolicy can be applied to each playback or recording stream via other API set.
+        /// (For example., <see cref="T:Tizen.Multimedia.Player"/>, <see cref="T:Tizen.Multimedia.WavPlayer"/>,
+        /// <see cref="T:Tizen.Multimedia.AudioPlayback"/>, <see cref="T:Tizen.Multimedia.AudioCapture"/>, etc.)
+        /// This method returns true only when the device is used for the stream which meets to the two conditions.
+        /// One is that the current AudioStreamPolicy sets a audio route path to the device and the other is that the playback
+        /// or recording stream from other API set should have already started to prepare or to play.(It depends on the API set.)
+        /// </remarks>
         /// <exception cref="ArgumentNullException"><paramref name="device"/> is null.</exception>
         /// <exception cref="InvalidOperationException">An internal error occurs.</exception>
         /// <exception cref="ObjectDisposedException">The <see cref="AudioStreamPolicy"/> has already been disposed of.</exception>
--- a/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
+++ b/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
@@ -305,6 +305,29 @@ namespace Tizen.Multimedia
         }
 
         /// <summary>
+        /// Checks if any stream from the current AudioStreamPolicy is using the device.
+        /// </summary>
+        /// <returns>true if any audio stream from the current AudioStreamPolicy is using the device; otherwise, false.</returns>
+        /// <param name="device">The device to be checked.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="device"/> is null.</exception>
+        /// <exception cref="InvalidOperationException">An internal error occurs.</exception>
+        /// <exception cref="ObjectDisposedException">The <see cref="AudioStreamPolicy"/> has already been disposed of.</exception>
+        /// <seealso cref="AudioManager.GetConnectedDevices()"/>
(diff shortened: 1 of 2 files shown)


src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
```diff
@@ -304,6 +304,29 @@ public void RemoveDeviceForStreamRouting(AudioDevice device)
...
+        /// Checks if any stream from the current AudioStreamPolicy is using the device.
+        /// </summary>
+        /// <returns>true if any audio stream from the current AudioStreamPolicy is using the device; otherwise, false.</returns>
+        /// <param name="device">The device to be checked.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="device"/> is null.</exception>
+        /// <exception cref="InvalidOperationException">An internal error occurs.</exception>
+        /// <exception cref="ObjectDisposedException">The <see cref="AudioStreamPolicy"/> has already been disposed of.</exception>
+        /// <seealso cref="AudioManager.GetConnectedDevices()"/>
+        /// <since_tizen> 6 </since_tizen>
+        public bool IsStreamOnDevice(AudioDevice device)
```
myroot: How about use `Has` prefix?

```suggestion
        public bool HasStreamOnDevice(AudioDevice device)
```

myroot: Q : If I call `AddDeviceForStreamRouting` with `AudioDevice`. 
 and I call this api with `AudioDevice` that added above line
 this API will return true?

sangchul1011: Thanks for your opinion. I also think your suggestion is more easier to understand.
BTW, this AudioStreamPolicy can have not only one stream but also several streams. In this case, what do you recommend to use "HasStream..." vs "HaveStreams..."?

Answer for your Question above:
It depends on the previous situation. It would return true in case of when user already created a stream(playback or recording API) with this AudioStreamPolicy.
Briefly described, this new method will return true only when actual stream has been made with the AudioStreamPolicy handle via Other API(e.g., in AudioPlayback/Player there's ApplyStreamPolicy() method) and the stream is using the device(is routed to the device).

myroot: > Thanks for your opinion. I also think your suggestion is more easier to understand.
> BTW, this AudioStreamPolicy can have not only one stream but also several streams. In this case, what do you recommend to use "HasStream..." vs "HaveStreams..."?

AudioStreamPolicy가 3인칭 단수이므로... Has가 문법에 맞는것 같네요

sangchul1011: Sorry for confusing. I meant that "HasStream.." vs "HasStreams..".

myroot: I think `HasStream..` is better. 
Because has stream is also including has streams, but has streams didn't including has stream

ex)
 Did he has streams?
 No, I he just has a stream.

sangchul1011: Okay, I'll revise it. Thanks.

sangchul1011: I realize there's a slight difference meaning between the two names.
The "IsStreamOnDevice" implies that a stream is now using the device but it does not imply that a stream is just headed to the device.

I think it'd be better to use IsStreamOnDevice or HasStreamOnDevice.
 - Is there any stream on the device?
 - Does it have any stream on the device?

Please let me know your opinion.

myroot: `HasRouteTo(AudioDevice)`
`HasStreamRouteTo(AudioDevice)`

I think, `HasRouteTo` imply that it has a route to device
and, `HasStreamRouteTo` imply that it using the device to route Stream
Because, Stream naming imply `using`, `going` 

If you don't agree to me, 
I prefer `HasStreamOnDevice`

sangchul1011: >I think, `HasRouteTo` imply that it has a route to device
I wonder what the complete sentence of 'HasRouteTo' is.
e.g.) HasStreamOnDevice -> Does it have any stream on the device

> and, `HasStreamRouteTo` imply that it using the device to route Stream
> Because, Stream naming imply `using`, `going`
 I don't actually agree that 'stream' implies 'using' the device now. It is just a kind of resource. (not stream'ing')

If you could share the complete sentence for the abbreviation, we can discuss more about it.

As of now, I think HasStreamOnDevice is the best.
 - Does it have any stream on the device? (purpose of this  API)

myroot: > I wonder what the complete sentence of 'HasRouteTo' is.
Audio Stream Policy has route to speaker
``` c#
 if (streamPolicy.HasRouteTo(speaker)) { ....
```

  player.State is Playing and has route to speaker
``` c#
 if (player.State == PlayerState.Playing && streamPolicy.HasRouteTo(speaker)) { ...
```
is equivalent
``` c#
 if (streamPolicy.HasStreamRotueTo(speaker)) { ...
```
Acutally, I don't think, `Stream is flowing` is not property of AudioStreamPolicy, 
it is a property of player. AudioStreamPolicy can't decide flow of stream

sangchul1011: > Acutally, I don't think, `Stream is flowing` is not property of AudioStreamPolicy,
> it is a property of player. AudioStreamPolicy can't decide flow of stream

AudioManager has device callbacks which has a parameter of AudioDevice. If it is invoked, there was no way to get to know this device has been used for a playback/recording stream made from AudioSteamPolicy handle, except for the management of device id in application side. That's why this API is created with two key factors - audioStreamPolicy, device id.

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
```diff
@@ -304,6 +304,29 @@ public void RemoveDeviceForStreamRouting(AudioDevice device)
                 ThrowIfError("Failed to remove device for stream routing");
         }
 
+        /// <summary>
+        /// Checks if any stream from the current AudioStreamPolicy is using the device.
+        /// </summary>
+        /// <returns>true if any audio stream from the current AudioStreamPolicy is using the device; otherwise, false.</returns>
```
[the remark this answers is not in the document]
sangchul1011: I'll add more explanation with remarks tag soon.

PR #725, opened 2019-02-21
[SystemSetting] Add setter for AccessibilityTtsEnabled

jinwangan:
Add set for AccessibilityTtsEnabled and update test case


jinwangan:
Internal test case result.
> Result : 85 Pass / 12 Not Support / 0 Failed

PR #731, opened 2019-02-27
[NUI] Fix ABI break

dongsug-song:
### Description of Change ###
[NUI] Fix ABI break

### API Changes ###
revert hidden APIs to public ones

### DIFF
--- a/src/Tizen.NUI/src/internal/DaliEnumConstants.cs
+++ b/src/Tizen.NUI/src/internal/DaliEnumConstants.cs
@@ -29,16 +29,16 @@ namespace Tizen.NUI
         /// <since_tizen> 3 </since_tizen>
         public enum TextureType
         {
-                /// <summary>
-                /// One 2D image
-                /// </summary>
-                /// <since_tizen> 3 </since_tizen>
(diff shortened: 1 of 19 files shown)


m-macias:
This change would require an ACR. Please don't merge it, before the ACR reaches the ACR Accept state. Additionally, what is the purpose of region directive in your commit? It neither adds any value to the user nor hides it. It is also misleading, because those methods are NOT deprecated by our standards. If you want to deprecate them, that would require an ACR and the usage of Obsolete attribute.

dongsug-song:
Dear Mr. m-macias
This is to fix ABI break from tizen5.0 (actually those APIs had been changed as hidden ones without ACR in tizen5.5) 
This is revert patch to fix ABI break.
please review again. 
Thank you.

m-macias:
Ok. I checked the history of your changes and it seems that you are removing the EditorBrowsableState attributes which were mostly introduced by #699 and #652 pull requests earlier this year. You can merge it, but be aware that any EditorBrowsable attribute addition or removal is detected as possible API change by our maintenance tools and would require an ACR processing.

PR #747, opened 2019-03-12
[Bluetooth][TCSACR-221] Modify return type to check operation is success or not

Wootak:
Modify return type to check operation is success or not


src/Tizen.Network.Bluetooth/Tizen.Network.Bluetooth/BluetoothGatt.cs
```diff
@@ -387,7 +387,7 @@ private static void UnregisterConnectionStateChangedEvent()
         /// <exception cref="NotSupportedException">Thrown when the BT/BTLE is not supported.</exception>
         /// <exception cref="InvalidOperationException">Thrown when the create GATT client fails.</exception>
         /// <since_tizen> 6 </since_tizen>
-        public Task ConnectAsync(bool autoConnect)
+        public Task<bool> ConnectAsync(bool autoConnect)
```
myroot: I can't understand why need to return bool
It will thrown exception when request was faild

Wootak: You are right. Thank you for your advice.

PR #748, opened 2019-03-13
[MediaPlayer] Modify the range of parameter for buffering API

aferin:
### Description of Change ###
[MediaPlayer] Modify the range of parameter for buffering API

### API Changes ###

 - ACR: http://suprem.sec.samsung.net/jira/browse/TCSACR-231

Changed:(descriptions)
 - PlayerBufferingTime

### DIFF
--- a/src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
@@ -34,10 +34,19 @@ namespace Tizen.Multimedia
         /// Initializes a new instance of the PlayerBufferingTime struct.
         /// </summary>
         /// <param name="preBufferMillisecond">A duration of buffering data that must be prerolled to start playback.</param>
-        /// <param name="reBufferMillisecond">A duration of buffering data that must be prerolled to resume playback
-        /// if player is paused for buffering internally.</param>
+        /// Except 0 and -1, setting at least 1000 milliseconds is recommended to ensure the normal buffering operation.
+        /// 0 : use platform default value which could be different depending on the streaming type and network status. (the initial value)
+        /// -1 : use current value. (since 5.5)
+        /// <param name="reBufferMillisecond">A duration of buffering data that must be prerolled to resume playback,
+        /// when player is internally paused for buffering.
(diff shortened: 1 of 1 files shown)


src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -34,10 +34,19 @@ public struct PlayerBufferingTime
         /// Initializes a new instance of the PlayerBufferingTime struct.
         /// </summary>
         /// <param name="preBufferMillisecond">A duration of buffering data that must be prerolled to start playback.</param>
-        /// <param name="reBufferMillisecond">A duration of buffering data that must be prerolled to resume playback
-        /// if player is paused for buffering internally.</param>
+        /// Except 0 and -1, setting at least 1000 milliseconds is recommended to ensure the normal buffering operation.
+        /// 0 : use platform default value which could be different depending on the streaming type and network status. (the initial value)
```
journey2w: The current sentence is not bad. 
How about the following sentence?
```
// 0 : use platform default value, which depends on the streaming type and network status. It is set as the initial value of this parameter.
```

It depend on you to change the sentence.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -34,10 +34,19 @@ public struct PlayerBufferingTime
...
+        /// -1 : use current value. (since 5.5)
+        /// <param name="reBufferMillisecond">A duration of buffering data that must be prerolled to resume playback,
+        /// when player is internally paused for buffering.
+        /// Except 0 and -1, setting at least 1000 milliseconds is recommended to ensure the normal buffering operation.
+        /// 0 : use platform default value which could be different depending on the streaming type and network status. (the initial value)
```
journey2w: I think from line 45 to line 47 are moved to here because it explains in case of **0** value.
```
        /// <para>0 means platform default value which could be different depending on the streaming type and network status.
        /// If the player state is <see cref="PlayerState.Playing"/> or <see cref="PlayerState.Paused"/>,
        /// this function will return correct time value instead of 0. (since 5.5)</para>
```

PR #751, opened 2019-03-14
Added feature not supported check with API's

abhishekvijay:
- new exception added for case when feature is not supported
- account feature value(true/false) check is added in each API


abhishekvijay:
Abandon pull request, creating new one.

PR #768, opened 2019-03-27
[System.Information] Remove RuntimeInfo key convert logic for TV product

kichankwon:
### Description of Change ###
No more convert runtime-info enum value for TV product


kichankwon:
- We don't merge this patch in Tizen 4.0 because it will be applied on 2019 TV product.
- There is no change about API

myroot:
> * We don't merge this patch in Tizen 4.0 because it will be applied on 2019 TV product.
> * There is no change about API

Could you explain why **the convert logic** needed in tizen4.0 and why now does not need it?
and does it change a **behavior** of API?

kichankwon:
1. History about conversion logic
  - Public CAPI and VD CAPI has different enum values
  - It makes a problem on TV because TizenFX uses Public CAPI internally
  - To map Public CSAPI enum value to VD CAPI on TV, we introduced conversion logic
  - However, VD sets their CAPI enum values same with Public's (since 5.5)
  - Therefore, we have to remove conversion logic
2. There is no change about API

PR #775, opened 2019-04-02
[Location]Modified Stop() for StopBatch() functionality

sovalin:
- Improved calling of StopBatch() instead of Stop() within batch period.

### DIFF
--- a/src/Tizen.Location/Tizen.Location/Locator.cs
+++ b/src/Tizen.Location/Tizen.Location/Locator.cs
@@ -351,10 +351,10 @@ namespace Tizen.Location
         /// <exception cref="NotSupportedException">Thrown when the location is not supported.</exception>
         public void Stop()
         {
-	    Log.Info(Globals.LogTag, "Stopping Location Manager");
+            Log.Info(Globals.LogTag, "Stopping Location Manager");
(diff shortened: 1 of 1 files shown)


src/Tizen.Location/Tizen.Location/Locator.cs
```diff
@@ -352,13 +352,25 @@ public void Start()
         public void Stop()
         {
             Log.Info(Globals.LogTag, "Stopping Location Manager");
-            int ret = Interop.Locator.Stop(_handle);
-            if (((LocationError)ret != LocationError.None))
+            if (_batchPeriod > 0 && _batchPeriod > _batchInterval)
```
myroot: extract method or variable
```suggestion
            bool isBatched = _batchPeriod > 0 && _batchPeriod > _batchInterval;
            if (isBatched) 
```

myroot: You need to mark StartBatch was invoked, 
property check is not enough. 

``` c#
  locator.Start(); // it is just Start()
  locator.BatchPeriod =200;
  locator.BatchInterval = 10;
  locator.Stop(); // it shoud call Stop() but in your implement will call BatchStop()
```

PR #782, opened 2019-04-09
[MediaContent] Add exception and change p/invoke function

hsgwon:
### Description of Change ###

Changed p/invoke function for Delete() method.
Throw InvalidOperationException if file still exists before calling Delete() method.
(PM confirmed for this change)

### API Changes ###
- TCSACR-239

### DIFF
--- a/src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoCommand.cs
+++ b/src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoCommand.cs
@@ -476,8 +476,16 @@ namespace Tizen.Content.MediaContent
         /// <privilege>http://tizen.org/privilege/content.write</privilege>
         /// <param name="mediaId">The media ID to delete.</param>
         /// <returns>true if the matched record was found and deleted, otherwise false.</returns>
-        /// <remarks>The <see cref="MediaDatabase.ScanFile(string)"/> or the <see cref="MediaDatabase.ScanFolderAsync(string)"/> can be used instead.</remarks>
-        /// <exception cref="InvalidOperationException">The <see cref="MediaDatabase"/> is disconnected.</exception>
+        /// <remarks>
+        /// The <see cref="MediaDatabase.ScanFile(string)"/> or the <see cref="MediaDatabase.ScanFolderAsync(string)"/> can be used instead.<br/>
+        /// Since API level 6, If file still exists in file system before calling this method,
+        /// <see cref="InvalidOperationException"/> will be thrown to keep db consistency.
+        /// </remarks>
+        /// <exception cref="InvalidOperationException">
+        ///     The <see cref="MediaDatabase"/> is disconnected.<br/>
+        ///     -or-<br/>
+        ///     File still exists in file system. (Since API level 6)
+        /// </exception>
         /// <exception cref="ObjectDisposedException">The <see cref="MediaDatabase"/> has already been disposed of.</exception>
         /// <exception cref="MediaDatabaseException">An error occurred while executing the command.</exception>
         /// <exception cref="ArgumentNullException"><paramref name="mediaId"/> is null.</exception>
@@ -496,7 +504,26 @@ namespace Tizen.Content.MediaContent
                 return false;
             }
 
-            CommandHelper.Delete(Interop.MediaInfo.Delete, mediaId);
+            Interop.MediaInfo.GetMediaFromDB(mediaId, out var handle).
+                ThrowIfError("Failed to delete MediaInfo.");
+
+            if (handle.IsInvalid)
+            {
+                return false;
+            }
+
+            var path = InteropHelper.GetString(handle, Interop.MediaInfo.GetFilePath);
+
+            // If we don't check file existence before calling `ScanFile` method,
+            // The inconsistency between DB and file system could be occurred.
+            if (File.Exists(path))
+            {
+                throw new InvalidOperationException("File still exists in file system. Remove it first.");
+            }
+
+            // Native 'delete' function was deprecated, so we need to use 'scan file' function instead of it.
+            Database.ScanFile(path);
+
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoCommand.cs
```diff
@@ -476,8 +476,16 @@ private static List<MediaInfo> QueryMedia(SelectArguments arguments)
...
+        /// <remarks>
+        /// The <see cref="MediaDatabase.ScanFile(string)"/> or the <see cref="MediaDatabase.ScanFolderAsync(string)"/> can be used instead.<br/>
+        /// Since API level 6, If file still exists in file system before calling this method,
+        /// <see cref="InvalidOperationException"/> will be thrown to keep db consistency.
+        /// </remarks>
+        /// <exception cref="InvalidOperationException">
+        ///     The <see cref="MediaDatabase"/> is disconnected.<br/>
+        ///     -or-<br/>
+        ///     File still exists in file system. (Since API level 6)
+        /// </exception>
```
myroot: I litte confused, 
This method return bool type, 
and it is returing false when fail to operate.  Which case returing false and which case throwing exception?

hsgwon: I guess  the initial developer of this API intended that the case of no record with given `mediaId` is not an abnormal behavior, so just return false. 
I think it's because the purpose of this method is that delete record with `mediaId` but it's already deleted.
But the other abnormal behavior(exceptions) like `invalid argument`, `already disposed`, `internal error while calling p/invoke`, `database is not connected` must be fixed by user.

myroot: If no data with mediaId, Which error code was returned by Interop.MediaInfo.GetMediaFromDB(mediaId, out var handle)?

``` c
2279	if (sqlite3_step(stmt) == SQLITE_ROW) {
2280		_media = (media_info_s*)calloc(1, sizeof(media_info_s));
2281
2282		if (_media == NULL) {
2283			media_content_error("OUT_OF_MEMORY(0x%08x)", MEDIA_CONTENT_ERROR_OUT_OF_MEMORY);
2284			SQLITE3_FINALIZE(stmt);
2285			return MEDIA_CONTENT_ERROR_OUT_OF_MEMORY;
2286		}
2287
2288		_media_info_item_get_detail(stmt, (media_info_h)_media);
2289
2290		*media = (media_info_h)_media;
2291	} else {
2292		ret = MEDIA_CONTENT_ERROR_INVALID_PARAMETER;
2293	}
```
It seems to retrun a MEDIA_CONTENT_ERROR_INVALID_PARAMETER error

src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoCommand.cs
```diff
@@ -496,7 +504,26 @@ public bool Delete(string mediaId)
                 return false;
             }
 
-            CommandHelper.Delete(Interop.MediaInfo.Delete, mediaId);
+            Interop.MediaInfo.GetMediaFromDB(mediaId, out var handle).
+                ThrowIfError("Failed to delete MediaInfo.");
+
+            if (handle.IsInvalid)
+            {
+                return false;
```
myroot: According to your command, it should throw exception

src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoCommand.cs
```diff
@@ -476,8 +476,16 @@ private static List<MediaInfo> QueryMedia(SelectArguments arguments)
         /// <privilege>http://tizen.org/privilege/content.write</privilege>
         /// <param name="mediaId">The media ID to delete.</param>
         /// <returns>true if the matched record was found and deleted, otherwise false.</returns>
-        /// <remarks>The <see cref="MediaDatabase.ScanFile(string)"/> or the <see cref="MediaDatabase.ScanFolderAsync(string)"/> can be used instead.</remarks>
-        /// <exception cref="InvalidOperationException">The <see cref="MediaDatabase"/> is disconnected.</exception>
+        /// <remarks>
```
myroot: And please add more description about returing false case

src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoCommand.cs
hsgwon: Yes, you're right. I've checked it with native developer and there's no case for null handle without exception.
I'll remove it.

src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoCommand.cs
hsgwon: I'll remove `handle.IsInvalid` case and so it's enough to describe false case with current description.

src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoCommand.cs
```diff
@@ -476,8 +476,16 @@ private static List<MediaInfo> QueryMedia(SelectArguments arguments)
...
+        /// The <see cref="MediaDatabase.ScanFile(string)"/> or the <see cref="MediaDatabase.ScanFolderAsync(string)"/> can be used instead.<br/>
+        /// Since API level 6, If file still exists in file system before calling this method,
+        /// <see cref="InvalidOperationException"/> will be thrown to keep db consistency.
```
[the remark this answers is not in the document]
hsgwon: I uploaded new PR to fix this.
: https://github.com/Samsung/TizenFX/pull/866

src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoCommand.cs
```diff
@@ -476,8 +476,16 @@ private static List<MediaInfo> QueryMedia(SelectArguments arguments)
...
+        /// </remarks>
+        /// <exception cref="InvalidOperationException">
+        ///     The <see cref="MediaDatabase"/> is disconnected.<br/>
+        ///     -or-<br/>
+        ///     File still exists in file system. (Since API level 6)
```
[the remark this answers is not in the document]
hsgwon: I uploaded new PR to fix this.
: https://github.com/Samsung/TizenFX/pull/866

PR #769, opened 2019-03-27
[MediaController] Add new API for Season, Episode, Resolution

hsgwon:
### Description of Change ###
Add new API for Season, Episode, Resolution

### API Changes ###
 - TCSACR-236

PR #783, opened 2019-04-11
[MediaPlayer] add APIs for controlling an audio pitch

aferin:
### Description of Change ###
[MediaPlayer] add APIs for controlling an audio pitch

### API Changes ###
 - ACR: http://suprem.sec.samsung.net/jira/browse/TCSACR-240

Added:
AudidoPitch class
- public bool IsEnabled
- public float GetValue()
- public void SetValue(float value)


src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
```diff
@@ -0,0 +1,125 @@
...
+    /// Provides the ability to control a audio pitch for <see cref="Multimedia.Player"/>.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class AudioPitch
+    {
+        /// <summary>
+        /// Gets the <see cref="Multimedia.Player"/> that owns this instance.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public Player Player { get; }
```
myroot: Is it really need?
usually AudioPitch will referenced by player, so it make a cross-reference

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// Gets the audio pitch value.
+        /// </summary>
+        /// <returns>The pitch value of an audio stream. The default value is 1.</returns>
+        /// <remarks>This function is used for audio content only.</remarks>
+        /// <exception cref="InvalidOperationException">The player is not in the valid state.</exception>
+        /// <exception cref="ObjectDisposedException">The player has already been disposed of.</exception>
+        /// <seealso cref="IsEnabled"/>
+        /// <seealso cref="SetValue(float)"/>
+        /// <since_tizen> 6 </since_tizen>
+        public float GetValue()
```
myroot: Why do you use `Value` naming?
Why dou you use method instead of property?

How about use property namded by Pitch
```suggestion
        public float Pitch { get { ... } set {...} }
```

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Diagnostics;
+using static Interop;
+
+namespace Tizen.Multimedia
+{
+    /// <summary>
+    /// Provides the ability to control a audio pitch for <see cref="Multimedia.Player"/>.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class AudioPitch
```
myroot: I recommend to use struct
and C# layer is working only data container

Described in code
``` c#
public struct AudioPitch {
    public bool IsEnabled {get;set;}
    public float Pitch {get;set;}
}

public class Player {
      public AudioPitch AudioPitch {get;set;}
}

player.AudioPitch = new AudioPitch {
    IsEnabled = true,
    Pitch = 3.3
};
```

hsgwon: Is there any difference between AudioPitch and SeriesInformation?
SeriesInformation was just data container for two value together.
(SeriesInformation : https://github.com/Samsung/TizenFX/pull/769 )

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
```diff
@@ -0,0 +1,125 @@
...
+        public void SetValue(float value)
+        {
+            Player.ValidateNotDisposed();
+
+            if (Player.AudioPitch.IsEnabled == false)
+                throw new InvalidOperationException("A audio pitch is not enabled.");
+
+            if (value < 0.5F || 2.0F < value)
+            {
+                throw new ArgumentOutOfRangeException(nameof(value), value, "Valid level is 1.0 to 10.0");
```
myroot: We can just ignore out of range values or clamp to the range

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
aferin: it is being used for checking the state of a player. i'll change it to private

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
aferin: ok i'll change it to property.
but i think naming still looks weird. because pitch will be written twice.

Player.AudioPitch.Pitch = 1.1;

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
aferin: hmm.. i don't get it. do you mean getting and throwing error from native api?
(there is a mistake in log..)

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
myroot: SeriesInformation is immutable object, but current AudioPitch is not

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
myroot: Ok, `Value` naming is reasonable for it

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
myroot:
``` c#
   if (value < 0.5F)
       value = 0.5F;
   if (value > 2.0F)
       value = 2.0F;
```

aferin: the error is returned in native API. and we guarantee the same operation with native one. so i think it would be better to return the same error.

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
hsgwon: I'm confused a little bit.
`Immutable object` is the condition to being `Struct`, Isn't it?

X AVOID defining a struct **unless** the type has all of the following characteristics:
```
- It logically represents a single value, similar to primitive types (int, double, etc.).
- It has an instance size under 16 bytes.
- **It is immutable.**
- It will not have to be boxed frequently.
```

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
myroot: We do not want same behavior with native, because we want improve usability of C# API more than native API

and if we clamp the value into the range, native API dose not return error

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
myroot: So, My means, AudioPitch should be immutable

myroot: In my opinion, AudioPitch is considered an immutable object.
because it is a entity class(data class) 
It is just set of data to apply Player object
and it has not operator to change internal state. 
So it should be a immutable object.

myroot: AudioPitch object can be assignable to all others player object
because it is just a data that contained a pitch value

myroot: If you do not think so,
This is no different than simply attaching it directly to the player.

like below 
``` c#
class Player { ...
 public float AudioPitch {get;set;}
 public bool AudioPitchEnabled {get;set;}
...
}
```

hsgwon: `SeriesInformation` is same condition you said the point of view( data container and immutable). 
I just wanna check it again whether `SeriesInformation` should be `Class` or `Struct`.
How about it? I wanna know exactly the criteria for future work.

myroot: Acutally, SeriesInformation also better to use struct
(SeriesInformation is a data container and immutable)

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
aferin: then.. we should have made all APIs like that when c# api released..
every APIs return Argument exception currently in player.
if i add like that, we would have two policies for a range without any obvious reasons.

myroot: I think it is not a big issue, 
Usually only method throw Argument exception
we also guide a avoid a exception on property

So, app developer did't confused about it.

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
hsgwon: Actually, `SeriesInformation` was `struct` originally but you said it should be `class` because usually use `class`. (https://github.com/Samsung/TizenFX/pull/769#discussion_r270718868)

So now, I see the criteria for choosing between `class` and `struct` clearly. 
Thank you.

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
hsgwon: I just wonder what is the better choice. 
User set the value as -1 and return without any exception. In this case, I'm worry about that user could be confused as it's working well. 
I agree with your opinion that user should check description before use method.
But AudioPitch allow just 0.5~2.0 value range, and I think out of range value is abnormal case and it's better to throw the exception for it.

It is method and it can throw any exception if needed.

sangchul1011: BTW, does the Tizen C# have a rule over MS's? 
Here's dotnet guideline. (https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property)
  ✓ DO preserve the previous value if a property setter throws an exception.
  X AVOID throwing exceptions from property getters.

If the exception is permitted in case of setting the property, user also can notified what they are trying to set the invalid value.
FYI, setting volume level property of AudioManager does not clamp the value rather it throws error also.

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            Debug.Assert(player != null);
+            Player = player;
+        }
+
+        /// <summary>
+        /// Gets or sets the audio pitch.
+        /// </summary>
+        /// <remarks>This function is used for audio content only. The default value is false.</remarks>
+        /// <remarks>To set, the player must be in the <see cref="PlayerState.Idle"/> state.</remarks>
```
m-macias: Please write it as a single <remarks> tag.

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
aferin: IsEnabled must be set before player is prepared. each property has a different precondition. nevertheless, should i change class to struct?

myroot: How about it?
``` c#
class Player { ...
 public float AudioPitch {get;set;}
 public bool AudioPitchEnabled {get;set;}
...
}
```
If it has a different precondition, do not put it in the same value object.

myroot: I'm worried about current implementation in below usage.
``` c#
AudioPitch pitch = player.AudioPitch;
.....
pitch.Value = 1.0;     // player internal state was changed
pitch.IsEnabled = false; // player internal state was changed
```
I think that control of player should be made directly to player object.
However, current implementation can change the properties of player objects through in different objects like the above example.
AudioPitch object did not represent a player object.
It make hard to debug when complexed issue was occurred

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
myroot: Ok, I understand that better throw exception in this case.
but the range was already checked by native layer and the range could be changed by platform.
so, I think that no need to check range in c# layer

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
aferin: hmm. ok i'll check if there is no expandability for AudioPotch more. if so, i'll put these properties into player class.

src/Tizen.Multimedia.MediaPlayer/Player/AudioPitch.cs
hsgwon: I think C# API can check the range to avoid useless P/Invoke call, if the range will not be changed any more.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -560,6 +560,81 @@ public bool ReplayGain
...
+
+                return value;
+            }
+
+            set
+            {
+                ValidateNotDisposed();
+
+                if (AudioPitchEnabled == false)
+                    throw new InvalidOperationException("An audio pitch is not enabled.");
```
myroot: Is it need?
Usually developer pitch value change before enabling pitch feature
``` c#
player.AudioPitch = 1.1;
player.AudioPitchEnabled = true;
```

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -560,6 +560,81 @@ public bool ReplayGain
...
+        /// Gets or sets the audio pitch.
+        /// </summary>
+        /// <value>The value indicating whether or not AudioPitch is enabled. The default is false.</value>
+        /// <remarks>This function is used for audio content only.</remarks>
+        /// <remarks>To set, the player must be in the <see cref="PlayerState.Idle"/> state.</remarks>
+        /// <exception cref="InvalidOperationException">The player is not in the valid state.</exception>
+        /// <exception cref="ObjectDisposedException">The player has already been disposed of.</exception>
+        /// <seealso cref="AudioPitch"/>
+        /// <since_tizen> 6 </since_tizen>
+        public bool AudioPitchEnabled
```
hsgwon: I just wonder `AudioPitchEnabled` can be set by user or internal restriction for audio only content?
If it can be set and controlled by user, why is it not just included in `AudioPitch` setter?

aferin: * @remarks Enabling pitch control could increase the CPU usage on some devices.
i forgot to add this description. i'll update it on the next patch

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
aferin: AudioPItchEnabled has to be decided before the plugin is made.
so it would be called once when user create a player. on the other hand value can be changed continuously if AudioPitch is enabled.

hsgwon: As described in `AudioPitchEnabled`, it have to set before prepare. It's because it have to be set before creating pipeline(Prepare state) in native.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
hsgwon: As discussed in offline, `AudioPitchEnable` have state restriction unlike `AudioPitch`, 
so it would be better that separate it.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
myroot: So, I don't understand, why need a check state of AudioPitchEnabled value in here?
User want to change AudioPitch value before enabling `AudioPitchEnabled`

hsgwon: As I said before, enabling audio pitch value should be done before Prepare state.
(Before creating gstreamer pipeline in NATIVE framework)
So `AudioPitchEnabled` should be set first before set `AudioPitch`.

With your example above, native framework will return invalid operation error.

hsgwon: Actually, if we don't check the value of `AudioPitchEnabled` in here, we get same error because native framework will return same error. But it have useless P/Invoke call.

myroot: It is bad to API usability.
if possible, I want to fix native api behavior
Anyway, if impossible to change , I approve it

hsgwon: I agree with your opinion in perspective of API usability, but it cannot be changed.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -560,6 +560,82 @@ public bool ReplayGain
             }
         }
 
+        /// <summary>
+        /// Gets or sets the audio pitch.
+        /// </summary>
+        /// <value>The value indicating whether or not AudioPitch is enabled. The default is false.</value>
+        /// <remarks>This function is used for audio content only.</remarks>
+        /// <remarks>To set, the player must be in the <see cref="PlayerState.Idle"/> state.</remarks>
```
hsgwon: Please write it as a single `<remarks>` tag.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -560,6 +560,82 @@ public bool ReplayGain
...
+                    ThrowIfFailed(this, "Failed to enable the audio pitch of the player");
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the audio pitch value.
+        /// </summary>
+        /// <value>The audio stream pitch value. The default is 1.</value>
+        /// <remarks>Enabling pitch control could increase the CPU usage on some devices.</remarks>
+        /// <remarks>This function is used for audio content only.</remarks>
```
hsgwon: Please write it as a single `<remarks>` tag.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
hsgwon: Please write it as a single `<remarks>` tag.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
hsgwon: Please write it as a single `<remarks>` tag.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -560,6 +560,82 @@ public bool ReplayGain
...
+        /// <seealso cref="AudioPitchEnabled"/>
+        /// <since_tizen> 6 </since_tizen>
+        public float AudioPitch
+        {
+            get
+            {
+                ValidateNotDisposed();
+
+                if (AudioPitchEnabled == false)
+                    throw new InvalidOperationException("An audio pitch is not enabled.");
```
hsgwon: Please add brace like below.
```C#
if (AudioPitchEnabled == false)
{
    throw new InvalidOperationException("An audio pitch is not enabled.");
}
```

hsgwon: Please refer below C# coding rule, M4.3.1.
http://suprem.sec.samsung.net/confluence/display/SE/4.3+Statements

PR #787, opened 2019-04-12
[Bluetooth][Non-ACR] Fix no data exception issue

Wootak:
If no data from scan result, return null instead of exception.

### DIFF
--- a/src/Tizen.Network.Bluetooth/Tizen.Network.Bluetooth/BluetoothLeAdapterImpl.cs
+++ b/src/Tizen.Network.Bluetooth/Tizen.Network.Bluetooth/BluetoothLeAdapterImpl.cs
@@ -201,11 +201,6 @@ namespace Tizen.Network.Bluetooth
             return ret;
         }
 
-        private string GetPacketTypeToString(BluetoothLePacketType packetType)
-        {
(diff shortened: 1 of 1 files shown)


src/Tizen.Network.Bluetooth/Tizen.Network.Bluetooth/BluetoothLeAdapterImpl.cs
```diff
@@ -201,29 +201,41 @@ internal int StopScan()
...
-            IList<string> list = new List<string>();
             IntPtr uuidListArray = IntPtr.Zero;
             int count = -1;
             BluetoothLeScanDataStruct scanDataStruct = BluetoothUtils.ConvertLeScanDataToStruct(scanData);
 
             int ret = Interop.Bluetooth.GetScanResultServiceUuid(ref scanDataStruct, packetType, ref uuidListArray, ref count);
-            if (ret != (int)BluetoothError.None)
+            if (ret == (int)BluetoothError.NoData)
+            {
+                Log.Info(Globals.LogTag, "No ServiceUuids in " + GetPacketTypeToString(packetType));
```
myroot: Why do you made a convert method?
It already convert to string like "BluetoothLeAdvertisingPacket" without your convert method

Wootak: I remove the convert method. Thank you.

PR #791, opened 2019-04-15
[Messaging] Added New API for messages_get_message_count

chauhan321:
Signed-off-by: kamaljeet <kamal.jc@samsung.com>

### Description of Change ###

I have added a new C# API  (GetMessageCount()) to call the CAPI (messages_get_message_count). 

<!--
Added:
 - public int GetMessageCount(int mbox, int messageType);

### DIFF
--- a/src/Tizen.Messaging/Tizen.Messaging.Messages/Message.cs
+++ b/src/Tizen.Messaging/Tizen.Messaging.Messages/Message.cs
@@ -131,6 +131,21 @@ namespace Tizen.Messaging.Messages
             }
         }
 
+        public int GetMessageCount(int mbox, int messageType)
+        {
+            int count;
+            Log.Error(Globals.LogTag, "GetMessageCount is Called");
+            int ret = Interop.Messages.GetMessageCount(_messageHandle, mbox, messageType, out count);
+            if (ret != (int)MessagesError.None)
+            {
+                Log.Error(Globals.LogTag, "Failed to get Message count, Error - " + (MessagesError)ret);
+                MessagesErrorFactory.ThrowMessagesException(ret, _messageHandle);
+                return 1;
+            }
+            Log.Error(Globals.LogTag, "GetMessageCount ends");
+            return 0;
+        }
+
         private void GetAllAddresses()
         {
             int count;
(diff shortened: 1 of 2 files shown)


src/Tizen.Messaging/Tizen.Messaging.Messages/Message.cs
```diff
@@ -131,6 +131,21 @@ private void AddAddress(MessagesAddress address)
             }
         }
 
+        public int GetMessageCount(int mbox, int messageType)
```
m-macias: All public members should have documentation.

src/Tizen.Messaging/Tizen.Messaging.Messages/Message.cs
```diff
@@ -131,6 +131,21 @@ private void AddAddress(MessagesAddress address)
...
+        {
+            int count;
+            Log.Error(Globals.LogTag, "GetMessageCount is Called");
```
m-macias: In my opinion, Log.Error should only be used if an exception appears, not as a standard logging behaviour.

src/Tizen.Messaging/Tizen.Messaging.Messages/Message.cs
```diff
@@ -131,6 +131,21 @@ private void AddAddress(MessagesAddress address)
...
+            int ret = Interop.Messages.GetMessageCount(_messageHandle, mbox, messageType, out count);
+            if (ret != (int)MessagesError.None)
+            {
+                Log.Error(Globals.LogTag, "Failed to get Message count, Error - " + (MessagesError)ret);
+                MessagesErrorFactory.ThrowMessagesException(ret, _messageHandle);
+                return 1;
```
m-macias: In C# we prefer an exception handling over error codes, which means that functions should focus on their purpose rather than returning success/failure value. The purpose of this method is to "get count", which you are currently NOT returning anyway (because count is assigned locally). Please rewrite this method to return count value, but don't use 'out' parameter as in native implementation.

src/Tizen.Messaging/Tizen.Messaging.Messages/Message.cs
```diff
@@ -131,6 +131,21 @@ private void AddAddress(MessagesAddress address)
             }
         }
 
+        public int GetMessageCount(int mbox, int messageType)
+        {
+            int count;
+            Log.Error(Globals.LogTag, "GetMessageCount is Called");
+            int ret = Interop.Messages.GetMessageCount(_messageHandle, mbox, messageType, out count);
```
myroot: `Interop.Messages.GetMessageCount` use with message service handle but, you passed a message handle

PLEASE, Test your code before submit.

myroot:
If do not give any feedback, I will close it on today

PR #794, opened 2019-04-16
[MediaPlayer] Modify the pre-condition of streaming function

aferin:
### Description of Change ###
[MediaPlayer] Modify the pre-condition of streaming function

Extend pre-condition include READY state so that the application could get buffering progress more earlier just right after calling Prepare.

### API Changes ###

 - ACR: http://suprem.sec.samsung.net/jira/browse/TCSACR-242

Changed :
Added PlayerState.Ready State
public DownloadProgress GetDownloadProgress()


aferin:
> Don't we need to do ACR process? It seems that behavior are changed.
> If you plan to do ACR process, please add `ACR required` label also.
i've added the label. i am writing ACR.

Issue #65, opened 2018-01-22
[STT] Fail to add/remove RecognitionResult EventHandler 


myroot:
You always set native event handler by SttSetRecognitionResultCB
and always unset natvie event handler by SttUnsetRecognitionResultCB

But, Event can be added more than 1. 
So, it is fail to below sequence
```c#
 var client = SttClient();
 client.RecognitionResult += EventHandler1;
 client.RecognitionResult += EventHandler2;
 // now EventHandler1 and EventHandler2 was working

 ....
 client.RecognitionResult -= EventHandler1;
 // now any event handler was not working
```
https://github.com/Samsung/TizenFX/blob/3566375f1189cdcd03f869fae78eabd77bbe3a3b/src/Tizen.Uix.Stt/Tizen.Uix.Stt/SttClient.cs#L370-L414

PR #800, opened 2019-04-18
[STT/TTS] Fix desctructor to call dispose (#596)

urmain:
### Description of Change ###
[STT/TTS] Fix desctructor to call dispose


myroot:
@urmain If you are owner of STT, please fix https://github.com/Samsung/TizenFX/issues/65 on Issue, it is a reported a year ago

PR #802, opened 2019-04-22
[STC] Add C# API wrapper for Stc CAPIs

lkasana:
Description of Change
Add C# API wrapper for Smart Traffic Control (Stc) CAPIs

API Changes
    ACR: TCSACR-217 (http://suprem.sec.samsung.net/jira/browse/TCSACR-217)

Older Pull Request(closed because of slow loading due to too many comments):
https://github.com/Samsung/TizenFX/pull/691


src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// Gets statistics information of each application that used network in between specified timestamps and matches the given rule, asynchronously.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="rule"> The StcRule object.</param>
+        /// <returns>A list of the NetworkStatistics objects.</returns>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public static Task<IEnumerable<NetworkStatistics>> GetAllStatisticsAsync(StcRule rule)
```
myroot: Which values are used to filter in StcRule?
Is the appid specified in the StcRule ignored?

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,288 @@
...
+        internal bool _disposed;
+
+        /// <summary>
+        /// Creates a StcRule object.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public StcRule()
```
myroot: How about make as pure C# object?
and consider using nullable type

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,288 @@
...
+        /// <summary>
+        /// Set the time interval for statistics rule.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="from">The beginning of the time interval.</param>
+        /// <param name="to">The end of the time interval.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        public void SetTimeInterval(DateTime from, DateTime to)
```
myroot: Why need it? you also provide `From` `To` property

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,288 @@
...
+
+        /// <summary>
+        /// To get Interface type for statistics rule.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <returns>Interface type.</returns>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
+        public NetworkInterface GetInterfaceType()
```
myroot: Why not property?

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,288 @@
...
+
+        /// <summary>
+        /// To set Interface type for statistics rule.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="interfaceType"> Interface type.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
+        public void SetInterfaceType(NetworkInterface interfaceType)
```
myroot: Which value should I set to select any interface type?

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
lkasana: appid is used in "GetStatisticsAsync", while GetAllStatisticsAsync does not include appid, as it gives stats for every app.

myroot: So, You need to remove appid property in StcRule data.
This API design is easy to make mistakes and easy to misreading API usage

lkasana: typedef struct {
    char app_id[STC_APP_ID_LEN];
    stc_time_interval_s interval;
    stc_iface_type_e iface_type;
    stc_time_period_e time_period;
 } stc_stats_rule_s;

To get statistics, filtered on the basis of different parameters(combinations), we set corresponding elements of StcRule and use appropriate API to get Statistics. This is how CAPI is designed, and also CAPI recently had changes related to these.
GetStatisticsAsync() filters on the basis of appId(and time interval) provided by given StcRule. So we needed AppId setter and getter in StcRule.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: (capi)SetTimeInterval(_ruleHandle, from, to): It needs both "from and to" parameters. So one approach i find suitable is to use a method as setter. The "From and To" properties are getters only.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: There is an 'Unknown' enum in NetworkInterface. And you told me to throw an exception in case of Unknown. Now property getters don't throw exceptions usually, so i switched to method.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: 'Wifi' can be selected as interface type.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
myroot: On `GetStatisticsAsync` method, AppId can be passed as a parameter like below
``` c#
 public static Task<NetworkStatistics> GetStatisticsAsync(string appId, StcRule rule)
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
myroot: Why do you provide a getter of `From`?
It is a value passed from user

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,288 @@
...
+
+        /// <summary>
+        /// To set Time period for statistics rule.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="timePeriod">Time period.</param>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
+        public void SetTimePeriod(TimePeriodType timePeriod)
```
myroot:
```suggestion
        public void SetTimePeriodType(TimePeriodType timePeriod)
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,288 @@
...
+            int ret = Interop.Stc.Rule.SetInterfaceType(_ruleHandle, (NativeNetworkInterface)interfaceType);
+            if (ret != (int)StcError.None)
+            {
+                Log.Error(Globals.LogTag, "Failed to set Interface type, Error - " + (StcError)ret);
+                StcErrorFactory.ThrowStcException(ret);
+            }
+        }
+
+        /// <summary>
+        /// To get Time period for statistics rule.
```
myroot: It is too confuse between `Interval` and `period`

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,288 @@
...
+            }
+            if (timePeriod == NativeTimePeriodType.Unknown)
+            {
+                throw new InvalidOperationException("Time period is Unknown.");
+            }
+            return (TimePeriodType)timePeriod;
+        }
+
+        /// <summary>
+        /// To set Time period for statistics rule.
```
myroot: It is too confuse between `Interval` and `period`
and there is no proper description about period

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
myroot: Why `Unknown` value are returing?
If a default value is existed as NetworkInterface.All, We can avoid error case

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: Yes, 'From' is set by user. But Core package provide getter API too, and hence i also replicated it. I was told to replicate the functionalities provided by Core STC package. As Core package recently had changes and may have more changes in future, so i thought such APIs may get used in next changes.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
myroot: This value is setted by User
So, it can't existed a error case

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
```diff
@@ -0,0 +1,323 @@
...
+        /// <summary>
+        /// Network interface type is  wired.
+        /// </summary>
+        Wired,
+        /// <summary>
+        /// Network interface type is bluetooth.
+        /// </summary>
+        Bluetooth,
+        /// <summary>
+        /// Network interface type is all.
```
myroot: Description is wrong

Type for any network interface

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,288 @@
...
+                if (ret != (int)StcError.None)
+                {
+                    Log.Error(Globals.LogTag, "Failed to set AppId, Error - " + (StcError)ret);
+                    StcErrorFactory.ThrowStcException(ret);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Set the time interval for statistics rule.
```
myroot: `Interval` seem be little ambiguus, IMO, Range is more proper for it

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
myroot: PLEASE, forget Core API, It is not Core API. We can't accept low usability API.
Please, thinks about API usability

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
lkasana: There is no need to send 'appId' separately, It could be set in the above rule parameter only.

myroot: Why do you think so?
I recommend to remvoe appid property from StcRule class. because it has no means some context(GetAllStatisticsAsync).
if appid is need in GetStatisticsAsync, we can pass as separatly parameter

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,313 @@
...
+        /// A method to destroy the managed NetworkStatistics objects.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected void Dispose(bool disposing)
```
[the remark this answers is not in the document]
WonyoungChoi: Please add XML comment for this protected member method. 
If you want to inherit the comment of super class, you can use the inheritdoc tag instead.
```
/// <inheritdoc /> 
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,313 @@
...
+            {
+                return;
+            }
+
+            if(disposing)
+            {
+                // destroy managed resources
+            }
+
+            int ret = Interop.Stc.Info.StatsDestroy(_infoHandle);
```
WonyoungChoi: Destroying a handle received from outside of this class is likely to cause problem. 
Please consider to use SafeHandle instead of disposing the _infoHandle here.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        /// <summary>
+        /// A property to get the application ID from statistics information.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <value>Application ID.</value>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
```
WonyoungChoi: Wrong comment. this property only provides getter. So when try set any value to this property. It makes a complication error.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        /// <summary>
+        /// A property to get interface name from statistics information.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <value>Interface name.</value>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
```
WonyoungChoi: Wrong comment. this property only provides getter. So when try set any value to this property. It makes a complication error.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        /// <summary>
+        /// A property to get "from" value(start) of time interval from statistics information.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <value>from(start) of time interval.</value>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
```
WonyoungChoi: Wrong comment. this property only provides getter. So when try set any value to this property. It makes a complication error.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        /// <summary>
+        /// A property to get "to" value(end) of time interval from statistics information.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <value>to(end) of time interval.</value>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown while setting this property when Stc is not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown while setting this value due to an invalid operation.</exception>
```
WonyoungChoi: Wrong comment. this property only provides getter. So when try set any value to this property. It makes a complication error.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
myroot: Why not `NetworkInterface.All` ?

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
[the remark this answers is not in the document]
lkasana: Description added.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,288 @@
...
+        /// A method to destroy the managed StcRule objects.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected void Dispose(bool disposing)
```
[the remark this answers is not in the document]
lkasana: Description added.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: When we create rule object, we use Core API "stc_stats_rule_create", it sets interfaceType to unknown('0') by default. And depending on the filtering options, the user might set it or let it remain to be unknown. So it is possible that C# layer may get "unknown" from CAPI. I think that we should not even throw an exception in such case, because it is perfectly valid to get Unknown as NetworkInterface of some Rule.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: Sorry, I misunderstood earlier. Actually I tried setting NetworkInterface.All (as the interface type of Rule) in Core APIs' testapp, but then it does not return any data. When I set the NetworkInterface to 'Wifi' or 'Unknown(Default value)', it gives some statistics of apps accordingly.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcEnumerations.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done for both: Rule handle and Statistics handle.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Corrected description of such properties. Thank you.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,283 @@
...
+            {
+                return;
+            }
+
+            if(disposing)
+            {
+                // destroy managed resources
+            }
+
+            if(_infoHandle != null && !_infoHandle.IsInvalid) {
```
myroot: _infoHandle is a managed resource, so it shoud be disposed only `disposing was true`

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,283 @@
...
+        protected void Dispose(bool disposing)
+        {
+            if (_disposed)
+            {
+                return;
+            }
+
+            if(disposing)
+            {
+                // destroy managed resources
```
myroot:
```suggestion
                _infoHandle?.Dispose();
                _infoHandle = null;
```
_infoHandle = null;

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
myroot: You can use nullable type to present `unknown` as `default`. So, I recommend to use pure c# code making a StcRule class. 

>How about make as pure C# object?
and consider using nullable type

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/nullable-types/using-nullable-types

lkasana:
Add support to handle multiple requests for GetStatisticsAsync and GetAllStatisticsAsync.

The implementation is referenced from WiFi C# package and Microsoft documentation.
Callback delegates are now stored in map.
'Task' is not needed to be stored according to:
'https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/lambda-expressions#variable-scope-in-lambda-expressions'

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done for both the handles(rule too). Thank you.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: Done. Patch pushed. Please review.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: Description is added

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,251 @@
...
+using System.Threading.Tasks;
+using System.Collections.Generic;
+
+namespace Tizen.Network.Stc
+{
+    /// <summary>
+    /// A class for managing the Stc Rules to match applications.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class StcRule : IDisposable
```
myroot: Make as a pure c# object

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
lkasana: I have made changes as per suggestion.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
myroot:
``` c#
    public class StcRule
    {
        public DateTime? From { get; set; }
        public DateTime? To { get; set; }

        public NetworkInterface? InterfaceType { get; set; }
        public TimePeriodType? TimePeriod { get; set; }
    }

```

myroot: IntPtr ConvertToNativeRule(StcRule2 rule)
        {
            Interop.Stc.SafeRuleHandle handle;
            Interop.Stc.Rule.Create(StcManagerImpl.Instance.GetSafeHandle(), out handle);
            if (rule.From.HasValue && rule.To.HasValue)
            {
                Interop.Stc.Rule.SetTimeInterval(_ruleHandle, rule.From.Value, rule.To.Value);
            }
            if (rule.InterfaceType.HasValue)
            {
                Interop.Stc.Rule.SetInterfaceType(_ruleHandle, (NativeNetworkInterface)rule.InterfaceType.Value);
            }
            if (rule.TimePeriod.HasValue)
            {
                Interop.Stc.Rule.SetTimePeriod(_ruleHandle, (NativeTimePeriodType)rule.TimePeriod.Value);
            }
            return handle;
        }

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
```diff
@@ -0,0 +1,218 @@
...
+
+                        statsList.Add(new NetworkStatistics(cloned));
+                        return CallbackRet.Continue;
+                    };
+
+                    int retTemp = Interop.Stc.ForeachAllStats(infoList, foreachAllStatsCb, IntPtr.Zero);
+                    if(retTemp != (int)StcError.None)
+                    {
+                        Log.Error(Globals.LogTag, "foreachAllStatus() failed , Error - " + (StcError)retTemp);
+                        StcErrorFactory.ThrowStcException(retTemp);
```
myroot: you should not throw exception in here
use task.SetException

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,281 @@
...
+        {
+            NativeNetworkInterface ifaceType;
+            int ret = Interop.Stc.Info.GetInterfaceType(_infoHandle, out ifaceType);
+            if (ret != (int)StcError.None)
+            {
+                Log.Error(Globals.LogTag, "Failed to get Interface type from info, Error - " + (StcError)ret);
+            }
+            if (ifaceType == NativeNetworkInterface.Unknown)
+            {
+                throw new InvalidOperationException("Interface Type is Unknown.");
```
myroot: Is it possible scenario that throw exception?
Could you explain which case could be occurred error?

Acutally, I want to change it to getter only property without excpetion

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,281 @@
...
+            }
+        }
+
+        /// <summary>
+        /// To get the network protocol type from statistics information.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <returns>NetworkProtocol type.</returns>
+        /// <exception cref="InvalidOperationException">Thrown when method fails due to an invalid operation/value.</exception>
+        public NetworkProtocol GetProtocol()
```
myroot: Is it possible scenario that throw exception?
Could you explain which case could be occurred error?

Acutally, I want to change it to getter only property without excpetion

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,281 @@
...
+            return (NetworkProtocol)protocol;
+        }
+
+        /// <summary>
+        /// To get if the app is running as a foreground process, from it's statistics information.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <returns>True, if process state is Foreground, otherwise False.</returns>
+        /// <exception cref="InvalidOperationException">Thrown when method fails due to an invalid operation/value.</exception>
+        public bool IsForegroundProcess()
```
myroot: Is it possible scenario that throw exception?
Could you explain which case could be occurred error?

Acutally, I want to change it to getter only property without excpetion

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
```diff
@@ -0,0 +1,206 @@
...
+                            Log.Error(Globals.LogTag, "StatsClone() failed , Error - " + (StcError)retValue);
+                            task.SetException(new InvalidOperationException("Error occurs during StatsClone(), " + (StcError)retValue));
+                        }
+
+                        statsList.Add(new NetworkStatistics(cloned));
+                        return CallbackRet.Continue;
+                    };
+
+                    int retTemp = Interop.Stc.ForeachAllStats(infoList, foreachAllStatsCb, IntPtr.Zero);
+                    if(retTemp != (int)StcError.None)
```
myroot:
``` c#
                    if(retTemp != (int)StcError.None)
                    {
                        Log.Error(Globals.LogTag, "foreachAllStatus() failed , Error - " + (StcError)retTemp);
                        task.SetException(new InvalidOperationException("Error occurs during foreachAllStatus(), " + (StcError)retTemp));
                    }
                    else
                    {
                        task.SetResult(statsList);
                    }
```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
```diff
@@ -0,0 +1,206 @@
...
+                    lock (_getStatsCb_map)
+                    {
+                        _getStatsCb_map.Remove(key);
+                    }
+
+                    return CallbackRet.Continue;
+                };
+            }
+
+            int ret = Interop.Stc.GetStats(GetSafeHandle(), rule.ConvertToNativeRule(appId), _getStatsCb_map[id], id);
```
myroot: I recommend to release SafeRuleHandle after use

``` c#
using (var ruleHandle = rule.ConvertToNativeRule(appId)) 
{
     int ret = Interop.Stc.GetStats(GetSafeHandle(), ruleHandle , _getStatsCb_map[id], id);
    ....
}

```

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,281 @@
...
+        /// <summary>
+        /// To get the interface type from statistics information.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <returns>Interface type.</returns>
+        /// <exception cref="InvalidOperationException">Thrown when method fails due to an invalid operation/value.</exception>
+        public NetworkInterface GetInterfaceType()
+        {
+            NativeNetworkInterface ifaceType;
+            int ret = Interop.Stc.Info.GetInterfaceType(_infoHandle, out ifaceType);
```
myroot: It never returing error if _inforHandle is not null. 
So you no need to check error code

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Threading.Tasks;
+using System.Collections.Generic;
+
+namespace Tizen.Network.Stc
+{
+    /// <summary>
+    /// A class for managing the Stc Rules to match applications.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class StcRule
```
myroot: Let's think about naming of `StcRule`
Why it is StcRule?
It is a class for represent a filter to use on retrieval of Statistics data. 
So, I think, StatisticsFilter is better name.
a case of PackageManager
https://github.com/Samsung/TizenFX/blob/f7c1f3b65cd7a6f8a5f86899750f5d49c7ccda64/src/Tizen.Applications.PackageManager/Tizen.Applications/PackageFilter.cs#L25

a case of ApplicationInfo
https://github.com/Samsung/TizenFX/blob/f7c1f3b65cd7a6f8a5f86899750f5d49c7ccda64/src/Tizen.Applications.Common/Tizen.Applications/ApplicationInfoFilter.cs#L26

Calendar
https://github.com/Samsung/TizenFX/blob/27893e385679cbc0a0421355866043e415e55977/src/Tizen.Pims.Calendar/Tizen.Pims.Calendar/CalendarFilter.cs#L26

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="appId"> Application's appId for which statistics information is required.</param>
+        /// <param name="rule"> The StcRule object.</param>
+        /// <returns>The Statistics information of the application that used network in between timestamps specified.</returns>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        /// <exception cref="ArgumentException">Thrown when the method is provided with invalid argument.</exception>
+        public static Task<NetworkStatistics> GetStatisticsAsync(string appId, StcRule rule)
```
myroot: I have a question, 
How does TimePeriod of StcRule affect the return value?
Is it valid for this API?

myroot:
If no more proceeding, I will close it, because it opened a month ago

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: As suggested in the discussion: Exception throw is removed, Unknown is passed to app as 'Others' and made it a Property.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: As suggested in the discussion: Exception throw is removed, Unknown is passed to app as 'Others' and made it a Property.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: As suggested in the discussion: Exception throw is removed, Unknown is passed to app as 'Others' and made it a Property.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
lkasana: Done as suggested.
Earlier I thought SafeHandle will take care of it, and dispose the ruleHandle when reference count is zero.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
lkasana: TimeInterval(From and To) indicates the Range of time for which we require statistical data. Whereas, 'TimePeriod' is to granulate that data in further pieces. That is the statistics data will be chunked in periods of Month or Week or Hour.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: Changed as per suggestion.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: This method is already changed to property, with no throw exceptions.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManagerImpl.cs
lkasana: Done.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcRule.cs
lkasana: Changes were done. Thank you.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
```diff
@@ -0,0 +1,271 @@
...
+                return protocol;
+            }
+        }
+
+        /// <summary>
+        /// A property to get the process state of application from statistics information.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <value>Monitored process state.</value>
+        public ProcessStateType ProcessState()
```
myroot: How about use property, and How about rename to `ApplicationState`
because, "Background", "Foreground" is a state of Application, of course process can has state like "Background", "Forground" with cgroup, but it seem not.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
myroot: According to your comment, The statistical data are split to some pieces.
So, the return value should be represented as collection of Network Statistics
Isn't it?

src/Tizen.Network.Stc/Tizen.Network.Stc/StcInfo.cs
lkasana: Done. I missed to make it a property.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="appId"> Application's appId for which statistics information is required.</param>
+        /// <param name="filter"> The StatisticsFilter object.</param>
+        /// <returns>The Statistics information of the application that used network in between timestamps specified.</returns>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        /// <exception cref="ArgumentException">Thrown when the method is provided with invalid argument.</exception>
+        public static Task<NetworkStatistics> GetStatisticsAsync(string appId, StatisticsFilter filter)
```
myroot: According to your comment, The statistical data are split to some pieces.
So, the return value should be represented as collection of Network Statistics
Isn't it?

lkasana: I shall clear it once more from code flow and from the core developer.

lkasana: Yes, you are right. We will get many callbacks for 'GetStatisticsAsync'. Inside callback we can continue to store the statistics in list, but we don't know when to set taskResult. That is we don't know when will we get the last callback.
This similar case was faced in 'stc_foreach_stats'(deprecated). Now this is replaced by 'stc_foreach_stats'(async) and 'stc_foreach_stats'(sync).

lkasana: According to Mr. Tak, we should only use 'GetAllStatisticsAsync' for every purpose. Because this API is newly designed to solve the issue of 'last callback reached or not'.
Related ACR: http://suprem.sec.samsung.net/jira/browse/ACR-1358

Now we can provide AppId in StatisticsFilter according to need.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="filter"> The StatisticsFilter object.</param>
+        /// <returns>A list of the NetworkStatistics objects.</returns>
+        /// <privilege>http://tizen.org/privilege/network.get</privilege>
+        /// <exception cref="NotSupportedException">Thrown when the Stc is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when the permission is denied.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when the method failed due to an invalid operation.</exception>
+        /// <exception cref="ArgumentException">Thrown when the method is provided with invalid argument.</exception>
+        public static Task<IEnumerable<NetworkStatistics>> GetAllStatisticsAsync(StatisticsFilter filter)
```
myroot: Now, we can remove `All` keyword

src/Tizen.Network.Stc/Tizen.Network.Stc/StcStatsFilter.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <since_tizen> 6 </since_tizen>
+        /// <value>Time period.</value>
+        public TimePeriodType? TimePeriod { get; set; }
+
+        internal Interop.Stc.SafeFilterHandle ConvertToNativeFilter()
+        {
+            Interop.Stc.SafeFilterHandle handle;
+            Interop.Stc.Filter.Create(StcManagerImpl.Instance.GetSafeHandle(), out handle);
+
+            Interop.Stc.Filter.SetAppId(handle, AppId);
```
myroot: you should set when appid is not null

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
myroot: AppId null is means `Any'?

lkasana: Renamed 'GetAllStatisticsAsync' to 'GetStatisticsAsync'.
Yes, Appid null means 'Any app'.

lkasana: As per today's discussion with Mr. Tak, we switched from 'GetStatisticsAsync' to 'GetAllStatisticsAsync'. But now from the code flow, it is found that 'GetAllStatisticsAsync' does not consider app_id in its internal logic. It does pass app_id to stc-manager. We will shortly fix this issue in Core Stc package.
That is, either
1) 'GetStatisticsAsync' should tell the end of callbacks.
2) or, 'GetAllStatisticsAsync' should consider app_id in its logic.

src/Tizen.Network.Stc/Tizen.Network.Stc/StcStatsFilter.cs
lkasana: Actually in Core package, app_id is not initialized in rule_create, app_id is declared as fixed size array. It is initialized in the setter method. So i have to set app_id in any case.

myroot:
It's too premature to expose the C# API. because Native API has many logical issue.
So. I close it
I can't add stc API in TizenFX

src/Tizen.Network.Stc/Tizen.Network.Stc/StcManager.cs
myroot: IMO, 1 is correct way. 
2 is not correct because it includes `All` in method name

PR #817, opened 2019-05-02
[WebView] Copy the managed string into unmanaged memory for eina_hash…

younghajung:
…_add

Signed-off-by: yh106.jung <yh106.jung@samsung.com>

### Description of Change ###

Copy the managed string into unmanaged memory for eina_hash_add

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.WebView/Interop/Interop.Eina.cs
+++ b/src/Tizen.WebView/Interop/Interop.Eina.cs
@@ -22,11 +22,11 @@ internal static partial class Interop
     internal static partial class Eina
     {
         [DllImport(Libraries.Eina)]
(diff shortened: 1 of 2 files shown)


myroot:
Why do you change it? Do you have any issue on Interop with string?

younghajung:
> Why do you change it? Do you have any issue on Interop with string?

Yes, when we called ewk_view_url_request_set, wrong string value had been passed.

myroot:
> 
> 
> > Why do you change it? Do you have any issue on Interop with string?
> 
> Yes, when we called ewk_view_url_request_set, wrong string value had been passed.

I got it. `eina_hash_add` does not hold a string object. it just store a `pointer`

PR #821, opened 2019-05-07
[NUI] Add GraphicsTypeManager for Density Independent Pixel

dongsug-song:
### Description of Change ###
[NUI] Add GraphicsTypeManager for Density Independent Pixel
[NUI] Fix KeyboardFocusManager custom algorithm reset

### API Changes ###
N/A


dongsug-song:
Hello, Mr. Fang
I added xaml Transition and XamlResource in View and Layer not to use ContentPage.
do you think there will be no performance issue for this adding?
could you check this?
thank you.

PR #826, opened 2019-05-09
[MediaPlayer] Change documentation for streaming APIs

aferin:
### Description of Change ###
following the comments, I changed the documentation for streaming APIs

https://github.com/Samsung/TizenFX/pull/748

### DIFF
--- a/src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
@@ -40,11 +40,11 @@ namespace Tizen.Multimedia
         /// <param name="reBufferMillisecond">A duration of buffering data that must be prerolled to resume playback,
         /// when player is internally paused for buffering.
         /// Except 0 and -1, setting at least 1000 milliseconds is recommended to ensure the normal buffering operation.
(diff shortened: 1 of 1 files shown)


src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -40,11 +40,11 @@ public struct PlayerBufferingTime
         /// <param name="reBufferMillisecond">A duration of buffering data that must be prerolled to resume playback,
         /// when player is internally paused for buffering.
         /// Except 0 and -1, setting at least 1000 milliseconds is recommended to ensure the normal buffering operation.
-        /// 0 : use platform default value which could be different depending on the streaming type and network status. (the initial value)
-        /// -1 : use current value. (since 5.5)
+        /// 0 : use platform default value, which depends on the streaming type and network status. It is set as the initial value of this parameter.
         /// <para>0 means platform default value which could be different depending on the streaming type and network status.
```
hsgwon: It seems it's duplicated above statement. Is this really needed that described here again?

aferin: hmm you are right. i think it is better to delete the below sentence.

PR #848, opened 2019-05-22
Split Tizen.NUI

AdunFang:
### API Changes ###

 - ACR:


WonyoungChoi:
Why so many API are changed to public from internal? These APIs should be kept as "internal". You can use InternalsVisibleTo attribute to refer the internal APIs each other. 
https://docs.microsoft.com/en-us/dotnet/standard/assembly/friend-assemblies?tabs=csharp

WonyoungChoi:
I merge this PR to split-nui branch first because this PR is too big to handle in Pull Request. 
Please make a PR to split-nui branch if you have more changes. 
After works in branch, I will merge it to master later.

PR #869, opened 2019-05-30
[Build] Suppress CS0618 warnings due to internal reference

WonyoungChoi:
### Description of Change ###
Suppress CS0618(Type or member is obsolete) warnings due to internal reference

### API Changes ###

 - ACR:


WonyoungChoi:
It is not necessary to mark obsolete attribute to the non-static member because the class is already obsoleted.

PR #870, opened 2019-05-30
[MediaPlayer] Change descriptions of audio pitch APIs

aferin:
### Description of Change ###
Changed descriptions of audio pitch APIs according to comments.
http://suprem.sec.samsung.net/jira/browse/TCSACR-240

### DIFF
--- a/src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
@@ -560,7 +560,7 @@ namespace Tizen.Multimedia
         }
 
         /// <summary>
-        /// Gets or sets the audio pitch.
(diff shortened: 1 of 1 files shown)


src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -560,7 +560,7 @@ public bool ReplayGain
         }
 
         /// <summary>
-        /// Gets or sets the audio pitch.
+        /// Enables or disables controlling the pitch of audio or gets the status of controlling the pitch of audio.
```
hsgwon: How about split this to two statement.
```
Enables ~~ of audio.
Gets ~~~ of audio.
```

aferin: ok i've patched it

PR #871, opened 2019-05-31
[NUI] Add Renderer Ranges

huiyueun:


src/Tizen.NUI/src/public/Renderer.cs
```diff
@@ -621,6 +621,38 @@ protected override void Dispose(DisposeTypes type)
             base.Dispose(type);
         }
 
+
+        /// <since_tizen> 5.5 </since_tizen>
```
WonyoungChoi: It should be <since_tizen> 6 </since_tizen>

PR #894, opened 2019-06-20
[NUI] Add inherit binding context support

elishateng:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -4030,6 +4030,7 @@ namespace Tizen.NUI.BaseComponents
                     };
                     ChildAdded(this, e);
                 }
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -4030,6 +4030,7 @@ public override void Add(View child)
                     };
                     ChildAdded(this, e);
                 }
+                BindableObject.SetInheritedBindingContext(child, this?.BindingContext);
```
dongsug-song: please add this "BindableObject.SetInheritedBindingContext(child, this?.BindingContext);" into Layer. 
=> Layer.Add(). I tested and it is needed.

elishateng: Got it.

PR #897, opened 2019-06-21
Add Tizen.Environment.SetEnvironmentVariable wrapper to setup environment variables

gbalykov:
### Description of Change ###

Dotnet-launcher sets app specific environment variable after `coreclr_initialize`. On older versions of CoreCLR, `SetEnvironmentVariableW` function was used for this, but on the newest master it is removed from unix exports. 

Now `Tizen.Environment.SetEnvironmentVariable` could be used instead, when delegate is created for this method. Delegate can't be created for `System.Environment.SetEnvironmentVariable` directly, because it is overloaded. 

Related issue: https://github.com/dotnet/coreclr/issues/23683

### API Changes ###

`Tizen.Environment.dll` is added. It contains `Tizen.Environment.SetEnvironmentVariable`.

cc @alpencolt

### DIFF
--- a/src/Tizen.Environment/Tizen/Environment.cs
+++ b/src/Tizen.Environment/Tizen/Environment.cs
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2019 Samsung Electronics Co., Ltd All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+
+namespace Tizen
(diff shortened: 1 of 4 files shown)


myroot:
TizenFX is not good lib to add this API.
This API is not need Application side, it only used loader.

src/Tizen.Environment/Tizen/Environment.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+
+namespace Tizen
+{
+    /// <summary>
+    /// The Environment class provides the method to set environment variables through delegate.
+    /// Delegate for System.Environment.SetEnvironmentVariable can't be created because it is overloaded, so
+    /// Tizen.Environment.SetEnvironmentVariable wrapper is added.
+    /// </summary>
+    /// <since_tizen> 5.5 </since_tizen>
```
m-macias: Current Platform version is 5.5, but current API version is 6 and that value should be used in the `<since_tizen>` tag. Please change it to `<since_tizen>6</since_tizen>`.

src/Tizen.Environment/Tizen/Environment.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class Environment
+    {
+        /// <summary>
+        /// Sets environment variable
+        /// </summary>
+        /// <returns></returns>
```
m-macias: Please remove the `<returns>` tag, if this method doesn't return anything.

PR #899, opened 2019-06-24
[NUI] Fix memory leak

AdunFang:
### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -1294,7 +1294,20 @@ public class View : Container, IResourcesProvider
...
 
-        internal readonly MergedStyle _mergedStyle;
+        private MergedStyle mergedStyle = null;
+        internal MergedStyle _mergedStyle
+        {
+            get
+            {
+                if (null == mergedStyle)
+                {
+                    mergedStyle = new MergedStyle(GetType(), this);
```
dongsug-song: is it become free and disposed automatically? Or is it required explicitly disposed in Dispose() ?  I think this would be OK for nui c# application, but could still have leak error in nui xaml application. could you make this clear?

PR #900, opened 2019-06-24
[NUI] Change GetDefaultWindow() to static func

tscholb:
### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/NUIApplication.cs
```diff
@@ -206,7 +206,7 @@ public Window Window
         /// <returns>The default Window.</returns>
         /// <since_tizen> 6 </since_tizen>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public Window GetDefaultWindow()
+        public static Window GetDefaultWindow()
```
dongsug-song: please add this comment =>         // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)

PR #903, opened 2019-06-26
[NUI] Adjust directory

AdunFang:
### Description of Change ###
- xaml_c visiting classes must be public => change these classes to public
- remove the unused code of Tizen.NUI.Xaml to reduce the size of Tizen.NUI.dll

### API Changes ###
None


dongsug-song:
Hello, Mr. Fang
This patch has too many changes. could you explain about the changes briefly?
Thank you.

AdunFang:
Hi, Mr. Song

Because xaml_c will inject the xaml to IL,

that means it will visitor the class which used in xaml,

so that these classes must be public.

This patch change these classes to public,
and remove the unused code of Tizen.NUI.Xaml to reduce the size of Tizen.NUI.dll.

PR #910, opened 2019-07-02
[NFC] Fix logical error for get NDEF message

jh8801jung:
Signed-off-by: Jung Jihoon <jh8801.jung@samsung.com>

### Description of Change ###
Fix for logical error when store & get the NDEF message.
(http://suprem.sec.samsung.net/jira/browse/TNEXT-17352?filter=-1)
(It is not needed ACR. It just bug fix)


jh8801jung:
When I checked the logic using SafeHandle, I found critical logic error when free memory of  Native NdefMessage / NdefRecord. (Double free will be occur.)
We will postpone this pull request code until we have a way in Native.

PR #917, opened 2019-07-02
[NUI] Fix ImageView bug

huiyueun:
Fix VD pin popup issue

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
@@ -41,6 +41,7 @@ namespace Tizen.NUI.BaseComponents
                 string resource = Tizen.Applications.Application.Current.DirectoryInfo.Resource;
                 url = url.Replace("*Resource*", resource);
             }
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -903,22 +908,28 @@ private void UpdateImageMap(PropertyMap fromMap)
 			
             SetProperty(ImageView.Property.IMAGE, new PropertyValue(imageMap));
         }
-
+		
         private void UpdateImage(int key, PropertyValue value)
         {
             PropertyMap temp = new PropertyMap();
 
+            if (_resourceUrl == "")
```
dongsug-song: could you check the difference of "" and String.Empty ?

huiyueun: The result of using String.Empty and "" is the same.
However, I found that String.Empty is more efficient in memory.
https://blog.naver.com/sinjoker/221223223812

I will modify "" to use String.Empty.

PR #934, opened 2019-07-17
Addition of Sensor Recorder Class

pvschakradhar:
Addition of sensor recorder class of the below mentioned ACR.

http://suprem.sec.samsung.net/jira/browse/TCSACR-218


src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+        }
+
+        /// <summary>
+        /// Queries the recorded data asynchronously.
+        /// </summary>
+        /// <param name="type">
+        /// Sensor type.
+        /// </param>
+        /// <since_tizen> 6 </since_tizen>
+        public async Task ReadAsync(int type)
```
myroot:
```suggestion
        public async Task<int> ReadAsync(int type)
```

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+                _sensorRecorderDataCallBack = (int _type, int data, int remains, int _error, Int64 _userData) =>
+                {
+                    RecorderDataEventArgs e = new RecorderDataEventArgs(_type, data, remains, _error, _userData);
+
+                    if (_error != (int)SensorError.None)
+                    {
+                        Log.Error(Globals.LogTag, "Error in Sensor Recorder Data aynchronously CallBack");
+                    }
+
+                    DataReceived?.Invoke(null, e);
```
myroot: Do not use Event to dispatch a result of async method

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+        }
+
+        /// <summary>
+        /// Dispose API for destroying the SensorProvider handle.
+        /// </summary>
+        /// <since_tizen> 3 </since_tizen>
+        /// <param name="disposing">The boolean value for destoying SensorProvider handle.</param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
```
myroot: You should only release managed resource when disposing was true. 
You should alway release unmanaged resource regardless of disposing value

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+        /// Gets the start and the end time of the time period of a given record data.
+        /// </summary>
+        /// <param name="data">
+        /// Record data handle.
+        /// </param>
+        /// <returns>
+        /// returns the start and the end time of the time period of a given record data.
+        /// </returns>
+        /// <since_tizen> 6 </since_tizen>
+        public Tuple<long, long> DataGetTime(int data)
```
myroot: Do not use Tuple, Can't know means of each values
Use struct

myroot: Why data handle is int type?

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Tizen.Sensor
+{
+    /// <summary>
+    /// The SensorRecorder is used for retrieving the recorded data from a particular sensor.
```
myroot: How to specify a particular sensor?

I can't know how to use this class

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+        /// Sets a recording option parameter.
+        /// </summary>
+        /// <param name="recorderOption">
+        /// Option parameter.
+        /// </param>
+        /// <param name="value">
+        /// Value.
+        /// </param>
+        /// <since_tizen> 6 </since_tizen>
+        public void OptionSetInt(int recorderOption, int value)
```
myroot: Do not use int for option
I can't know which options are usable.

This API is too low level to use

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+        ///  Sets an integer-type query parameter.
+        /// </summary>
+        /// <param name="recorderQuery">
+        /// Query parameter.
+        /// </param>
+        /// <param name="value">
+        /// Value.
+        /// </param>
+        /// <since_tizen> 6 </since_tizen>
+        public void QuerySetInt(int recorderQuery, int value)
```
myroot: What is recorderQuery and why is it a int type?

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+        }
+
+        /// <summary>
+        ///  Queries the recorded data synchronously.
+        /// </summary>
+        /// <param name="type">
+        /// Sensor type.
+        /// </param>
+        /// <since_tizen> 6 </since_tizen>
+        public void Read(int type)
```
myroot: Why no result? Usually read operation expect a result of read

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+                _sensorRecorderDataCallBack = (int _type, int data, int remains, int _error, Int64 _userData) =>
+                {
+                    RecorderDataEventArgs e = new RecorderDataEventArgs(_type, data, remains, _error, _userData);
+
+                    if (_error != (int)SensorError.None)
+                    {
+                        Log.Error(Globals.LogTag, "Error in Sensor Recorder Read synchronously CallBack");
+                    }
+
+                    DataReceived?.Invoke(null, e);
```
myroot: Do use event to pass a result of read

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+                    {
+                        Log.Error(Globals.LogTag, "Error in Sensor Recorder Data aynchronously CallBack");
+                    }
+
+                    DataReceived?.Invoke(null, e);
+                    Log.Debug(Globals.LogTag, "Recorder data callback recieved");
+                    return true;
+                };
+
+                await Task.Run(() => AsyncReader(type, QueryHandle, _sensorRecorderDataCallBack, IntPtr.Zero)).ConfigureAwait(false);
```
myroot: Wrong implementation. 
App developer expect `Task` is complete when read operation was complete(data is received), but it is just end of request. it does not means response of request

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+        }
+
+        /// <summary>
+        ///  Starts to record a given sensor type.
+        /// </summary>
+        /// <param name="type">
+        /// A sensor type to be recorded.
+        /// </param>
+        /// <since_tizen> 6 </since_tizen>
+        public void Start(int type)
```
myroot: Why sensor type is int type?

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+                }
+            }
+            else
+            {
+                Log.Error(Globals.LogTag, "Recorder query is not created");
+            }
+        }
+
+        /// <summary>
+        ///  Queries the recorded data synchronously.
```
myroot: Method name is `Read` but description is not

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+        /// Checks whether it is supported to record a given sensor type.
+        /// </summary>
+        /// <param name="type">
+        /// A sensor type to check.
+        /// </param>
+        /// <returns>
+        /// returns true if it is supported to record a given sensor type.
+        /// </returns>
+        /// <since_tizen> 6 </since_tizen>
+        public bool IsSupported(int type)
```
myroot: Why not static?

```suggestion
        public bool IsRecordingSupported(int type)
```

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,504 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Tizen.Sensor
+{
+    /// <summary>
+    /// The SensorRecorder is used for retrieving the recorded data from a particular sensor.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class SensorRecorder : IDisposable
```
myroot: SensorRecorder should be static because it can record only one sensor at one time.
or use singleton pattern

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
pvschakradhar: okay, you mean to directly return the result in return value. after the await operation is completed ?

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
pvschakradhar: There are 2 options 
SENSOR_RECORDER_OPTION_RETENTION_PERIOD
SENSOR_RECORDER_OPTION_INTERVAL
should I replace int with enum ?

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
pvschakradhar: its an Enum for filtering and aggregation parameters for querying sensor records. I will replace it with enum.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
pvschakradhar: I am using Event to dispatch the result.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
pvschakradhar: I am using event to pass the result. Are you suggesting to directly return the result in return value ?

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
pvschakradhar: Will it be fine if i replace it with enum SensorType ?

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
pvschakradhar: replacing it with singleton pattern would be better it seems.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
pvschakradhar: Are you suggesting to return the response of request. like below code snippet ?

public async Task<int> ReadAsync(int type)
{
	int data;

        await Task.Run(() => data = AsyncReader(type, QueryHandle, _sensorRecorderDataCallBack, IntPtr.Zero)).ConfigureAwait(false);

	return data;
}

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
myroot: @pvschakradhar Could you explain work flow of SensorRecoder?
I can't understand how to work it and how to use it 
I need information of relation of Sensor and SensorRecorder

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
myroot: Sure

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
myroot: My missing, 
Do not use event. 
You should just return value

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
myroot: No, 

``` c#
               TaskCompletionSource<int> tcs = new TaskCompletionSource<int>();

                _sensorRecorderDataCallBack = (int _type, int data, int remains, int _error, Int64 _userData) =>
                {
                    if (_error != (int)SensorError.None)
                    {
                        Log.Error(Globals.LogTag, "Error in Sensor Recorder Data aynchronously CallBack");
                        tcs.SetException(new Exception("Error in Sensor .... " ));
                    } else {
                         tcs.SetResult(data);
                    }
                 }
   ....
                 return tcs.Task;
```

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
myroot: No, SensorType is not enough, 
because Sensor could be more than one
I recommend to use `Sensor` object

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
pvschakradhar: SensorRecorder is separate class, which is used for fetching the data stored in Data Base. There is no direct relation between Sensor and sensor recorder. Data Base differentiates the information as per sensor type.
So SensorRecorder can be made Static class or singleton class.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
pvschakradhar: data handle is void* in C API. In C# API it's just used to send back that value to C API again. Internally it's being typecasted.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
myroot: How to provide a state of recording?
The recording state should separated by each sensor instance.

for example, we should provide a way to check that specific sensor is now recording state

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
myroot: Usually, void* transfer to IntPtr in C#, but we do not recommend directly exposing a IntPtr.
You should provider a wrapper class to represent a data handle

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
pvschakradhar: I analysed in depth and found that data is being received in callback function with the sensor type parameter. So with this type parameter, sensor data can be separated. Kindly suggest your opinion on this. Is it better to use separate instance per sensor or static/singleton class for all sensors.

But I found sensor data(struct ctx_sensor_rec_data_s)
is coming as a void pointer, Need to handle receiving the complete data from C API to C# API.
Here is the structure of a data in C Layer:
struct ctx_sensor_rec_data_s {
		std::vector<std::string> keys;
		std::shared_ptr<Tuple> tuple;
	};

src/Tizen.Sensor/Tizen.Sensor/RecorderPlugins/PressureRecorder.cs
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+namespace Tizen.Sensor
+{
+    /// <summary>
+    /// The PressureRecorder class is used to start/stop record PressureSensor events.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public sealed class PressureRecorder : SensorRecorderDataManager
```
myroot: Do not make a subclass of SensorRecorderDataManager

pvschakradhar: yes, we redesigned the classes in the following way:
https://wormhole.sec.samsung.net/confluence/display/SRIDTizenServiceFW/Sensor+Recorder+Class+Diagram

pvschakradhar: please check this link for explanation regarding design
https://wormhole.sec.samsung.net/confluence/display/SRIDTizenServiceFW/Sensor+Recorder+Class+Diagram+Explanation

myroot: I have no permission to access

pvschakradhar: please check it now.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorderDataManager.cs
```diff
@@ -0,0 +1,576 @@
...
+            }
+
+            return new RecordTimePeriod(startTime, endTime);
+        }
+
+        /// <summary>
+        ///  Queries the recorded data synchronously and updates the data handle
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public void Update(Query query)
```
m-macias: Parameters of methods must have valid documentation. Method has 1 parameter, 0 are documented.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorderDataManager.cs
```diff
@@ -0,0 +1,576 @@
...
+                Log.Error(Globals.LogTag, "Error in asynchronously updating data handle");
+                throw SensorErrorFactory.CheckAndThrowException(error, "SensorRecorderDataManager.AsyncUpdater Failed");
+            }
+        }
+
+        /// <summary>
+        /// Queries the recorded data asynchronously and updates the data handle
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public async Task UpdateAsync(Query query)
```
m-macias: Parameters of methods must have valid documentation. Method has 1 parameter, 0 are documented.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,265 @@
...
+                throw SensorErrorFactory.CheckAndThrowException(error, "SensorRecorder.IsSupported Failed");
+            }
+            return isSupported;
+        }
+
+        /// <summary>
+        ///  Starts to record a given sensor type.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public void Start(RecordOption option)
```
m-macias: Parameters of methods must have valid documentation. Method has 1 parameter, 0 are documented.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,265 @@
...
+        /// Value.
+        /// </param>
+        /// <since_tizen> 6 </since_tizen>
+        void SetInterval(int value);
+
+        /// <summary>
+        /// gets the option object.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        RecordOption getOption();
```
m-macias: Method which returns non-void value must have exactly one 'returns' section.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorder.cs
```diff
@@ -0,0 +1,265 @@
...
+            {
+                Log.Error(Globals.LogTag, "Recorder option is not created");
+            }
+        }
+
+        /// <summary>
+        /// gets the OptionHandle object.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public RecordOption getOption()
```
m-macias: Method which returns non-void value must have exactly one 'returns' section.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorderDataManager.cs
```diff
@@ -0,0 +1,576 @@
...
+        /// Value.
+        /// </param>
+        /// <since_tizen> 6 </since_tizen>
+        void SetInterval(int value);
+
+        /// <summary>
+        /// gets the QueryHandle object.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        Query getQuery();
```
m-macias: Method which returns non-void value must have exactly one 'returns' section.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorderDataManager.cs
```diff
@@ -0,0 +1,576 @@
...
+            {
+                Log.Error(Globals.LogTag, "Recorder query is not created");
+            }
+        }
+
+        /// <summary>
+        /// gets the QueryHandle object.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public Query getQuery()
```
m-macias: Method which returns non-void value must have exactly one 'returns' section.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorderDataManager.cs
m-macias: Method which returns non-void value must have exactly one 'returns' section.

src/Tizen.Sensor/Tizen.Sensor/SensorRecorderDataManager.cs
pvschakradhar: added

pvschakradhar:
please follow the below links for ACR and design
ACR : http://suprem.sec.samsung.net/jira/browse/TCSACR-218
Design (Sensor Recorder in that page) : http://suprem.sec.samsung.net/confluence/pages/viewpage.action?pageId=34308642


Explanation of class diagram:

The RecordOption class is designed using builder pattern, It allows user to set respective option parameters.

The SensorRecorder class is used for controlling the sensor recorder(start,stop). SensorRecorder class uses the optionHandle in RecordOption class, in its Start() API.

The Query class is designed using builder pattern, It allows user to set respective query parameters.

The SensorRecorderDataManager class is used for retrieving the data from the database using API's DataGetTime(), DataGetDouble(). These API's use dataHandle internally, which is updated with the help of API's Update() and UpdateAsync().

The queryHandle in Query class is used in Update() and UpdateAsync() API's.

src/Tizen.Sensor/Tizen.Sensor/RecorderPlugins/PressureRecorder.cs
myroot: Still no permission

myroot:
image

It is a class diagram of sensor recorder that you design

Q1. Is `SensorRecordedDataManager` static class or not?

1. `IsSupported` of `SensorRecordedDataManager` is nonsense, 
1. `IsSupported` of `SensorRecorder` is nonsense, 
2. There is no `SensorRecorder` constructor, but it need
3. Need to specify the return type
4. In design, only specify a public members

pvschakradhar:
Sorry for the inconvenience, I am unable to attach class diagram here.
I updated the class diagram with only public members in design page (http://suprem.sec.samsung.net/confluence/pages/viewpage.action?pageId=34308642).

Q1. Is SensorRecordedDataManager static class or not?

we didn't go for static implementation because we are maintaining a private variable dataHandle.

with a given query, when update(Qyery query) API is called then dataHandle is being recieved in callback.
and this dataHandle is internally being used in APIs DataGetDouble(RecorderData key), DataGetTime().

user gets respective information about data from DataGetDouble(RecorderData key), DataGetTime() API's.

PR #935, opened 2019-07-18
[NUI] Add to set MIME type to the input panel and its callback

Seoyeon2Kim:

### Description of Change ###
 - Added an API to set the allowed MIME type to deliver to the input panel.
 - Added a callback to receive the content of the input method URI. 

### API Changes ###
Added:
 - public event EventHandler<ContentReceivedEventArgs> ContentReceived { add; remove; }
 - public void SetMIMEType(string mimeType);
 - public class ContentReceivedEventArgs : EventArgs {}

### DIFF
--- a/src/Tizen.NUI/src/internal/ContentReceivedSignalType.cs
+++ b/src/Tizen.NUI/src/internal/ContentReceivedSignalType.cs
@@ -1,0 +1,208 @@
+/*
+ * Copyright(c) 2019 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI/src/internal/ContentReceivedSignalType.cs
```diff
@@ -0,0 +1,208 @@
...
+                isDisposeQueued = true;
+                DisposeQueue.Instance.Add(this);
+            }
+        }
+
+        /// <summary>
+        /// Dispose
+        /// </summary>
+        /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
dongsug-song: it looks no need to add "[EditorBrowsable(EditorBrowsableState.Never)]" on methods or properties of "internal class". This "[EditorBrowsable(EditorBrowsableState.Never)]" needs only for HiddenAPI (InhouseAPI) which has public access modifier.

Seoyeon2Kim: Oh you're right. Removed the tags.
Thank you for reviewing it.

PR #936, opened 2019-07-18
[NUI] Add KeyString in Key.cs

Seoyeon2Kim:
Change-Id: I5aa3680b81cc62d6718aa0f21a08bc7c74b8826b

### Description of Change ###
- Provide to use 'KeyString' for a clearer meaning instead of 'KeyPressed'

### API Changes ###
Added:
 - public string KeyString { get; }

### DIFF
--- a/src/Tizen.NUI/src/public/Key.cs
+++ b/src/Tizen.NUI/src/public/Key.cs
@@ -167,6 +167,20 @@ namespace Tizen.NUI
         }
 
         /// <summary>
+        /// Get the actual string returned that should be used for input editors.
+        /// </summary>
+        /// <returns>The key string</returns>
+        /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public string KeyString
+        {
+            get
+            {
+                return keyString;
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Key.cs
```diff
@@ -166,6 +166,20 @@ public string KeyPressed
             }
         }
 
+        /// <summary>
+        /// Get the actual string returned that should be used for input editors.
+        /// </summary>
+        /// <returns>The key string</returns>
+        /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public string KeyString
```
dongsug-song: Hello, private keyString is set/get property, but I can not understand why this KeyString is only get property?

dongsug-song: yes... I think it would be better to make them as same for set/get properties. one is only get but the other is set/get... this looks not good... 
OR if you think "private keyString" will be used even in the future, then it need be "internal keyString", I think. because we can give internal API access permission to specific module by using friend-assembly.

Seoyeon2Kim: Added 'Set' property in KeyString, too. :)

PR #938, opened 2019-07-22
[MediaCodec] Fix the bug of callback registration

hsgwon:
### Description of Change ###
Fix the bug of callback registration

### DIFF
--- a/src/Tizen.Multimedia.MediaCodec/MediaCodec/MediaCodec.cs
+++ b/src/Tizen.Multimedia.MediaCodec/MediaCodec/MediaCodec.cs
@@ -618,6 +618,9 @@ namespace Tizen.Multimedia.MediaCodec
         {
             _bufferStatusCb = (statusCode, _) =>
             {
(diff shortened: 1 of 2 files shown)


src/Tizen.Multimedia.MediaCodec/MediaCodec/MediaCodec.cs
```diff
@@ -617,11 +618,7 @@ private void RegisterBufferStatusChanged()
         {
             _bufferStatusCb = (statusCode, _) =>
             {
-                Debug.Assert(Enum.IsDefined(typeof(MediaCodecStatus), statusCode),
-                    $"{ statusCode } is not defined in MediaCodecStatus!");
```
aferin: even if the type is changed to enum, the range of it will not be checked. are there any reasons to delete it?

hsgwon: It's p/invoke function so, it's hard to set wrong value be native framework.
But, even if so, it's better to check it. I'll rollback it.

PR #396, opened 2018-08-17
[MediaController] Add new APIs for sending command and its callback

hsgwon:
### Description of Changes ###

Add new APIs for sending command and its callback and deprecate old one.

### API Changes ###

- TCSACR-176
   (It's related Native ACR-1235.)

Added:

- Class
   - CommandCompletedEventArgs
   - CustomCommandReceivedEventArgs
   - PlaybackActionCommandReceivedEventArgs 
   - PlaybackPositionCommandReceivedEventArgs 
   - PlaylistCommandReceivedEventArgs 
   - RepeatModeCommandReceivedEventArgs 
   - ShuffleModeCommandReceivedEventArgs 
 
- Enum
   - MediaControlPlaybackCommand.Toggle
   - MediaControlPlaybackState.MovingToNext
   - MediaControlPlaybackState.MovingToPrevious
   - MediaControlRepeatMode.OneMedia
   - MediaControlShuffleMode.Off / On

- Event
   - PlaybackActionCommandReceived      // server
   - PlaybackPositionCommandReceived   // server
   - PlaylistCommandReceived                   // server
   - ShuffleModeCommandReceived         // server
   - RepeatModeCommandReceived         // server
   - CustomCommandReceived                 // server

- Method
   - SendCommandReply                           // server
   - SendCommand                                    // client. 7 overload

Deprecated:
- Class
   - PlaybackCommandReceivedEventArgs

- Event
   - PlaybackCommandReceived

- Method
   - SendPlaybackCommand

PR #944, opened 2019-07-24
[MediaPlayer] Add API to be able to cancel preparing it

aferin:
### API Changes ###
 - ACR: http://suprem.sec.samsung.net/jira/browse/TCSACR-253

Added:
 - public event EventHandler<EventArgs> Prepared;
 - public async Task PrepareAsync(CancellationToken cancellationToken)

### DIFF
--- a/src/Tizen.Multimedia.MediaPlayer/Player/Player.Events.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/Player.Events.cs
@@ -74,11 +74,6 @@ namespace Tizen.Multimedia
         public event EventHandler<BufferingProgressChangedEventArgs> BufferingProgressChanged;
         private NativePlayer.BufferingProgressCallback _bufferingProgressCallback;
 
-        /// <summary>
-        /// Occurs when the player is prepared.
-        /// </summary>
-        /// <since_tizen> 6 </since_tizen>
-        public event EventHandler<EventArgs> Prepared;
         private NativePlayer.PrepareCallback _prepareCallback;
 
         internal event EventHandler<MediaStreamBufferStatusChangedEventArgs> MediaStreamAudioBufferStatusChanged;
--- a/src/Tizen.Multimedia.MediaPlayer/Player/Player.Events.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/Player.Events.cs
@@ -74,6 +74,13 @@ namespace Tizen.Multimedia
         public event EventHandler<BufferingProgressChangedEventArgs> BufferingProgressChanged;
         private NativePlayer.BufferingProgressCallback _bufferingProgressCallback;
 
+        /// <summary>
+        /// Occurs when the player is prepared.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public event EventHandler<EventArgs> Prepared;
+        private NativePlayer.PrepareCallback _prepareCallback;
+
         internal event EventHandler<MediaStreamBufferStatusChangedEventArgs> MediaStreamAudioBufferStatusChanged;
         private NativePlayer.MediaStreamBufferStatusCallback _mediaStreamAudioBufferStatusChangedCallback;
 
--- a/src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
@@ -384,6 +384,9 @@ namespace Tizen.Multimedia
             {
                 ValidatePlayerState(PlayerState.Preparing);
 
+                // a user can get the state before finally block is called.
+                ClearPreparing();
+
                 Log.Warn(PlayerLog.Tag, $"preparing will be cancelled.");
                 NativePlayer.Unprepare(Handle).ThrowIfFailed(this, "Failed to unprepare the player");
 
@@ -392,9 +395,6 @@ namespace Tizen.Multimedia
 
             _prepareCallback = _ =>
             {
-                // a user can get the state before finally block is called.
-                ClearPreparing();
-
                 Log.Warn(PlayerLog.Tag, $"prepared callback is called.");
                 taskCompletionSource.TrySetResult(true);
             };
--- a/src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
@@ -390,14 +390,18 @@ namespace Tizen.Multimedia
                 taskCompletionSource.TrySetCanceled();
             });
 
-            _prepareCallback = _ => taskCompletionSource.TrySetResult(true);
+            _prepareCallback = _ =>
+            {
+                Log.Warn(PlayerLog.Tag, $"prepared callback is called.");
+                taskCompletionSource.TrySetResult(true);
+            };
 
             try
             {
                 NativePlayer.PrepareAsync(Handle, _prepareCallback, IntPtr.Zero).
                     ThrowIfFailed(this, "Failed to prepare the player");
 
-                await taskCompletionSource.Task.ConfigureAwait(false) ;
+                await taskCompletionSource.Task.ConfigureAwait(false);
             }
             finally
             {
--- a/src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
@@ -366,6 +366,8 @@ namespace Tizen.Multimedia
         /// <since_tizen> 6 </since_tizen>
         public virtual async Task PrepareAsync(CancellationToken cancellationToken)
         {
+            ValidateNotDisposed();
+
             var taskCompletionSource = new TaskCompletionSource<bool>();
 
             if (_source == null)
--- a/src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
@@ -346,6 +346,69 @@ namespace Tizen.Multimedia
         }
 
         /// <summary>
+        /// Prepares the cancellable media player for playback, asynchronously.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token to cancel preparing.</param>
+        /// <seealso cref="CancellationToken"/>
+        /// <returns>A task that represents the asynchronous prepare operation.</returns>
+        /// <remarks>To prepare the player, the player must be in the <see cref="PlayerState.Idle"/> state,
+        /// and a source must be set.
+        /// The state must be <see cref="PlayerState.Preparing"/> to cancel preparing.
+        /// When preparing is cancelled, a state will be changed to <see cref="PlayerState.Idle"/> from <see cref="PlayerState.Preparing"/>.
+        /// <exception cref="ObjectDisposedException">The player has already been disposed of.</exception>
+        /// <exception cref="InvalidOperationException">
+        ///     Operation failed; internal error.
+        ///     -or-<br/>
+        ///     The player is not in the valid state.
+        ///     </exception>
+        /// <seealso cref="PrepareAsync()"/>
+        /// <seealso cref="Unprepare()"/>
+        /// <seealso cref="Prepared"/>
+        /// <since_tizen> 6 </since_tizen>
+        public async Task PrepareAsync(CancellationToken cancellationToken)
+        {
+            var taskCompletionSource = new TaskCompletionSource<bool>();
+
+            if (_source == null)
+            {
+                throw new InvalidOperationException("No source is set.");
+            }
+
+            ValidatePlayerState(PlayerState.Idle);
+
+            OnPreparing();
+
+            SetPreparing();
+
+            // register a callback to handle cancellation token anytime it occurs
+            cancellationToken.Register(() =>
+            {
+                ValidatePlayerState(PlayerState.Preparing);
+
+                Log.Warn(PlayerLog.Tag, $"preparing will be cancelled.");
+                NativePlayer.Unprepare(Handle).ThrowIfFailed(this, "Failed to unprepare the player");
+
+                taskCompletionSource.TrySetCanceled();
+
+                ClearPreparing();
+            });
+
+            _prepareCallback = _ =>
+            {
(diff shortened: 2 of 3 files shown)


src/Tizen.Multimedia.MediaPlayer/Player/Player.Events.cs
```diff
@@ -74,6 +74,13 @@ public partial class Player
         public event EventHandler<BufferingProgressChangedEventArgs> BufferingProgressChanged;
         private NativePlayer.BufferingProgressCallback _bufferingProgressCallback;
 
+        /// <summary>
+        /// Occurs when the player is prepared.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public event EventHandler<EventArgs> Prepared;
```
hsgwon: Did you intend to just notify `preparing is done` to user?
If so, I think it doesn't needed. 
The usage of this method is the same with existing `PrepareAsync()` method.
I think the only difference is that `Cancel` is available or not.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -345,6 +345,69 @@ public virtual Task PrepareAsync()
...
+        /// <exception cref="InvalidOperationException">
+        ///     Operation failed; internal error.
+        ///     -or-<br/>
+        ///     The player is not in the valid state.
+        ///     </exception>
+        /// <seealso cref="PrepareAsync()"/>
+        /// <seealso cref="Unprepare()"/>
+        /// <seealso cref="Prepared"/>
+        /// <since_tizen> 6 </since_tizen>
+        public async Task PrepareAsync(CancellationToken cancellationToken)
```
hsgwon: It's not `virtual` method otherwise existing `public virtual Task PrepareAsync()` method.
Is it intended?

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -345,6 +345,69 @@ public virtual Task PrepareAsync()
...
+
+                ClearPreparing();
+            });
+
+            _prepareCallback = _ =>
+            {
+                Log.Warn(PlayerLog.Tag, $"prepared callback is called.");
+                Prepared?.Invoke(this, EventArgs.Empty);
+
+                ClearPreparing();
```
myroot:
``` c#
ClearPreparing();
Prepared?.Invoke(this, EventArgs.Empty);
```
You need to clear preparing flag before send event
because event name is `Prepared`

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -345,6 +345,69 @@ public virtual Task PrepareAsync()
...
+            cancellationToken.Register(() =>
+            {
+                ValidatePlayerState(PlayerState.Preparing);
+
+                Log.Warn(PlayerLog.Tag, $"preparing will be cancelled.");
+                NativePlayer.Unprepare(Handle).ThrowIfFailed(this, "Failed to unprepare the player");
+
+                taskCompletionSource.TrySetCanceled();
+
+                ClearPreparing();
```
myroot: First clear preparing flag and send notify
``` c#
ClearPreparing();
taskCompletionSource.TrySetCanceled();
```

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -345,6 +345,69 @@ public virtual Task PrepareAsync()
...
+            }
+
+            ValidatePlayerState(PlayerState.Idle);
+
+            OnPreparing();
+
+            SetPreparing();
+
+            // register a callback to handle cancellation token anytime it occurs
+            cancellationToken.Register(() =>
```
myroot: cancellationToken can be null

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -345,6 +345,69 @@ public virtual Task PrepareAsync()
...
+        /// <exception cref="ObjectDisposedException">The player has already been disposed of.</exception>
+        /// <exception cref="InvalidOperationException">
+        ///     Operation failed; internal error.
+        ///     -or-<br/>
+        ///     The player is not in the valid state.
+        ///     </exception>
+        /// <seealso cref="PrepareAsync()"/>
+        /// <seealso cref="Unprepare()"/>
+        /// <seealso cref="Prepared"/>
+        /// <since_tizen> 6 </since_tizen>
```
myroot: I wonder why do not replace implementation of `public virtual Task PrepareAsync()` method with this method like below
``` c#
 public virtual Task PrepareAsync() {
     return PrepareAsync(null);
 }
```

src/Tizen.Multimedia.MediaPlayer/Player/Player.Events.cs
myroot: I also agree to hsgwon.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -345,6 +345,69 @@ public virtual Task PrepareAsync()
...
+            {
+                Log.Warn(PlayerLog.Tag, $"prepared callback is called.");
+                Prepared?.Invoke(this, EventArgs.Empty);
+
+                ClearPreparing();
+
+                taskCompletionSource.TrySetResult(true);
+            };
+
+            NativePlayer.PrepareAsync(Handle, _prepareCallback, IntPtr.Zero).ThrowIfFailed(this, "Failed to prepare the player");
```
myroot: If exception was thrown on this line, preparing flag was not cleared

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
hsgwon: CancellationToken is `struct` type. So, if you call like `PrepareAsync(null)`, it's build break.
VS warn you with following message.
```
cannot convert from `<null>` to CancellationToken.
```

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
hsgwon: CancellationToken is `struct` type. So, if you call like `PrepareAsync(null)`, it's build break.
VS warn you with following message.
```
cannot convert from `<null>` to CancellationToken.
```

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
hsgwon: How about following changes?
--------------------------------------
```csharp
public virtual async Task PrepareAsync(CancellationToken cancellationToken)
{
    ...

    // register a callback to handle cancellation token anytime it occurs
    cancellationToken.Register(() =>
    {
        ValidatePlayerState(PlayerState.Preparing);

        NativePlayer.Unprepare(Handle).
            ThrowIfFailed(this, "Failed to unprepare the player");

        tcs.TrySetCanceled();
    });

    _prepareCallback = _ => tcs.TrySetResult(true);

    try
    {
        // Or, make following code to `void PrepareAsync()` mehtod.
        NativePlayer.PrepareAsync(Handle, _prepareCallback, IntPtr.Zero).
            ThrowIfFailed(this, "Failed to prepare the player");

        await tcs.Task;
    }
    finally
    {
        ClearPreparing();
    }
}
```

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
myroot: Yes, you are right, it can't be null but CancellationToken provide a empty token to use as default
``` c#
 return PrepareAsync(CancellationToken.None);
```
https://docs.microsoft.com/ko-kr/dotnet/api/system.threading.cancellationtoken.none?view=netframework-4.8#System_Threading_CancellationToken_None

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
myroot: It seem to good

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
hsgwon: Right, it's better to change it like your comment, if there's no reason why it remained separately.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Events.cs
aferin: ok i'll delete it.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
aferin: as far as i know, virtual keyword allows it to be overridden. will it be needed for it?
i also wonder the reason why PrepareAsync() has a virtual keyword.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
aferin: ok but now i've deleted Prepared event.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
aferin: ok i'll fix it

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
aferin: mr.gwon is right for it.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
aferin: actually i was trying to combine both methods. but i couldn't be sure if it is totally the same or not, because of "return Task.Factory.StartNew" from what i've read, it is dividing thread.
doens't have any issues about it?

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
aferin: oh i was considering it but forgot to change it. thanks for reminding me.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
aferin: ok. i'll change the codes based on this one.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -345,6 +345,66 @@ public virtual Task PrepareAsync()
...
+            });
+
+            _prepareCallback = _ => taskCompletionSource.TrySetResult(true);
+
+            try
+            {
+                NativePlayer.PrepareAsync(Handle, _prepareCallback, IntPtr.Zero).
+                    ThrowIfFailed(this, "Failed to prepare the player");
+
+                await taskCompletionSource.Task;
```
myroot:
```suggestion
                await taskCompletionSource.Task.ConfigureAwait(false) ;
```

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
myroot: We need to check the thread of _prepareCallback that called by PrepareAsync
if _prepareCallback was called on main loop thread. it cause behavior break.

As is below code is working
``` c#
 player.PrepareAsync().Wait();
```
But, 
if _prepareCallback was only execute on main loop thread, 
PrepareAsync(CancellationToken cancellationToken) is not working on same behavior
``` c#
 player.PrepareAsync(CancellationToken.None).Wait();
```

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
hsgwon: I just want to know what is difference between this and the below comment of last year PR.

https://github.com/Samsung/TizenFX/pull/396#discussion_r217556747

myroot: The code after `await` is important, if the code should be run on main thread, We should use `ConfigureAwait(true)`, if not We can(and recommend to) use `ConfigureAwait(false)`
in this case, `ClearPreparing` can run on worker thread

ConfigureAwait(true) is returing on the current thread. 
ConfigureAwait(false) is returing on the working thread.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
aferin: while Prepare() in PrepareAsync() is executed on a new thread, _prepareCallback in PrepareAsync(CancellationToken) is called on the main thread.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -345,6 +345,75 @@ public virtual Task PrepareAsync()
                 TaskScheduler.Default);
         }
 
+        /// <summary>
+        /// Prepares the cancellable media player for playback, asynchronously.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token to cancel preparing.</param>
+        /// <seealso cref="CancellationToken"/>
+        /// <returns>A task that represents the asynchronous prepare operation.</returns>
```
journey2w: Suggestion:
The task that represents the asynchronous prepare operation.

src/Tizen.Multimedia.MediaPlayer/Player/Player.cs
```diff
@@ -345,6 +345,75 @@ public virtual Task PrepareAsync()
...
+        /// <remarks>To prepare the player, the player must be in the <see cref="PlayerState.Idle"/> state,
+        /// and a source must be set.
+        /// The state must be <see cref="PlayerState.Preparing"/> to cancel preparing.
+        /// When preparing is cancelled, a state will be changed to <see cref="PlayerState.Idle"/> from <see cref="PlayerState.Preparing"/>.</remarks>
+        /// <exception cref="ObjectDisposedException">The player has already been disposed of.</exception>
```
journey2w: Suggestion:

The player has already been disposed.

PR #956, opened 2019-08-01
[NUI]Change GetWindowList() to static

tscholb:
### API Changes ###

 - ACR:


tscholb:
[NUI]Change GetWindowList() to static

PR #958, opened 2019-08-02
[NUI] Change the name of Tizen.NUI.CommonUI as Tizen.NUI.Components

Lynettejin90:
### API Changes ###

 - ACR:


dongsug-song:
Hello, Ms. Jin 
there are many changes, so the API Checker seems to not work normally.
so I checked locally with attached API_sorted_this_patch.csv file with current TizenFX master.
there was no API break, so I will merge this patch.
thank you.

PR #960, opened 2019-08-05
[NUI] Add method to destroy a window

tscholb:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window.cs
+++ b/src/Tizen.NUI/src/public/Window.cs
@@ -997,6 +997,14 @@ namespace Tizen.NUI
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Window.cs
```diff
@@ -996,6 +996,14 @@ public Layer GetLayer(uint depth)
             }
         }
 
+        /// <summary>
+        /// Destroy the window immediately.
+        /// </summary>
+        public void Destroy()
```
dongsug-song: please add "[EditorBrowsable(EditorBrowsableState.Never)]" for hidden api

PR #962, opened 2019-08-06
[Tizen.Network] Add Tizen.Network

chleun-moon:
### Description of Change ###
Add Tizen.Network namespace to contains enumerations commonly used in sub namespaces (Tizen.Network.xxx)

### API Changes ###
 - ACR: http://suprem.sec.samsung.net/jira/browse/TCSACR-257

Add new Enumerations to be used Tizen.Network.Connection, Tizen.Network.WiFi and Tizen.Network.NetworkMonitor (to be added newly).

### DIFF
--- a/src/Tizen.Network/Tizen.Network/NetworkEnumerations.cs
+++ b/src/Tizen.Network/Tizen.Network/NetworkEnumerations.cs
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2019 Samsung Electronics Co., Ltd All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+namespace Tizen.Network
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+	public enum ConnectionType
+	{
+		/// <summary>
+		/// Disconnected
+		/// </summary>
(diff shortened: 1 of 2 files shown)


src/Tizen.Network/Tizen.Network/NetworkEnumerations.cs
```diff
@@ -0,0 +1,81 @@
...
+ * limitations under the License.
+*/
+
+namespace Tizen.Network
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+	public enum ConnectionType
```
myroot: It is duplication of Tizen.Network.Connection.ConnectionType

src/Tizen.Network/Tizen.Network/NetworkEnumerations.cs
```diff
@@ -0,0 +1,81 @@
...
+		/// </summary>
+		/// <since_tizen> 6 </since_tizen>
+		NetProxy,
+	}
+
+    /// <summary>
+    /// Enumeration for the proxy method type.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public enum ProxyType
```
myroot: It is duplication of Tizen.Network.Connection.ProxyType

src/Tizen.Network/Tizen.Network/NetworkEnumerations.cs
chleun-moon: Yes. ConnectionType and ProxyType are defined in Tizen.Network.Connection. They are used in Tizen.Network.WiFi and will be used in Tizen.Network.NetworkMonitor to be newly added. So we want to add ConnectionType and ProxyType in Tizen.Network namespace and deprecate from Tizen.Network.Connection namespace. What do you think about it?

myroot: I don't recommend it, it cause too many API changes.
If Tizen.Network.Connection.ConnectionType is deprecated, the following APIs are also deprecated
`public static ConnectionState WiFiState`
`public static ConnectionState BluetoothState`
`public ConnectionState State`

I recommend to use Tizen.Network.Connection.ConnectionType in othres

chleun-moon: Ok. I'll cancel this PR as your recommend.
Thanks.

PR #967, opened 2019-08-07
[NUI] integration from dalihub

dongsug-song:
### Description of Change ###
[NUI] integration from dalihub

### API Changes ###
None


dongsug-song:
After this sync patch is merged, then back sync to dalihub will be done

PR #974, opened 2019-08-13
[MediaPlayer] add APIs to use AudioOffload

aferin:
### API Changes ###
 - ACR: http://suprem.sec.samsung.net/jira/browse/TCSACR-262

Added:
- public class AudioOffload
- public bool IsEnabled
- public bool IsActivated
- public IEnumerable<MediaFormatAudioMimeType> SupportedFormats
- public AudioOffload AudioOffload

Changed:  : the error case is added. when audio offload is set, invalid operation occurs for the below APIs
(AudioEffect.cs)
public EqualizerBand this[int index]
public void Clear()
public int Count
public Range BandLevelRange
public bool IsAvailable

(EqualizerBand.cs)
public int Level
public int Frequency
public int FrequencyRange

(Player.Properties.cs)
public AudioLatencyMode AudioLatencyMode
public bool ReplayGain
public bool AudioPitchEnabled
public float AudioPitch

(Player.cs)
public void SetPlaybackRate(float rate)
public void EnableExportingAudioData(AudioMediaFormat format, PlayerAudioExtractOption option)

(PlayerTrackInfo.cs)
public int GetCount()
public string GetLanguageCode(int index)
public int Selected

### DIFF
--- a/src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
@@ -42,7 +42,7 @@ namespace Tizen.Multimedia
 
             _count = Count;
             _bandLevelRange = BandLevelRange;
-            _bands = new EqualizerBand[_count];
+            _bands = new EqualizerBand[Count];
         }
 
         /// <summary>
--- a/src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
@@ -35,7 +35,7 @@ namespace Tizen.Multimedia
             Player = owner;
 
             _isAvailable = IsAvailable;
-            if (_isAvailable == false)
+            if (IsAvailable== false)
             {
                 return;
             }
--- a/src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
@@ -26,22 +26,16 @@ namespace Tizen.Multimedia
     public class AudioEffect
     {
         private readonly EqualizerBand[] _bands;
-        private readonly int _count;
-        private readonly bool _isAvailable;
-        private readonly Range _bandLevelRange;
 
         internal AudioEffect(Player owner)
         {
             Player = owner;
 
-            _isAvailable = IsAvailable;
             if (IsAvailable== false)
             {
                 return;
             }
 
-            _count = Count;
-            _bandLevelRange = BandLevelRange;
             _bands = new EqualizerBand[Count];
         }
 
@@ -64,7 +58,7 @@ namespace Tizen.Multimedia
             get
             {
                 Player.ValidateNotDisposed();
-                Player.AudioOffload.CheckEnabled();
+                Player.AudioOffload.CheckDisabled();
 
                 if (index < 0 || Count <= index)
                 {
@@ -92,7 +86,7 @@ namespace Tizen.Multimedia
         public void Clear()
         {
             Player.ValidateNotDisposed();
-            Player.AudioOffload.CheckEnabled();
+            Player.AudioOffload.CheckDisabled();
 
             Native.EqualizerClear(Player.Handle).
                 ThrowIfFailed(Player, "Failed to clear equalizer effect");
@@ -109,7 +103,7 @@ namespace Tizen.Multimedia
         {
             get
             {
-                Player.AudioOffload.CheckEnabled();
+                Player.AudioOffload.CheckDisabled();
 
                 Native.GetEqualizerBandsCount(Player.Handle, out var count).
                     ThrowIfFailed(Player, "Failed to initialize the AudioEffect");
@@ -129,7 +123,7 @@ namespace Tizen.Multimedia
         {
             get
             {
-                Player.AudioOffload.CheckEnabled();
+                Player.AudioOffload.CheckDisabled();
 
                 Native.GetEqualizerLevelRange(Player.Handle, out var min, out var max).
                     ThrowIfFailed(Player, "Failed to initialize the AudioEffect");
@@ -149,7 +143,7 @@ namespace Tizen.Multimedia
         {
             get
             {
-                Player.AudioOffload.CheckEnabled();
+                Player.AudioOffload.CheckDisabled();
 
                 Native.EqualizerIsAvailable(Player.Handle, out var available).
                     ThrowIfFailed(Player, "Failed to initialize the AudioEffect");
--- a/src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
@@ -26,36 +26,23 @@ namespace Tizen.Multimedia
     public class AudioEffect
     {
         private readonly EqualizerBand[] _bands;
+        private readonly int _count;
+        private readonly bool _isAvailable;
+        private readonly Range _bandLevelRange;
 
         internal AudioEffect(Player owner)
         {
             Player = owner;
 
-            bool available = false;
-
-            Native.EqualizerIsAvailable(Player.Handle, out available).
-                ThrowIfFailed(Player, "Failed to initialize the AudioEffect");
-
-            IsAvailable = available;
-
-            if (IsAvailable == false)
+            _isAvailable = IsAvailable;
+            if (_isAvailable == false)
             {
                 return;
             }
 
-            int count = 0;
-            Native.GetEqualizerBandsCount(Player.Handle, out count).
-                ThrowIfFailed(Player, "Failed to initialize the AudioEffect");
-
-            int min = 0;
-            int max = 0;
-            Native.GetEqualizerLevelRange(Player.Handle, out min, out max).
-                ThrowIfFailed(Player, "Failed to initialize the AudioEffect");
-
-            Count = count;
-            BandLevelRange = new Range(min, max);
-
-            _bands = new EqualizerBand[count];
+            _count = Count;
+            _bandLevelRange = BandLevelRange;
+            _bands = new EqualizerBand[_count];
         }
 
         /// <summary>
@@ -68,12 +55,16 @@ namespace Tizen.Multimedia
         ///     -or-<br/>
         ///     <paramref name="index"/> is equal to or greater than <see cref="Count"/>.
         /// </exception>
+        /// <exception cref="InvalidOperationException">
+        ///     If audio offload is enabled by calling <see cref="AudioOffload.IsEnabled"/>. (Since tizen 6.0)
+        /// </exception>
         /// <since_tizen> 3 </since_tizen>
         public EqualizerBand this[int index]
         {
             get
             {
                 Player.ValidateNotDisposed();
+                Player.AudioOffload.CheckEnabled();
 
                 if (index < 0 || Count <= index)
                 {
@@ -94,10 +85,14 @@ namespace Tizen.Multimedia
         /// Clears the equalizer effect.
         /// </summary>
         /// <exception cref="ObjectDisposedException">The <see cref="Player"/> has already been disposed of.</exception>
+        /// <exception cref="InvalidOperationException">
+        ///     If audio offload is enabled by calling <see cref="AudioOffload.IsEnabled"/>. (Since tizen 6.0)
+        /// </exception>
         /// <since_tizen> 3 </since_tizen>
         public void Clear()
         {
             Player.ValidateNotDisposed();
+            Player.AudioOffload.CheckEnabled();
 
             Native.EqualizerClear(Player.Handle).
                 ThrowIfFailed(Player, "Failed to clear equalizer effect");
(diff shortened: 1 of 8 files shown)


src/Tizen.Multimedia.MediaPlayer/Player/AudioOffload.cs
```diff
@@ -0,0 +1,161 @@
...
+        /// Provides a means to retrieve audio offload information.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        internal AudioOffload(Player player)
+        {
+            Debug.Assert(player != null);
+            Player = player;
+        }
+
+        internal void CheckEnabled()
```
myroot: It is right?
As a naming, it seem to check enable state, so enabled is normal, others is abnormal.

But, implementation is opposite

src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
```diff
@@ -26,36 +26,23 @@ namespace Tizen.Multimedia
...
-            bool available = false;
-
-            Native.EqualizerIsAvailable(Player.Handle, out available).
-                ThrowIfFailed(Player, "Failed to initialize the AudioEffect");
-
-            IsAvailable = available;
-
-            if (IsAvailable == false)
+            _isAvailable = IsAvailable;
+            if (_isAvailable == false)
```
myroot:
```suggestion
            if (IsAvailable== false)
```

src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
```diff
@@ -26,36 +26,23 @@ namespace Tizen.Multimedia
     public class AudioEffect
     {
         private readonly EqualizerBand[] _bands;
+        private readonly int _count;
+        private readonly bool _isAvailable;
+        private readonly Range _bandLevelRange;
```
myroot: _count, _isAvailable, _bandLevelRange

Why need it as filed member?
nobody use it, except constructor
I think, it should be a local variable

src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
```diff
@@ -26,36 +26,23 @@ namespace Tizen.Multimedia
...
-            Native.GetEqualizerLevelRange(Player.Handle, out min, out max).
-                ThrowIfFailed(Player, "Failed to initialize the AudioEffect");
-
-            Count = count;
-            BandLevelRange = new Range(min, max);
-
-            _bands = new EqualizerBand[count];
+            _count = Count;
+            _bandLevelRange = BandLevelRange;
+            _bands = new EqualizerBand[_count];
```
myroot:
```suggestion
            _bands = new EqualizerBand[Count];
```

src/Tizen.Multimedia.MediaPlayer/Player/EqualizerBand.cs
```diff
@@ -29,23 +29,19 @@ public class EqualizerBand
     {
         private readonly AudioEffect _owner;
         private readonly int _index;
+        private readonly int _frequency;
+        private readonly int _frequencyRange;
```
myroot: Why need it as filed member?
nobody use it, except constructor
I think, it should be a local variable

src/Tizen.Multimedia.MediaPlayer/Player/EqualizerBand.cs
```diff
@@ -29,23 +29,19 @@ public class EqualizerBand
...
 
         internal EqualizerBand(AudioEffect owner, int index)
         {
             Debug.Assert(owner != null);
 
             _owner = owner;
             _index = index;
+            _frequency = Frequency;
+            _frequencyRange = FrequencyRange;
```
myroot: Why need it?

src/Tizen.Multimedia.MediaPlayer/Player/AudioOffload.cs
aferin: then i'll change the name to CheckDisabled.

src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
aferin: that's right. i'll fix it.

src/Tizen.Multimedia.MediaPlayer/Player/EqualizerBand.cs
aferin: i'll fix it.

src/Tizen.Multimedia.MediaPlayer/Player/EqualizerBand.cs
aferin: it isn't needed

src/Tizen.Multimedia.MediaPlayer/Player/EqualizerBand.cs
```diff
@@ -56,12 +48,16 @@ internal EqualizerBand(AudioEffect owner, int index)
...
+        /// <exception cref="InvalidOperationException">
+        ///     If audio offload is enabled by calling <see cref="AudioOffload.IsEnabled"/>. (Since tizen 6.0)
+        /// </exception>
         /// <since_tizen> 3 </since_tizen>
         public int Level
         {
             get
             {
                 _owner.Player.ValidateNotDisposed();
+                _owner.Player.AudioOffload.CheckDisabled();
```
myroot: Why do to check disabling? it seem to no need to check disable becuase it just get

myroot:
Do you really need to check disabling of  AudioOffload?
It cause to a one introp call for each function call related with audio.

src/Tizen.Multimedia.MediaPlayer/Player/EqualizerBand.cs
aferin: it is related to an internal policy of player. that is added just for a double check.

aferin:
that operation is already applied in native API.
(http://suprem.sec.samsung.net/jira/browse/ACR-1420)

from what i heard.. when the audio offload is enabled, all APIs that i've fixed will not work.

myroot:
I'm wondering it is really need to check the audio offload state before using the audio related API.
It cause some performance degradation

aferin:
if audio offload is enabled, those APIs won't work as an user expects.
so we should let the user know it. if error isn't returned, the user could think APIs work well.

myroot:
> 
> 
> if audio offload is enabled, those APIs won't work as an user expects.
> so we should let the user know it. if error isn't returned, the user could think APIs work well.

Why not return error code on native layer?
because, interop call is a little expensive operation

aferin:
i think it would be better that every c# API returns error code on c# layer.
ValidatePlayerState is already used for checking the state of a player in most of APIs. it also uses inter op.
and i think checking an availability after state is a bit weird.

src/Tizen.Multimedia.MediaPlayer/Player/AudioOffload.cs
```diff
@@ -47,7 +47,7 @@ internal AudioOffload(Player player)
 
         internal void CheckDisabled()
         {
-            if (Player.AudioOffload.IsEnabled)
+            if (Features.IsSupported(PlayerFeatures.AudioOffload) && Player.AudioOffload.IsEnabled)
```
hsgwon: When `AudioOffload feature` is not supported, CheckDisabled() returns `InvalidOperationException` with 'offload is enabled' message, not `NotSupportedException`.
Is it proper exception and message?

aferin: i put this code to check if target has the AudioOffload feature and it is enabled at the same time.
when CheckDisabled is called, it doesnt mean an user calls AudioOffload API directly.
CheckDisabled is called in other Audio APIs to use them.
let's say an user wants to call SetPlaybackRate. even if that target doesnt have the feature, SetPlaybackRate has to be called without any problems.
otherwise, if AudiOOffload exists and is enabled the user can't use this API.

aferin:
IsEnabled can be set only in idle state. so calling inter op is not a necessary. i'll upload a new patch soon.

src/Tizen.Multimedia.MediaPlayer/Player/AudioOffload.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// <since_tizen> 6 </since_tizen>
+    public class AudioOffload
+    {
+        private IList<MediaFormatAudioMimeType> _supportedFormat;
+
+        /// <summary>
+        /// Gets the <see cref="Multimedia.Player"/> that owns this instance.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public Player Player { get; }
```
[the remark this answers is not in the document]
aferin: no i'll change it to internal.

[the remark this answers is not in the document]
hsgwon: Actually, `private` or `private readonly` is enough, if there's no reference in the same assembly.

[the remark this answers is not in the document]
aferin: ok i'll fix it.

src/Tizen.Multimedia.MediaPlayer/Player/AudioOffload.cs
```diff
@@ -0,0 +1,188 @@
...
+                _enabled = value;
+            }
+        }
+
+        /// <summary>
+        /// Get a state whether or not the audio offload is activated.
+        /// </summary>
+        /// <value>The value indicating whether or not AudioOffload is activated.</value>
+        /// <remarks>
+        /// Audio offload could be inactivated depending on the audio device capability even if the audio offload feature is supported.
```
journey2w: Is "even if" correct in this sentence?
Reference:  https://www.englishgrammar.org/even-if-and-even-though/

What about the following sentence?
        /// It depends on the audio device capability whether audio offload can activated even though the audio offload feature is supported.

src/Tizen.Multimedia.MediaPlayer/Player/AudioOffload.cs
```diff
@@ -0,0 +1,188 @@
...
+                _player.ValidatePlayerState(PlayerState.Ready, PlayerState.Playing, PlayerState.Paused);
+
+                NativePlayer.IsAudioOffloadActivated(_player.Handle, out var value).
+                    ThrowIfFailed(_player, "Failed to get whether the audio offload of the player is enabled or not");
+                return value;
+            }
+        }
+
+        /// <summary>
+        /// Retrieves all formats for audio offload.
```
journey2w: Does this sentence show your intention?

What about the following sentence?
 /// Retrieves the supported audio formats by audio offload.
or
 /// Retrieves the supported audio formats for audio offload.

src/Tizen.Multimedia.MediaPlayer/Player/AudioOffload.cs
```diff
@@ -0,0 +1,188 @@
...
+        /// </summary>
+        /// <returns>
+        /// It returns a list contained all formats for audio offload.
```
journey2w: What about the following sentence?

/// It returns a list containing supported audio formats by audio offload.
or
/// It returns a list containing supported audio formats for audio offload.

src/Tizen.Multimedia.MediaPlayer/Player/EqualizerBand.cs
```diff
@@ -90,14 +87,42 @@ public int Level
         /// <summary>
         /// Gets the frequency in the dB.
```
journey2w: Suggestion:

/// Gets the frequency in dB.

src/Tizen.Multimedia.MediaPlayer/Player/EqualizerBand.cs
```diff
@@ -90,14 +87,42 @@ public int Level
...
+
+                Native.GetEqualizerBandFrequency(_owner.Player.Handle, _index, out var frequency).
+                ThrowIfFailed(_owner.Player, "Failed to initialize equalizer band");
+
+                return frequency;
+            }
+        }
 
         /// <summary>
         /// Gets the frequency range in the dB.
```
journey2w: Suggestion: 

/// Gets the frequency range in dB.

src/Tizen.Multimedia.MediaPlayer/Player/AudioOffload.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        /// <summary>
+        /// Enables or disables the audio offload.
+        /// </summary>
+        /// <value>The value indicating whether or not AudioOffload is enabled. The default value is false.</value>
+        /// <remarks><para>The player lets the hardware decode and render the sound if the audio offload is enabled.
+        /// This will reduce the power consumption, but will disable the ability to handle output PCM.
+        /// Please check the below list of functions which will not work if offloading is enabled.</para>
+        /// <para>If audio offload is enabled, the following functions will return <see cref="InvalidOperationException"/>
+        /// and they will not work at all even if they were called before offload is enabled.<br/>
```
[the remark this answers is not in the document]
aferin: the first sentence looks quite different. i'll update descriptions following your suggestion except it.

PR #976, opened 2019-08-14
[AudioManager] Add 'Network' audio device type

sangchul1011:

### Description of Change ###
Add Network type to Tizen.Multimedia.AudioDeviceType enumeration.

### API Changes ###
 - ACR: TCSACR-263

### DIFF
--- a/src/Tizen.Multimedia/AudioManager/AudioDeviceType.cs
+++ b/src/Tizen.Multimedia/AudioManager/AudioDeviceType.cs
@@ -65,6 +65,11 @@ namespace Tizen.Multimedia
         /// <summary>
         /// Bluetooth voice (SCO).
         /// </summary>
-        BluetoothVoice
+        BluetoothVoice,
+
+        /// <summary>
+        /// Network.
+        /// </summary>
+        Network
     }
 }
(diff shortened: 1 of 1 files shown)


src/Tizen.Multimedia/AudioManager/AudioDeviceType.cs
```diff
@@ -65,6 +65,11 @@ public enum AudioDeviceType
         /// <summary>
         /// Bluetooth voice (SCO).
         /// </summary>
-        BluetoothVoice
+        BluetoothVoice,
+
+        /// <summary>
+        /// Network.
```
myroot: Description is too short to understand,
What kind of network protocol was used?

sangchul1011: This device type will be used when ACM service which is from VD/MCD/public spk profile is activated internally. One another example, PulseAudio - tizen audio backend - supports RAOP protocol and we can use the protocol only with internal test application.(for now, not sure about exporting public API). It would be represented via this new device type.

You can easily think this kind of device is used when audio data is transmitted via wifi network.

myroot: Would you please update the description to summarize this?

myroot: It is android reference for you
https://developer.android.com/reference/android/media/AudioDeviceInfo#TYPE_IP

PR #977, opened 2019-08-19
[Tizen.Applications.ComponentBased][TCSACR-265][Add] Add ComponentBased application APIs

gotoextreme:
Add ComponentBased application APIs

API Changes
ACR: TCSACR-265

BaseComponent.cs
+ API signature
[namespace]
+ Tizen.Applications.ComponentBased.Common
[class]
BaseComponent
[property]
+ DeviceOrientationChanged <<event>>
+ LocaleChanged <<event>>
+ LowBattery <<event>>
+ LowMemory <<event>>
+ RegionFormatChanged <<event>>
+ SuspendedStateChanged <<event>>
[method]
+ Finish()
+ GetComponentType() : ComponentType
+ OnRestoreContents(c)
+ OnSaveContent(c)
+ SendLaunchRequestAsync(control, replyAfterLaunching) : Task<AppControlResult>
[enum]
ComponentType.Frame
ComponentType.Service
 

FrameComponent.cs
[namespace]
+ Tizen.Applications.ComponentBased.Common
[class]
FrameComponent
[property]
+ CurrentDisplayStatus : DisplayStatus <<readonly>>
+ Window : IWindow <<readonly>>
[method]
+ GetComponentType() : ComponentType
+ OnCreate() : IWindow
+ OnDestroy()
+ OnPause()
+ OnResume()
+ OnStart(appControl, restarted)
+ OnStop()
ServiceComponent.cs
[namespace]
+ Tizen.Applications.ComponentBased.Common
[class]
ServiceComponent
[method]
+ GetComponentType() : ComponentType
+ OnCreate() : bool
+ OnDestroy()
+ OnStartCommand(appControl, restarted)
 

CBApplicationBase.cs
[namespace]
+ Tizen.Applications.ComponentBased.Common
[class]
CBApplicationBase
[method]
Dispose(disposing)
OnExit()
OnFini()
OnInit(args)
OnRun()
+ CBApplicationBase(typeInfo)
+ Exit()
+ RegisterComponent(compType, compId)
+ Run(args)
 
 

IWindow.cs
[namespace]
+ Tizen.Applications.ComponentBased.Common
[class]
IWindow
[method]
{abstract} + GetRaw() : System.IntPtr
{abstract} + GetResId() : int
 

CBBasicApplication.cs
[namespace]
+ Tizen.Applications.ComponentBased.Default
[class]
CBBasicApplication
[method]
+ CBBasicApplication(typeInfo)
OnExit()
OnFini()
OnInit(args)
OnRun()
 

ElmWindow.cs
[namespace]
+ Tizen.Applications.ComponentBased.Default
[class]
ElmWindow
[method]
+ ElmWindow(win)
+ GetRaw() : System.IntPtr
+ GetResId() : int


src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/ElmWindow.cs
```diff
@@ -0,0 +1,55 @@
...
+using System;
+using Tizen.Applications.ComponentBased.Common;
+
+namespace Tizen.Applications.ComponentBased.Default
+{
+    /// <summary>
+    /// Window class for basic application type
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class ElmWindow : IWindow
```
myroot: X DO NOT use abbreviations or contractions as part of identifier names.

```suggestion
    public class ElementaryWindow: IWindow
```
```suggestion
    public class EFLWindow: IWindow
```

myroot: Did you need create a new window class?
Why not use ElmSharp.Window?

src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/ElmWindow.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            _win = win;
+        }
+
+        /// <summary>
+        /// Gets the raw handle of window
+        /// </summary>
+        /// <returns>The native handle of window</returns>
+        /// <since_tizen> 6 </since_tizen>
+        public IntPtr GetRaw()
```
myroot:
```suggestion
        public IntPtr GetHandle()
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/IWindow.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// Gets window resource ID
+        /// </summary>
+        /// <returns></returns>
+        int GetResId();
+
+        /// <summary>
+        /// Gets window raw pointer
+        /// </summary>
+        /// <returns></returns>
+        IntPtr GetRaw();
```
myroot: `Handle` is used to represent a native handle
and I recomment to use Property
```suggestion
        IntPtr Handle {get;}
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/IWindow.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// Interface for window
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public interface IWindow
+    {
+        /// <summary>
+        /// Gets window resource ID
+        /// </summary>
+        /// <returns></returns>
+        int GetResId();
```
myroot: I recomment to use property
```suggestion
        int ResourceId {get;}
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
```diff
@@ -0,0 +1,186 @@
...
+    public abstract class BaseComponent
+    {
+        internal IntPtr Handle;
+        private const string LogTag = "Tizen.Applications.BaseComponent";
+
+        /// <summary>
+        /// Component types.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public enum ComponentType
```
myroot: Do not use nested enum type

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
```diff
@@ -0,0 +1,186 @@
...
+            ComponentId = compId;
+            Parent = parent;
+        }
+
+        /// <summary>
+        /// Overrides this method if you want to specify a type of this component.
+        /// Default component type is Service type.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public virtual ComponentType GetComponentType()
```
myroot: Why don't you use Property?
```suggestion
        public virtual ComponentType ComponentType { get; }
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseType.cs
```diff
@@ -0,0 +1,173 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Tizen.Applications.CoreBackend;
+using static Interop.CBApplication;
+using static Tizen.Applications.CoreBackend.DefaultCoreBackend;
+
+namespace Tizen.Applications.ComponentBased.Common
+{
+    internal class BaseType
```
myroot: Naming is not good, it is too ambiguous.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/CBApplicationBase.cs
```diff
@@ -0,0 +1,206 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Tizen.Applications.ComponentBased.Common
+{
+    /// <summary>
+    /// The class for supporting multi-components based application model.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class CBApplicationBase : Application
```
myroot: X DO NOT use abbreviations or contractions as part of identifier names.
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions

```suggestion
    public class ComponentBasedApplication : Application
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/CBApplicationBase.cs
```diff
@@ -0,0 +1,206 @@
...
+        public CBApplicationBase(IDictionary<Type, string> typeInfo)
+        {
+            _callbacks.OnInit = new Interop.CBApplication.CBAppInitCallback(OnInitNative);
+            _callbacks.OnFini = new Interop.CBApplication.CBAppFiniCallback(OnFiniNative);
+            _callbacks.OnRun = new Interop.CBApplication.CBAppRunCallback(OnRunNative);
+            _callbacks.OnExit = new Interop.CBApplication.CBAppExitCallback(OnExitNative);
+            _callbacks.OnCreate = new Interop.CBApplication.CBAppCreateCallback(OnCreateNative);
+            _callbacks.OnTerminate = new Interop.CBApplication.CBAppTerminateCallback(OnTerminateNative);
+
+            foreach (Type t in typeInfo.Keys)
```
myroot:
```suggestion
            foreach (var component in typeInfo) 
            {
                 RegisterComponent(component.Key, component.Value);            
            }
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/CBApplicationBase.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// <since_tizen> 6 </since_tizen>
+        public void RegisterComponent(Type compType, string compId)
+        {
+            foreach (BaseType type in ComponentTypes)
+            {
+                if (compType.BaseType == type._classType)
+                    throw new InvalidOperationException("Already exist type");
+            }
+
+            if (compType.BaseType == typeof(BaseComponent))
```
myroot: Do not directly compare with type info, becuase it will fail to sub classing case
```suggestion
            if (typeof(FrameComponent).IsAssignableFrom(compType))           
            else if (typeof(ServiceComponent).IsAssignableFrom(compType))
            else if (typeof(BaseComponent).IsAssignableFrom(compType))
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseType.cs
```diff
@@ -0,0 +1,173 @@
...
+    {
+        internal readonly Type _classType;
+        internal readonly string _compId;
+        internal IList<BaseComponent> _compInstances = new List<BaseComponent>();
+        private Interop.CBApplication.BaseLifecycleCallbacks _callbacks;
+        private BaseComponent.ComponentType _compType;
+        private CBApplicationBase _parent;
+
+        internal BaseType(Type ctype, string id, BaseComponent.ComponentType comp_type, CBApplicationBase parent)
```
myroot: What is the responsibility of this class?
Implementation and naming is not matched

Implementations looks like a data of some runtime instance information
But, naming of `Type` is a static information. 

and I don't know why need it, it should be a part of BaseComponent

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
```diff
@@ -0,0 +1,186 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Tizen.Applications.ComponentBased.Common
+{
+    /// <summary>
+    /// This is a base class for components
```
myroot: Please write more details information about components

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// It will be created after OnCreate method is invoked.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public string Id { get; internal set; }
+
+        /// <summary>
+        /// A component ID
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public string ComponentId { get; internal set; }
```
myroot: Why not private?
```suggestion
        public string ComponentId { get; private set; }
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public event EventHandler<SuspendedStateEventArgs> SuspendedStateChanged;
+
+        /// <summary>
+        /// A component instance ID.
+        /// It will be created after OnCreate method is invoked.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public string Id { get; internal set; }
```
myroot: Who is set Id ?

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponent.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// The class for showing UI module
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public abstract class FrameComponent : BaseComponent
+    {
+        /// <summary>
+        /// Enumeration for display status.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public enum DisplayStatus
```
myroot: Do not use nested enum type

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponent.cs
```diff
@@ -0,0 +1,115 @@
...
+            /// </summary>
+            Off = Interop.CBApplication.NativeDisplayStatus.Off
+        }
+
+        /// <summary>
+        /// Gets the display status of a component.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Thrown when component type is already added to the component.</exception>
+        /// <since_tizen> 6 </since_tizen>
+        public DisplayStatus CurrentDisplayStatus
```
myroot:
```suggestion
        public DisplayStatus DisplayStatus
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponent.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public IWindow Window { get; internal set; }
+
+        /// <summary>
+        /// Overrides this method if want to handle behavior when the component is launched.
+        /// </summary>
+        /// <returns>Window object to use</returns>
+        /// <since_tizen> 6 </since_tizen>
+        public virtual IWindow OnCreate()
```
myroot: It should be abstract
```suggestion
        public abstract IWindow OnCreate();
```

src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/CBBasicApplication.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Collections.Generic;
+using Tizen.Applications.ComponentBased.Common;
+
+namespace Tizen.Applications.ComponentBased.Default
+{
+    /// <summary>
+    /// Basic type application which will support ElmWindow for FrameComponent
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class CBBasicApplication : CBApplicationBase
```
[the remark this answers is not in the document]
gotoextreme: Done

src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/CBBasicApplication.cs
```diff
@@ -0,0 +1,83 @@
+﻿using ElmSharp;
+using System;
+using System.Collections.Generic;
+using Tizen.Applications.ComponentBased.Common;
+
+namespace Tizen.Applications.ComponentBased.Default
+{
+    /// <summary>
+    /// Basic type application which will support ElmWindow for FrameComponent
```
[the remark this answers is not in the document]
gotoextreme: Done

src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/CBBasicApplication.cs
```diff
@@ -0,0 +1,83 @@
...
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class CBBasicApplication : CBApplicationBase
+    {
+        private const string LogTag = "Tizen.Applications.CBBasicApplication";
+
+        /// <summary>
+        /// Initializes the CBBasicApplication class.
+        /// </summary>
+        /// <param name="typeInfo">The compnent type information.</param>
```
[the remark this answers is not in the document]
gotoextreme: Done

src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/ElmWindow.cs
gotoextreme: Done

gotoextreme: We need a common interface for handling window information.
It can be ElmSharp.Window, Dali window and etc..
ComponentBased application base provides a common interface for handling window information to remove EFL dependency.

src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/ElmWindow.cs
gotoextreme: Removed

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/IWindow.cs
gotoextreme: Removed

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/IWindow.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseType.cs
gotoextreme: I changed it into BaseFactory.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/CBApplicationBase.cs
gotoextreme: It is a base class of other applications, therefore I think ComponentBasedApplicationBase is a proper name.
I changed it into ComponentBasedApplicationBase

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/CBApplicationBase.cs
gotoextreme: done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/CBApplicationBase.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseType.cs
gotoextreme: This is an Abstract Factory pattern.
BaseType is responsible for creating base-component instances.
And BaseComponent class is a class that defines the behavior of base-component instance.  
By dividing these two behaviors, it is easier to maintain.
But, the name of this class is confusing. Therefore, let's change a name into BaseFactory.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
gotoextreme: BaseFactory makes a BaseComponent Instance and assigns an ID.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponent.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponent.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponent.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplicationBase.cs
```diff
@@ -0,0 +1,208 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Tizen.Applications.ComponentBased.Common
+{
+    /// <summary>
+    /// The class for supporting multi-components based application model.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class ComponentBasedApplicationBase : Application
```
myroot: Why do you need this middle layer between `Application` and `ComponentBasedApplication`?
This class only has one subclass(ComponentBasedApplication)

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplicationBase.cs
```diff
@@ -0,0 +1,208 @@
...
+        private bool _disposedValue = false;
+
+        /// <summary>
+        /// Initializes the ComponentBasedApplicationBase class.
+        /// </summary>
+        /// <param name="typeInfo">The component type information.
+        /// The key should be a class type of BaseComponent subclass.
+        /// The value should be a component id which is declared in tizen-manifest.xml.
+        /// </param>
+        /// <exception cref="OutOfMemoryException">Thrown when failed because of out of memory.</exception>
```
myroot: Do not throw OutOfMemoryException on your code. 

https://docs.microsoft.com/ko-kr/dotnet/standard/design-guidelines/using-standard-exception-types#outofmemoryexception

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplicationBase.cs
```diff
@@ -0,0 +1,208 @@
...
+            else if (typeof(FrameComponent).IsAssignableFrom(compType))
+            {
+                Log.Info(LogTag, "Add frame comp");
+                _componentFactories.Add(new FrameFactory(compType, compId, this));
+            }
+            else if (typeof(ServiceComponent).IsAssignableFrom(compType))
+            {
+                Log.Info(LogTag, "Add service comp");
+                _componentFactories.Add(new ServiceFactory(compType, compId, this));
+            }
```
myroot: You should throw exception if comType is not subtype of `BaseComponent`

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Text;
+using Tizen.Applications.CoreBackend;
+using static Interop.CBApplication;
+using static Tizen.Applications.CoreBackend.DefaultCoreBackend;
+
+namespace Tizen.Applications.ComponentBased.Common
+{
+    internal abstract class ComponentFactoryBase
+    {
+        internal readonly Type _classType;
```
myroot: Do not directly open a field. 

If you need to open this information you should provide as property

```suggestion
        public Type ComponentType { get; }
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplicationBase.cs
```diff
@@ -0,0 +1,208 @@
...
+        /// </summary>
+        /// <param name="compType">Class type</param>
+        /// <param name="compId">Component ID</param>
+        /// <exception cref="InvalidOperationException">Thrown when component type is already added.</exception>
+        /// <since_tizen> 6 </since_tizen>
+        public void RegisterComponent(Type compType, string compId)
+        {
+            foreach (ComponentFactoryBase factory in _componentFactories)
+            {
+                if (compType.BaseType == factory._classType)
```
myroot: If it does not allow a same type, you do better to use `Dictionary<Type, ComponentFactoryBase>` instead of `List<ComponentFactoryBase>`

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseFactory.cs
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+using static Interop.CBApplication;
+
+namespace Tizen.Applications.ComponentBased.Common
+{
+    internal class BaseFactory : ComponentFactoryBase
```
myroot: Naming is too common, 
What is purpose of this class?
And Why do you need `base` naming?

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
```diff
@@ -0,0 +1,131 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Tizen.Applications.CoreBackend;
+using static Interop.CBApplication;
+using static Tizen.Applications.CoreBackend.DefaultCoreBackend;
+
+namespace Tizen.Applications.ComponentBased.Common
+{
+    internal abstract class ComponentFactoryBase
```
myroot: I think `BaseComponentFactory` is better

BaseComponentFactory
FrameComponentFactory : BaseComponentFactory
ServiceComponentFactory : BaseComponentFactory

Because, class hierarchy should be matched
BaseComponent
FrameComponent : BaseComponent
ServiceComponent : BaseComponent

myroot: And this class is not factory
It violation of single responsibility principle

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplicationBase.cs
```diff
@@ -0,0 +1,208 @@
...
+                    throw new InvalidOperationException("Already exist type");
+            }
+
+            if (typeof(BaseComponent).IsAssignableFrom(compType))
```
myroot: It is wrong. 
You should check FrameComponent and ServiceComponent before BaseComponent. 

FrameComponent and ServiceComponent also assignable to BaseComponent
So, your code does work, It always add to base comp

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplicationBase.cs
gotoextreme: ComponentBasedApplicationBase has to manage a totally different lifecycle with Application which is needed by FrameComponent, ServiceComponent etc..

And ComponentBasedApplication is EFL base application. Dali base application can be added in the future.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
gotoextreme: Why do you think it is not an Abstract Factory pattern?
In my opinion, it can be an Abstract Factory pattern's factory.
It provides Bind function so that the ComponentBasedApplication can create instances of Components.
This class allows ComponentBasedApplication to create concreate FrameComponent or ServiceComponent without dependency.

> 
https://en.wikipedia.org/wiki/Abstract_factory_pattern

The client doesn't know (or care) which concrete objects it gets from each of these internal factories, since it uses only the generic interfaces of their products

> 

And about single responsibility, if you agree that it can be an Abstract Factory pattern,
this class does not violate the single responsibility since it providing a contract for the creation of polymorphic objects without binding the interface to a specific implementation.

https://stackoverflow.com/questions/46790893/abstract-factory-violate-single-responsibility-principle

myroot: It violation of single responsibility principle
Absract Factory class only create a some object, but your code dispatch event

myroot: I never seen Factory pattern without create method.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseFactory.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        private ComponentBasedApplicationBase _parent;
+        private Interop.CBApplication.BaseLifecycleCallbacks _callbacks;
```
myroot: Who is create _callbacks?

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseFactory.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        internal BaseFactory(Type ctype, string id, ComponentType compType, ComponentBasedApplicationBase parent) : base(ctype, id, compType, parent)
+        {
+            _callbacks.OnDeviceOrientationChanged = new Interop.CBApplication.BaseDeviceOrientationChangedCallback(OnDeviceOrientationChangedCallback);
```
myroot: `_callbacks` is accessed before create, did you test it?

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplicationBase.cs
myroot: As your comment. 
ComponentBasedApplicationBase -> ComponentBasedApplication 
ComponentBasedApplication  -> EFLComponentBasedApplication

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
pjh9216: It look like state pattern.  However, I don't know why it violate SRP. BaseComponentFactory just handles events. Because it is internal classes, it is not relate to ACR. why don't you discuss it later?

src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/ElmWindow.cs
myroot: Hm. but, ElmWindow has EFL dependency
I also agree create IWindow interface to remove UI F/W dependency. 
but, `ElmWindow` could be sub class of ElmSharp.Window
``` c# 
 public class ElmWindow : ElmSharp.Window, IWindow {
```

gotoextreme: Yes, but we only need resource ID of the window. Therefore, I changed it into ElmWindowInfo.
I think we don't need to inherit ElmSharp.Window.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplicationBase.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplicationBase.cs
gotoextreme: Done.
I added else statement and throw an exception.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplicationBase.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplicationBase.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/EFLWindowInfo.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <since_tizen> 6 </since_tizen>
+        public int ResourceId
+        {
+            get
+            {
+                if (_resId == 0)
+                {
+                    int err = Interop.EflCBApplication.GetResourceId(_win.RealHandle, out _resId);
+                    if (err != 0)
+                        Log.Info(LogTag, "elm_config_accel_preference_set is not called");
```
myroot: Is correct message?

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// <exception cref="ArgumentException">Thrown when failed because of the argument is invalid.</exception>
+        /// <exception cref="Exceptions.AppNotFoundException">Thrown when the application to run is not found.</exception>
+        /// <exception cref="Exceptions.LaunchRejectedException">Thrown when the launch request is rejected.</exception>
+        /// <privilege>http://tizen.org/privilege/appmanager.launch</privilege>
+        /// <since_tizen> 6 </since_tizen>
+        public Task<AppControlResult> SendLaunchRequestAsync(AppControl control, AppControlReplyCallback replyAfterLaunching)
+        {
+            int ret = Interop.AppControl.SetCallerInstanceId(control.SafeAppControlHandle, Id);
+            if (ret != 0)
+                throw new ArgumentException("Failed to set id");
```
myroot: Not correct exception, id is not argument on this api

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
myroot: Anyway, it is not factory, please rename

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
```diff
@@ -0,0 +1,209 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Tizen.Applications.ComponentBased.Common
+{
+    /// <summary>
+    /// The class for supporting multi-components based application model.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class ComponentBasedApplication : Application
```
myroot: How about consider change to abstract class?

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
```diff
@@ -0,0 +1,209 @@
...
+        /// <since_tizen> 6 </since_tizen>
+        protected virtual void OnFinished()
+        {
+        }
+
+        /// <summary>
+        /// This method will be called to start main-loop
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        protected virtual void OnRun()
```
myroot: It should be abstract, 
because if not override this method, application is not working

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
```diff
@@ -0,0 +1,209 @@
...
+        /// </summary>
+        /// <param name="compType">Class type</param>
+        /// <param name="compId">Component ID</param>
+        /// <exception cref="InvalidOperationException">Thrown when component type is already added.</exception>
+        /// <since_tizen> 6 </since_tizen>
+        public void RegisterComponent(Type compType, string compId)
+        {
+            if (_componentFactories.ContainsKey(compType))
+            {
+                throw new InvalidOperationException("Already exist type");
```
myroot: I think, ArgumentException is more correct, 
because, argument is cause error

in Dictionary, 
ArgumentException
An element with the same key already exists in the Dictionary<TKey,TValue>.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
```diff
@@ -0,0 +1,209 @@
...
+            }
+            else if (typeof(BaseComponent).IsAssignableFrom(compType))
+            {
+                Log.Info(LogTag, "Add base component");
+                BaseComponent b = Activator.CreateInstance(compType) as BaseComponent;
+                _componentFactories.Add(compType, new BaseFactory(compType, compId, b.ComponentType, this));
+            }
+            else
+            {
+                throw new ArgumentException("Invalid type");
```
myroot:
```suggestion
                throw new ArgumentException("compType must be sub type of BaseComponent", "compType");
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
```diff
@@ -0,0 +1,127 @@
...
+    {
+        public IList<BaseComponent> ComponentInstances { get; }
```
myroot: Do not directly expose List

if you want to provide way to adding on the ComponentInstances, 

You can provide a method to add
``` c#
  private IList<BaseComponent> ComponentInstances { get; }
  public IEnumerable<BaseComponent> Instances => ComponentInstances;
  protected void AddComponent(BaseComponent comp) {
      ComponentInstances.Add(comp);  
  }

  protected void RemoveComponent(BaseComponent comp) {
      ComponentInstances.Remove(comp);  
  }
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
```diff
@@ -0,0 +1,127 @@
...
+        public Type ComponentClassType { get; }
+        public ComponentType ComponentType { get; }
+        public string ComponentId { get; }
+        public ComponentBasedApplication Parent { get; }
```
myroot: It seem to circular dependency, 

BaseComponentFactoryBase -> ComponentBaseApplication
 :  BaseComponentFactoryBase.Parent 

ComponentBaseApplication -> BaseComponentFactoryBase 
 :  ComponentBasedApplication._componentFactories

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ServiceFactory.cs
```diff
@@ -0,0 +1,86 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Tizen.Applications.ComponentBased.Common
+{
+    internal class ServiceFactory : ComponentFactoryBase
+    {
+        private Interop.CBApplication.ServiceLifecycleCallbacks _callbacks;
+        private ComponentBasedApplication _parent;
```
myroot: It already provide on base class
`public ComponentBasedApplication Parent { get; }`

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
gotoextreme: Well, I thought Bind function create objects indirectly, and thus it could be an abstract factory pattern. Also, native-c implementation using a standard factory pattern for this feature, so it made me confusing I guess.

Anyway, I will fix it. How about BaseComponentStateManger than?

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
```diff
@@ -0,0 +1,209 @@
...
+        private IntPtr OnCreateNative(IntPtr data)
+        {
+            Log.Debug(LogTag, "On create");
+            IntPtr h = IntPtr.Zero;
+            foreach(KeyValuePair<Type, ComponentFactoryBase> entry in _componentFactories)
+            {
+                h = entry.Value.Bind(h);
+            }
+
+            return h;
```
myroot: What means of return value?
It is only returning a last returned value from Bind
Is that ok?

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
```diff
@@ -0,0 +1,209 @@
...
+        public override void Exit()
+        {
+            Interop.CBApplication.BaseExit();
+        }
+
+        private IntPtr OnCreateNative(IntPtr data)
+        {
+            Log.Debug(LogTag, "On create");
+            IntPtr h = IntPtr.Zero;
+            foreach(KeyValuePair<Type, ComponentFactoryBase> entry in _componentFactories)
```
myroot:
```suggestion
            foreach (KeyValuePair<Type, ComponentFactoryBase> entry in _componentFactories)
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
myroot: Yes, `BaseComponentStateManager` is goog to understand. 
FrameComponentStateManager : ComponentStateManger
ServiceComponentStateManager : ComponentStateManger

And I want to match a hierarchy between Component and this

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponent.cs
```diff
@@ -0,0 +1,98 @@
...
+
+                return (DisplayStatus)status;
+            }
+        }
+
+        /// <summary>
+        /// Gets the frame component's window information.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public IWindowInfo WindowInfo { get; internal set; }
```
myroot: How change like below, current code is violation of encapsulation

``` c#
        public IWindowInfo WindowInfo { get; private set; }
        public void OnCreate()
        {
            WindowInfo = CreateWindowInfo();
        }
        public abstract IWindowInfo CreateWindowInfo();

     // caller
     fc.OnCreate();
     if (fc.WindowInfo == null)
                return IntPtr.Zero;
     Interop.CBApplication.BaseFrameCreateWindow(out winHandle, fc.WindowInfo.ResourceId, IntPtr.Zero);
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
```diff
@@ -0,0 +1,209 @@
...
+            {
+                Log.Info(LogTag, "Add frame component");
+                _componentFactories.Add(compType, new FrameFactory(compType, compId, this));
+            }
+            else if (typeof(ServiceComponent).IsAssignableFrom(compType))
+            {
+                Log.Info(LogTag, "Add service component");
+                _componentFactories.Add(compType, new ServiceFactory(compType, compId, this));
+            }
+            else if (typeof(BaseComponent).IsAssignableFrom(compType))
```
myroot: I don't understadn why do you allow create of BaseComponent?
I think, Component should be one of FrameComponent or ServiceComponent
because ComponentType is only exist Frame or Service
``` c#
    public enum ComponentType
    {
        /// <summary>
        /// The frame component.
        /// </summary>
        Frame = Interop.CBApplication.NativeComponentType.Frame,

        /// <summary>
        /// The service component.
        /// </summary>
        Service = Interop.CBApplication.NativeComponentType.Service
    }
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
```diff
@@ -0,0 +1,189 @@
...
+            get
+            {
+                return ComponentType.Service;
+            }
+        }
+
+        /// <summary>
+        /// Overrides this method if want to handle behavior to restore the previous status.
+        /// </summary>
+        /// <param name="c">Contents</param>
```
myroot: Need to remarks about life time of Bundle

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
gotoextreme: Yes, returned value contains native-handle that contains native-component-factory map.
Bind function adds native-component-factory to the handle, therefore returned value contains multiple native-component-factory.

For further information, you can see native code here. https://review.tizen.org/gerrit/#/admin/projects/platform/core/appfw/component-based-application

Bind function will call native API component_based_app_base_add_frame_component.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
gotoextreme: What do you mean to match a hierarchy between Component and this?
Isn't it already have the same hierarchy? Do I have to change something?

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponent.cs
gotoextreme: Hmm... this way, we cannot explicitly force to make a window in OnCreate callback.
Maybe we need specific documentation.

But, as you say current code violate encapsulation so, I will discuss it with scrum product owner @pjh9216
He is on vacation until next Monday.

myroot: > Hmm... this way, we cannot explicitly force to make a window in OnCreate callback.
Why do you think so?
`public void OnCreate()` is not virtual, so can't modify on subclass
And CreateWindowInfo method was newly added and it inforce creation of WindowInfo
`public abstract IWindowInfo CreateWindowInfo();`

myroot: If you want to provide the way to add some code when created, 
You can provide a event like `Created`

myroot:
I have question, 
We regist a component with below API
`public void RegisterComponent(Type compType, string compId)`

Is will be created more than one a `compType` instance on single application process?

gotoextreme:
> I have question,
> We regist a component with below API
> `public void RegisterComponent(Type compType, string compId)`
> 
> Is it will be created more than one a `compType` instance on single application process?

Yes, multiple instances can be created based on the `compType` on a single application process.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
```diff
@@ -0,0 +1,189 @@
...
+    /// <summary>
+    /// This is a base-component class. The base-component includes only a few basic overridable lifecycle functions.
+    /// If you want to make a new component that different with FrameComponent or ServiceComponent,
+    /// you can inherit this class and implement your own lifecycle.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public abstract class BaseComponent
+    {
+        internal IntPtr Handle;
+        private const string LogTag = "Tizen.Applications.BaseComponent";
```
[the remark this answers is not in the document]
gotoextreme: Removed

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
gotoextreme: I updated codes. Now, component-based applications cannot create BaseComponent instance.
Only sub-class of FrameComponent and ServiceComponent can be registered by component-based applications.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponent.cs
gotoextreme: CreateWindowInfo is added and OnCreate does not return IWindow now.
Also, WindowInfo property is removed. The WindowInfo property is unnecessary.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ServiceFactory.cs
gotoextreme: Removed

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
gotoextreme: I changed it into protected.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentFactoryBase.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
gotoextreme: Changed it into InvalidOperationException

src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/EFLWindowInfo.cs
gotoextreme: Fixed

src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/EFLWindowInfo.cs
```diff
@@ -0,0 +1,49 @@
...
+        private const string LogTag = "Tizen.Applications.EFLWindow";
+        private EvasObject _win;
+        private int _resId;
+
+        /// <summary>
+        /// Initializes the EFLWindow class.
+        /// </summary>
+        /// <param name="win">The window object of component.</param>
+        /// <since_tizen> 6 </since_tizen>
+        public EFLWindowInfo(EvasObject win)
```
myroot: Why not use `Window` type? `EvasObject` could not be `Window`
Are you don't care is it not window?
```suggestion
        public EFLWindowInfo(Window win)
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// <since_tizen> 6 </since_tizen>
+        public override void Exit()
+        {
+            Interop.CBApplication.BaseExit();
+        }
+
+        private IntPtr OnCreateNative(IntPtr data)
+        {
+            Log.Debug(LogTag, "On create");
+            IntPtr h = IntPtr.Zero;
```
myroot: I hope to use a readable and understandable variable naming
```suggestion
            IntPtr nativeComponentFactoryMap = IntPtr.Zero;
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// </summary>
+        /// <param name="disposing">If true, disposes any disposable objects. If false, does not dispose disposable objects.</param>
+        /// <since_tizen> 6 </since_tizen>
+        protected override void Dispose(bool disposing)
+        {
+            if (!_disposedValue)
+            {
+                _disposedValue = true;
+            }
+            base.Dispose(disposing);
```
myroot:
``` c#
            if (!_disposedValue)
            {
                _disposedValue = true;
                base.Dispose(disposing);
            }
```
or
no need to override it

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponentStateManager.cs
```diff
@@ -0,0 +1,130 @@
...
+            if (fc == null)
+                return IntPtr.Zero;
+
+            string id;
+            Interop.CBApplication.GetInstanceId(context, out id);
+            fc.Bind(context, ComponentId, id, Parent);
+
+            IntPtr winHandle;
+            IWindowInfo win = fc.CreateWindowInfo();
+            if (win == null)
```
myroot: How about add a log message?

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponentStateManager.cs
```diff
@@ -0,0 +1,130 @@
...
+                return IntPtr.Zero;
+
+            bool result = fc.OnCreate();
+            if (!result)
+            {
```
myroot: How about add a log message?

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponentStateManager.cs
```diff
@@ -0,0 +1,130 @@
...
+            Interop.CBApplication.GetInstanceId(context, out id);
+            fc.Bind(context, ComponentId, id, Parent);
+
+            IntPtr winHandle;
+            IWindowInfo win = fc.CreateWindowInfo();
+            if (win == null)
+                return IntPtr.Zero;
+
+            bool result = fc.OnCreate();
+            if (!result)
```
myroot: result is only used on `if` so, we can remove it
```suggestion
            if (!fc.OnCreate())
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ServiceComponentStateManager.cs
```diff
@@ -0,0 +1,85 @@
...
+            ServiceComponent sc = Activator.CreateInstance(ComponentClassType) as ServiceComponent;
+            if (sc == null)
+                return false;
+
+            string id;
+            Interop.CBApplication.GetInstanceId(context, out id);
+            sc.Bind(context, ComponentId, id, Parent);
+
+            bool result = sc.OnCreate();
+            if (!result)
```
myroot:
```suggestion
            if (!sc.OnCreate())
```

gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponentStateManager.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponentStateManager.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/FrameComponentStateManager.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
gotoextreme: Done

src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/EFLWindowInfo.cs
gotoextreme: Changed it into Window

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
```diff
@@ -0,0 +1,201 @@
...
+        protected virtual void OnExit()
+        {
+        }
+
+        /// <summary>
+        /// Releases any unmanaged resources used by this object. Can also dispose any other disposable objects.
+        /// </summary>
+        /// <param name="disposing">If true, disposes any disposable objects. If false, does not dispose disposable objects.</param>
+        /// <since_tizen> 6 </since_tizen>
+        protected override void Dispose(bool disposing)
```
myroot: It seem no need to override. 
becuase _disposedValue does not used

gotoextreme: Removed

PR #985, opened 2019-08-22
[MediaVision] Add APIs for inference detectors and classifier

hsgwon:
### Description of Change ###
Add new inference APIs for:
- FaceDetector
- FacialLandmarkDetector
- ImageClassfier
- ObjectDetector

### API Changes ###

 - TCSACR-268


src/Tizen.Multimedia.Vision/MediaVision/FacialLandmarkDetector.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// A task that represents the asynchronous detect operation.<br/>
+        /// If there's no detected facial landmark, the return value will be null.
+        /// </returns>
+        /// <exception cref="ArgumentNullException"><paramref name="source"/> or <paramref name="config"/> is null.</exception>
+        /// <exception cref="InvalidOperationException">Internal error.</exception>
+        /// <exception cref="NotSupportedException">The feature is not supported.</exception>
+        /// <exception cref="UnauthorizedAccessException">The caller has no required privilege.</exception>
+        /// <seealso cref="InferenceModelConfiguration"/>
+        /// <since_tizen> 6 </since_tizen>
+        public static async Task<Point[]> DetectAsync(MediaVisionSource source,
```
myroot: How about use `IEnumerable<Point>` or `IList<Point>`

src/Tizen.Multimedia.Vision/MediaVision/ImageClassificationResult.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// Gets the number of detected objects.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public int Number { get; }
+
+        /// <summary>
+        /// Gets the indices of detected objects.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public int[] Indices { get; }
```
myroot:
```suggestion
        public IList<int> Indices { get; }
```

src/Tizen.Multimedia.Vision/MediaVision/ImageClassifier.cs
```diff
@@ -0,0 +1,87 @@
...
+            }
+
+            var tcs = new TaskCompletionSource<ImageClassificationResult>();
+
+            using (var cb = ObjectKeeper.Get(GetCallback(tcs)))
+            {
+                InteropInference.ClassifyImage(source.Handle, config.GetHandle(), cb.Target).
+                    Validate("Failed to detect face.");
+
+                return await tcs.Task;
```
myroot:
```suggestion
                return tcs.Task;
```

src/Tizen.Multimedia.Vision/MediaVision/InferenceModelConfiguration.cs
```diff
@@ -0,0 +1,503 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the size of inference model's tensor.
+        /// </summary>
+        /// <remarks>Both width and height of tensor should be greater than 0.</remarks>
+        /// <exception cref="ArgumentOutOfRangeException">The value is invalid.</exception>
+        /// <since_tizen> 6 </since_tizen>
+        public Size TensorSize
```
myroot: Is it also valid on OpenCV backend?

src/Tizen.Multimedia.Vision/MediaVision/ImageClassifier.cs
hsgwon: Is there any reason for this? 
All existing vision APIs have same form.

src/Tizen.Multimedia.Vision/MediaVision/InferenceModelConfiguration.cs
hsgwon: Yes, it's the general word as I heard from native owner.

src/Tizen.Multimedia.Vision/MediaVision/ImageClassificationResult.cs
hsgwon: I refered https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1819-properties-should-not-return-arrays?view=vs-2017.
I return copied array so I think it's OK.

Is there any other reason you recommend `IList` rather than array itself?

src/Tizen.Multimedia.Vision/MediaVision/ImageClassifier.cs
hsgwon: If we return it immediately without `await`, `cb` is likely disposed by `using` before it's actually called.

src/Tizen.Multimedia.Vision/MediaVision/ImageClassificationResult.cs
myroot: ✓ DO prefer using collections over arrays in public APIs. The Collections section provides details about how to choose between collections and arrays.

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/arrays

src/Tizen.Multimedia.Vision/MediaVision/ImageClassifier.cs
myroot: Ok, you are right

src/Tizen.Multimedia.Vision/MediaVision/ImageClassificationResult.cs
hsgwon: How about `ReadOnlyCollection<T>`?

✓ DO use ReadOnlyCollection<T>, a subclass of ReadOnlyCollection<T>, or in rare cases IEnumerable<T> for properties or return values representing read-only collections.
In general, prefer ReadOnlyCollection<T>. 
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections\

myroot: Yes, `ReadOnlyCollection<T>` could be better

hsgwon:
> src/Tizen.Multimedia.Vision/MediaVision/FaceRecognitionModel.cs
> src/Tizen.Multimedia.Vision/MediaVision/FaceTrackingModel.cs
> File mode was changed but I don't care

Thank you for your review :)

src/Tizen.Multimedia.Vision/MediaVision/FaceDetector.cs
```diff
@@ -103,5 +109,89 @@ private static InteropFace.DetectedCallback GetCallback(TaskCompletionSource<Rec
...
+        /// <summary>
+        /// Detects faces on the source image using inference engine set in <paramref name="config"/>.<br/>
+        /// Each time when DetectAsync is called, a set of the detected faces at the media source are received asynchronously.
+        /// </summary>
+        /// <remarks>
+        /// If there's no detected face, empty collection will be returned.
+        /// </remarks>
+        /// <feature>http://tizen.org/feature/vision.inference</feature>
+        /// <feature>http://tizen.org/feature/vision.inference.face</feature>
+        /// <param name="source">The source of the media where faces will be detected.</param>
```
journey2w: Suggestion:

        /// <param name="source">The media source where faces are detected.</param>

src/Tizen.Multimedia.Vision/MediaVision/FacialLandmarkDetector.cs
```diff
@@ -0,0 +1,137 @@
...
+namespace Tizen.Multimedia.Vision
+{
+    /// <summary>
+    /// Provides the ability to detect facial landmarks on image source using inference engine.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public static class FacialLandmarkDetector
+    {
+        /// <summary>
+        /// Detects facial landmarks on the source image using inference engine set in <paramref name="config"/>.<br/>
```
journey2w: Suggestion:

        /// Detects facial landmarks in the image source using inference engine set in <paramref name="config"/>.<br/>

PR #987, opened 2019-08-23
[AudioManager] Add properties for the preferred device

sangchul1011:
### Description of Change ###
Add two properties to AudioStreamPolicy.
These properties are to set a specific built-in device when the system has multiple devices of the same built-in device type.

### API Changes ###
 - ACR: TCSACR-269

### DIFF
--- a/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
+++ b/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
@@ -344,18 +344,7 @@ namespace Tizen.Multimedia
             }
             set
             {
-                int deviceId;
-
-                if (value == null)
-                {
-                    deviceId = 0;
-                }
-                else
-                {
-                    deviceId = value.Id;
-                }
-
-                Interop.AudioStreamPolicy.SetPreferredDevice(Handle, AudioDeviceIoDirection.Input, deviceId).
+                Interop.AudioStreamPolicy.SetPreferredDevice(Handle, AudioDeviceIoDirection.Input, value?.Id ?? 0).
                     ThrowIfError("Failed to set preferred input device");
 
                 _inputDevice = value;
@@ -400,18 +389,7 @@ namespace Tizen.Multimedia
             }
             set
             {
-                int deviceId;
-
-                if (value == null)
-                {
-                    deviceId = 0;
-                }
-                else
-                {
-                    deviceId = value.Id;
-                }
-
-                Interop.AudioStreamPolicy.SetPreferredDevice(Handle, AudioDeviceIoDirection.Output, deviceId).
+                Interop.AudioStreamPolicy.SetPreferredDevice(Handle, AudioDeviceIoDirection.Output, value?.Id ?? 0).
                     ThrowIfError("Failed to set preferred output device");
 
                 _outputDevice = value;
--- a/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
+++ b/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
@@ -30,6 +30,7 @@ namespace Tizen.Multimedia
         private Interop.AudioStreamPolicy.FocusStateChangedCallback _focusStateChangedCallback;
         private static AudioDevice _inputDevice = null;
         private static AudioDevice _outputDevice = null;
+        private const string Tag = "Tizen.Multimedia.AudioStreamPolicy";
 
         /// <summary>
         /// Initializes a new instance of the <see cref="AudioStreamPolicy"/> class with <see cref="AudioStreamType"/>.
@@ -327,8 +328,12 @@ namespace Tizen.Multimedia
         {
             get
             {
+                /* This P/Invoke intends to validate if the core audio system
+                 * is normal. Otherwise, it'll throw an error here. */
                 Interop.AudioStreamPolicy.GetPreferredDevice(Handle, out var inDeviceId, out _).
                     ThrowIfError("Failed to get preferred input device");
+
+                Log.Debug(Tag, $"preferred input device id:{inDeviceId}");
 
                 return _inputDevice;
             }
@@ -362,8 +367,12 @@ namespace Tizen.Multimedia
         {
             get
             {
+                /* This P/Invoke intends to validate if the core audio system
+                 * is normal. Otherwise, it'll throw an error here. */
                 Interop.AudioStreamPolicy.GetPreferredDevice(Handle, out _, out var outDeviceId).
                     ThrowIfError("Failed to get preferred output device");
+
+                Log.Debug(Tag, $"preferred output device id:{outDeviceId}");
 
                 return _outputDevice;
             }
--- a/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
+++ b/src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
@@ -305,6 +305,72 @@ namespace Tizen.Multimedia
         }
 
         /// <summary>
+        /// Gets or sets the preferred input device id.
+        /// </summary>
+        /// <value>
+        /// A device id which is greater than or equal to 0.<br/>
+        /// The default is 0 which means any device is not set on this property.
+        /// </value>
+        /// <remarks>
+        /// This property is to set a specific built-in device when the system has multiple devices of the same built-in device type.
+        /// When there's only one device for a built-in device type in the system, nothing will happen even if this property is set successfully.
+        /// </remarks>
+        /// <exception cref="ArgumentException">A device is not for input.</exception>
+        /// <exception cref="InvalidOperationException">An internal error occurs.</exception>
+        /// <exception cref="AudioPolicyException">A device is not supported by this <see cref="AudioStreamPolicy"/> instance.</exception>
+        /// <exception cref="ObjectDisposedException">The <see cref="AudioStreamPolicy"/> has already been disposed of.</exception>
+        /// <seealso cref="AudioManager.GetConnectedDevices()"/>
+        /// <since_tizen> 6 </since_tizen>
+        public int PreferredInputDeviceId
+        {
+            get
+            {
+                Interop.AudioStreamPolicy.GetPreferredDevice(Handle, out var inDeviceId, out _).
+                    ThrowIfError("Failed to get preferred input device");
+
+                return inDeviceId;
+            }
+            set
+            {
+                Interop.AudioStreamPolicy.SetPreferredDevice(Handle, AudioDeviceIoDirection.Input, value).
+                    ThrowIfError("Failed to set preferred input device");
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the preferred output device id.
+        /// </summary>
+        /// <value>
+        /// A device id which is greater than or equal to 0.<br/>
+        /// The default is 0 which means any device is not set on this property.
+        /// </value>
+        /// <remarks>
+        /// This property is to set a specific built-in device when the system has multiple devices of the same built-in device type.
+        /// When there's only one device for a built-in device type in the system, nothing will happen even if this property is set successfully.
+        /// </remarks>
+        /// <exception cref="ArgumentException">A device is not for output.</exception>
+        /// <exception cref="InvalidOperationException">An internal error occurs.</exception>
+        /// <exception cref="AudioPolicyException">A device is not supported by this <see cref="AudioStreamPolicy"/> instance.</exception>
+        /// <exception cref="ObjectDisposedException">The <see cref="AudioStreamPolicy"/> has already been disposed of.</exception>
+        /// <seealso cref="AudioManager.GetConnectedDevices()"/>
+        /// <since_tizen> 6 </since_tizen>
+        public int PreferredOutputDeviceId
+        {
+            get
+            {
+                Interop.AudioStreamPolicy.GetPreferredDevice(Handle, out _, out var outDeviceId).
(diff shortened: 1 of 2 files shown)


src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
```diff
@@ -304,6 +304,72 @@ public void RemoveDeviceForStreamRouting(AudioDevice device)
...
+        /// This property is to set a specific built-in device when the system has multiple devices of the same built-in device type.
+        /// When there's only one device for a built-in device type in the system, nothing will happen even if this property is set successfully.
+        /// </remarks>
+        /// <exception cref="ArgumentException">A device is not for input.</exception>
+        /// <exception cref="InvalidOperationException">An internal error occurs.</exception>
+        /// <exception cref="AudioPolicyException">A device is not supported by this <see cref="AudioStreamPolicy"/> instance.</exception>
+        /// <exception cref="ObjectDisposedException">The <see cref="AudioStreamPolicy"/> has already been disposed of.</exception>
+        /// <seealso cref="AudioManager.GetConnectedDevices()"/>
+        /// <since_tizen> 6 </since_tizen>
+        public int PreferredInputDeviceId
```
myroot: Is it a `AudioDevice`?
How about use `AudioDevice` type instead of int?
because, App developer can't know about Id, app developer should get a AudioDevice first.

and others API also use `AudioDevice` 
` public void RemoveDeviceForStreamRouting(AudioDevice device)`

```suggestion
        public AudioDevice PreferredInputDevice
```

sangchul1011: Here' a reason why I use integer value instead of the AudioDevice.

AudioDevice instance can get from new List() via inside of GetConnectedDevices() as an item. Hence, if we use AudioDevice as the return value, the app developer should not call FreeDeviceList() between setting PreferredInput[Output]Device property and getting PreferredInput[Output]Device.

Device id of integer value, on the other hand, doesn't need to care about the lifecycle of AudioDevice. What the app. developer has to care about it is to get new AudioDevice instance and check the id one of its property(audioDevice.Id).

In a situation like this, I'd like to adopt the latter.

myroot: Does not matter call FreeDeviceList(), 
`FreeDeviceList` is just release a native list object, and a AudioDevice object is created on this
https://github.com/Samsung/TizenFX/blob/7d5b33bd6d146e0504a4398f9b97b44a85718156/src/Tizen.Multimedia/AudioManager/AudioManager.cs#L82
and this lifecycle is managed by AudioDevice

sangchul1011: I checked again you're right that there's no FreeDeviceList() for CSharp API, but just for interal operation.

I'll revise it with doing like below.
 Set the property with handle -> invoke interop. setter method with audioDevice.Id and save the instance.
 Get the property with handle -> check the id from the saved instance with interop. getter method. and return the saved instance.

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
```diff
@@ -304,6 +306,118 @@ public void RemoveDeviceForStreamRouting(AudioDevice device)
...
+                {
+                    return null;
+                }
+
+                Interop.AudioStreamPolicy.GetPreferredDevice(Handle, out _, out var outDeviceId).
+                    ThrowIfError("Failed to get preferred output device");
+
+                if (outDeviceId != _outputDevice.Id)
+                {
+                    throw new InvalidOperationException("output device id does not match");
```
myroot: I don't understand, Why you need throw exception? 
When does this happen?

I expect, a AudioDevice instance should be returned, and it's device id is equal to _outputDevice.Id

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
```diff
@@ -304,6 +306,118 @@ public void RemoveDeviceForStreamRouting(AudioDevice device)
...
+        /// <exception cref="InvalidOperationException">An internal error occurs.</exception>
+        /// <exception cref="AudioPolicyException">A device is not supported by this <see cref="AudioStreamPolicy"/> instance.</exception>
+        /// <exception cref="ObjectDisposedException">The <see cref="AudioStreamPolicy"/> has already been disposed of.</exception>
+        /// <seealso cref="AudioManager.GetConnectedDevices()"/>
+        /// <since_tizen> 6 </since_tizen>
+        public AudioDevice PreferredInputDevice
+        {
+            get
+            {
+                if (_inputDevice == null)
```
hsgwon: I think you don't need to check `null`, just return it.
Because, if `_inputDevice` is `null`, you also return `null`.

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
```diff
@@ -304,6 +306,118 @@ public void RemoveDeviceForStreamRouting(AudioDevice device)
...
+                {
+                    return null;
+                }
+
+                Interop.AudioStreamPolicy.GetPreferredDevice(Handle, out var inDeviceId, out _).
```
hsgwon: Is this pinvoke call really needed?
Because, current preferred input device is captured in `_inputDevice` already.

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
```diff
@@ -304,6 +306,118 @@ public void RemoveDeviceForStreamRouting(AudioDevice device)
...
+                    ThrowIfError("Failed to get preferred input device");
+
+                if (inDeviceId != _inputDevice.Id)
+                {
+                    throw new InvalidOperationException("output device id does not match");
+                }
+
+                return _inputDevice;
```
hsgwon: I think below code is enough, 
if you want to save the preferred `AudioDevice` instance to `_inputDevice`.

```csharp
public AudioDevice PreferredInputDevice
{
      get => _inputDevice;
      set
      { ... }
}
```

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
```diff
@@ -304,6 +306,118 @@ public void RemoveDeviceForStreamRouting(AudioDevice device)
...
+                if (value == null)
+                {
+                    deviceId = 0;
+                }
+                else
+                {
+                    deviceId = value.Id;
+                }
+
+                Interop.AudioStreamPolicy.SetPreferredDevice(Handle, AudioDeviceIoDirection.Input, deviceId).
```
hsgwon: How about the following code to simplify all above code?

```csharp
Interop.AudioStreamPolicy.SetPreferredDevice(Handle, AudioDeviceIoDirection.Input, value?.Id ?? 0).
      ThrowIfError("Failed to set preferred input device");
```

hsgwon: How about the following code to simplify all above code.

```csharp
Interop.AudioStreamPolicy.SetPreferredDevice(Handle, AudioDeviceIoDirection.Input, value?.Id ?? 0).
      ThrowIfError("Failed to set preferred input device");
```

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
sangchul1011: As you know, setting the property is performed via native API only with device id and it affects sound daemon routing path. Don't we need to check again via native API it in case of getting the property? (e.g. If the daemon crash happens after setting the preferred device, it would have invalid value. Of course, it is abnormal case.)

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
sangchul1011: Sure, if we decide not to call the method below, it can be removed.

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
sangchul1011: Okay, it looks good. I'll do that.

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
myroot: Yes, I think, no need to check from native.
If set operation was failed, exception should be occurred in setter.
but if not occurred exception, API must return a value that is set by user

sangchul1011: I think we have a misunderstanding. I meant the pinvoke in "getter" property could give different value in such cases I wrote above, so I added the condition to check in the getter property.

myroot: Is it a possible?
A value is applied to instance of AudioPolice that is created by c# code
How to different value was return from native?

sangchul1011: It could be possible via native API. (I described it in the previous comments regarding inconsistent with system status for exception case.)

myroot: If multimedia systems are in inconsistent state, Native call will be failed. so, `ThrowIfError("Failed to get preferred output device");`  exception should be raised, and do not reached here

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
myroot: I also agree it

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
sangchul1011: I've been preparing the patch like that including core service. It could be removed after all.

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
```diff
@@ -304,6 +306,76 @@ public void RemoveDeviceForStreamRouting(AudioDevice device)
...
+        /// <exception cref="InvalidOperationException">An internal error occurs.</exception>
+        /// <exception cref="AudioPolicyException">A device is not supported by this <see cref="AudioStreamPolicy"/> instance.</exception>
+        /// <exception cref="ObjectDisposedException">The <see cref="AudioStreamPolicy"/> has already been disposed of.</exception>
+        /// <seealso cref="AudioManager.GetConnectedDevices()"/>
+        /// <since_tizen> 6 </since_tizen>
+        public AudioDevice PreferredInputDevice
+        {
+            get
+            {
+                Interop.AudioStreamPolicy.GetPreferredDevice(Handle, out var inDeviceId, out _).
```
hsgwon: `GetPreferredDevice()` method is called but its return value is not used in `PreferredInputDevice`'s return value.
You can say `GetPreferredDevice()` is only for checking mismatch in native side, but I think it's not usual way.

How about creating instance `AudioDevice` based on its output?
or Rename `GetPreferredDevice()` to `ValidatePreferredDeviceId` or something like that?

sangchul1011: 1. Creating instance of AudioDevice is not possible now. We can only get the instance from the device list.

2. The native getter API should take at least one out parameter otherwise it'll return invalid param error. Just renaming is not changing this situation. It might need one another internal native API without out parameter. Rather, I'd like to use this public getter API and add a log with the result of out parameter. How about this?

hsgwon: Okay. And please add some additional information for pinvoke call, as we're discussed.

sangchul1011: The revision has been done.
@myroot Could you plz review again? Thanks in advance.

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
```diff
@@ -304,6 +307,84 @@ public void RemoveDeviceForStreamRouting(AudioDevice device)
...
 
+        /// <summary>
+        /// Gets or sets the preferred input device.
+        /// </summary>
+        /// <value>
+        /// The <see cref="AudioDevice"/> instance.<br/>
+        /// The default is null which means any device is not set on this property.
+        /// </value>
+        /// <remarks>
+        /// This property is to set a specific built-in device when the system has multiple devices of the same built-in device type.
```
journey2w: Suggestion:
        /// This property is to set a specific device among built-in audio input devices on the system.
        /// When there is only one input device on the system, nothing happens even if this property is set successfully.

src/Tizen.Multimedia/AudioManager/AudioStreamPolicy.cs
```diff
@@ -304,6 +307,84 @@ public void RemoveDeviceForStreamRouting(AudioDevice device)
...
+
+        /// <summary>
+        /// Gets or sets the preferred output device.
+        /// </summary>
+        /// <value>
+        /// The <see cref="AudioDevice"/> instance.<br/>
+        /// The default is null which means any device is not set on this property.
+        /// </value>
+        /// <remarks>
+        /// This property is to set a specific built-in device when the system has multiple devices of the same built-in device type.
```
journey2w: Suggestion:
        /// This property is to set a specific device among built-in audio output devices on the system.
        /// When there is only one output device on the system, nothing happens even if this property is set successfully.

PR #990, opened 2019-08-27
[MediaPlayer] fix APIs to get the list of really supported formats

aferin:
### Description of Change ###
fix APIs to get the list of really supported formats

### DIFF
--- a/src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
@@ -259,15 +259,13 @@ namespace Tizen.Multimedia
                 Log.Error(PlayerLog.Tag, "The source is not set as a source to a player yet.");
                 throw new InvalidOperationException("The source is not set as a source to a player yet.");
             }
-            if (_audioMediaFormat == null && _videoMediaFormat == null)
-            {
-                throw new ArgumentNullException(string.Concat(nameof(_audioMediaFormat), " and ", nameof(_videoMediaFormat)));
-            }
+
             if (packet == null)
             {
                 Log.Error(PlayerLog.Tag, "packet is null");
                 throw new ArgumentNullException(nameof(packet));
             }
+
             if (packet.IsDisposed)
             {
                 Log.Error(PlayerLog.Tag, "packet is disposed");
@@ -291,6 +289,7 @@ namespace Tizen.Multimedia
                 Log.Error(PlayerLog.Tag, "Video is not configured with the current source.");
                 throw new ArgumentException("Video is not configured with the current source.");
             }
+
             if (packet.Format.Type == MediaFormatType.Audio && _audioMediaFormat == null)
             {
                 Log.Error(PlayerLog.Tag, "Audio is not configured with the current source.");
--- a/src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
@@ -33,8 +33,8 @@ namespace Tizen.Multimedia
     {
         private readonly MediaFormat _audioMediaFormat;
         private readonly MediaFormat _videoMediaFormat;
-        private static PlayerHandle _playerHandle;
-        private static IntPtr _handle;
+        private static List<MediaFormatAudioMimeType> _supportedAudioFormats;
+        private static List<MediaFormatVideoMimeType> _supportedVideoFormats;
 
         /// <summary>
         /// Gets all supported audio types.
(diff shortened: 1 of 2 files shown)


src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
```diff
@@ -42,7 +44,39 @@ public static IEnumerable<MediaFormatAudioMimeType> SupportedAudioTypes
...
             get
             {
-                yield return MediaFormatAudioMimeType.Aac;
+                NativePlayer.Create(out _playerHandle).ThrowIfFailed(null, "Failed to create player");
+                _handle = _playerHandle.DangerousGetHandle();
+                Debug.Assert(_handle != IntPtr.Zero);
+
+                try
+                {
+                    List<MediaFormatAudioMimeType> audioFormats = new List<MediaFormatAudioMimeType>();
```
hsgwon: How about the following?
```csharp
private static List<MediaFormatAudioMimeType> _supportedAudioFormats;
public static IEnumerable<MediaFormatAudioMimeType> SupportedAudioTypes
{
     get
      {
            if (_supportedAudioFormats== null)
            {
                  _supportedAudioFormats= new List<MediaFormatAudioMimeType>();
                  ...
            }
            
            return _videoFormats;
      }
```

src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
```diff
@@ -42,7 +44,39 @@ public static IEnumerable<MediaFormatAudioMimeType> SupportedAudioTypes
...
+
+                    NativePlayer.SupportedMediaFormatCallback callback = (int format, IntPtr userData) =>
+                    {
+                        if (!Enum.IsDefined(typeof(MediaFormatAudioMimeType), format))
```
hsgwon: How about creating `audioFormat` and `videoFormat` at once?
For example,
```csharp
if (Enum.IsDefined(typeof(MediaFormatAudioMimeType), format))
{
      audioFormats.Add((MediaFormatAudioMimeType)format);
}
else
{
      videoFormats.Add((MediaFormatVideoMimeType)format);
}
```

If so, we can reduce pinvoke call.

src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
```diff
@@ -33,6 +33,8 @@ public sealed class MediaStreamSource : MediaSource
     {
         private readonly MediaFormat _audioMediaFormat;
         private readonly MediaFormat _videoMediaFormat;
+        private static PlayerHandle _playerHandle;
+        private static IntPtr _handle;
```
hsgwon: I think it doesn't need to make `_playerHandle` and  `_handle` to `field`. `Local variable` in property getter is enough.

src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
hsgwon: And if you modify like above, you can make this code block(line 53~71) private common method, 
which can be used both properties.

src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
aferin: that's right. i'll fix it.

src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
aferin: actually there are much more formats in native header than c# header. i am worried `else` could get wrong with what i intend to.
if these methods were not divided, it could have covered every types at once..

src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
aferin: ok i'll keep it into local variable

src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
hsgwon: If you worried about `container`, `text`, `unknown` format, you can check it also like the following.
```csharp
if (Enum.IsDefined(typeof(MediaFormatAudioMimeType), format))
{
      audioFormats.Add((MediaFormatAudioMimeType)format);
}
else if (Enum.IsDefined(typeof(MediaFormatVideoMimeType), format))
{
      videoFormats.Add((MediaFormatVideoMimeType)format);
}
else
{
      Log.Debug(PlayerLog.Tag, "It's not audio/video format. skipped : " + format.ToString());
}
```

src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
hsgwon: `_supportedAudioFormats` should be captured by `field`, not local variable.

aferin: right i meant that. i fixed it to a field variable.

src/Tizen.Multimedia.MediaPlayer/Player/MediaStreamSource.cs
aferin: then i'll combine it to one method

PR #1002, opened 2019-09-08
[Applications.Common] Add a new property of AppControl

hjhun:

### Description of Change ###

This property is for getting and setting a component ID.

<class>
AppControl
<property>
+ string ComponentID { get; set; }

### API Changes ###

 - ACR: TCSACR-277

### DIFF
--- a/src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
@@ -58,6 +58,7 @@ namespace Tizen.Applications
         private string _category = null;
         private string _applicationId = null;
         private ExtraDataCollection _extraData = null;
+        private string _componentId = null;
 
         /// <summary>
         /// Initializes the instance of the AppControl class.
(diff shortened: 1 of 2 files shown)


src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
```diff
@@ -444,6 +445,48 @@ public ExtraDataCollection ExtraData
...
+        /// </value>
+        /// <example>
+        /// <code>
+        /// AppControl appControl = new AppControl();
+        /// appControl.ComponentId = "org.tizen.frame-component";
+        /// Log.Debug(LogTag, "ComponentId: " + appControl.ComponentId);
+        /// </code>
+        /// </example>
+        /// <since_tizen> 6 </since_tizen>
+        public string ComponentId
```
myroot: What is the role of this property?
How is it related to ApplicationId?
Why don't you extend AppControl to ComponentControl?

hjhun: To send a send launch request to execute a component of the component-based-application, the component ID is necessary.
To execute the component properly, the AppControl should have the application ID and the component ID.

myroot: Is Component ID not globally unique?

myroot: How about extend?
``` c#
 class ComponentControl : AppControl 
```

hjhun: Component ID is unique globally.
It's an additional property of AppControl.
If the component ID doesn't exist, the application(component-based-application) is launched.

myroot: > To execute the component properly, the AppControl should have the application ID and the component ID.

I can't understand why need appid? you said  `Component ID is unique globally`

hjhun: As I mentioned, the component Id is the additional property. If it does not exist on AppControl, the application is launched. If it exists, the component is launched.
Setting the component Id is not mandatory.

pjh9216: The component ID is not required to launch ComponetBasedApplication. If it is not added, the main component will be launched.
If we provide extended ComponentControl, developer may be confused about which class to use when launching only with the App ID without the component ID.
In fact, the caller app may not know that the target app is a ComponentBasedApplication.

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
```diff
@@ -444,6 +445,48 @@ public ExtraDataCollection ExtraData
             }
         }
 
+        /// <summary>
+        /// Gets and sets the component ID.
```
myroot: Please update description

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
myroot: Ok, I understand, but ComponentId is not proper to AppControl's property, because it meansingless in AppControl object in most cases.

myroot: If we provide it as property, developer may be confused "What is it and which value should be set?"

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
hjhun: I've updated the description.

myroot: Do you done?
you need to update ` Gets and sets the component ID.`

src/Tizen.Applications.Common/Tizen.Applications/AppControl.cs
```diff
@@ -444,6 +445,52 @@ public ExtraDataCollection ExtraData
             }
         }
 
+        /// <summary>
+        /// Gets and sets the component ID of the component-based application.
```
myroot:
```suggestion
        /// Gets and sets the component ID to explicitly launch a component.
```

PR #1012, opened 2019-09-16
[Tizen.Multimedia.Camera] Fixed exception bug

hsgwon:
### Description of Change ###

Fixed missing exception description and exception check code.


src/Tizen.Multimedia.Camera/Camera/Camera.cs
```diff
@@ -50,6 +50,8 @@ public class Camera : IDisposable, IDisplayable<CameraError>
         /// Initializes a new instance of the <see cref="Camera"/> class.
         /// </summary>
         /// <param name="device">The camera device to access.</param>
+        /// <exception cref="InvalidOperationException">Invalid CameraDevice type.</exception>
```
myroot: According to description, It seem to ArgumentException,

hsgwon: You're right. I'll fix it.

PR #1020, opened 2019-09-17
[NUI] Changed the APIs related to MultiWIndow to public

tscholb:
### Description of Change ###
[NUI] Changed the APIs related to MultiWIndow to public

### API Changes ###
http://suprem.sec.samsung.net/jira/browse/TCSACR-279

### DIFF
--- a/src/Tizen.NUI/src/public/Window.cs
+++ b/src/Tizen.NUI/src/public/Window.cs
@@ -74,7 +74,6 @@ namespace Tizen.NUI
         /// <param name="isTranslucent">Whether Window is translucent.</param>
         /// <returns>A new Window.</returns>
         /// <since_tizen> 6 </since_tizen>
-        [EditorBrowsable(EditorBrowsableState.Never)]
         public Window(Rectangle windowPosition = null , bool isTranslucent = false) : this(Interop.Window.Window_New__SWIG_0(Rectangle.getCPtr(windowPosition), "", isTranslucent), true)
         {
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Window.cs
```diff
@@ -89,7 +88,6 @@ public Window(Rectangle windowPosition = null , bool isTranslucent = false) : th
         /// <param name="isTranslucent">Whether Window is translucent.</param>
         /// <returns>A new Window.</returns>
         /// <since_tizen> 6 </since_tizen>
-        [EditorBrowsable(EditorBrowsableState.Never)]
```
journey2w: Line 84: 
The method description is same as Window(Rectangle windowPosition = null , bool isTranslucent = false).
It would be better to change the description.
Suggestion:
/// Creates a new Window with a specific name.

Line 137:
Do you have any reason why you use a passive voice?
You wrote descriptions on line 137, 215, and etc....
If you use active voice, is the meaning changed?
If not, I think it would be better to use active voice.

Suggestion:
        /// Emits the event when the screen is touched and when the touch ends too.<br />
       /// If there are multiple touchpoints then it is emitted when the first touch occurs and when the last finger is lifted too.<br />
        
Line 140:
I don't understand what " /// An interrupted event will also be emitted (if it occurs)." means.
Does it mean " /// Even though incoming events are interrupted, the event occurs." ?
If so, please change it.


Line 215 and others: Passive voice-> Active voice

 /// Emits emits when the window resized.

PR #1035, opened 2019-09-25
[NUI] Modify GetWindowList for getting Native Ptr

tscholb:
- Modify GetWindowList for getting Native Ptr
- Set Window name when creating window using window name

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Application.cs
+++ b/src/Tizen.NUI/src/internal/Application.cs
@@ -1391,7 +1391,7 @@ namespace Tizen.NUI
             List<Window> WindowList = new List<Window>();
             for( uint i = 0; i < ListSize; ++i )
             {
-                Window currWin = new Window(Interop.Application.Application_GetWindowsFromList(i), true);
+                Window currWin = Registry.GetManagedBaseHandleFromNativePtr(Interop.Application.Application_GetWindowsFromList(i)) as Window;
                 if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
                 WindowList.Add(currWin);
             }
--- a/src/Tizen.NUI/src/public/Window.cs
+++ b/src/Tizen.NUI/src/public/Window.cs
@@ -92,6 +92,7 @@ namespace Tizen.NUI
         [EditorBrowsable(EditorBrowsableState.Never)]
         public Window(string name, Rectangle windowPosition = null, bool isTranslucent = false) : this(Interop.Window.Window_New__SWIG_0(Rectangle.getCPtr(windowPosition), name, isTranslucent), true)
         {
+            this._windowTitle = name;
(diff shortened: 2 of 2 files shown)


src/Tizen.NUI/src/internal/Application.cs
```diff
@@ -1391,7 +1391,7 @@ internal static List<Window> GetWindowList()
             List<Window> WindowList = new List<Window>();
             for( uint i = 0; i < ListSize; ++i )
             {
-                Window currWin = new Window(Interop.Application.Application_GetWindowsFromList(i), true);
+                Window currWin = Registry.GetManagedBaseHandleFromNativePtr(Interop.Application.Application_GetWindowsFromList(i)) as Window;
                 if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
```
dongsug-song: seems to be better to add NULL check of currWin

src/Tizen.NUI/src/internal/Application.cs
```diff
@@ -1391,7 +1391,7 @@ internal static List<Window> GetWindowList()
             List<Window> WindowList = new List<Window>();
```
dongsug-song: seems better to use foreach than for

src/Tizen.NUI/src/internal/Application.cs
```diff
@@ -1391,7 +1391,7 @@ internal static List<Window> GetWindowList()
...
             for( uint i = 0; i < ListSize; ++i )
             {
-                Window currWin = new Window(Interop.Application.Application_GetWindowsFromList(i), true);
+                Window currWin = Registry.GetManagedBaseHandleFromNativePtr(Interop.Application.Application_GetWindowsFromList(i)) as Window;
                 if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
                 WindowList.Add(currWin);
             }
```
dongsug-song: when WindowList is created in line 1391, the ListSize will be Zero.
then after new currWin is added, the WindowList.Size will be increased.
ListSize is different from WindowList.Size ?
could you explain?

src/Tizen.NUI/src/public/Window.cs
```diff
@@ -92,6 +92,7 @@ public Window(Rectangle windowPosition = null , bool isTranslucent = false) : th
         [EditorBrowsable(EditorBrowsableState.Never)]
         public Window(string name, Rectangle windowPosition = null, bool isTranslucent = false) : this(Interop.Window.Window_New__SWIG_0(Rectangle.getCPtr(windowPosition), name, isTranslucent), true)
         {
+            this._windowTitle = name;
```
dongsug-song: it sets this._windowTitle which seems to be a private member variable.
then by this set, the name will be set to native Window?

PR #1036, opened 2019-09-25
[MediaPlayer] Add API for setting audio codec type

aferin:
### API Changes ###
 - ACR: http://suprem.sec.samsung.net/jira/browse/TCSACR-281

Added:
 - public enum CodecType
 - public CodecType AudioCodec
 - public class NotAvailableException : InvalidOperationException

### DIFF
--- a/src/Tizen.Network.IoTConnectivity/Tizen.Network.IoTConnectivity/RemoteResource.cs
+++ b/src/Tizen.Network.IoTConnectivity/Tizen.Network.IoTConnectivity/RemoteResource.cs
@@ -17,6 +17,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Concurrent;
 using System.Net;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
@@ -37,8 +38,8 @@ namespace Tizen.Network.IoTConnectivity
         private bool _cacheEnabled = false;
         private ResourceOptions _options;
 
-        private int _responseCallbackId = 1;
-        private static Dictionary<IntPtr, Interop.IoTConnectivity.Client.RemoteResource.ResponseCallback> _responseCallbacksMap = new Dictionary<IntPtr, Interop.IoTConnectivity.Client.RemoteResource.ResponseCallback>();
+        private static int _responseCallbackId = 1;
+        private static IDictionary<IntPtr, Interop.IoTConnectivity.Client.RemoteResource.ResponseCallback> _responseCallbacksMap = new ConcurrentDictionary<IntPtr, Interop.IoTConnectivity.Client.RemoteResource.ResponseCallback>();
 
         private Interop.IoTConnectivity.Client.RemoteResource.CachedRepresentationChangedCallback _cacheUpdatedCallback;
         private Interop.IoTConnectivity.Client.RemoteResource.StateChangedCallback _stateChangedCallback;
@@ -404,10 +405,8 @@ namespace Tizen.Network.IoTConnectivity
             _responseCallbacksMap[id] = (IntPtr resource, int err, int requestType, IntPtr responseHandle, IntPtr userData) =>
             {
                 IntPtr responseCallbackId = userData;
-                lock(_responseCallbacksMap)
-                {
-                    _responseCallbacksMap.Remove(responseCallbackId);
-                }
+
+                _responseCallbacksMap.Remove(responseCallbackId);
 
                 if (responseHandle != IntPtr.Zero)
                 {
@@ -459,10 +458,9 @@ namespace Tizen.Network.IoTConnectivity
             _responseCallbacksMap[id] = (IntPtr resource, int err, int requestType, IntPtr responseHandle, IntPtr userData) =>
             {
                 IntPtr responseCallbackId = userData;
-                lock (_responseCallbacksMap)
-                {
-                    _responseCallbacksMap.Remove(responseCallbackId);
-                }
+
+                _responseCallbacksMap.Remove(responseCallbackId);
+
                 if (err == (int)(IoTConnectivityError.Iotivity))
                 {
                     RemoteResponse response = new RemoteResponse();
@@ -519,10 +517,9 @@ namespace Tizen.Network.IoTConnectivity
             _responseCallbacksMap[id] = (IntPtr resource, int err, int requestType, IntPtr responseHandle, IntPtr userData) =>
             {
                 IntPtr responseCallbackId = userData;
-                lock (_responseCallbacksMap)
-                {
-                    _responseCallbacksMap.Remove(responseCallbackId);
-                }
+
+                _responseCallbacksMap.Remove(responseCallbackId);
+
                 if (responseHandle != IntPtr.Zero)
                 {
                     try
@@ -570,10 +567,9 @@ namespace Tizen.Network.IoTConnectivity
             _responseCallbacksMap[id] = (IntPtr resource, int err, int requestType, IntPtr responseHandle, IntPtr userData) =>
             {
                 IntPtr responseCallbackId = userData;
-                lock (_responseCallbacksMap)
-                {
-                    _responseCallbacksMap.Remove(responseCallbackId);
-                }
+
+                _responseCallbacksMap.Remove(responseCallbackId);
+
                 if (err == (int)(IoTConnectivityError.Iotivity))
                 {
                     RemoteResponse response = new RemoteResponse();


src/Tizen.Multimedia.MediaPlayer/Player/PlayerError.cs
```diff
@@ -166,5 +170,21 @@ public ResourceLimitException(string message) : base(message)
         {
         }
     }
+
+    /// <summary>
+    /// The exception that is thrown when there it is not available.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class NotAvailableException : InvalidOperationException
```
hsgwon: Does `NotAvailableException` means some kind of invalid operation?
If not, how about derive it from `Exception`?

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -667,6 +676,57 @@ public float AudioPitch
...
         }
 
+        /// <summary>
+        /// Gets or sets the default codec type of the audio decoder.
+        /// </summary>
+        /// <value>A <see cref="CodecType"/> that specifies the type.
+        /// The default codec type could be different depending on the device capability.</value>
+        /// <remarks>
+        /// <para>To set, the player must be in the <see cref="PlayerState.Idle"/> state.</para>
+        /// <para><see cref="CodecType.Sw"/> is supported basically.
```
hsgwon: Does this means software audio codec should be supported in all platform, right?

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -667,6 +676,57 @@ public float AudioPitch
...
+        /// If H/W audio codec type is not supported in some cases, S/W audio codec type could be used instead.</para>
+        /// <para>The availability could be changed depending on the codec capability.
+        /// If an application wants to use the H/W audio codec type as default,
+        /// following functions should be called after the codec type is set. :<br/>
+        /// <see cref="AudioEffect.IsAvailable"/><br/>
+        /// <see cref="EnableExportingAudioData"/><br/>
+        /// <see cref="DisableExportingAudioData"/><br/>
```
hsgwon: In the above description, `following functions should be called after the codec type is set.` is described.
If you describe `EnableExportingAudioData` and `DisableExportingAudioData` together here,
user can confuse whether it should be called the at the same time or not.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -667,6 +676,57 @@ public float AudioPitch
...
 
+        /// <summary>
+        /// Gets or sets the default codec type of the audio decoder.
+        /// </summary>
+        /// <value>A <see cref="CodecType"/> that specifies the type.
+        /// The default codec type could be different depending on the device capability.</value>
+        /// <remarks>
+        /// <para>To set, the player must be in the <see cref="PlayerState.Idle"/> state.</para>
+        /// <para><see cref="CodecType.Sw"/> is supported basically.
+        /// If H/W audio codec type is not supported in some cases, S/W audio codec type could be used instead.</para>
```
hsgwon: I think this statement is not necessary, if software codec is supported in all platform as described above statement.

src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
```diff
@@ -82,12 +82,17 @@ internal AudioEffect(Player owner)
         /// <exception cref="InvalidOperationException">
         ///     If audio offload is enabled by calling <see cref="AudioOffload.IsEnabled"/>. (Since tizen 6.0)
         /// </exception>
+        /// <exception cref="NotAvailableException"><see cref="IsAvailable"/> returns false. (Since tizen 6.0)</exception>
+        /// <seealso cref="IsAvailable"/>
         /// <since_tizen> 3 </since_tizen>
         public void Clear()
         {
             Player.ValidateNotDisposed();
```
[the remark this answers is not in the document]
aferin: i also think it is not needed for this class.
but it had already been in description so i added the code in last patch.

ObjectDisposedException is located at the top of each API except for NotSupportedException.
the below order was made under discussion.
Check feature -> Disposed -> parameter -> Offload -> state

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
aferin: as far as i understand it has to be but there still has small possibility that it isn't.
i got the vague answer so need to check again.

src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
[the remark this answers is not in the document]
hsgwon: `ObjectDisposedException is located at the top of each API except for NotSupportedException.`
If you want to check dispose before CheckDisabled(), it's ok.

But, If so, you should add it other property in the same class, as I said before.
`Count`, `BandLevelRange`, `IsAvailable`

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
hsgwon: The statement is strictly described as it's `supported basically`. 
So if not, you should describe it here.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
aferin: that sentence meant "even though an user set hw codec, it can be chagned to sw codec."

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
aferin: i dont think so. the meaning of enable and disable is clear.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -667,6 +676,57 @@ public float AudioPitch
...
+        /// <exception cref="ObjectDisposedException">The player has already been disposed of.</exception>
+        /// <exception cref="ArgumentException">The value is not valid.</exception>
+        /// <exception cref="InvalidOperationException">
+        ///     The player is not in the valid state.
+        ///     -or-<br/>
+        ///     Operation failed; internal error.
+        /// </exception>
+        /// <exception cref="CodecNotSupportedException">The selected codec is not supported.</exception>
+        /// <since_tizen> 6 </since_tizen>
+        public CodecType AudioCodec
```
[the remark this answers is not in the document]
aferin: right. that is better. i'll change the name.

src/Tizen.Multimedia.MediaPlayer/Player/PlayerEnums.cs
```diff
@@ -323,4 +323,23 @@ public enum PlayerAudioExtractOption
...
+    /// Specifies the types of a codec for <see cref="Player"/>.
+    /// </summary>
+    /// <seealso cref="Player.AudioCodec"/>
+    /// <since_tizen> 6 </since_tizen>
+    public enum CodecType
+    {
+        /// <summary>
+        /// An optional flag for using the H/W codec.
+        /// </summary>
+        Hw,
```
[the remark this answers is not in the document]
aferin: ok. i'll fix it.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
hsgwon: You mean it can be changed automatically to software codec in case of hardware codec is not supported, right?
If so, when does 'CodecNotSupportedException` is thrown as described below?

src/Tizen.Multimedia.MediaPlayer/Player/PlayerError.cs
aferin: it is close to internal error. Exception might be better.

src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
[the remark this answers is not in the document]
aferin: i'll rearrange ObjectDisposedException throughout the files later.
some of them are described but aren't there.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
aferin: i just remembered this meaning.
`
The default codec type could be different depending on the device capability.
S/W codec type is supported basically.
`
these sentences are a set.
the default codec type can be hw or sw depending on the device. but sw is supported basically.

aferin: `basically` might be put, in case sw codec is not supported

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
aferin: when the both hw codec and sw codec are not supported, it'll be thrown.

hsgwon: software codec can be not supported, right?
If so, please fix the statement above.
`<see cref="CodecType.Sw"/> is supported basically.`

hsgwon: Software codec can be also not supported, right?
If so, please fix the statement above.
<see cref="CodecType.Sw"/> is supported basically.

PR #1040, opened 2019-09-26
[ComponentBased][TCSACR-282] Add Release method to IWindowInfo interface

gotoextreme:
IWindowInfo.cs
[namespace]
+ Tizen.Applications.ComponentBased.Common
[class]
IWindowInfo
[property]
+ ResourceId: int<<readonly>>
[method]
+ Dispose()

 

EFLWindowInfo.cs
[namespace]
+ Tizen.Applications.ComponentBased.Default
[class]
EFLWindowInfo
[method]
+ EFLWindowInfo(Window win)
[property]
+ ResourceId: int<<readonly>>
[method]
+ Dispose()


gotoextreme:
> How about use IDisposable interface?

Done

src/Tizen.Applications.ComponentBased.Default/Tizen.Applications.ComponentBased.Default/EFLWindowInfo.cs
```diff
@@ -44,6 +45,35 @@ public int ResourceId
...
+        protected virtual void Dispose(bool disposing)
+        {
+            if (_disposed)
+                return;
+
+            if (disposing)
+            {
+            }
+
+            _win.Unrealize();
```
myroot: _win is managed object, and managed object should be released when only disposing was true

PR #1043, opened 2019-09-30
[MediaPlayer] Add API for setting audio codec type

aferin:
### API Changes ###
 - ACR: http://suprem.sec.samsung.net/jira/browse/TCSACR-281

Added:
 - public enum CodecType
 - public CodecType AudioCodec
 - public class NotAvailableException : InvalidOperationException

### DIFF
--- a/src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
+++ b/src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
@@ -31,7 +31,7 @@ namespace Tizen.Multimedia
         {
             Player = owner;
 
-            if (IsAvailable== false)
+            if (IsAvailable == false)
             {
                 return;
             }
@@ -82,11 +82,16 @@ namespace Tizen.Multimedia
         /// <exception cref="InvalidOperationException">
         ///     If audio offload is enabled by calling <see cref="AudioOffload.IsEnabled"/>. (Since tizen 6.0)
         /// </exception>
+        /// <exception cref="NotAvailableException"><see cref="IsAvailable"/> returns false. (Since tizen 6.0)</exception>
+        /// <seealso cref="IsAvailable"/>
         /// <since_tizen> 3 </since_tizen>
         public void Clear()
         {
             Player.ValidateNotDisposed();
             Player.AudioOffload.CheckDisabled();
+
+            if (IsAvailable == false)
+                throw new NotAvailableException("The function is not available.");
 
             Native.EqualizerClear(Player.Handle).
                 ThrowIfFailed(Player, "Failed to clear equalizer effect");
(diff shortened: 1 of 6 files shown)


src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
```diff
@@ -135,9 +150,13 @@ public Range BandLevelRange
         /// <summary>
         /// Gets the value whether the AudioEffect is available or not.
         /// </summary>
+        /// <remarks>This function returns the availability of the audio effect function group and
+        /// it could be unavailable depending on the platform capabilities.</remarks>
```
journey2w: /// <remarks>This function returns the availability of the audio effect function group.
=> What is audio effect function group? Is there any specific reason to use the phrase?

Moreover, if you want to use a conjunction, but is better. I think, however, a conjunction is not needed.

How about the following phrase?
/// <remarks>This function returns the availability of the audio effects.
/// Depending on the platform capabilities, it could be unavailable. </remarks>

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -667,6 +676,56 @@ public float AudioPitch
             }
         }
 
+        /// <summary>
+        /// Gets or sets the default codec type of the audio decoder.
+        /// </summary>
+        /// <value>A <see cref="CodecType"/> that specifies the type.
```
journey2w: It is a noun clause. The clause is awkward.

How about writing a sentence?
/// <value>A <see cref="CodecType"/> specifies the type.

src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -667,6 +676,56 @@ public float AudioPitch
...
+        /// The default codec type could be different depending on the device capability.</value>
+        /// <remarks>
+        /// <para>To set, the player must be in the <see cref="PlayerState.Idle"/> state.</para>
+        /// <para>If H/W audio codec type is not supported in some cases, S/W audio codec type could be used instead.</para>
+        /// <para>The availability could be changed depending on the codec capability.
+        /// If an application wants to use the H/W audio codec type as default,
+        /// following functions should be called after the codec type is set. :<br/>
```
journey2w: Suggestion:
        /// The following functions should be called after the codec type is set. :<br/>

src/Tizen.Multimedia.MediaPlayer/Player/AudioEffect.cs
```diff
@@ -135,9 +150,13 @@ public Range BandLevelRange
...
         /// <exception cref="InvalidOperationException">
         ///     If audio offload is enabled by calling <see cref="AudioOffload.IsEnabled"/>. (Since tizen 6.0)
         /// </exception>
+        /// <exception cref="NotAvailableException">The function is not available depending on the audio codec type. (Since tizen 6.0)</exception>
```
journey2w: This sentence is not bad.
However, how about the following sentence?

 /// <exception cref="NotAvailableException">Depending on the audio codec type, the function is not available. (Since tizen 6.0)

If you think the suggested phrase is better, please use it for others.

PR #1045, opened 2019-09-30
Patch for Email Emulator TCT issues

chauhan321:
### Description of Change ###
Email Emulator TCT were getting failed , proper handling of the callback result is achieved in the async function.

### API Changes ###
NA

### DIFF
--- a/src/Tizen.Messaging/Tizen.Messaging.Email/EmailSender.cs
+++ b/src/Tizen.Messaging/Tizen.Messaging.Email/EmailSender.cs
@@ -15,6 +15,7 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace Tizen.Messaging.Email
@@ -25,13 +26,16 @@ namespace Tizen.Messaging.Email
     /// <since_tizen> 3 </since_tizen>
     public static class EmailSender
     {
+	static private Dictionary<int, Interop.Email.EmailSentCallback> _sendCbMap = new Dictionary<int, Interop.Email.EmailSentCallback>();
+	static private int _callbackId = 0;
+
         /// <summary>
         /// Sends the email message.
         /// </summary>
         /// <param name="email">The email message.</param>
         /// <returns> Failure if the email sending activity failed, otherwise Success.</returns>
-        /// <since_tizen> 3 </since_tizen>
-        public static async Task<EmailSendResult> SendAsync(EmailMessage email)
+	/// <since_tizen> 3 </since_tizen>
+	public static async Task<EmailSendResult> SendAsync(EmailMessage email)
         {
             var task = new TaskCompletionSource<EmailSendResult>();
             int ret = (int)EmailError.None;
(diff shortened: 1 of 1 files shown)


src/Tizen.Messaging/Tizen.Messaging.Email/EmailSender.cs
```diff
@@ -40,35 +44,39 @@ public static async Task<EmailSendResult> SendAsync(EmailMessage email)
...
+            int id = _callbackId++;
+            _sendCbMap[id] = (IntPtr handle, int result, IntPtr userData) =>
             {
-                task.SetResult((EmailSendResult)result);
+                task?.SetResult((EmailSendResult)result);
+                _sendCbMap.Remove((int)userData);
             };
 
-            ret = Interop.Email.SetCb(email._emailHandle, _emailSendingCallback, IntPtr.Zero);
+            ret = Interop.Email.SetCb(email._emailHandle, _sendCbMap[id], (IntPtr)id);
```
myroot: SetCb is set a global callback for recevice result of sent.
but, it could be overwittten by next method call that was called before complete a previous request.

chauhan321: All the callbacks are differentiated with id , so they will not interfere each other.

myroot: Ok, it stored with mail id and it can save a multiple mail id

src/Tizen.Messaging/Tizen.Messaging.Email/EmailSender.cs
```diff
@@ -40,35 +44,39 @@ public static async Task<EmailSendResult> SendAsync(EmailMessage email)
             email.FillHandle();
             email.Save();
 
-            Interop.Email.EmailSentCallback _emailSendingCallback = (IntPtr handle, int result, IntPtr userData) =>
+            int id = _callbackId++;
```
myroot: you do better to use some lock guard 

```suggestion
            lock(lockObjectForId) {
                  int id = _callbackId++;
                  ...
            }
```

src/Tizen.Messaging/Tizen.Messaging.Email/EmailSender.cs
```diff
@@ -40,35 +44,39 @@ public static async Task<EmailSendResult> SendAsync(EmailMessage email)
...
             if (ret != (int)EmailError.None)
             {
                 Log.Error(EmailErrorFactory.LogTag, "Failed to send email, Error code: " + (EmailError)ret);
+                _sendCbMap.Remove(id);
                 throw EmailErrorFactory.GetException(ret);
             }
 
-            var sendResult = await task.Task;
-
+            var sendResult = await task.Task.ConfigureAwait(false);
```
myroot: Email API was not safe for multi-thread, so, you should return to main thread

```suggestion
            var sendResult = await task.Task.ConfigureAwait(true);
```

PR #1060, opened 2019-10-07
[UIKit] Introduce Tizen.UIKit project

jaehyun0cho:
Tizen.UIKit project is newly introduced.

### API Changes ###

 - ACR:


WonyoungChoi:
@jaehyun0cho  This PR has too many build warnings more than 1600. Please resolve these warnings first.

WonyoungChoi:
@jaehyun0cho And please hide unnecessary APIs such like native binding methods (P/Invoke calls).

WonyoungChoi:
@jaehyun0cho 
**Wrong Directory Structure**
All modules in TizenFX have the same directory structure for consistency. 
Please consider changing the directory structure as follows:
```
[Module Root]/
    [Qualified Namespace]/Foo.cs   
```
For example:
```
Tizen.UIKit/
  Eina/Accessor.cs
  Eina/Array.cs
  ...
  UIKit/Bind.cs
  UIKit/Factory.cs
  ...
  UIKit.Eo/EoWrapper.cs
  UIKit.Eo/Globals.cs
  UIKit.Eo/NativeModule.cs
  ...
  UIKit.Canvas/Group.cs
  UIKit.Canvas/Object.cs
  ...
```

myroot:
Why did use suffix a `kit`? This is usually used on apple

https://developer.apple.com/documentation/UIKit
https://developer.apple.com/documentation/mapkit
https://developer.apple.com/documentation/appkit
https://developer.apple.com/documentation/watchkit
...
https://developer.apple.com/documentation

src/Tizen.UIKit/src/UIKit.Input/Interface.cs
```diff
@@ -0,0 +1,1182 @@
...
+/// 
+/// A &quot;pointer&quot; refers to the main pointing device, which could be a mouse, trackpad, finger, pen, etc... In other words, the finger id in any pointer event will always be 0.
+/// 
+/// A &quot;finger&quot; refers to a single point of input, usually in an absolute coordinates input device, and that can support more than one input position at at time (think multi-touch screens). The first finger (id 0) is sent along with a pointer event, so be careful to not handle those events twice. Note that if the input device can support &quot;hovering&quot;, it is entirely possible to receive move events without down coming first.
+/// 
+/// A &quot;key&quot; is a key press from a keyboard or equivalent type of input device. Long, repeated, key presses will always happen like this: down...up,down...up,down...up (not down...up or down...down...down...up).</summary>
+/// <since_tizen> 6 </since_tizen>
+[UIKit.Input.InterfaceConcrete.NativeMethods]
+[UIKit.Wrapper.BindingEntity]
+public interface IInterface :
```
m-macias: Please don't name your interface IInterface, because this name doesn't say what to expect from it to the user.

src/Tizen.UIKit/src/UIKit.Input/Key.cs
```diff
@@ -0,0 +1,981 @@
...
+    /// This is the field you typically use to uniquely identify a keyboard symbol, in keyboard shortcuts for example.</summary>
+    /// <value>Symbol name produced by key event.</value>
+    public System.String KeySym {
+        get { return GetKeySym(); }
+        set { SetKeySym(value); }
+    }
+
+    /// <summary>A UTF8 string if this keystroke has produced a visible string to be added.</summary>
+    /// <value>Visible string produced by this key event, in UTF8.</value>
+    public System.String String {
```
m-macias: You could just use `string` instead of `System.String`.

woohyun:
> Why did use suffix a `kit`? This is usually used on apple
> 
> https://developer.apple.com/documentation/UIKit
> https://developer.apple.com/documentation/mapkit
> https://developer.apple.com/documentation/appkit
> https://developer.apple.com/documentation/watchkit
> ...
> https://developer.apple.com/documentation

@myroot 
During we have tried to find out the best name of the namespace, we thought "kit" was not something for Apple, and could be used in common purpose. 
If you have better idea on this ~ please share. This one was the best among several candidates.

myroot:
>  we thought "kit" was not something for Apple, and could be used in common purpose.

But it is a truth that UIKit is the name of iOS UI frameworks.

I recommend to use just  `UI`

woohyun:
> > we thought "kit" was not something for Apple, and could be used in common purpose.
> 
> But it is a truth that UIKit is the name of iOS UI frameworks.
> 
> I recommend to use just `UI`

But the thing is that - we have internal namespace with UI. 
I don't think "Ui.Ui.Button" looks good.

myroot:
> 
> 
> > > we thought "kit" was not something for Apple, and could be used in common purpose.
> > 
> > 
> > But it is a truth that UIKit is the name of iOS UI frameworks.
> > I recommend to use just `UI`
> 
> But the thing is that - we have internal namespace with UI.
> I don't think "Tizen.Ui.Ui.Button" looks good.

You can change a middle name like below
"Tizen.UI.View.Button"
"Tizen.UI.Widget.Button"

rookiejava:
Apart from creating unnecessary misunderstanding by using the same name as other platforms, `Kit` suffix looks awkward for me in terms of **consistency** with other Tizen .NET APIs (As you know, we usually do not use any suffix, for example, `Tizen.Application`, `Tizen.Multimedia`, `Tizen.Location`,...  NOT `Tizen.ApplicationKit`, `Tizen.MultimediaKit`, `Tizen.LocationKit`,...). I couldn't seen any use of `Kit` suffix in Tizen .NET project.

woohyun:
@myroot @rookiejava 
As you know well, EFL# is a set if auto-binding codes from Eo classes.
Changing "middle name" costs a lot because we need to change all the class names manually.

As I talked - if there would be better namespace instead of UIKit, please give a recommendation.
It was voted among my team members, but I also think there could be better name for this.

jaehyun0cho:
Hello, @m-macias .
Thank you for the comments. We are resolving the issues that you commented.

I have a question about https://github.com/Samsung/TizenFX/pull/1060#discussion_r333872886
Could you explain which indentation you mentioned in the above case?
e.g. tab is missing before "public class GroupMemberAddedEventArgs"?

jaehyun0cho:
Now the following modification for each comment has been done and will be updated to this pull request.
For the rest of the comments, we are still finding ways to fix or modify because the codes are automatically generated.

1. Renames the namespace UIKit to CoreUI
https://github.com/Samsung/TizenFX/pull/1060#issuecomment-540941008
https://github.com/Samsung/TizenFX/pull/1060#issuecomment-543106054
https://github.com/Samsung/TizenFX/pull/1060#issuecomment-543518572

2. Renames the namespace UIKit.Ui to CoreUI.UI
https://github.com/Samsung/TizenFX/pull/1060#issuecomment-543511773

3. Adds Apache license header at the top of the .cs file
https://github.com/Samsung/TizenFX/pull/1060#discussion_r333889099

4. Adds documentation to the public code
https://github.com/Samsung/TizenFX/pull/1060#discussion_r333881430

5. Changes public to internal if the method or class should not be visible to outside
https://github.com/Samsung/TizenFX/pull/1060#discussion_r333874867
https://github.com/Samsung/TizenFX/pull/1060#discussion_r333872886

6. Uses correct directory structures based on namespaces
https://github.com/Samsung/TizenFX/pull/1060#issuecomment-539303805

jaehyun0cho:
The updated version of this pull request is https://github.com/Samsung/TizenFX/pull/1253

PR #1085, opened 2019-10-16
[NUI] Refine Attributes so that user can use Attribute to create the …

AdunFang:
…control in BaseComponents

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
+++ b/src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
@@ -25,12 +25,12 @@ namespace Tizen.NUI.Components
     /// </summary>
     /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
     [EditorBrowsable(EditorBrowsableState.Never)]
-    public class ViewAttributes : Attributes
+    public class ViewAttributes : Attributes, global::System.IDisposable
     {
         static public T CreateView<T>(string style) where T : View, new()
         {
             T ret = new T();
-            (Tizen.NUI.Components.StyleManager.Instance.GetAttributes("style") as ViewAttributes).ApplyToView(ret);
+            (Tizen.NUI.Components.StyleManager.Instance.GetAttributes(style) as ViewAttributes).ApplyToView(ret);
             return ret;
         }
 
@@ -100,6 +100,11 @@ namespace Tizen.NUI.Components
             if (attributes.MinimumSize != null)
             {
                 MinimumSize = new Size2D(attributes.MinimumSize.Width, attributes.MinimumSize.Height);
+            }
+
+            if (attributes.MaximumSize != null)
+            {
+                MaximumSize = new Size2D(attributes.MaximumSize.Width, attributes.MaximumSize.Height);
             }
 
             if (attributes.SizeModeFactor != null)
@@ -211,6 +216,16 @@ namespace Tizen.NUI.Components
             set;
         }
         /// <summary>
+        /// View MaximumSize
+        /// </summary>
+        /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Size2D MaximumSize
+        {
+            get;
+            set;
+        }
+        /// <summary>
         /// View SizeModeFactor
         /// </summary>
         /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
@@ -302,6 +317,10 @@ namespace Tizen.NUI.Components
             {
                 view.MinimumSize = MinimumSize;
             }
+            if (MaximumSize != null)
+            {
+                view.MaximumSize = MaximumSize;
+            }
             if (PositionUsesPivotPoint != null)
             {
                 view.PositionUsesPivotPoint = PositionUsesPivotPoint.Value;
@@ -336,15 +355,8 @@ namespace Tizen.NUI.Components
 
         protected virtual void RegisterStateChangeCallback(View view)
         {
-            view.FocusGained += (object sender, global::System.EventArgs e) =>
-            {
-                ApplyToView(view, ControlStates.Focused);
-            };
-
-            view.FocusLost += (object sender, global::System.EventArgs e) =>
-            {
-                ApplyToView(view, ControlStates.Normal);
-            };
+            view.FocusGained += FocusGained;
+            view.FocusLost += FocusLost;
         }
 
         public void DisApplyToView(View view)
@@ -352,6 +364,9 @@ namespace Tizen.NUI.Components
             if (appliedViews.ContainsKey(view))
             {
                 appliedViews.Remove(view);
+
+                view.FocusGained -= FocusGained;
+                view.FocusLost -= FocusLost;
             }
         }
 
@@ -363,6 +378,29 @@ namespace Tizen.NUI.Components
             }
         }
 
+        public void Dispose()
+        {
+            StyleManager.Instance.ThemeChangedEvent -= OnThemeChangedEvent;
+        }
+
         private Dictionary<View, int> appliedViews = new Dictionary<View, int>();
+
+        private void FocusGained(object sender, global::System.EventArgs e)
+        {
+            View view = sender as View;
+            if (null != view)
+            {
+                ApplyToView(view, ControlStates.Focused);
+            }
+        }
+
+        private void FocusLost(object sender, global::System.EventArgs e)
+        {
+            View view = sender as View;
+            if (null != view)
+            {
+                ApplyToView(view, ControlStates.Normal);
+            }
+        }
     }
 }
--- a/src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
+++ b/src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
@@ -45,7 +45,7 @@ namespace Tizen.NUI.Components
         }
 
         /// <summary>
-        /// Constructs a ViewAttributes that is a copy of 
+        /// Constructs a ViewAttributes that is a copy of Attributes
         /// </summary>
         /// <param name="attributes">Construct Attributes</param>
         /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
--- a/src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
+++ b/src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
@@ -14,7 +14,9 @@
  * limitations under the License.
  *
  */
+using System.Collections.Generic;
 using System.ComponentModel;
+using Tizen.NUI.BaseComponents;
 
 namespace Tizen.NUI.Components
 {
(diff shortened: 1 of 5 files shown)


src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
```diff
@@ -264,5 +280,89 @@ public override Attributes Clone()
...
+            {
+                ApplyToView(view, ControlStates.Normal);
+            };
+        }
+
+        public void DisApplyToView(View view)
+        {
+            if (appliedViews.ContainsKey(view))
+            {
+                appliedViews.Remove(view);
```
rabbitfor: I wonder if there's no need to unregister FocusGained and FocusLost callback from the view here.

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
```diff
@@ -25,14 +27,25 @@ namespace Tizen.NUI.Components
...
+
         /// <summary>
         /// Construct ViewAttributes.
         /// </summary>
         /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public ViewAttributes() : base() { }
+        public ViewAttributes() : base()
+        {
+            StyleManager.Instance.ThemeChangedEvent += OnThemeChangedEvent;
```
rabbitfor: IMHO, if you don't subscribe OnThemeChangedEvent from the StyleManager instance, the ViewAttribute instance will never be able to be collected by GC.

rabbitfor:
I have a question about the concept of `Theme` in the StyleManager.
In the situation below,

```C#
// Register button styles of 2 different themes to StyleManager.
// The default theme is "Summer"
StyleManager.Instance.RegisterStyle("BasicButton", "Summer", typeof(SummerBasicButtonStyle), true);
StyleManager.Instance.RegisterStyle("BasicButton", "Winter", typeof(WinterBasicButtonStyle));

// Create button with the style "BasicButton"
Button button = new Button("BasicButton");
root.Add(button);

// Summer style button is created
```

Then when the user changes the theme in StyleManager,

```
StyleManager.Instance.Theme = "Winter";
```

The button keeps the Summer style or it changes its style to Winter?

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
```diff
@@ -25,14 +27,25 @@ namespace Tizen.NUI.Components
     [EditorBrowsable(EditorBrowsableState.Never)]
     public class ViewAttributes : Attributes
     {
+        static public T CreateView<T>(string style) where T : View, new()
+        {
+            T ret = new T();
+            (Tizen.NUI.Components.StyleManager.Instance.GetAttributes("style") as ViewAttributes).ApplyToView(ret);
```
rabbitfor: `GetAttributes("style")` work?

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
```diff
@@ -264,5 +280,89 @@ public override Attributes Clone()
...
+                ApplyToView(view, ControlStates.Focused);
+            };
+
+            view.FocusLost += (object sender, global::System.EventArgs e) =>
+            {
+                ApplyToView(view, ControlStates.Normal);
+            };
+        }
+
+        public void DisApplyToView(View view)
```
hyunjushin: DisApply is not commoly used......   Please explain role of the function...

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
AdunFang: Will remove it in Dispose().

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
AdunFang: If don't unregister the callback, when some Views are disposed, the changing of theme will try to apply the attribute to disposed View, it will cause issue.

rabbitfor: I don't understand. I was talking about unregistering `FocusGained` and `FocusLost` callback from the `view`. But what you said is about he reason of removing `view` from the `appliedViews`.

rabbitfor: I was talking about unregistering `FocusGained` and `FocusLost` callbacks from the `view`. 
I think you should unregister  `FocusGained` and `FocusLost` callbacks from the `view` here.

But what you commented is the reason of removing `view` from the `appliedViews`.

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
AdunFang: It shall not be string, has been fixed.

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
AdunFang: I'm sorry I misunderstood you， the FocusGained and FocusLoast will be removed when Disapply.

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
AdunFang: User need to call DisApplyToView() before call the Dispose().

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
```diff
@@ -363,6 +378,29 @@ private void OnThemeChangedEvent(object sender, StyleManager.ThemeChangeEventArg
             }
         }
 
+        public void Dispose()
+        {
+            StyleManager.Instance.ThemeChangedEvent -= OnThemeChangedEvent;
```
rabbitfor: Your current code requests a user to call Dispose() explicitly. But I think there's better way.
How about this?
```C#
// Register OnThemeChanged callback to StyleManager
public virtual void ApplyToView(View view, ControlStates state = ControlStates.Normal)
{
  ...
  if (!appliedViews.ContainsKey(view))
  {
    if (!appliedViews.Count)
    {
      StyleManager.Instance.ThemeChangedEvent += OnThemeChangedEvent;
    }

    appliedViews.Add(view, 0);
    RegisterStateChangeCallback(view);
  }
}
```
```C#
// Unregister OnThemeChanged callback from StyleManager
public void DisApplyToView(View view)
{
  if (appliedViews.ContainsKey(view))
  {
    appliedViews.Remove(view);

    if (!appliedViews.Count)
    {
      StyleManager.Instance.ThemeChangedEvent -= OnThemeChangedEvent;
    }
  }
}
```

rabbitfor: By the way, why the attribute wants to receive ThemeChanged event from the StyleManager in the first place?
From the code, when theme has changed, the attribute just apply its properties again to the view. Can you explain why you need it?

rabbitfor: Sorry to bother you, but can you answer [this](https://github.com/Samsung/TizenFX/pull/1085#issuecomment-543152739) also?

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
```diff
@@ -264,5 +280,89 @@ public override Attributes Clone()
...
+
+        private void OnThemeChangedEvent(object sender, StyleManager.ThemeChangeEventArgs e)
+        {
+            foreach (KeyValuePair<View, int> keyValuePair in appliedViews)
+            {
+                ApplyToView(keyValuePair.Key);
+            }
+        }
+
+        private Dictionary<View, int> appliedViews = new Dictionary<View, int>();
```
agnelovaz: Why is a Dictionary used here instead of a List,  does the value (int) useful?  I only see the value set to 0 in all cases.
List has a Contains()  method, which you could use instead of ContainsKey().

agnelovaz: Would viewsWithAttributes be more descriptive than appliedViews?  (I had to check to see what was being applied)

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
agnelovaz: If a View is removed then automatically Disposal will not run as there would be a handle to the View in this class, right?
Will the user know to call this API (DisApplyToView), seems like something that should be handled internally.

agnelovaz: This API is not actually removing the attributes from the View, it just removes the View from a container.  So the naming is not right.

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
```diff
@@ -264,5 +280,89 @@ public override Attributes Clone()
...
+                view.HeightResizePolicy = HeightResizePolicy.Value;
+            }
+            if (SizeModeFactor != null)
+            {
+                view.SizeModeFactor = SizeModeFactor;
+            }
+
+            if (!appliedViews.ContainsKey(view))
+            {
+                appliedViews.Add(view, 0);
```
agnelovaz: Wonder if a better solution than appliedViews would be storing a flag with the View?  Then would not have to worry about managing the lifecyle of Views.

dongsug-song:
@AdunFang 
Hello, could you check the other people's review and revise the patch. Thank you.

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
AdunFang: @rabbitfor 
"Does the button keep the Summer style or does it change its style to Winter?"
The style will change to "Winter"

@agnelovaz 
"Will the user know to call this API (DisApplyToView), seems like something that should be handled internally."

Because we don't want to let BaseCompoents.View depend the Tizen.NUI.Components, so we can't remove the view-handle in the View.Dispose().

Or we can add an event about dispose in the Basecomponents, so we don't need the method DisApplyToView. 

Do you think this is feasible?

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
AdunFang: "Register OnThemeChanged in Constructor -> ApplyToView
Unregister OnThemeChanged in Dispose -> DisApplyToView"

The Attributes need to register the OnThemeChanged only once, but ApplyToView() maybe called many times, so I think it is better to register it in the constructor.

The Style and Attributes will let the View perceive the change of style in theme, because we don't want to let BaseComponents.View depend the Tizen.NUI.Components, so that we must register the OnThemeChanged and apply different style.

The current code of OnThemeChanged() has problem, I'm modifying it.

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
AdunFang: The Dictionary will be replaced by HashSet.

AdunFang: Ok, the name will be changed.

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
rabbitfor: I understand your point, but from my code above,
it checks `if (!appliedViews.Count)` before register OnThemeChanged, after then, it increments `Count` by adding view to `appliedViews`.
So I believe that it won't happen that the OnThemeChanged  is registered multiple times.

And I hope that the Attributes class is not IDisposable, because users are not likely to call Dispose() method.

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
agnelovaz: I would be worried about emitting a signal from a Dispose() function as I would expect the emitting object to have now been garbage collected.  So too late?  You could try it and share the result.

src/Tizen.NUI.Components/Attributes/ViewAttributes.cs
agnelovaz: Can you check the HashSet doesn't prevent a View from being Disposed by keeping a handle.
If so you might want to use WeakHandles.

hyunjushin:
please reopen after refactoring this
"- Applying BinableObject Pattern to bind View and Attributes
Move Background and Shadow properties to View from Components
In Button, provide Icon and Text property as ImageAttributes and TextAttributes repectively.
"

PR #1086, opened 2019-10-17
Revert "[Tizen.Network.Connection] Init/deinit without ManagedThreadId"

WonyoungChoi:
Reverts Samsung/TizenFX#1077


WonyoungChoi:
CAPI must be applied first

PR #1094, opened 2019-10-17
Deprecate of Tizen.Account.OAuth2

pvschakradhar:
ACR link : http://suprem.sec.samsung.net/jira/browse/TCSACR-289


WonyoungChoi:
This PR was merged before ACR approval because it should be included into M2 release.

PR #1099, opened 2019-10-22
[NUI] revert the patch for flexibleview animation

huayongxu:
### API Changes ###

 - ACR:


dongsug-song:
this is revert patch, so I merge.

PR #1104, opened 2019-10-24
[NUI] Refine codes to reduce code duplicaton

elishateng:
### API Changes ###

 - ACR:


dongsug-song:
@elishateng 
Hello, this has "ACR Required" tag. please update the TCSACR link in the comment, otherwise, this patch can not be merged... thank you.

PR #1107, opened 2019-10-30
[NUI] Open classes of components

AdunFang:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Attributes/DropDownAttributes.cs
+++ b/src/Tizen.NUI.Components/Attributes/DropDownAttributes.cs
@@ -37,7 +37,7 @@ namespace Tizen.NUI.Components
             SpaceBetweenButtonTextAndIcon = 0;
             Space = new Vector4(0, 0, 0, 0);
             ListRelativeOrientation = DropDown.ListOrientation.Left;
-            ListMargin = new Vector4(0, 0, 0, 0);
+            ListMargin = new Extents(0, 0, 0, 0);
             FocusedItemIndex = 0;
         }
         /// <summary>
@@ -76,7 +76,7 @@ namespace Tizen.NUI.Components
 
             if (attributes.ListMargin != null)
             {
-                ListMargin = new Vector4(attributes.ListMargin.X, attributes.ListMargin.Y, attributes.ListMargin.Z, attributes.ListMargin.W);
+                ListMargin = new Extents(attributes.ListMargin.Start, attributes.ListMargin.End, attributes.ListMargin.Top, attributes.ListMargin.Bottom);
             }
 
             if (attributes.ListSize != null)
@@ -172,7 +172,7 @@ namespace Tizen.NUI.Components
         /// <since_tizen> 6 </since_tizen>
         /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public Vector4 ListMargin
+        public Extents ListMargin
         {
             get;
             set;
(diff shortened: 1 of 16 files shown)


src/Tizen.NUI.Components/Controls/DropDown.cs
```diff
@@ -530,77 +505,19 @@ public ListOrientation ListRelativeOrientation
...
-        public int ListBottomMargin
-        {
-            get
-            {
-                return (int)dropDownAttributes.ListMargin.W;
-            }
-            set
-            {
-                dropDownAttributes.ListMargin.W = value;
+                dropDownAttributes.ListMargin = new Extents(value.Start, value.End, value.Top, value.Bottom);
```
zhouleonlei: OK

m-macias:
The removal of EditorBrowsableState.Never attribute requires an ACR on Jira. Please don't merge it.

dongsug-song:
@AdunFang 
@elishateng 
Hello, 
"ACR Required" tag was attached, please check if some hidden API is changed to be opened as public.
you need to do TCSACR and update the ACR link here.
before ACR is accepted, we can not merge the patch, as you know.
thank you.

hyunjushin:
CheckBox is inherited from SelectButton. But SelectButton looks also hidden.

dongsug-song:
@AdunFang 
Hello,
could you check the updated review comments and answer?

src/Tizen.NUI.Components/Attributes/DropDownAttributes.cs
```diff
@@ -306,12 +306,12 @@ public ImageAttributes CheckImageAttributes
         }
 
         /// <summary>
-        /// Right space from check image.
+        /// Gap of Check image to boundary.
```
agnelovaz: Is the boundary the border of the Control? If so then should it state which axis?  Could this not be done with Padding?

src/Tizen.NUI.Components/Controls/CheckBoxGroup.cs
```diff
@@ -153,7 +145,6 @@ public bool IsCheckedAll()
         /// <param name="state">The boolean state of the check box</param>
         /// <since_tizen> 6 </since_tizen>
         /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
-        [EditorBrowsable(EditorBrowsableState.Never)]
         public void CheckingAll(bool state)
```
agnelovaz: public void CheckAll(bool state)

src/Tizen.NUI.Components/Controls/CheckBoxGroup.cs
```diff
@@ -31,15 +31,13 @@ namespace Tizen.NUI.Components
     /// </code>
     /// <since_tizen> 6 </since_tizen>
     /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
-    [EditorBrowsable(EditorBrowsableState.Never)]
     public class CheckBoxGroup : SelectGroup
```
agnelovaz: Do we need to make this Public now?  Wondering if a SelectGroup is really needed, instead Buttons could be grouped by the parent container they are added too.   Currently grouped buttons need adding to a parent View and also added to a SelectGroup. I'm assuming that grouping buttons would need a visual (a box) indication not just conceptual group.

src/Tizen.NUI.Components/Controls/InputField.cs
```diff
@@ -537,7 +517,6 @@ protected virtual bool OnTextFieldKeyEvent(object source, KeyEventArgs e)
         /// <param name="h">Input Field' height.</param>
         /// <since_tizen> 6 </since_tizen>
         /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
-        [EditorBrowsable(EditorBrowsableState.Never)]
         protected void SetTextFieldSize2D(int w, int h)
```
agnelovaz: Should this have been renamed to remove 2D?

hyunjushin:
please reopen after refactoring this
"- Applying BinableObject Pattern to bind View and Attributes
Move Background and Shadow properties to View from Components
In Button, provide Icon and Text property as ImageAttributes and TextAttributes repectively.
"

PR #1114, opened 2019-11-01
[NUI] NaturalSize returns size set instead of Zero for CustomViews

agnelovaz:
Change-Id: I3e900c09df94d19b93d35d5bf947a99cefd5337b

### Description of Change ###
Previously CustomViews returned zero for NaturalSize.

Now return the size set if NaturalSize not overridden.

In the case of Controls derived from CustomView that have their size explicitly set via the View API, that size is returned as the NaturalSize, not the size that may have been set internally by native code.

### API Changes ###

 - ACR:


adeelkazmi:
Use the commit style of other commits in these repo, i.e. [NUI] ...

PR #1122, opened 2019-11-08
[NoACR][MachineLearning.Inference] NotSupportedException support

again4you:
If machine_learning.inference is false or does not exist, or NNStreamer
installed on target device, NotSupportedException occurs when calling
the C# API of MachineLearning Inference.


again4you:
### Submit v3
* Caching the checked result for performance improvement
* TC is failed when feature key is set but nnstreamer is not installed.

again4you:
### Submit v4
* Remove unnecessary `DllNotFound` Exception handler

PR #1124, opened 2019-11-14
[NUI] Fix the crash when setting MinimumSize, MaximumSize

dongsug-song:
### Description of Change ###
[NUI] Fix the crash when setting MinimumSize, MaximumSize

### API Changes ###
Nothing


dongsug-song:
no more review done, so I merge.

PR #1128, opened 2019-11-18
[WebView] Add Chromium.SetArguments

younghajung:
This patch adds Chromium.SetArguments that can be used to set
Chromium's commandline switches.

### DIFF
--- a/src/Tizen.WebView/Tizen.WebView/Chromium.cs
+++ b/src/Tizen.WebView/Tizen.WebView/Chromium.cs
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+using System;
+using System.ComponentModel;
+
 namespace Tizen.WebView
 {
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/Tizen.WebView/Tizen.WebView/Chromium.cs
```diff
@@ -41,5 +44,17 @@ public static int Shutdown()
...
         }
+
+        /// <summary>
+        /// Sets argument count and argument array for Chromium.
+        /// </summary>
+        /// <param name="argc">Count of arguments</param>
+        /// <param name="argv">Argument array. The first value of array must be program's name.</param>
+        /// <since_tizen> 6 </since_tizen>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static void SetArguments(int argc, string[] argv)
```
myroot: we don't need argc

myroot:
```suggestion
        public static void SetArguments(string[] args)
```

younghajung: Good point! Thanks.

PR #1136, opened 2019-11-20
[NUI] Padding bug in LayoutGroup fixed

agnelovaz:
#516 # Description of Change ###
LayoutGroup measuring child without using the parent's padding.
Resulted in LinearLayout not reducing the childs size when padding applied to parent
Issue #2 of Layouts bug reported.

### API Changes ###

 - ACR:


agnelovaz:
Will also be merged into API6

dongsug-song:
no more review done, so merged

PR #1142, opened 2019-11-22
[NUI] Prevent Height and Width specs being overwritten

agnelovaz:
Change-Id: I80a4e1166f35e4c09c116916262ac965de54b048

### Description of Change ###
When setting the WidthSpecification or HeightSpecification with an exact value
the Size.Width or Size.Height is set but this in turns triggers a callback which
sets Size(width,height) but uses the stored height or width value which may not be
the set Width or Height Specification hence sets them to 0;

### API Changes ###

 - ACR:


agnelovaz:
Will need to be pushed to API6 too,

dongsug-song:
no more review done, so merged

PR #1153, opened 2019-11-29
[Sensor] Addition of initial sensor values

pvschakradhar:
Updating initial data values of all the sensors

### DIFF
--- a/src/Tizen.Sensor/Tizen.Sensor/Plugins/Accelerometer.cs
+++ b/src/Tizen.Sensor/Tizen.Sensor/Plugins/Accelerometer.cs
@@ -118,7 +118,7 @@ namespace Tizen.Sensor
         }
 
         /// <summary>
-        /// Read initial accelerometer data.
+        /// Read accelerometer data synchronously.
         /// </summary>
         protected override void ReadData()
         {
--- a/src/Tizen.Sensor/Tizen.Sensor/Plugins/Accelerometer.cs
+++ b/src/Tizen.Sensor/Tizen.Sensor/Plugins/Accelerometer.cs
@@ -117,7 +117,7 @@ namespace Tizen.Sensor
             return count;
         }
 
-        private void GetInitialCount()
+        private void GetInitialData()
         {
             Interop.SensorEventStruct sensorData;
             int error = Interop.SensorListener.ReadData(ListenerHandle, out sensorData);
@@ -148,7 +148,7 @@ namespace Tizen.Sensor
                 DataUpdated?.Invoke(this, new AccelerometerDataUpdatedEventArgs(sensorData.values));
             };
 
-            GetInitialCount();
+            GetInitialData();
 
             int error = Interop.SensorListener.SetEventCallback(ListenerHandle, Interval, _callback, IntPtr.Zero);
             if (error != (int)SensorError.None)
--- a/src/Tizen.Sensor/Tizen.Sensor/Plugins/Accelerometer.cs
+++ b/src/Tizen.Sensor/Tizen.Sensor/Plugins/Accelerometer.cs
@@ -117,7 +117,10 @@ namespace Tizen.Sensor
             return count;
         }
 
-        private void GetInitialData()
+        /// <summary>
+        /// Read initial accelerometer data.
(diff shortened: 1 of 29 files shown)


src/Tizen.Sensor/Tizen.Sensor/Plugins/Accelerometer.cs
```diff
@@ -117,6 +117,22 @@ private static int GetCount()
             return count;
         }
 
+        private void GetInitialCount()
```
seseki17: I think "Count" is not a proper word for this function name. is it right?

pvschakradhar: Yes. I will replace it with Data.
"GetInitialData()", will this be fine ?

src/Tizen.Sensor/Tizen.Sensor/Sensor.cs
```diff
@@ -47,6 +47,10 @@ public abstract class Sensor : IDisposable
         private IntPtr _sensorHandle = IntPtr.Zero;
         private IntPtr _listenerHandle = IntPtr.Zero;
 
+        /// <summary>
+        /// Read initial sensor data.
```
myroot: Read a sensor data synchronously.

pvschakradhar: comment updated, Thank you

src/Tizen.Sensor/Tizen.Sensor/Sensor.cs
```diff
@@ -47,6 +47,10 @@ public abstract class Sensor : IDisposable
         private IntPtr _sensorHandle = IntPtr.Zero;
         private IntPtr _listenerHandle = IntPtr.Zero;
 
+        /// <summary>
+        /// Read a sensor data synchronously.
+        /// </summary>
+        protected virtual void ReadData() { }
```
myroot: `abstract` is more correct to enforce implementation it on base class
```suggestion
        protected abstract void ReadData() ;
```

pvschakradhar: yes, true that.

bwikbs:
At this point, This patch requires ACR. And our intention is not to release ``ReadData`` API to public.
So.. plz, change ``protected`` to ``internal``

myroot:
> At this point, This patch requires ACR. And our intention is not to release `ReadData` API to public.
> So.. plz, change `protected` to `internal`
`internal protected` is more proper
https://docs.microsoft.com/ko-kr/dotnet/csharp/language-reference/keywords/protected-internal

pvschakradhar:
> > At this point, This patch requires ACR. And our intention is not to release `ReadData` API to public.
> > So.. plz, change `protected` to `internal`
> 
> `protected internal` is more proper
> https://docs.microsoft.com/ko-kr/dotnet/csharp/language-reference/keywords/protected-internal

but while opting for protected internal,  we still need to release ReadData API to public since it possesses both protected and internal behaviour

myroot:
> > > At this point, This patch requires ACR. And our intention is not to release `ReadData` API to public.
> > > So.. plz, change `protected` to `internal`
> > 
> > 
> > `protected internal` is more proper
> > https://docs.microsoft.com/ko-kr/dotnet/csharp/language-reference/keywords/protected-internal
> 
> but while opting for protected internal, we still need to release ReadData API to public since it possesses both protected and internal behaviour


No, protected internal is not public api

myroot:
> 
> 
> > > > At this point, This patch requires ACR. And our intention is not to release `ReadData` API to public.
> > > > So.. plz, change `protected` to `internal`
> > > 
> > > 
> > > `protected internal` is more proper
> > > https://docs.microsoft.com/ko-kr/dotnet/csharp/language-reference/keywords/protected-internal
> > 
> > 
> > but while opting for protected internal, we still need to release ReadData API to public since it possesses both protected and internal behaviour
> 
> No, protected internal is not public api

My fault, It should be `internal`, sorry

bwikbs:
@pvschakradhar Thanks :smile:

PR #1179, opened 2019-12-11
[NUI] Add checking if the Timer is used in main thread

dongsug-song:
### Description of Change ###
[NUI] Add checking if the Timer is used in main thread

### API Changes ###
Nothing


dongsug-song:
same patch has been reviewed and merged in master branch.

PR #1192, opened 2019-12-13
[NUI] Refine Toast for jira GRE-1605/1628

zhouleonlei:
### Description of Change ###
1. Make Toast inherit from TextLabelStyle 
2. Remove Toast from the Window after hidden

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Toast.cs
+++ b/src/Tizen.NUI.Components/Controls/Toast.cs
@@ -76,7 +76,6 @@ namespace Tizen.NUI.Components
         private Window window = null;
 		protected TextLabel[] textLabels = null;
         private string strText = null;
-        private NPatchVisual toastBackground = null;
         private Timer timer = null;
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI.Components/Controls/Toast.cs
```diff
@@ -76,19 +75,9 @@ public static Toast FromText(string text, uint duration)
 
         private Window window = null;
 		protected TextLabel[] textLabels = null;
-
-        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        protected TextLabel textLabel = null;
-
         private string strText = null;
         private NPatchVisual toastBackground = null;
```
rabbitfor: View has Background. (Property.Map)
I think it is not necessary to have npatch visual in Toast.

PR #1220, opened 2019-12-19
[NUI] Add NoOrientationPreference

sparrow74:
### Description of Change ###
[NUI] Add NoOrientationPreference

Add NoOrientationPreference to WindowOrientation enum.
It is used to unset the preferred orientation with SetPreferredOrientation.

### API Changes ###
Nothing


dongsug-song:
it is same PR in master branch, so merge now.

PR #1224, opened 2019-12-19
[NUI] Revise NUI samples

dongsug-song:
### Description of Change ###
[NUI] Revise NUI samples

### API Changes ###
Nothing


dongsug-song:
@zhouleonlei 
Hello,
I removed VD resources, it shouldn't be here. (I will remove them also in nui-dali windows-backend.)
I changed some and want to ask some question.

the "dali-theme.json" seems to be used, and there is {APPLICATION_RESOURCE_PATH} macro in that file. I don't know where this is set. is this dali json theme really needed?

more over, we need to revise all other samples which are biased to FHUB. we should not use the word of FHUB, DA or VD. Of course, we got a permission to use DA's resources but this is public sample so we must not mention about commercial product. 
please check in your side also.
thank you.

zhouleonlei:
> @zhouleonlei
> Hello,
> I removed VD resources, it shouldn't be here. (I will remove them also in nui-dali windows-backend.)
> I changed some and want to ask some question.
> 
> the "dali-theme.json" seems to be used, and there is {APPLICATION_RESOURCE_PATH} macro in that file. I don't know where this is set. is this dali json theme really needed?
> 
> more over, we need to revise all other samples which are biased to FHUB. we should not use the word of FHUB, DA or VD. Of course, we got a permission to use DA's resources but this is public sample so we must not mention about commercial product.
> please check in your side also.
> thank you.

Hello Mr.Song
"dali-theme.json"  is useful for Demo background.
And APPLICATION_RESOURCE_PATH is in Dali:
const char* APPLICATION_RESOURCE_PATH_KEY = "APPLICATION_RESOURCE_PATH";
  mThemeBuilderConstants[ APPLICATION_RESOURCE_PATH_KEY ] = Application::GetResourcePath();
But the images's path of dali-theme.json are missed.

Thanks.

PR #1240, opened 2019-12-26
[NUI.Components] Fix BackgroundImage doesn't works issue

elishateng:
### Description of Change ###

1. Fix BackgroundImage doesn't works issue

### API Changes ###

 - ACR:


dongsug-song:
By Mr. Fang =>

The reason is:
In .NetCore, if the static member hasn't been called, it will not be visited.
For example:
If the code is:

```

public class A
{
    public static List<string> list = new new List<string>;
}

A a = new A();


```
the A.list will not be created.

In our case, when user create a Button, because they haven't visit the static bindable propertys of Control, so these propertys will not be created, then when the ButtonStyle is applied to the Button, the background will not be set.

PR #1246, opened 2019-12-27
[NUI.Components] Fix Background Image not works issue

elishateng:
### Description of Change ###

1. Fix Background Image not works issue

### API Changes ###

 - ACR:


AdunFang (commented):
It is OK.

PR #1250, opened 2019-12-27
[NUI] Add to get current window orientation.

sparrow74:
Add the API to get current window orientation.
It is to return the logical orientation


dongsug-song:
same as https://github.com/Samsung/TizenFX/pull/1249

PR #1273, opened 2020-01-06
[NUI] Fixed the type conversion issue which cause TCT failed

zhouleonlei:
### Description of Change ###
Fixed the type conversion issue which cause TCT failed


AdunFang (commented):
It is OK.

PR #1288, opened 2020-01-09
[NUI] Implement switch's Old background API

zhouleonlei:
### Description of Change ###
Implement switch's Old background API


AdunFang (commented):
It is OK.

PR #1345, opened 2020-01-29
[NUI] Add GaussianBlurView

neostom432:
### Description of Change ###
Add GaussianBlurView.

For now user can use this class internally, because APIs are all hidden.

### API Changes ###
NONE

### DIFF
--- a/src/Tizen.NUI/src/internal/GaussianBlurView.cs
+++ b/src/Tizen.NUI/src/internal/GaussianBlurView.cs
@@ -252,13 +252,13 @@ namespace Tizen.NUI
         }
 
 
-        private void SetBackgroundColor(Vector4 color)
+        internal void SetBackgroundColor(Vector4 color)
         {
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/internal/GaussianBlurView.cs
```diff
@@ -55,179 +81,205 @@ protected override void Dispose(DisposeTypes type)
...
             {
-                get
-                {
-                    return _gaussianBlurView;
-                }
-                set
-                {
-                    _gaussianBlurView = value;
-                }
+                FinishedSignal().Disconnect(_finishedCallback);
```
dongsug-song: this disconnecting signal should be placed above the delete_GaussianBlurView(swigCPtr), because the signal must disconnected before the native object is actually deleted. please check again.

neostom432: DONE

PR #1348, opened 2020-01-30
[NUI] Fix AlphaMaskUrl setting order problem

dongsug-song:
### Description of Change ###
[NUI] Fix AlphaMaskUrl setting order problem
- adding private variable of _alphaMaskUrl.
- if _alphaMaskUrl is set once, AlphaMaskUrl property is added always in PropertyMap of UpdateImage()

### API Changes ###
nothing


dongsug-song:
same patch has been merged in API6

PR #1350, opened 2020-01-30
[NUI] Add ScrollToIndex to ScrollableBase

neostom432:
### Description of Change ###
Scrolls to the item at the specified index.

### API Changes ###
NONE

### DIFF
--- a/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
+++ b/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
@@ -381,6 +381,43 @@ namespace Tizen.NUI.Components
             mScrollingChild = new View();
         }
 
+
+        /// <summary>
+        /// Scrolls to the item at the specified index.
+        /// </summary>
+        /// <param name="index">Index of item.</param>
+        /// <since_tizen> 6 </since_tizen>
+        /// This may be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI.Components/Controls/ScrollableBase.cs
```diff
@@ -381,6 +381,43 @@ public override void OnChildRemove(View view)
...
+
+            float targetPosition = 0;
+
+            for (int i = 0; i < index; i++)
+            {
+                // SetSize doesn't change specification so check both
+                float expectedWidth = Math.Max(mScrollingChild.Children[i].WidthSpecification, mScrollingChild.Children[i].Size.Width);
+                float expectedHeight = Math.Max(mScrollingChild.Children[i].HeightSpecification, mScrollingChild.Children[i].Size.Height);
+
+                targetPosition += ScrollingDirection == Direction.Vertical ? expectedHeight : expectedWidth;
```
rabbitfor: How about make them inline.

I think current code calculates both width and height values unnecessarily.
```
targetPosition += ScrollingDirection == Direction.Vertical ? Math.Max(mScrollingChild.Children[i].HeightSpecification, mScrollingChild.Children[i].Size.Height) : Math.Max(mScrollingChild.Children[i].WidthSpecification, mScrollingChild.Children[i].Size.Width);
```

neostom432: Done.

And use position instead of size for calculating scrolling position.

PR #1379, opened 2020-02-07
[NUI] Fix visual bugs

rabbitfor:
* Fix ths issue that the VisualMap does not tranform
* Fix TextVisual crash issue

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Visuals/TextVisual.cs
+++ b/src/Tizen.NUI/src/public/Visuals/TextVisual.cs
@@ -23,10 +23,11 @@ namespace Tizen.NUI
     /// <since_tizen> 3 </since_tizen>
     public class TextVisual : VisualMap
     {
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Visuals/TextVisual.cs
```diff
@@ -23,10 +23,11 @@ namespace Tizen.NUI
     /// <since_tizen> 3 </since_tizen>
     public class TextVisual : VisualMap
     {
+        static private float defaultPointSize = 12;
```
hyunjushin: is this similar with DALi?

rabbitfor: I changed value 12 to 18, which is used in DALi as a default value for TextLabel

PR #1411, opened 2020-02-14
[MessagePort] Fix Listen Method

hjhun:

### Description of Change ###

To avoid throwing an exception in the callback function, this patch adds
an exception handling about Creating Bundle().


swift-kim:
This shouldn't be fixed in this way. You don't know the actual cause of the crash. At least there's no need to use try-catch. You can check if message is 0 or not and return.

hjhun:
> This shouldn't be fixed in this way. You don't know the actual cause of the crash. At least there's no need to use try-catch. You can check if message is 0 or not and return.

I agree.  But, you know, the message cannot be zero. The native code checks whether the message is nullptr or not. This is a workaround for the issue. I know this is not a solution for the issue.
The app developer has to be analyzed the memory corruption issue. But, he doesn't check it.
And, I think the callback function should not throw the exception directly.

Issue #1356, opened 2020-02-02
Expose existing keys via other means than an Exception in Data manager.


Mrnikbobjeff:
DataManager acts as an Interop wrapper for ckm. Ckm itself exposes the existence of keys via an error code. Currently, you either have to check all aliases on every key set to check whether the alias already exists, and remove it if it does before setting it again OR catch the exception, delete the key and try again. I would request three changes in total:
1. Change the return type to be a HashSet for the GetAliases method. This is achievable by writing a new class which simply takes the type of collection to be used as a template parameter. Have the current implementation extend that class with the template argument type list. This way existing code is not changed for other uses, but we can easily plug a new collection type in where appropriate. Also make sure to not decay the return type as is the case now.
2. Expose a way to override existing keys. Ideally by exposing an overload allowing exactly that.
3. Expose a way to check for key existence. Ckm has error codes for both existing an nonexistent keys. Simply add functions use these error codes to return a book.

PR #1415, opened 2020-02-16
[SecureRepository] Changed return type to HashSet for faster alias existence checks.

Mrnikbobjeff:
### Description of Change ###
#1356 was the ticket spawning this pr. DataManager and a couple of other classes expose an IEnumerable<string>, which is always a List currently. I believe all 3 occurances can be changed to HashSet, as the aliases should always be unique. If the general change would be accepted I would push the remaining 2 changes for KeyManager and CertificateManager which follow the same principle. Also, do I have to update the since_tizen documentation tag when changing the return type to a derived type?

### API Changes ###
Changed:
 - IReadonlyCollection<string> Tizen.Security.SecureRepository.DataManager.GetAliases(string key) 
The actual return type is HashSet.

### DIFF
--- a/src/Tizen.Security.SecureRepository/Tizen.Security.SecureRepository/DataManager.cs
+++ b/src/Tizen.Security.SecureRepository/Tizen.Security.SecureRepository/DataManager.cs
@@ -73,7 +73,7 @@ namespace Tizen.Security.SecureRepository
         /// <since_tizen> 3 </since_tizen>
         /// <returns>All aliases of data, which the client can access.</returns>
         /// <exception cref="ArgumentException">No alias to get.</exception>
-        static public IEnumerable<string> GetAliases()
+        static public IReadOnlyCollection<string> GetAliases()
         {
             IntPtr ptr = IntPtr.Zero;
 
@@ -82,7 +82,7 @@ namespace Tizen.Security.SecureRepository
                 Interop.CheckNThrowException(
                     Interop.CkmcManager.GetDataAliasList(out ptr),
                     "Failed to get data aliases");
-                return new SafeAliasListHandle(ptr).Aliases;
+                return new SafeAliasContainerHandle<HashSet<string>>(ptr).Aliases;
             }
             finally
             {
--- a/src/Tizen.Security.SecureRepository/Tizen.Security.SecureRepository/DataManager.cs
+++ b/src/Tizen.Security.SecureRepository/Tizen.Security.SecureRepository/DataManager.cs
@@ -73,7 +73,7 @@ namespace Tizen.Security.SecureRepository
         /// <since_tizen> 3 </since_tizen>
         /// <returns>All aliases of data, which the client can access.</returns>
         /// <exception cref="ArgumentException">No alias to get.</exception>
-        static public IReadOnlyCollection<string> GetAliases()
+        static public IEnumerable<string> GetAliases()
         {
             IntPtr ptr = IntPtr.Zero;
 
--- a/src/Tizen.Security.SecureRepository/Tizen.Security.SecureRepository/SafeAliasListHandle.cs
+++ b/src/Tizen.Security.SecureRepository/Tizen.Security.SecureRepository/SafeAliasListHandle.cs
@@ -21,12 +21,12 @@ using static Interop;
 
 namespace Tizen.Security.SecureRepository
 {
-    internal class SafeAliasListHandle
+    internal class SafeAliasContainerHandle<T> where T : ICollection<string>, new() 
     {
-        public SafeAliasListHandle(IntPtr ptr)
+        public SafeAliasContainerHandle(IntPtr ptr)
         {
             var cur = ptr;
-            var aliases = new List<string>();
+            var aliases = new T();
             while (cur != IntPtr.Zero)
             {
                 var ckmcAliasList = Marshal.PtrToStructure<Interop.CkmcAliasList>(cur);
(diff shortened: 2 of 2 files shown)


src/Tizen.Security.SecureRepository/Tizen.Security.SecureRepository/DataManager.cs
```diff
@@ -73,7 +73,7 @@ static public byte[] Get(string alias, string password)
         /// <since_tizen> 3 </since_tizen>
         /// <returns>All aliases of data, which the client can access.</returns>
         /// <exception cref="ArgumentException">No alias to get.</exception>
-        static public IEnumerable<string> GetAliases()
+        static public IReadOnlyCollection<string> GetAliases()
```
myroot: Why do you change the return type?

src/Tizen.Security.SecureRepository/Tizen.Security.SecureRepository/SafeAliasListHandle.cs
```diff
@@ -21,12 +21,12 @@
 
 namespace Tizen.Security.SecureRepository
 {
-    internal class SafeAliasListHandle
+    internal class SafeAliasContainerHandle<T> where T : ICollection<string>, new()
```
myroot: Why do you make as generic type? it just use for once, it seem to not need a generic. 
You just change HashSet from List, it is enough

Mrnikbobjeff: Currently it is used in two places, as the base class for the SafeAliasListHandle which is still used as outlined in the pr description, and in DataManager where I already changed the behaviour. I did it this way because I wanted the other usages to still work if they return the same values multiple times. I assume KeyManager and CertificateManager also return unique values? Then this is indeed not needed and I can adapt it.

src/Tizen.Security.SecureRepository/Tizen.Security.SecureRepository/DataManager.cs
Mrnikbobjeff: Even though the IEnumerable.Count() function will not enumerate all items, I believe it would be be beneficial to communicate to the caller that calling count will not do so. IReadOnlyCollection is literally IEnumerable<T> with the Count property, I can change it back if it not convincing enough.

myroot: You can't change public API

myroot:
Why do you want to change to HashSet from List?
even though HashSet is faster than List when to find a duplicate item.
but, in your code, does not insert a duplicate item. so it is not an advantage.

And a public API can't change return type, So, GetAliases can't returning as HashSet type even though it was hashset.

Mrnikbobjeff:
HashSet offers o(1) checks to verify that an alias exists. A list does not. Even if we do not insert the same item multiple times the retrieval is still faster. Even if we only return ienumerable the contains function will be o(1). So no matter what I insert or whether I return IEnumerable<T>, HashSet or IReadonlyCollection, HashSet will always perform better here.

myroot:
> 
> 
> HashSet offers o(1) checks to verify that an alias exists. A list does not. Even if we do not insert the same item multiple times the retrieval is still faster. Even if we only return ienumerable the contains function will be o(1). So no matter what I insert or whether I return IEnumerable, HashSet or IReadonlyCollection, HashSet will always perform better here.


I also agree to you about that HashSet is more efficiency when to check existence if we provide as ICollection, but we only provide as IEnumerable type and it does not provide `Contains` method. so no way to get benefit of HashSet

myroot:
If you change the implemetation of `AliasExists` method with `SafeAliasContainerHandle<HashSet>` like below
``` c#
static public bool AliasExists(string alias) {
            IntPtr ptr = IntPtr.Zero;
            try
            {
                Interop.CheckNThrowException(
                    Interop.CkmcManager.GetDataAliasList(out ptr),
                    "Failed to get data aliases");
                return (new SafeAliasContainerHandle<HashSet<string>>(ptr).Aliases).Contains(alias);
            }
            finally
            {
                if (ptr != IntPtr.Zero)
                    Interop.CkmcTypes.AliasListAllFree(ptr);
            }
}
```
I also agree to change to `SafeAliasContainerHandle<HashSet>`

myroot:
> > HashSet offers o(1) checks to verify that an alias exists. A list does not. Even if we do not insert the same item multiple times the retrieval is still faster. Even if we only return ienumerable the contains function will be o(1). So no matter what I insert or whether I return IEnumerable, HashSet or IReadonlyCollection, HashSet will always perform better here.
> 
> I also agree to you about that HashSet is more efficiency when to check existence if we provide as ICollection, but we only provide as IEnumerable type and it does not provide `Contains` method. so no way to get benefit of HashSet


I'm missing `Linq`. I agree to you. It more efficiency to check contains with linq `Contains` method

Mrnikbobjeff:
Regarding your first comment, I would advise you to check the Linq implementation for contains. Like many other Linq operations it checks for a faster path, e.g. checking whether the ienumerable is actually an icollection. You can see the referensource implementation in the link at the end. Regarding your second comment, I would very strongly advise against that approach. The whole reason I created this pr was to have a faster way to check for key existence without much overhead. If we were to marshal all keys every time we want to check for key existence we will have terrible performance losses for absolutely no gain. This is the way the people at Xamarin.Essentials have to handle it currently, but the situation was so unsatisfactory.

src/Tizen.Security.SecureRepository/Tizen.Security.SecureRepository/DataManager.cs
Mrnikbobjeff: I would address the issue tomorrow. I would also just change the return type back to innumerable as the count() method is special cases for collection anyway so there is not much gain there

PR #1432, opened 2020-02-24
[NUI] Support Nullable type in Xaml

AdunFang:
### Description of Change ###
Support Nullable type in Xaml,
so that user can descript selector in Xaml.

### API Changes ###

 - ACR:


dongsug-song:
@AdunFang 
Hello, can you make PR on API7 branch also?
thank you.

PR #1437, opened 2020-02-27
[NUI] Use padding & margin to meausre size in Layout

neostom432:
### Description of Change ###
Previously, padding of parent and margin of child are not concerned when measure child size.

For now, When child use MatchParent,
the size of child will be ParentSize - ParentPadding - childMargin.

And when child use WrapContent,
the size of child will be GrandchildSize + childPadding.

Also Weight will be use only ParentSize - ParentPadding - childMargin size for measuring.

### API Changes ###
NONE

### DIFF
--- a/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
+++ b/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
@@ -95,8 +95,8 @@ namespace Tizen.NUI.Components
                 }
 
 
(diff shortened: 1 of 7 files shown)


dongsug-song:
@neostom432 
Hello,
in my humble opinion, if sample app or TC(test case) code is added, we can check future defects more easily.
so it would be better to add sample app even later on.
thank you.

PR #1455, opened 2020-03-10
[NUI] Modify Window.Instance to Application.Instance.GetWindow()

huiyueun:

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window.cs
+++ b/src/Tizen.NUI/src/public/Window.cs
@@ -209,6 +209,11 @@ namespace Tizen.NUI
         {
             get
             {
+                if (Application.Instance is ComponentApplication)
+                {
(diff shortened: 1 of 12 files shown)


src/Tizen.NUI/src/public/Window.cs
```diff
@@ -209,6 +209,11 @@ public static Window Instance
         {
             get
             {
+                if (Application.Instance is ComponentApplication)
+                {
+                    Tizen.Log.Error("NUI", "You can't use 'Window.Intance' in ComponentApplication");
+                    throw NDalicPINVOKE.SWIGPendingException.Retrieve();
```
idkiller: Why can not get the window of the ComponentApplication?
How to NUI API User get windows of ComponentApplication?

PR #1457, opened 2020-03-10
[NUI] Check the previous width and height specifications

Seoyeon2Kim:

### Description of Change ###
- If new WidthSpecification and HeightSpecification values are same
 as the previous ones, then no need to request layout again.
- So, added to check that the new size width and height are different from the old ones.

### API Changes ###
- N/A


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -1941,24 +1947,28 @@ public int HeightSpecification
             set
             {
                 _heightPolicy = value;
-                if (_heightPolicy >= 0)
+                if( _oldHeightPolicy != _heightPolicy )
```
neostom432: The policy is under 0 means that user set LayoutParamPolicies.

In this case, size should not be set.

So, Please check if policy is under 0 too.
( _oldHeightPolicy !=  _heightPolicy && _heightPolicy >= 0)

neostom432: Never mind. I checked again and found that it was on your code already

PR #1460, opened 2020-03-12
[NUI] Make AccessibilityManager APIs as public ones

Seoyeon2Kim:

### Description of Change ###
- Move AccessibilityManager class from internal to public
- Added doxygen and make AccessibilityManager work in NUI
- This class will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.

### API Changes ###
- N/A


dongsug-song:
@Seoyeon2Kim 
Could you add some sample test code?

- example : 
https://github.com/Samsung/TizenFX/commit/597a56299982130f326547308433e73475ef8825
test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/DimTest.cs

thank you.

src/Tizen.NUI/src/public/AccessibilityManager.cs
```diff
@@ -31,11 +41,20 @@ internal AccessibilityManager(global::System.IntPtr cPtr, bool cMemoryOwn) : bas
...
         }
 
+        /// This will not be public opened.
+        [EditorBrowsable(EditorBrowsableState.Never)]
         protected override void ReleaseSwigCPtr(System.Runtime.InteropServices.HandleRef swigCPtr)
         {
             Interop.AccessibilityManage.delete_AccessibilityManager(swigCPtr);
         }
 
+        internal static AccessibilityManager GetAccessibilityManagerFromPtr(global::System.IntPtr cPtr)
```
dongsug-song: Please check if this API is really used. because this is "internal" so user can not use this. if this is not used in NUI. please remove it.

src/Tizen.NUI/src/public/AccessibilityManager.cs
```diff
@@ -503,412 +506,665 @@ private void OnFocusOvershot(IntPtr currentFocusedView, AccessibilityManager.Foc
...
         }
 
+        /// <summary>
+        /// Gets the singleton of AccessibilityManager object.
+        /// </summary>
+        /// <returns> A handle to the AccessibilityManager </returns>
+        /// <since_tizen> 6 </since_tizen>
+        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public static AccessibilityManager Get()
```
dongsug-song: can you check this AccessibilityManager is really sigleton? if so, please remove the "Get()" and use "Instance" like "FocusManager"

src/Tizen.NUI/src/public/AccessibilityManager.cs
```diff
@@ -503,412 +506,665 @@ private void OnFocusOvershot(IntPtr currentFocusedView, AccessibilityManager.Foc
...
+        /// and the focus order of the original view and all the views followed in the focus chain will be increased accordingly.
+        /// If the focus order assigned to the view is 0, it means that view's focus order is undefined
+        /// (e.g. the view has a  description but with no focus order being set yet) and therefore that view is not focusable.
+        /// </summary>
+        /// <param name="view"> the view to be set with</param>
+        /// <param name="order"> the focus order to be set with</param>
+        /// <since_tizen> 6 </since_tizen>
+        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public void SetFocusOrder(View view, uint order)
```
dongsug-song: looks like all the APIs are very similar to FocusManager's APIs. could you explain why?

src/Tizen.NUI/src/public/AccessibilityManager.cs
```diff
@@ -503,412 +506,665 @@ private void OnFocusOvershot(IntPtr currentFocusedView, AccessibilityManager.Foc
...
+        /// Moves the focus to the specified view.
+        /// Only one view can be focused at the same time. The view must have a defined focus order
+        /// and must be focusable, visible and in the window.
+        /// </summary>
+        /// <param name="view"> the view to be set with</param>
+        /// <returns> Whether the focus is successful or not </returns>
+        /// <since_tizen> 6 </since_tizen>
+        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public bool SetCurrentFocusView(View view)
```
dongsug-song: looks like all the APIs are very similar to FocusManager's APIs. could you explain why?

src/Tizen.NUI/src/public/AccessibilityManagerEventArgs.cs
```diff
@@ -0,0 +1,835 @@
...
+        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public class StatusChangedEventArgs : EventArgs
+        {
+            private AccessibilityManager _accessibilityManager;
+
+            /// <since_tizen> 6 </since_tizen>
+            /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
+            [EditorBrowsable(EditorBrowsableState.Never)]
+            public AccessibilityManager AccessibilityManager
```
dongsug-song: this seems to be an old code pattern. in EventHandler, the first argument is object which is sender. (in this case, AccessibilityManager) so you don't need to pass AccessibilityManger instance to event subscriber through the EventArgument. This pattern is old style, and it is seen in may places. please remove this.

src/Tizen.NUI/src/public/AccessibilityManagerEventArgs.cs
```diff
@@ -0,0 +1,835 @@
...
+            public class ActionScrollEventArgs : EventArgs
+            {
+              private AccessibilityManager _accessibilityManager;
+              private TouchEvent _touchEvent;
+
+              public AccessibilityManager AccessibilityManager
+              {
+                get
+                {
+                  return _accessibilityManager;
```
dongsug-song: this seems to be an old code pattern. in EventHandler, the first argument is object which is sender. (in this case, AccessibilityManager) so you don't need to pass AccessibilityManger instance to event subscriber through the EventArgument. This pattern is old style, and it is seen in may places. please remove this.

src/Tizen.NUI/src/public/AccessibilityManagerEvent.cs
```diff
@@ -795,6 +938,14 @@ internal partial class AccessibilityManager
...
         }
 
+        /// <summary>
+        /// This is emitted when accessibility action is received to focus and read from the first item on the top continuously
+        /// (by three fingers single tap).
+        /// </summary>
+        /// <returns> The signal to connect to </returns>
+        /// <since_tizen> 6 </since_tizen>
+        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
dongsug-song: Could you check this Event code pattern ? this is old style. you can refer the View event code... => tizenfx/src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs

=> no lock(), added some condition to check native signal status.

src/Tizen.NUI/src/public/AccessibilityManagerEvent.cs
```diff
@@ -826,6 +977,14 @@ internal partial class AccessibilityManager
...
 
+        /// <summary>
+        /// This is emitted when accessibility action is received to move the focus to and read from the next item continuously
+        /// (by three fingers double tap).
+        /// </summary>
+        /// <returns> The signal to connect to </returns>
+        /// <since_tizen> 6 </since_tizen>
+        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public event DaliEventHandlerWithReturnType<object, ActionReadFromNextEventArgs, bool> ActionReadFromNext
```
dongsug-song: Could you check this Event code pattern ? this is old style. you can refer the View event code... => tizenfx/src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs

=> no lock(), added some condition to check native signal status.

src/Tizen.NUI/src/public/AccessibilityManager.cs
Seoyeon2Kim: Okay, removed  it.
It was used in the signal internally of the previous patch, but now it's unnecessary. So, I removed it in the next patch.

src/Tizen.NUI/src/public/AccessibilityManager.cs
Seoyeon2Kim: Yes, it is "Manager", so should be Singleton class. Changed to Instance. Thanks.

src/Tizen.NUI/src/public/AccessibilityManager.cs
Seoyeon2Kim: When Accessibility is activated, focus is required as well. Of course, it does not use FocusManager internally, and it has its own focus implementation. The reason for the same API names is to be for the naming unity in DALi, which is C# binding.

src/Tizen.NUI/src/public/AccessibilityManager.cs
Seoyeon2Kim: When Accessibility is activated, focus is required as well. Of course, it does not use FocusManager internally, and it has its own focus implementation. The reason for the same API names is to be for the naming unity in DALi, which is C# binding.

src/Tizen.NUI/src/public/AccessibilityManagerEventArgs.cs
Seoyeon2Kim: Okay, Removed the old coding style and changed to the latest one. (similar to other classes)
Please  check it again.

src/Tizen.NUI/src/public/AccessibilityManagerEventArgs.cs
Seoyeon2Kim: Okay, Removed the old coding style and changed to the latest one. (similar to other classes)
Please  check it again.

src/Tizen.NUI/src/public/AccessibilityManagerEvent.cs
Seoyeon2Kim: Removed 'lock()' in all signals and changed the event code pattern.
I refer to ViewEvent.cs and FocusManager.cs  code.

src/Tizen.NUI/src/public/AccessibilityManagerEvent.cs
Seoyeon2Kim: Removed 'lock()' in all signals and changed the event code pattern.
I refer to ViewEvent.cs and FocusManager.cs  code.

PR #1393, opened 2020-02-11
[ElmSharp] Add ElmSharp.Wearable preloading

JoonghyunCho:
### Description of Change ###
This PR suggests to invoke ElmSharp.Wearable preloading method by using reflection on `ElmSharp`.

`ElmSharp.Wearable` preloads widgets which enhance the application startup time.

  - `CircleDateTimeSelector` :  ▽24.13ms on `CircleDateTimeSelectorRenderer`
  - `CircleSurface`, `CircleProgressBar`, `CircleSlider`, `MoreOption` : ▽5ms or ▽26ms(with `ToolbarItem`) on `CirclePageRenderer`
  - `CircleScroller` : ▽6.53ms on `CircleScrollViewRenderer`
  - `CircleSpinner` : ▽9.21ms on `CircleStepperRenderer`

### API Changes ###
- N/A

PR #1464, opened 2020-03-16
[ElmSharp] Add ElmSharp.Wearable preloading

JoonghyunCho:
### Description of Change ###
This PR adds invoking `ElmSharp.Wearable` preloading method by using reflection on `ElmSharp`, and adds `ElmSharp.PreloadedWindow.BaseLayout` and `ElmSharp.PreloadedWindow.BaseConformant`.

`ElmSharp.Wearable` preloads widgets which enhance the application startup time.

  - `CircleDateTimeSelector` :  ▽24.13ms on `CircleDateTimeSelectorRenderer`
  - `CircleSurface`, `CircleProgressBar`, `CircleSlider`, `MoreOption` : ▽5ms or ▽26ms(with `ToolbarItem`) on `CirclePageRenderer`
  - `CircleScroller` : ▽6.53ms on `CircleScrollViewRenderer`
  - `CircleSpinner` : ▽9.21ms on `CircleStepperRenderer`

### API Changes ###
- N/A

### DIFF
--- a/src/ElmSharp/ElmSharp/PreloadedWindow.cs
+++ b/src/ElmSharp/ElmSharp/PreloadedWindow.cs
@@ -35,7 +35,7 @@ namespace ElmSharp
 
             if (Elementary.GetProfile() == "wearable")
             {
-                PreloadElmSharpWearable();
+                WarmupWearableWidgets();
             }
         }
 
@@ -77,7 +77,7 @@ namespace ElmSharp
             //TODO: Consider to call Image.LoadAsync()
         }
 
-        public void PreloadElmSharpWearable()
+        public void WarmupWearableWidgets()
         {
             try
             {
--- a/src/ElmSharp/ElmSharp/PreloadedWindow.cs
+++ b/src/ElmSharp/ElmSharp/PreloadedWindow.cs
@@ -15,15 +15,20 @@
  */
 
 using System;
+using System.ComponentModel;
 using System.Reflection;
 
 namespace ElmSharp
 {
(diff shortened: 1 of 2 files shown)


JoonghyunCho:
Applying https://github.com/Samsung/TizenFX/pull/1393 to API7.

src/ElmSharp/ElmSharp/PreloadedWindow.cs
```diff
@@ -69,6 +96,21 @@ public void WarmupWidgets()
...
+        {
+            try
+            {
+                Assembly assem = Assembly.Load("ElmSharp.Wearable");
+                var type = assem.GetType("ElmSharp.Wearable.Preload");
+                type.GetMethod("WarmupWidgets", BindingFlags.Public | BindingFlags.Static).Invoke(null, new object[] { this });
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine(e.ToString());
```
darkleem: Catching all Exception results in the Console output only and the program is alive, no matter what the exception occurs. Is that what you intended?
I understand that there are no instances where exceptions are not forced to occur at the API level unless they are special.

JoonghyunCho: Yes. This reflection call will be invoked by a candidate process, and it calls preload feature in `ElmSharp.Wearable` which is technically not mandatory for launching an application.
An application should not be affected even if this call fails. At the same time, I thought it would be better to show the failure reason on console log.

sung-su:
LGTM

PR #1475, opened 2020-03-18
[NUI] Add Page related apis for Window

elishateng:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window.cs
+++ b/src/Tizen.NUI/src/public/Window.cs
@@ -34,6 +34,7 @@ namespace Tizen.NUI
         private Layer _rootLayer;
         private string _windowTitle;
         private List<Layer> _childLayers = new List<Layer>();
+        private Dictionary<Type, Page> pages = new Dictionary<Type, Page>();
         private LayoutController localController;
 
         internal Window(global::System.IntPtr cPtr, bool cMemoryOwn) : base(Interop.Window.Window_SWIGUpcast(cPtr), cMemoryOwn)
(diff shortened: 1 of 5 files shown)


src/Tizen.NUI/src/public/Window.cs
```diff
@@ -1117,6 +1118,64 @@ public void SetAvailableOrientations( List<Window.WindowOrientation> orientation
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
         }
 
+        /// <summary>
+        /// Adds a page to the window.
+        /// </summary>
+        /// <param name="page">Page to add.</param>
+        /// <since_tizen> 8 </since_tizen>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public void AddPage(Page page)
```
idkiller: Why add only one page per type? is it by design?

src/Tizen.NUI/src/public/XamlBinding/ContentPage.cs
```diff
@@ -110,6 +110,7 @@ public ContentPage(Window win)
             Root.HeightResizePolicy = ResizePolicyType.FillToParent;
 
             win.Add(Root);
+            win.AddPage(this);
```
idkiller: Because of this part, creating a new instance of the same type will trow an exception. is it intended?

src/Tizen.NUI/src/public/XamlBinding/ContentPage.cs
```diff
@@ -158,6 +159,7 @@ protected override void Dispose(DisposeTypes type)
                 Root.Dispose();
                 Root = null;
             }
+            Window.Instance.RemovePage(this);
```
idkiller: This code makes this class unusable before Window.Instance is initialized 
it cannot be used in static constructor or initialize of static member.
Is this by design?

src/Tizen.NUI/src/public/Window.cs
```diff
@@ -1117,6 +1118,64 @@ public void SetAvailableOrientations( List<Window.WindowOrientation> orientation
...
+        /// <summary>
+        /// Remove a page from the window.
+        /// </summary>
+        /// <param name="page">Page to remove.</param>
+        /// <since_tizen> 8 </since_tizen>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public void RemovePage(Page page)
+        {
+            if (pages.ContainsValue(page))
+            {
```
idkiller: Why check a value?
just use pages.Remove(page.GetType()) with null check.

src/Tizen.NUI/src/public/Window.cs
elishateng: Hello, Mr. Lee, Page is the base class of ContentPage.

elishateng:
> Only one page per type can be created and only contentPage has logic, but child of TemplatedPage or Page can be created also. (Not sealed and also can exist on xaml )

Hello, Mr. Lee, I use List to replace Dictionary, please review it again.

elishateng:
> Only one page per type can be created and only contentPage has logic, but child of TemplatedPage or Page can be created also. (Not sealed and also can exist on xaml )

Hello, Mr. Lee, Currently TemplatedPage and Page can't use in xaml, so I just change ContentPage.

src/Tizen.NUI/src/public/Window.cs
idkiller: I know...
In the previous review, I asked about this method implementation accepting only one instance per Type as a Dictionary.
It seems to be okay now.

src/Tizen.NUI/src/public/XamlBinding/Page.cs
```diff
@@ -211,6 +211,31 @@ public bool SendBackButtonPressed()
             return OnBackButtonPressed();
         }
 
+        /// <summary>
+        /// Equality operator
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static bool operator ==(Page x, Page y)
```
idkiller: == is same as ReferenceEquals.
why overload it???

src/Tizen.NUI/src/public/XamlBinding/Page.cs
```diff
@@ -211,6 +211,31 @@ public bool SendBackButtonPressed()
...
+            }
+            return false;
+        }
+
+        /// <summary>
+        /// Inequality operator. Returns Null if either operand is Null
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static bool operator !=(Page x, Page y)
```
idkiller: == is same as ReferenceEquals.
why overload it???

src/Tizen.NUI/src/public/XamlBinding/Page.cs
elishateng: Because Page inherit BaseHandle and BaseHandle overload ==/!=,  it will compare from native handle in ==/!= . We overload ==/!= to avoid throw exception when user check if page is null.

hyunjushin:
please discuss about  ContentPage refactoring with SSong.

dongsug-song:
@elishateng 
Hello,
As we have discussed on https://code.sec.samsung.net/jira/browse/GRE-1777 , I close this PR without merging.

AdunFang (commented):
It is OK.

PR #1479, opened 2020-03-19
[Feedback] Change struct array to Dictionary<string, int>

sanghyeok-oh:
### Description of Change ###
[Feedback][refactor] Change struct array to Dictionary<string, int>

### DIFF
--- a/src/Tizen.System.Feedback/Feedback/Feedback.cs
+++ b/src/Tizen.System.Feedback/Feedback/Feedback.cs
@@ -21,14 +21,6 @@ using System.Collections.Generic;
 namespace Tizen.System
 {
     /// <summary>
-    /// The class for constants.
-    /// </summary>
-    internal static class Constants
-    {
-        internal const int NumberOfPattern = 39;
-    }
-
-    /// <summary>
     /// The Feedback API provides functions to control haptic and sound.
     /// The Feedback API provides the way to play and stop feedback, and get the information whether a specific pattern is supported.
     /// Below is the supported pattern string:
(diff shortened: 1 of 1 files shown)


src/Tizen.System.Feedback/Feedback/Feedback.cs
```diff
@@ -316,26 +258,15 @@ public bool IsSupportedPattern(FeedbackType type, String pattern)
...
-            {
-                if (String.Compare(pattern, Pattern[i].PatternString) == 0)
-                {
-                    found = true;
-                    break;
-                }
-            }
-
-            if (!found)
+            if (!Pattern.ContainsKey(pattern))
```
idkiller: Using ContainsKey and Item operator at the same time is redundant, because we use ContainsKey inside Items operator.
I recommend using TryGetValue.
```C#
if (Pattern.TryGetValue(pattern, out int value))
{
    if (type == FeedbackType.All)
        res = (Interop.Feedback.FeedbackError)Interop.Feedback.Play(value);
    else
        res = (Interop.Feedback.FeedbackError)Interop.Feedback.PlayType((Interop.Feedback.FeedbackType)type, value);
}
else throw new ArgumentException("Invalid Arguments");
```

src/Tizen.System.Feedback/Feedback/Feedback.cs
```diff
@@ -250,24 +204,12 @@ public Feedback()
...
```
idkiller: Using ContainsKey and Item operator at the same time is redundant, because we use ContainsKey inside Items operator.
I recommend using TryGetValue.

```C#
if (Pattern.TryGetValue(pattern, out int value))
{
    res = res = (Interop.Feedback.FeedbackError)Interop.Feedback.IsSupportedPattern((Interop.Feedback.FeedbackType)type, value, out supported);
}
else throw new ArgumentException("Invalid Arguments");
```

src/Tizen.System.Feedback/Feedback/Feedback.cs
sanghyeok-oh: Thank you for point this, TryGetValue() seems more efficient way to access element of Dictionary. :)

src/Tizen.System.Feedback/Feedback/Feedback.cs
sanghyeok-oh: Thank you for point this, TryGetValue() seems more efficient way to access element of Dictionary. :)

src/Tizen.System.Feedback/Feedback/Feedback.cs
```diff
@@ -250,24 +204,13 @@ public Feedback()
...
-                if (String.Compare(pattern, Pattern[i].PatternString) == 0)
-                {
-                    found = true;
-                    break;
-                }
-            }
-
-            if (!found)
+            if (!Pattern.TryGetValue(pattern, out number))
                 throw new ArgumentException("Invalid Arguments");
```
WonyoungChoi: Provide move detail message and param name:
```C#
throw new ArgumentException($"Not supported pattern string : {pattern}", nameof(pattern));
```

src/Tizen.System.Feedback/Feedback/Feedback.cs
```diff
@@ -316,26 +259,16 @@ public bool IsSupportedPattern(FeedbackType type, String pattern)
...
```
WonyoungChoi: Provide move detail message and param name:
```C#
throw new ArgumentException($"Not supported pattern string : {pattern}", nameof(pattern));
```

src/Tizen.System.Feedback/Feedback/Feedback.cs
sanghyeok-oh: Done. :)

src/Tizen.System.Feedback/Feedback/Feedback.cs
sanghyeok-oh: Done. :)

PR #1481, opened 2020-03-23
[ElmSharp] Add internal Window auxiliary hint APIs

arosis78:
### Description of Change ###
Add internal Window auxiliary hint APIs

### API Changes ###
Added:
 - int AddAuxiliaryHint(string hint, string value)
 - bool DeleteAuxiliaryHint(int id)
 - bool SetAuxiliaryHintValue(int id, string value)
 - string GetAuxiliaryHintValue(int id)
 - int GetAuxiliaryHintId(string hint)

### DIFF
--- a/src/ElmSharp/ElmSharp/Window.cs
+++ b/src/ElmSharp/ElmSharp/Window.cs
@@ -1317,7 +1317,10 @@ namespace ElmSharp
         /// </summary>
         internal static void DisableWindowEffect(Window window)
         {
-            Interop.Elementary.elm_win_aux_hint_add(window.Handle, "wm.policy.win.effect.disable", "1");
+            if (window != null)
+            {
+                Interop.Elementary.elm_win_aux_hint_add(window.Handle, "wm.policy.win.effect.disable", "1");
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/ElmSharp/ElmSharp/Window.cs
```diff
@@ -1311,5 +1311,27 @@ static void Preload()
...
+        /// <summary>
+        /// Enable window effect. For internal use only.
+        /// </summary>
+        internal static void EnableWindowEffect(Window window)
+        {
+            if (window != null)
+            {
+                Interop.Elementary.elm_win_aux_hint_add(window.Handle, "wm.policy.win.effect.disable", "0");
+            }
+        }
```
rookiejava: Instead of tightly coupled with specific value (`wm.policy.win.effect.disable`), how about providing a method that can have parameter like the native platform for flexibility?

e.g. 
```cs
Window.AddAuxiliaryHint(string key, string value)
```

Or, if you want keep current API signature and if this change is only meaningful for wearable profile, how about adding it to ElmSharp.Wearable?

arosis78: Good Idea. I added all window auxiliary hint APIs.

arosis78:
aux_hint APIs are sent auxiliary hint to the window, but aux_msg APIs are recieved message from the window.
I think that aux_msg APIs are not need to add in ElmSharp now.

rookiejava:
please reflect the changes to `API7` and `master` . Thanks!

PR #1497, opened 2020-03-30
[NUI] Add SetPositionByLayout Property

neostom432:
### Description of Change ###
If true, set position by layouting result.
If false, will not set position in SetFrame because it means it's parent like
RecyclerList will set position.

### API Changes ###
NONE

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
@@ -74,6 +74,14 @@ namespace Tizen.NUI
          public bool LayoutWithTransition{get; set;}
 
         /// <summary>
+        /// [Draft] Set position by layouting result
+        /// </summary>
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
```diff
@@ -73,6 +73,14 @@ public class LayoutItem
         [EditorBrowsable(EditorBrowsableState.Never)]
          public bool LayoutWithTransition{get; set;}
 
+        /// <summary>
+        /// [Draft] Set position by layouting result
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
```
dongsug-song: This looks hidden API, so please remove <since_tizen> tag and the comment of ///This will be public opened ...

neostom432: DONE

PR #1499, opened 2020-03-30
[NUI] Remove ClearBackgound operate when setting BackgroundImage

elishateng:
### Description of Change ###

1. Remove ClearBackgound operate when setting BackgroundImage, otherwise blink issue exists.

### API Changes ###

 - ACR:


AdunFang (commented):
It is OK.

PR #1516, opened 2020-04-03
[NUI.Components] Fix focus state can't reach issue

elishateng:
### Description of Change ###

1. HasFocus always return true during OnFocusGained or OnFocusLost.

### API Changes ###

 - ACR:


AdunFang (commented):
It is OK.

PR #1521, opened 2020-04-08
[NUI] Add dotnet pre-initiailze

huiyueun:

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/NUIApplication.cs
+++ b/src/Tizen.NUI/src/public/NUIApplication.cs
@@ -374,6 +374,17 @@ namespace Tizen.NUI
         {
             base.OnCreate();
         }
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI/src/public/NUIApplication.cs
```diff
@@ -374,6 +374,17 @@ protected override void OnCreate()
         {
             base.OnCreate();
         }
+
+        /// <summary>
+        /// This is used to improve application launch performance.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        static public void PreLoad(string[] args)
```
JongHeonChoi, line 382: Called only the Preload method with no parameter in launcher
```
static public void Preload()
```

PR #1524, opened 2020-04-08
[AudioIO] Add new volume property to set/get recording volume

hsgwon:
### Description of Change ###

Add new volume property to set/get recording volume

### API Changes ###

 - TCSACR-336

### DIFF
--- a/src/Tizen.Multimedia.AudioIO/AudioIO/AudioCapture.cs
+++ b/src/Tizen.Multimedia.AudioIO/AudioIO/AudioCapture.cs
@@ -177,6 +177,41 @@ namespace Tizen.Multimedia
         public AudioSampleType SampleType { get; }
 
         /// <summary>
+        /// Gets or sets the volume of the audio input data stream.
+        /// </summary>
+        /// <value>
+        /// The default value is 1.0.<br/>
+        /// The valid range is greater than or equal to 0.0 and less than or equal to 2.0.<br/>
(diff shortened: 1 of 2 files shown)


src/Tizen.Multimedia.AudioIO/AudioIO/AudioCapture.cs
```diff
@@ -176,6 +176,41 @@ internal void ValidateState(params AudioIOState[] desiredStates)
...
+                var ret = AudioInput.GetVolume(_handle, out double volume);
+                MultimediaDebug.AssertNoError((int)ret);
+
+                return volume;
+            }
+            set
+            {
+                ValidateNotDisposed();
+
+                var ret = AudioInput.SetVolume(_handle, value < 0.0 ? 0.0 : value > 2.0 ? 2.0 : value);
```
hsgwon, line 209: @SeungbaeShin , I just want to avoid the possibility of occurring invalid range input value.
If it's done in native fw already, I'll remove this.

myroot, line 209:
```suggestion
                var ret = AudioInput.SetVolume(_handle, Math.Min(Math.Max(value, 0.0), 2.0));
```

hsgwon: OK I'll change it.

PR #1539, opened 2020-04-13
[NUI] Interrupt pan when popup is popped

neostom432:
### Description of Change ###
Currently, panning is working even if there is popup over component.
As workaround, add PanGestureDetector to Popup to prevent panning.

Additionally, add View to ScrollableBase which is to interrupt touching during scrolling.

### API Changes ###
NONE

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Popup.cs
+++ b/src/Tizen.NUI.Components/Controls/Popup.cs
@@ -200,6 +200,7 @@ namespace Tizen.NUI.Components
         private ButtonGroup btGroup = null;
         private Window window = null;
         private Layer container = new Layer();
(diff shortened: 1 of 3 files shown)


adeelkazmi:
You can do the same as we do with the native popup in C#. There we add a popup to a layer which is on the top and the layer is set to consume all touch.

https://github.com/dalihub/dali-core/blob/master/dali/public-api/actors/layer.h#L480

This is currently only internal in Layer.cs: https://github.com/dalihub/TizenFX/blob/master/src/Tizen.NUI/src/public/Layer.cs#L606

But a property could be created to do this instead of the roundabout way of doing things in this patch.

PR #1540, opened 2020-04-13
[Packagemanager] Do not catch the exception thrown by event handler of app

jeremy-jang:
### Description of Change ###
Do not catch the exception thrown by event handler of app
    
The applications should be responsible its exception, they should
determine handle or not the exception.

### API Changes ###
N/A


jeremy-jang:
I agree. I added a new commit that not catching the exception thrown by app.

PR #1543, opened 2020-04-16
[NUI] Change default layout to AbsoluteLayout

neostom432:
### Description of Change ###
We used LayoutItem and LayoutGroup as default layout but many developer wants to use AbsoluteLayout as default.
So changed default layout to AbsoluteLayout.

### API Changes ###
NONE

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/LayoutGroup.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LayoutGroup.cs
@@ -175,16 +175,7 @@ namespace Tizen.NUI
                 {
                     // If child of this layout is a pure View then assign it a LayoutGroup
                     // If the child is derived from a View then it may be a legacy or existing container hence will do layouting itself.
-                    if (child.GetType() == typeof(View))
-                    {
-                        child.Layout = new LayoutGroup();
-                    }
-                    else
-                    {
-                        // Adding child as a leaf, layouting will not propagate past this child.
-                        // Legacy containers will be a LayoutItems too and layout their children how they wish.
-                        child.Layout = new LayoutItem();
-                    }
+                    child.Layout = new AbsoluteLayout();
                 }
             }
             else
--- a/src/Tizen.NUI/src/internal/Layouting/LayoutController.cs
+++ b/src/Tizen.NUI/src/internal/Layouting/LayoutController.cs
@@ -160,17 +160,10 @@ namespace Tizen.NUI
             for (uint i = 0; i < rootNode.ChildCount; i++)
             {
                 View view = rootNode.GetChildAt(i);
-                if (rootNode.Layout == null )
+                if (view.Layout == null )
                 {
-                    if (rootNode.GetType() == typeof(View))
-                    {
-                        rootNode.Layout = new LayoutGroup();
-                        AutomaticallyAssignLayouts(rootNode);
-                    }
-                    else
-                    {
-                        rootNode.Layout = new LayoutItem();
-                    }
+                    view.Layout = new AbsoluteLayout();
+                    AutomaticallyAssignLayouts(rootNode);
                 }
             }
         }
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -2156,7 +2156,7 @@ namespace Tizen.NUI.BaseComponents
                 if (value?.Owner != null)
                 {
                     // Previous owner of the layout gets a default layout as a replacement.
-                    value.Owner.Layout = new LayoutGroup();
+                    value.Owner.Layout = new AbsoluteLayout();
 
                     // Copy Margin and Padding to replacement LayoutGroup.
                     value.Owner.Layout.Margin = value.Margin;


dasg34:
I agree with that.
However, i'm not sure if we need to keep `LayoutGroup` that is very similar to `AbsoluteLayout`.
|| LayoutGroup | AbsoluteLayout |
|---|---|---|
| `Margin` support? | O | X |
| `LayoutParamPolicies` support? | ?(it works when children not exist) | O |

I think `AbsoluteLayout` can be integrated with `LayoutGroup` if you have a plan to support `Margin` for `AbsoluteLayout`. Then, `AbsoluteLayout` will become wrapper of `LayoutGroup` and we can use `LayoutGroup` as a default Layout.
What do you think?

src/Tizen.NUI/src/public/Layouting/LayoutGroup.cs
```diff
@@ -175,16 +175,7 @@ private void AddChildToLayoutGroup(View child)
                 {
                     // If child of this layout is a pure View then assign it a LayoutGroup
                     // If the child is derived from a View then it may be a legacy or existing container hence will do layouting itself.
-                    if (child.GetType() == typeof(View))
```
dasg34, line 178: If you don't check whether child is pure `View` or not, Controls's children set Size and Position twice.
e.g. in case of `Components.Button`
1. `AbsoluteLayout.Onlayout`
2. `Control.OnRelayout` - `Button.OnUpdate` - `UpdateUIContent`

neostom432: Yes, but this also makes something wrong when user just inherit View to make their own component.

class V : View
{
}
....
V myView = new V(); 

myView cannot have LayoutGroup because return value of GetType is V, not View.

And about components thing, we are planning to change inner layout logic to using Layout feature so before that, sizing logic will be called twice but after that, it won't.

neostom432:
> I agree with that.
> However, i'm not sure if we need to keep `LayoutGroup` that is very similar to `AbsoluteLayout`.
> 	LayoutGroup 	AbsoluteLayout
> `Margin` support? 	O 	X
> `LayoutParamPolicies` support? 	?(it works when children not exist) 	O
> 
> I think `AbsoluteLayout` can be integrated with `LayoutGroup` if you have a plan to support `Margin` for `AbsoluteLayout`. Then, `AbsoluteLayout` will become wrapper of `LayoutGroup` and we can use `LayoutGroup` as a default Layout.
> What do you think?

AbsoluteLayout also support Margin/Padding.
I also thought about what you said but if we do that, AbsoluteLayout and LayoutGroup will be same.

So I decided to make absolute layout as default.

dasg34:
> > I agree with that.
> > However, i'm not sure if we need to keep `LayoutGroup` that is very similar to `AbsoluteLayout`.
> > LayoutGroup 	AbsoluteLayout
> > `Margin` support? 	O 	X
> > `LayoutParamPolicies` support? 	?(it works when children not exist) 	O
> > I think `AbsoluteLayout` can be integrated with `LayoutGroup` if you have a plan to support `Margin` for `AbsoluteLayout`. Then, `AbsoluteLayout` will become wrapper of `LayoutGroup` and we can use `LayoutGroup` as a default Layout.
> > What do you think?
> 
> AbsoluteLayout also support Margin/Padding.
> I also thought about what you said but if we do that, AbsoluteLayout and LayoutGroup will be same.
> 
> So I decided to make absolute layout as default.

What i said about Margin support is AbsoluteLayout doesn't respect children Margin. but it is not relative to make AbsoluteLayout default layout. so, i will approve this patch.

 ---

#### [LayoutGroup]
layoutgroup


#### [AbsoluteLayout]
AbsoluteLayout


#### [Test Code]
```
View view = new View();
view.HeightSpecification = LayoutParamPolicies.MatchParent;
view.WidthSpecification = LayoutParamPolicies.MatchParent;
//view.Layout = new LayoutGroup();
view.Layout = new AbsoluteLayout();
view.BackgroundColor = Color.Green;
Window.Instance.GetDefaultLayer().Add(view);

Button btn2 = new Button();
btn2.Margin = new Extents(50, 50, 50, 50);
btn2.Size2D = new Size2D(300, 300);
view.Add(btn2);
```

neostom432:
> > > I agree with that.
> > > However, i'm not sure if we need to keep `LayoutGroup` that is very similar to `AbsoluteLayout`.
> > > LayoutGroup 	AbsoluteLayout
> > > `Margin` support? 	O 	X
> > > `LayoutParamPolicies` support? 	?(it works when children not exist) 	O
> > > I think `AbsoluteLayout` can be integrated with `LayoutGroup` if you have a plan to support `Margin` for `AbsoluteLayout`. Then, `AbsoluteLayout` will become wrapper of `LayoutGroup` and we can use `LayoutGroup` as a default Layout.
> > > What do you think?
> > 
> > 
> > AbsoluteLayout also support Margin/Padding.
> > I also thought about what you said but if we do that, AbsoluteLayout and LayoutGroup will be same.
> > So I decided to make absolute layout as default.
> 
> What i said about Margin support is AbsoluteLayout doesn't respect children Margin. but it is not relative to make AbsoluteLayout default layout. so, i will approve this patch.
> #### [LayoutGroup]
> 
> layoutgroup
> #### [AbsoluteLayout]
> 
> AbsoluteLayout
> #### [Test Code]
> 
> ```
> View view = new View();
> view.HeightSpecification = LayoutParamPolicies.MatchParent;
> view.WidthSpecification = LayoutParamPolicies.MatchParent;
> //view.Layout = new LayoutGroup();
> view.Layout = new AbsoluteLayout();
> view.BackgroundColor = Color.Green;
> Window.Instance.GetDefaultLayer().Add(view);
> 
> Button btn2 = new Button();
> btn2.Margin = new Extents(50, 50, 50, 50);
> btn2.Size2D = new Size2D(300, 300);
> view.Add(btn2);
> ```

Oh... I didn't know that... We should check child margin in AbsoluteLayout.
I will fix this. Thanks to notice :)

PR #1546, opened 2020-04-16
[NUI] Add hidden class for internal purpose

neostom432:
### Description of Change ###
Some class need to be changed for internal purpose.
Add Never tag to hide API.

### API Changes ###
NONE


dongsug-song:
@neostom432 
Hello, 
if possible, please add sample code or sample link.
Thank you.

PR #1550, opened 2020-04-18
[NUI][Style] Fix style loading issue when extending control classes

woohyun:
When a NUI.Component.Control class was extended, the default style
was not loaded well.
This was because searching style logic was executed with not whole base classes
but only with the inherited class.

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Control.cs
+++ b/src/Tizen.NUI.Components/Controls/Control.cs
@@ -53,7 +53,16 @@ namespace Tizen.NUI.Components
         [EditorBrowsable(EditorBrowsableState.Never)]
         public Control() : base()
         {
-            ViewStyle viewStyle = StyleManager.Instance.GetComponentStyle(this.GetType());
+            var cur_type = this.GetType();
+            ViewStyle viewStyle = null;
+
+            do
+            {
+                if (cur_type.Equals(typeof(Tizen.NUI.Components.Control))) break;
+                viewStyle = StyleManager.Instance.GetComponentStyle(cur_type);
+                cur_type = cur_type.BaseType;
+            }
+            while (viewStyle == null);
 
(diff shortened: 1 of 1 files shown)


dasg34:
Public API `Instance.GetComponentStyle` still return null style for extending control classes.
```
MyButton btn = new MyButton();
var style = StyleManager.Instance.GetComponentStyle(btn.GetType()); // style == null
```
What do you think of moving this implementation to `SyleManager.GetComponentStyle`?

 ---

Hmm, It seems that `SyleManager.GetComponentStyle` is not only used for control classes.
DropDown.cs:779
```
itemDataStyle = (DropDownItemStyle)StyleManager.Instance.GetComponentStyle(this.GetType());
```

rabbitfor:
> Public API `Instance.GetComponentStyle` still return null style for extending control classes.
> 
> ```
> MyButton btn = new MyButton();
> var style = StyleManager.Instance.GetComponentStyle(btn.GetType()); // style == null
> ```
> 
> What do you think of moving this implementation to `SyleManager.GetComponentStyle`?
> 
> Hmm, It seems that `SyleManager.GetComponentStyle` is not only used for control classes.
> DropDown.cs:779
> 
> ```
> itemDataStyle = (DropDownItemStyle)StyleManager.Instance.GetComponentStyle(this.GetType());
> ```

How about provide a flag to StyleManager.Instance.GetComponentStyle()?
```
ViewStyle GetComponentStyle(Type componentType, bool lookupBaseClasses = true)
{
   ...
}
```
I think we need to providde both cases,
* Get Component style when the type is matching exactly.
* Get Component style and look up super classes when failed

hyunjushin:
@woohyun 
 I'm not clear if inheriting style from base class for custom winset is common or not. How about other platform?

PR #1553, opened 2020-04-20
[NUI] Refactorying FittingMode for ImageView

tscholb:
FittingMode does not need the required size anymore.
and new fittingMode is added for ImageView.

- CENTER : Image fills inside using original width & height.
- FILL   : Image filles whole width & height. this fitting mode don't  maintain aspect ratio

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/VisualConstants.cs
+++ b/src/Tizen.NUI/src/public/VisualConstants.cs
@@ -148,7 +148,17 @@ namespace Tizen.NUI
         /// <summary>
         /// Image rows: Limit loaded image resolution to row height using the FitHeight mode.
         /// </summary>
-        FitHeight
+        FitHeight,
+        /// <summary>
+        /// Thumbnail gallery grid: Limit loaded image resolution to screen tile using the Center mode.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        Center,
+        /// <summary>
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/VisualConstants.cs
```diff
@@ -148,7 +148,15 @@ public enum FittingModeType
         /// <summary>
         /// Image rows: Limit loaded image resolution to row height using the FitHeight mode.
         /// </summary>
-        FitHeight
+        FitHeight,
```
dongsug-song, line 151: Hello, 
This seems to be required ACR. (adding API)
because FittingModeType is public API.
could you check again?

src/Tizen.NUI/src/public/VisualConstants.cs
```diff
@@ -298,7 +306,23 @@ public enum VisualFittingModeType
         /// <summary>
         /// The visual should be stretched to fill, not preserving aspect ratio.
         /// </summary>
-        Fill
+        Fill,
```
dongsug-song, line 309: Hello, 
This seems to be required ACR. (adding API)
because VisualFittingModeType is public API.
could you check again?

dongsug-song:
@tscholb 
Hello, 
Please update TCSACR link here. 
ACR needs be accepted first, otherwise this PR can not be merged.
thank you.

tscholb:
This type will be temporary provided as EditorBrowsable.
i will modify additionally after ACR

PR #1567, opened 2020-04-24
[NUI] Remove binding between style and view

elishateng:
### Description of Change ###

1. Remove binding between style and view
2. Refactor Button, DropDown codes after removing binding.

### API Changes ###

 - ACR:


AdunFang (commented):
It is OK.

PR #1585, opened 2020-05-04
[NUI] remove input method dependency from View.ControlState

dasg34:
### Description of Change ###
`ConstrolState` can be changed by API or Input(`Tizen.NUI.Key` and
`Tizen.NUI.Touch`). it means `ConstrolState` is not always related to
Input method. some Controls need it to use state changed effect. however,
not all controls require that. if we need some property to change state,
we can add it to Style Extension classes.

Also, this patch rename `View.StateChagedEvent` recommended by Framework Design Guidelines.
(https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/)

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Extension/ButtonExtension.cs
+++ b/src/Tizen.NUI.Components/Controls/Extension/ButtonExtension.cs
@@ -26,6 +26,12 @@ namespace Tizen.NUI.Components.Extension
     public abstract class ButtonExtension
     {
         /// <summary>
+        /// The Touch info to get current touch position.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected Touch TouchInfo { get; private set; }
+
+        /// <summary>
         /// Called immediately after the Button creates the text part.
         /// </summary>
         /// <param name="button">The Button instance that the extension currently applied to.</param>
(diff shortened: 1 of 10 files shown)


dasg34:
Here is simple ripple effect class based on this patch.
https://github.com/dasg34/TizenFX/commit/67d1691028d960debf811d31e8018da076a85d71

src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
```diff
@@ -1060,22 +1060,18 @@ public ResourceLoadingStatusType Status
...
         {
             /// <summary>
             /// Create an instance with mandatory fields.
             /// </summary>
             /// <param name="previousState">The previous control state.</param>
             /// <param name="currentState">The current control state.</param>
-            /// <param name="inputMethod">Indicates the input method that triggered this change.</param>
-            /// <param name="inputData">The input method data that depends on the inputMethod.</param>
             [EditorBrowsable(EditorBrowsableState.Never)]
-            public ControlStateChangedInfo(ControlStates previousState, ControlStates currentState, InputMethodType inputMethod, object inputData)
```
rabbitfor, line 1073: I think ControlStateChangedEvent structure need to include input method information.

src/Tizen.NUI.Components/Controls/Extension/ButtonExtension.cs
```diff
@@ -66,9 +72,8 @@ public virtual ImageView OnCreateOverlayImage(Button button, ImageView overlayIm
         /// </summary>
         /// <param name="button">The Button instance that the extension currently applied to.</param>
         /// <param name="previousState">The previous contol state of the Button.</param>
-        /// <param name="touchInfo">The touch information in case the state has changed by touching.</param>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public virtual void OnControlStateChanged(Button button, ControlStates previousState, Touch touchInfo)
+        public virtual void OnControlStateChanged(Button button, ControlStates previousState)
```
rabbitfor, line 76: How about `OnControlStateChanged(Button button, ControlStateChangedEventArgs args)`

rabbitfor:
Can I ask why did you remove Input method information from the ControlStateChangedInfo (ControlStateChangedEventArgs)?

PR #1586, opened 2020-05-04
[NUI] remove redundant private properties for Tizen.NUI.Key

dasg34:
### Description of Change ###

This patch will reduce method call.
e.g.
```
Tizen.NUI.Key.set_keyString()
Tizen.NUI.Key.set_KeyString()
```
==>
```
Tizen.NUI.Key.set_keyString()
```

### DIFF
--- a/src/Tizen.NUI/src/public/Key.cs
+++ b/src/Tizen.NUI/src/public/Key.cs
@@ -99,12 +99,16 @@ namespace Tizen.NUI
         {
             get
             {
-                return keyPressedName;
+                string ret = Interop.Key.Key_keyPressedName_get(swigCPtr);
+                if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+                return ret;
             }
             set
             {
-                keyPressedName = value;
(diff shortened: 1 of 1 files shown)


dasg34:
`KeyPressed` is **public** property that have a only getter. but, `keyPressed` is **private** property that have getter and setter.
I'm not sure which one is better, remove **private** `keyPressed` setter OR add **public** `KeyPressed setter?

dasg34:
> Could you run Tizen.NUI.Tests first?
> 
> These are public APIs and It used in TV so should check TCT.

Do you mean `Tizen.NUI.Samples`? I can't find `Tizen.NUI.Tests` in this repo.
BTW, there are no logical changes and no public API changes.

neostom432:
> > Could you run Tizen.NUI.Tests first?
> > These are public APIs and It used in TV so should check TCT.
> 
> Do you mean `Tizen.NUI.Samples`? I can't find `Tizen.NUI.Tests` in this repo.
> BTW, there are no logical changes and no public API changes.

I meant Test cases for Tizen( in review.tizen.org ).
As you said, if there are no logical changes, good to go.

PR #1592, opened 2020-05-08
[NUI] Fix some API to use GetManagedBaseHandleFromNativePtr

tscholb:
- Fix GetWindow() using GetManagedBaseHandleFromNativePtr

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Application.cs
+++ b/src/Tizen.NUI/src/internal/Application.cs
@@ -1351,7 +1351,12 @@ namespace Tizen.NUI
 
         public Window GetWindow()
         {
-            Window ret = new Window(Interop.Application.Application_GetWindow(swigCPtr), true);
+            Window ret = Registry.GetManagedBaseHandleFromNativePtr(Interop.Application.Application_GetWindow(swigCPtr)) as Window;
+            if(ret == null)
+            {
+                ret = new Window(Interop.Application.Application_GetWindow(swigCPtr), true);
+            }
+
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/internal/Application.cs
```diff
@@ -1351,7 +1351,10 @@ internal bool AddIdle(SWIGTYPE_p_Dali__CallbackBase callback)
 
         public Window GetWindow()
         {
-            Window ret = new Window(Interop.Application.Application_GetWindow(swigCPtr), true);
+            Window ret = Registry.GetManagedBaseHandleFromNativePtr(Interop.Application.Application_GetWindow(swigCPtr)) as Window;
+            if(ret == null)
```
dongsug-song, line 1355: Please use brace
if(null == ret){ 
    xxx
}
OR, you can do auto formatting in VS by "ctrl + k + d"
thank you.

src/Tizen.NUI/src/internal/WidgetImpl.cs
```diff
@@ -310,7 +310,12 @@ private bool SwigDerivedClassHasMethod(string methodName, global::System.Type[]
 
         private void SwigDirectorOnCreate(string contentInfo, global::System.IntPtr window)
         {
-            OnCreate(contentInfo, new Window(window, true));
+            Window ret = Registry.GetManagedBaseHandleFromNativePtr(window) as Window;
+            if (ret == null)
```
dongsug-song, line 314: is it ok not to add new sub window in NUI window list?
does GetWindowList() get list from native side?

tscholb: GetWindowList() get list from native side, so i think it is fine.

PR #1594, opened 2020-05-09
[NUI] Fix the issue that size receives 2 parameters from xaml

elishateng:
### Description of Change ###

1. Fix the issue that size receives 2 parameters from xaml

### API Changes ###

 - ACR:


AdunFang (commented):
It is OK.

PR #1600, opened 2020-05-12
[NUI] Add Title in Wearable component

JoogabYun:
### Description of Change ###

 The Title inherits TextLabel.
 The Title can has a fadeOut effects on the left and right side.

### API Changes ###

 - ACR: NONE

### DIFF
--- a/src/Tizen.NUI.Wearable/src/public/Title.cs
+++ b/src/Tizen.NUI.Wearable/src/public/Title.cs
@@ -1,0 +1,349 @@
+﻿/*
+ * Copyright(c) 2020 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+using System;
+using System.ComponentModel;
+using System.Runtime.InteropServices;
+using Tizen.NUI.BaseComponents;
+using Tizen.NUI.Binding;
+
+namespace Tizen.NUI.Components
+{
+    /// <summary>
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI.Wearable/src/public/Title.cs
```diff
@@ -0,0 +1,342 @@
...
+            {
+                leftImage = new ImageView();
+                leftImage.PositionUsesPivotPoint = true;
+                leftImage.ParentOrigin = Tizen.NUI.ParentOrigin.TopLeft;
+                leftImage.PivotPoint = Tizen.NUI.PivotPoint.TopLeft;
+                leftImage.WidthResizePolicy = ResizePolicyType.FillToParent;
+                leftImage.HeightResizePolicy = ResizePolicyType.FillToParent;
+                this.Add(leftImage);
+            }
+            if (null == rightImage)
```
neostom432, line 219: Could you explain any case that these inner items are existing before initializing?

src/Tizen.NUI.Wearable/src/public/Title.cs
```diff
@@ -0,0 +1,342 @@
...
+                {
+                    rightImage.Unparent();
+                    rightImage.Dispose();
+                    rightImage = null;
+                }
+            }
+            base.Dispose(type);
+        }
+
+        private void UpdateImage()
```
neostom432, line 166: When user change size of Title Control bigger, is blur updated?

src/Tizen.NUI.Wearable/src/public/Title.cs
JoogabYun: There is not. This is an unnecessary check code!

src/Tizen.NUI.Wearable/src/public/Title.cs
JoogabYun: Yes, The width is fixed and the height follows the Title Control height.

src/Tizen.NUI.Wearable/src/public/Title.cs
```diff
@@ -0,0 +1,329 @@
...
+            }
+
+            this.Ellipsis = false;
+            this.HorizontalAlignment = HorizontalAlignment.Center;
+            this.VerticalAlignment = VerticalAlignment.Center;
+        }
+
+        private void Initialize()
+        {
+            leftImage = new ImageView();
```
neostom432, line 209: Minor comment
- we usually add m in front of member variable.

PR #1616, opened 2020-05-20
[NUI] Make a color visual of View's shadow to center aligned

rabbitfor:

### API Changes ###

 - ACR:


rabbitfor:
Please note that, this is a workaround patch to fix the issue that the shadow blurness spreads out from the left-top corner.

PR #1621, opened 2020-05-22
[MediaController] Fix metadata bug

hsgwon:
### Description of Change ###

Fix metadata bug when native handle is null.

### DIFF
--- a/src/Tizen.Multimedia.Remoting/MediaController/MediaControlMetadata.cs
+++ b/src/Tizen.Multimedia.Remoting/MediaController/MediaControlMetadata.cs
@@ -15,7 +15,6 @@
  */
 
 using System;
-using System.Diagnostics;
 using Native = Interop.MediaControllerPlaylist;
 
 namespace Tizen.Multimedia.Remoting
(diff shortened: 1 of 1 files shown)


src/Tizen.Multimedia.Remoting/MediaController/MediaControlMetadata.cs
```diff
@@ -36,7 +35,13 @@ public MediaControlMetadata()
...
         internal MediaControlMetadata(IntPtr handle)
         {
-            Debug.Assert(handle != IntPtr.Zero);
+            // If native framework return null handle,
+            // it means server doesn't set metadata yet and it's not error.
+            // So we need to return empty metadata instance as native framework does.
+            if (handle == IntPtr.Zero)
+            {
+                return;
+            }
```
WonyoungChoi, line 45: Not related this change, but you can delete lines to set "Season", "Episode", "Resolution" from this constructor. 
These properties return values from EncodedXXXX property in their getter methods.

hsgwon: Right, I removed those lines and uploaded new commit.
Please check it again.

PR #1632, opened 2020-05-28
[NUI] Scrollbar : Separate the concept of viewport size and view size.

rabbitfor:
* Previously, viewSize represents viewport size, but in some cases they are not the same value.
* View size : Implement OnRelayout of Scrollbar to respond to response to size changes.
* Viewport size : Receive by user when initialize.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Scrollbar.cs
+++ b/src/Tizen.NUI.Components/Controls/Scrollbar.cs
@@ -104,13 +104,13 @@ namespace Tizen.NUI.Components
             return ((ScrollbarStyle)((Scrollbar)bindable).viewStyle)?.TrackPadding;
         });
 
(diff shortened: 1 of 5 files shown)


src/Tizen.NUI.Components/Controls/Scrollbar.cs
```diff
@@ -214,30 +214,72 @@ public Extents TrackPadding
 
         /// <inheritdoc/>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public override void Initialize(float contentLength, Size viewSize, float currentPosition, bool isHorizontal = false)
+        public override void Initialize(float contentLength, float viewportLength, float currentPosition, bool isHorizontal = false)
```
neostom432, line 217: If possible,
it will be better if user can set this parameter through property.

PR #1634, opened 2020-05-28
[NUI] circular text

JoogabYun:
### Description of Change ###

circular text

### API Changes ###

 - ACR: NONE


hyunjushin:
@JoogabYun  should application knows renderer and texture to use circular text?

JoogabYun:
> should application knows renderer and texture to use circular text?

Currently, to show the PixelBuffer that renders circular text on the screen, User need to implement a renderer.

PixelBuffer pixelBuffer = TextUtils.Render(textParameters, ref embeddedItemLayout)l

PR #1638, opened 2020-05-29
[NUI] Add default DropdownItemStyle

zhouleonlei:
### Description of Change ###
Add default DropDown item style


liuyu199206:
LGTM

PR #1639, opened 2020-05-29
[NUI] SetProperty first then SetValue

zhouleonlei:
### Description of Change ###
Call TrySetProperty first and then TrySetValue to keep the code logic consistent whether it is through xaml or code.


liuyu199206:
LGTM

AdunFang (commented):
It is OK.

PR #1645, opened 2020-06-02
[NUI] Fix LinearListLayoutManager

neostom432:
### Description of Change ###
Change Layout and Recycle rule for linear layout.

### API Changes ###
NONE

### DIFF
--- a/src/Tizen.NUI.Components/Controls/RecyclerView/LinearLayoutManager.cs
+++ b/src/Tizen.NUI.Components/Controls/RecyclerView/LinearLayoutManager.cs
@@ -29,6 +29,31 @@ namespace Tizen.NUI.Components
     public class LinearListLayoutManager : LayoutManager
     {
         private float mLayoutOriginPosition = 0;
+        private int firstVisibleItemIndex = -1;
+        private int lastVisibleItemIndex = -1;
+
+        private bool IsItemVisible(float scrollPosition, RecycleItem item)
+        {
+            bool result = false;
+            View list = Container.GetParent() as View;
+
+            Vector2 visibleArea = new Vector2( Math.Abs(scrollPosition), 
+                Math.Abs(scrollPosition) + (LayoutOrientation == Orientation.Horizontal?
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI.Components/Controls/RecyclerView/LinearLayoutManager.cs
```diff
@@ -29,6 +29,31 @@ namespace Tizen.NUI.Components
     public class LinearListLayoutManager : LayoutManager
```
woohyun, line 29: I'm a bit confused with the class name because it's different from the file name (= LinearLayoutManager).

neostom432: FlexibleView already has LinearLayoutManager and GridLayoutManager but they are not public and not used.
So, We will remove FlexibleView first and after that, will change LinearListLayoutManager to LinearLayoutManger.

woohyun: I got it. Thanks for answering :)

src/Tizen.NUI.Components/Controls/RecyclerView/LinearLayoutManager.cs
```diff
@@ -74,47 +118,36 @@ public override void Layout(float scrollPosition)
...
-
-            firstVisibleItemPosition += 180;
-
+          
             bool checkFront = (mPrevScrollPosition - scrollPosition) > 0;
 
             if(checkFront)
             {
-                if(firstVisibleItemPosition < Math.Abs(scrollPosition))
+                if(firstVisibleItemIndex > 3)
```
hyunjushin, line 126: what is the meaning of 3?

neostom432: I want to keep only two items at tip of each side.
So if index is over 2, recycle item.

It should be set by property later.

PR #1646, opened 2020-06-02
[NUI] Made the value of Size be consistent with Specification

zhouleonlei:
### Description of Change ###
Made the value of Size be consistent with Specification for the defect of adding layout


AchoWang:
Looks good to me.

AdunFang (commented):
It is OK.

PR #1649, opened 2020-06-03
[NUI] Add ApplyStyle for Pagination

zhouleonlei:
### Description of Change ###
Add the function of ApplyStyle for Pagination

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Pagination.cs
+++ b/src/Tizen.NUI.Components/Controls/Pagination.cs
@@ -240,6 +240,20 @@ namespace Tizen.NUI.Components
         }
 
         /// <summary>
+        /// Apply style to pagination.
+        /// </summary>
+        /// <param name="viewStyle">The style to apply.</param>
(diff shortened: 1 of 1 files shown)


zhouleonlei:
> LGTM.
> But basically, we don't have to delcare paginationStyle in Pagination class because it is duplicate with View.viewStyle. So I think like other components, such as Button, it need to provide API that cast ViewStyle to PaginationStyle.
> 
> ```
> /// <summary>
>         /// Return a copied Style instance of Button
>         /// </summary>
>         /// <remarks>
>         /// It returns copied Style instance and changing it does not effect to the Button.
>         /// Style setting is possible by using constructor or the function of ApplyStyle(ViewStyle viewStyle)
>         /// </remarks>
>         /// <since_tizen> 8 </since_tizen>
>         public new ButtonStyle Style => ViewStyle as ButtonStyle;
> ```

OK. I will modify it

AdunFang (commented):
It is OK.

PR #1651, opened 2020-06-04
[API7][Security] Fix PrivacyPrivilegeManager.RequestPermissions crash issue

myroot:
### Description of Change ###

Refer : https://github.com/Samsung/TizenFX/pull/1650

### API Changes ###

 - ACR:


TomaszSwierczek:
I will not guarantee that the patch is okay, as I cannot test it (I have limited resources due to remote work from home atm), but it looks like it fixes the most important problem of not storing any reference to callback function using a static map. It looks okay. If it works, then go ahead.

PR #1652, opened 2020-06-04
[NUI] Fix ScrollTo to move pan position to the argument position

jaehyun0cho:
Previously, ScrollTo moved pan position including the ScrollableBase's
current position.
e.g. ScrollTo(360, true); //moves by 360
     ScrollTo(0, true); //moves by 360 again based on current position

The argument position is the new pan position. So the new position of
ScrollableBase becomes (-position).
To move ScrollableBase's position to (-position), it moves by
(-position - currentPosition).

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
+++ b/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
@@ -591,7 +591,9 @@ namespace Tizen.NUI.Components
             float currentPositionX = mScrollingChild.CurrentPosition.X != 0 ? mScrollingChild.CurrentPosition.X : mScrollingChild.Position.X;
             float currentPositionY = mScrollingChild.CurrentPosition.Y != 0 ? mScrollingChild.CurrentPosition.Y : mScrollingChild.Position.Y;
             float delta = ScrollingDirection == Direction.Horizontal ? currentPositionX : currentPositionY;
-            delta -= position;
+            // The argument position is the new pan position. So the new position of ScrollableBase becomes (-position).
+            // To move ScrollableBase's position to (-position), it moves by (-position - currentPosition).
+            delta = -position - delta;
 
             ScrollBy(delta, animate);
         }


src/Tizen.NUI.Components/Controls/ScrollableBase.cs
```diff
@@ -591,7 +591,9 @@ public void ScrollTo(float position, bool animate)
             float currentPositionX = mScrollingChild.CurrentPosition.X != 0 ? mScrollingChild.CurrentPosition.X : mScrollingChild.Position.X;
             float currentPositionY = mScrollingChild.CurrentPosition.Y != 0 ? mScrollingChild.CurrentPosition.Y : mScrollingChild.Position.Y;
             float delta = ScrollingDirection == Direction.Horizontal ? currentPositionX : currentPositionY;
-            delta -= position;
+            // The argument position is the new pan position. So the new position of ScrollableBase becomes (-position).
```
neostom432, line 594: I was thought developer treats ScrollPosition as positive number even if the real value(current position of container) is negative.

So I assumed that user passes positive number to ScrollTo API.

What do you think?

jaehyun0cho: > I was thought developer treats ScrollPosition as positive number even if the real value(current position of container) is negative.
> 
> So I assumed that user passes positive number to ScrollTo API.
> 
> What do you think?

I agree with you. And that is why I made this patch.
As the comment on this patch, previously ScrollableBase.ScrollTo()'s position was not the destination position. But it worked like a vector and I think that the calculation was not correct as well.
e.g.
ScrollTo(360, true); //moves to pan position 360 (ScrollablaBase's position -360)
ScrollTo(-720, true); //moves to position 0 (ScrollableBase's position 0) because the calculation is delta = -360 - (-720) = 360 and ScrollBy(360, true); moves ScrollableBase's position by 360 from the current position (-360).

I referred how Scrollbar.ScrollTo() works and I think it would be right if ScrollableBase.ScrollTo() works like Scrollbar.ScrollTo().

neostom432: Understand.  It was wrong.

Thanks to fix it :)

PR #1654, opened 2020-06-04
[NUI] add internal visibility for NUI.Components.DA temporarily.

huayongxu:
### API Changes ###

 - ACR:


dongsug-song:
This will be removed after NUI.Components.DA is replaced by NUI.Components.

AdunFang (commented):
It is OK.

PR #1656, opened 2020-06-04
[NUI] Fix to enable ScrollTo before touch scrolling

jaehyun0cho:
Previously, ScrollTo did not work unless touch scrolling had been done.
Because maxScrollDistance was not set in BoundScrollPosition.

Now, maxScrollDistance is calculated in BoundScrollPosition if
maxScrollDistance is not set in BoundScrollPosition.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
+++ b/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
@@ -610,6 +610,11 @@ namespace Tizen.NUI.Components
             }
             else
             {
+                if (Math.Equals(maxScrollDistance, 0.0f))
+                {
+                    maxScrollDistance = CalculateMaximumScrollDistance();
+                }
+
                 targetPosition = Math.Min(0, targetPosition);
                 targetPosition = Math.Max(-maxScrollDistance, targetPosition);
             }


src/Tizen.NUI.Components/Controls/ScrollableBase.cs
```diff
@@ -610,6 +610,11 @@ private float BoundScrollPosition(float targetPosition)
             }
             else
             {
+                if (Math.Equals(maxScrollDistance, 0.0f))
```
neostom432, line 613: I think we should calculate maxScrollDistance when size of mScrollingChild is set or changed.

How about listening Relayout event of mScrollingChild and set it in callback?

jaehyun0cho: Yes, I think it's a good idea!
Could you make the patch?
I abandon this patch when you make the patch.

neostom432: God it. I'll request review after making my patch :)

neostom432: https://github.com/Samsung/TizenFX/pull/1665 << please review my patch

PR #1658, opened 2020-06-04
[Bluetooth][Non-ACR] Add NULL check logic for Marshal.PtrToStringAnsi

dh79pyun:
This patchset is to avoid ArgumentNullException error. There are some
cases to come NULL variable from Native API.

BluetoothAdapterImpl.cs: RegisterDiscoveryStateChangedEvent(258) > Discovery state changed callback is called
Unhandled exception.
System.ArgumentNullException: Value cannot be null. (Parameter 'ptr')
at System.Runtime.InteropServices.Marshal.PtrToStringAnsi(IntPtr ptr, Int32 len)
at Tizen.Network.Bluetooth.BluetoothUtils.ConvertStructToDiscoveredDevice(BluetoothDiscoveredDeviceStruct structDevice)
at Tizen.Network.Bluetooth.BluetoothAdapterImpl.<RegisterDiscoveryStateChangedEvent>b__45_0(Int32 result, BluetoothDeviceDiscoveryState state, IntPtr deviceInfo, IntPtr user

Signed-off-by: DoHyun Pyun <dh79.pyun@samsung.com>

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.Network.Bluetooth/Tizen.Network.Bluetooth/BluetoothStructs.cs
+++ b/src/Tizen.Network.Bluetooth/Tizen.Network.Bluetooth/BluetoothStructs.cs
@@ -244,7 +244,9 @@ namespace Tizen.Network.Bluetooth
             resultDevice.RemoteDeviceService = uuidList;
             resultDevice.RemoteDeviceCount = device.ServiceCount;
             resultDevice.RemoteManufLength = device.ManufacturerDataLength;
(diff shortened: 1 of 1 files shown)


myroot:
Is `ManufacturerData` a string? it seems not string

dh79pyun:
> Is `ManufacturerData` a string? it seems not string

You are right. ManufactureData is not a string. I should be byte array.
But, because we already publish it as string type for C# Public. So we will do the change ACR for it.
Before that change ACR, we should make the patchset to avoid the crash for application.

PR #1663, opened 2020-06-08
[NUI] Replace paginationStyle with Style

zhouleonlei:
### Description of Change ###
Replace paginationStyle with Style


AdunFang (commented):
It is OK.

PR #1668, opened 2020-06-09
[NUI] remove margin, padding calculation for AbsoluteLayout child size

dasg34:
Since child position property and dali position property that decides
real position of controls are shared as same property(View.Position),
AbsoluteLayout cannot respect margin, padding for child position. so,
we don't need to calculate size with margin, padding.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/AbsoluteLayout.cs
+++ b/src/Tizen.NUI/src/public/Layouting/AbsoluteLayout.cs
@@ -13,32 +13,16 @@
  * limitations under the License.
  *
  */
-using System;
 using Tizen.NUI.BaseComponents;
 
 namespace Tizen.NUI
 {
     /// <summary>
     /// [Draft] This class implements a absolute layout, allowing explicit positioning of children.
-    ///  Positions are from the top left of the layout and can be set using the Actor::Property::POSITION and alike.
+    ///  Positions are from the top left of the layout and can be set using the View.Position and alike.
     /// </summary>
     public class AbsoluteLayout : LayoutGroup
     {
-        /// <summary>
-        /// Struct to store Measured states of height and width.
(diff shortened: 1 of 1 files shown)


dasg34:
Since Margin, Padding calculation is implemented in `LayoutGroup`,  you can reproduce the bug commit message said with the following code.
When you tried to click button, it will be moved continuously.
```
Window.Instance.WindowSize = new Size2D(500, 700);

View root = new View();
root.WidthSpecification = LayoutParamPolicies.WrapContent;
root.HeightSpecification = LayoutParamPolicies.WrapContent;
root.BackgroundColor = new Color(0.8f, 0.8f, 0.8f, 1.0f);
Window.Instance.GetDefaultLayer().Add(root);

LayoutGroup layoutGroup = new LayoutGroup();
root.Layout = layoutGroup;

Button btn = new Button();
btn.WidthSpecification = 300;
btn.Margin = new Extents(10, 0, 0, 0);
btn.HeightSpecification = 100;
btn.BackgroundColor = new Color(0.9f, 0.5f, 0.3f, 1.0f);
root.Add(btn);

Timer timer = new Timer(100);
timer.Tick += (object source, Timer.TickEventArgs e) =>
{
    layoutGroup.RequestLayout();
    return true;
};
timer.Start();
```

dasg34:
> In web,
> They keep calculating margin and padding while using position:absolute.
> 
> I'm not sure ignoring margin and padding is good way.
> Do other Platform also ignore padding and margin in AbsoulteLayout too?

Xamarin supports margin, padding. also, i'd like to support margin, padding for NUI. but it is not possible because of the bug i said above.

neostom432:
Can merge after resolving conflicts.

dasg34:
Updated :)

Issue #1670, opened 2020-06-09
[NFC] Native Callback GC issue


myroot:
https://github.com/Samsung/TizenFX/blob/98dcbeeefd4172dbe9fab93fafe101200ce6255e/src/Tizen.Network.Nfc/Tizen.Network.Nfc/NfcTag.cs#L192

https://github.com/Samsung/TizenFX/blob/98dcbeeefd4172dbe9fab93fafe101200ce6255e/src/Tizen.Network.Nfc/Tizen.Network.Nfc/NfcTag.cs#L225

https://github.com/Samsung/TizenFX/blob/98dcbeeefd4172dbe9fab93fafe101200ce6255e/src/Tizen.Network.Nfc/Tizen.Network.Nfc/NfcTag.cs#L260

https://github.com/Samsung/TizenFX/blob/98dcbeeefd4172dbe9fab93fafe101200ce6255e/src/Tizen.Network.Nfc/Tizen.Network.Nfc/NfcTag.cs#L295

It could be released by GC before complete async operation

WonyoungChoi:
Similar issue: #1650

PR #1674, opened 2020-06-09
[NFC] Fix Native callback GC issue

myroot:
### Description of Change ###

 This PR fix https://github.com/Samsung/TizenFX/issues/1670

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.Network.Nfc/Tizen.Network.Nfc/NfcTag.cs
+++ b/src/Tizen.Network.Nfc/Tizen.Network.Nfc/NfcTag.cs
@@ -29,6 +29,26 @@ namespace Tizen.Network.Nfc
     {
         private bool disposed = false;
         private IntPtr _tagHandle = IntPtr.Zero;
+        int _requestId = 0;
+
+        Dictionary<int, TaskCompletionSource<byte[]>> _transceiveTaskSource = new Dictionary<int, TaskCompletionSource<byte[]>>();
+        Interop.Nfc.TagTransceiveCompletedCallback _nativeTransceiveCallback;
+
+        Dictionary<int, TaskCompletionSource<NfcError>> _voidTaskSource = new Dictionary<int, TaskCompletionSource<NfcError>>();
+        Interop.Nfc.VoidCallback _nativeVoidCallback;
+
+        Dictionary<int, TaskCompletionSource<NfcNdefMessage>> _readNdefTaskSource = new Dictionary<int, TaskCompletionSource<NfcNdefMessage>>();
+        Interop.Nfc.TagReadCompletedCallback _nativeTagReadCallback;
+
+        /// <summary>
+        /// Constructor of NfcTag
+        /// </summary>
+        public NfcTag()
+        {
+            _nativeTransceiveCallback = TransceiveCompletedCallback;
+            _nativeVoidCallback = VoidCallback;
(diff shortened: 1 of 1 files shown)


src/Tizen.Network.Nfc/Tizen.Network.Nfc/NfcTag.cs
```diff
@@ -29,6 +29,26 @@ public class NfcTag : IDisposable
...
+        Interop.Nfc.TagReadCompletedCallback _nativeTagReadCallback;
+
+        /// <summary>
+        /// Constructor of NfcTag
+        /// </summary>
+        public NfcTag()
+        {
+            _nativeTransceiveCallback = TransceiveCompletedCallback;
+            _nativeVoidCallback = VoidCallback;
+            _nativeTagReadCallback = ReadNdefCallback;
```
WonyoungChoi, line 50: Are these member variables necessary? How about passing the methods directly?
For example,
```
Interop.Nfc.Tag.Transceive(_tagHandle, buffer, buffer.Length, TransceiveCompletedCallback, (IntPtr)requestId)
```

myroot: Good comment, I already test to use a static method to pass native callbak, but it was implicitly converted to temperal delegate and that is released by GC.

WonyoungChoi: Interesting! If it is true, please put some comment into lines to avoid regression errors later. And how about setting these methods in member declaration lines?
```csharp
Interop.Nfc.TagTransceiveCompletedCallback _nativeTransceiveCallback = TransceiveCompletedCallback;
```
We can delete "ACR Required" label before confirm of @m-macias.

myroot: Done

myroot: Add a comment, 
but we can't assign in declaration line, it is CS0236 error(A field initializer cannot reference the nonstatic field, method, or property 'field')

myroot: We can consider change to static like below and share dictionary with all NfcTag object
``` c#
        static int _requestId = 0;

        static Dictionary<int, TaskCompletionSource<byte[]>> _transceiveTaskSource = new Dictionary<int, TaskCompletionSource<byte[]>>();
        static Interop.Nfc.TagTransceiveCompletedCallback _nativeTransceiveCallback = TransceiveCompletedCallback;

        static Dictionary<int, TaskCompletionSource<NfcError>> _voidTaskSource = new Dictionary<int, TaskCompletionSource<NfcError>>();
        static Interop.Nfc.VoidCallback _nativeVoidCallback = VoidCallback;

        static Dictionary<int, TaskCompletionSource<NfcNdefMessage>> _readNdefTaskSource = new Dictionary<int, TaskCompletionSource<NfcNdefMessage>>();
        static Interop.Nfc.TagReadCompletedCallback _nativeTagReadCallback = ReadNdefCallback;
```

WonyoungChoi:
I removed "ACR Required" label because this PR just provides a constructor to replace the default constructor. It does not make ABI break.

PR #1679, opened 2020-06-10
[NUI] replace viewStlye with Style

dasg34:
Use `Style` property instead of `viewStyle` field.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Scrollbar.cs
+++ b/src/Tizen.NUI.Components/Controls/Scrollbar.cs
@@ -36,12 +36,12 @@ namespace Tizen.NUI.Components
             var instance = ((Scrollbar)bindable);
             var thickness = (float?)newValue;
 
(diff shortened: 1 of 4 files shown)


rabbitfor:
Component's Style API is from View.ViewStyle property.
And it seems that the purpose of View.ViewStyle is to assure viewStyle value in case it is null.

```C#
public new CircularSliderStyle Style => ViewStyle as CircularSliderStyle;
```
```C#
public ViewStyle ViewStyle
{
    get
    {
        if (null == viewStyle)
        {
            ApplyStyle(GetViewStyle());
        }
        return viewStyle;
    }
}

protected virtual ViewStyle GetViewStyle()
{
    return CreateViewStyle();
}

protected virtual ViewStyle CreateViewStyle()
{
    return new ViewStyle();
}
```

So I think null checking such as `Style?` is redundent. What do you think?

dasg34:
> So I think null checking such as `Style?` is redundent. What do you think?

It makes sense.

PR #1680, opened 2020-06-10
[NUI] make viewStyle field private

dasg34:
Fixes CA1051

Depends on #1679

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -70,7 +70,7 @@ public partial class View : Container, IResourcesProvider
         /// The Style instance binded with this View.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
```
idkiller, line 72: EditorBrowsable is not need to private member.

rabbitfor:
Please fix build errors

PR #1701, opened 2020-06-12
[NUI] replace field declaration with property

woohyun:
This would fix CA1051 warnings which indicates that
"Do not declare visible instance fields".

### API Changes ###

 - ACR:


WonyoungChoi:
@woohyun @hyunjushin Public APIs were changed without ACR. Please revert and fix again.

PR #1706, opened 2020-06-14
[NUI] throw exception when profile not found

dasg34:
fixes CA1031

### API Changes ###

 - ACR:


rabbitfor:
If we change this to throw an exception, running Tizen.NUI in linux pc will not be available anymore.

dasg34:
Ok, i will abandon this patch. we need other solution..

PR #1710, opened 2020-06-15
[WidgetControl] Add property for setup app id

sukhyungkang:
Signed-off-by: SukHyung, Kang <shine.kang@samsung.com>

### Description of Change ###
Add new property for setup app id

### API Changes ###
 - ACR: https://code.sec.samsung.net/jira/browse/TSACR-325

Changed:
 - Add
    public string SetupAppId

### DIFF
--- a/src/Tizen.Applications.WidgetControl/Tizen.Applications/WidgetControl.cs
+++ b/src/Tizen.Applications.WidgetControl/Tizen.Applications/WidgetControl.cs
@@ -436,6 +436,47 @@ namespace Tizen.Applications
         }
 
         /// <summary>
+        /// Gets setup app ID of the widget.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/widget.viewer</privilege>
+        /// <feature>http://tizen.org/feature/shell.appwidget</feature>
+        /// <exception cref="UnauthorizedAccessException">Thrown when an application does not have the required privileges to access this method.</exception>
+        /// <exception cref="NotSupportedException">Thrown when the required features are not supported.</exception>
+        /// <exception cref="InvalidOperationException">Thrown in case of failed conditions.</exception>
+        public string SetupAppId
+        {
+            get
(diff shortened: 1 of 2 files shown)


src/Tizen.Applications.WidgetControl/Tizen.Applications/WidgetControl.cs
```diff
@@ -435,6 +435,34 @@ public bool IsNoDisplay
             }
         }
 
+        /// <summary>
+        /// Gets setup app ID of the widget.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
```
m-macias, line 441: Current API version is API8. Current Platform version is tizen_6.0. It should be:
```suggestion
        /// <since_tizen> 8 </since_tizen>
```

src/Tizen.Applications.WidgetControl/Tizen.Applications/WidgetControl.cs
```diff
@@ -435,6 +435,34 @@ public bool IsNoDisplay
...
+                Interop.WidgetService.ErrorCode err =
+                    (Interop.WidgetService.ErrorCode)Internals.Errors.ErrorFacts.GetLastResult();
+                switch (err)
+                {
+                    case Interop.WidgetService.ErrorCode.PermissionDenied:
+                        throw new UnauthorizedAccessException();
+
+                    case Interop.WidgetService.ErrorCode.NotSupported:
+                        throw new NotSupportedException("Not supported");
+                }
```
idkiller, line 460: If you put null in the constructor, Id can be null.
At that time, WIDGET_ERROR_INVALID_PARAMETER occurs.

Also, when _get_app_id_of_setup_app is called from widget_service_get_app_id_of_setup_app, other errors appear to occur.

So, it seems that the case for the invalid parameter and the default case must also be handled.

sukhyungkang: exceptions are added

PR #1711, opened 2020-06-15
[NUI] Modify Component-application window

huiyueun:

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/ComponentApplication.cs
+++ b/src/Tizen.NUI/src/internal/ComponentApplication.cs
@@ -8,12 +8,10 @@ namespace Tizen.NUI
     class ComponentApplication : Application
     {
         private global::System.Runtime.InteropServices.HandleRef swigCPtr;
-        private List<NUIFrameComponent> _frameComponentList;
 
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/internal/ComponentApplication.cs
```diff
@@ -152,31 +150,19 @@ internal ApplicationSignal CreateNativeSignal()
...
         {
-            _frameComponentList.Add(nuiFrameComponent);
-        }
+            IntPtr handle = IntPtr.Zero;
 
-        internal int GetFrameComponentCount()
-        {
-            return _frameComponentList.Count;
-        }
+            if (_applicationCreateNativeEventHandler != null)
```
dongsug-song, line 159: I think this can be simplified like this? how do you think?

handle = _applicationCreateNativeEventHandler?.Invoke();

huiyueun: handle = _applicationCreateNativeEventHandler?.Invoke();
->
Error	CS0266	Cannot implicitly convert type 'System.IntPtr?' to 'System.IntPtr'. An explicit conversion exists (are you missing a cast?)

PR #1723, opened 2020-06-19
[NUI] change to protected property from field

dasg34:
Fixes CA1051

### API Changes ###

 - ACR:


src/Tizen.NUI.Components/Controls/RecyclerView/RecycleLayoutManager.cs
```diff
@@ -101,6 +92,13 @@ public enum Orientation
         [EditorBrowsable(EditorBrowsableState.Never)]
         public int DataCount{get; set;}
 
+        /// <summary>
+        /// FIXME
```
rabbitfor, line 96: The last scrolled position which is calculated by ScrollableBase. The value should be updated in Recycle() method.

rabbitfor:
Can you fix the conflicts? :)

PR #1727, opened 2020-06-19
[NUI] Change View.ControlState setter to protected

rabbitfor:

### Description of Change ###
Setting ControlState is now allowed only in the derived class of View.

### API Changes ###

 - ACR:


rabbitfor:
> I think `ControlState` of `CircularProgress` should be updated in propertyChanged callback of `IsEnabledProperty`. but it can be handled in another patch.

I updated the patch.
Now the the ControlState can be Disabled in CircularProgress.privateIsEnabled setter.

dasg34:
👍

PR #1728, opened 2020-06-19
[NUI] Split large files

elishateng:
### Description of Change ###

Split large files (ViewStyle, TextLabel, Popup) that large than 1000 lines.
(Don't change codes, just split files.)

### API Changes ###

 - ACR:


myroot:
It is just cheating

elishateng:
> It is just cheating

Hello, The SAM tool requires that the number of lines of code cannot exceed 500 lines, otherwise the LOC score will be cut.

myroot:
> 
> 
> > It is just cheating
> 
> Hello, The SAM tool requires that the number of lines of code cannot exceed 500 lines, otherwise the LOC score will be cut.


You should reduce a line of code on class, this changes just split a file not class. it is just cheating. does not improve code quality on CLOC.
on SAM-Tools, does not support partial class, it calculate just last size of class on same class name

myroot:
https://github.sec.samsung.net/RS7-Architectural-Refactoring/SAM-Tools/wiki/Metrics-in-SAM#lines-of-code-loc


1. Definition
 * A software metric used to measure the size of a computer program by counting the number of lines in the text of the program's source code.
 * In C++ language, the Struct, Enum, and Union are considered as Class.
 * Code PLOC is used to calculate this metric.
 * Unit : **Class**

myroot:
* Keep classes small (https://www.ibm.com/developerworks/library/j-perry-writing-good-java-code/index.html)
  - So far you've created a few classes. After generating getter/setter pairs for even the small number (by the standards of a real-world Java class) of attributes, the Person class has 150 lines of code. At that size, Person is a small class. It's not uncommon (and it's unfortunate) to see classes with 50 or 100 methods and a thousand lines or more of source. Some classes might be that large out of necessity, but most likely they need to be refactored. Refactoring is changing the design of existing code without changing its results. I recommend that you follow this best practice.
  - In general, a class represents a conceptual entity in your application, and a class's size should reflect only the functionality to do whatever that entity needs to do. Keep your classes tightly focused to do a small number of things and do them well.
  - Keep only the methods that you need. If you need several helper methods that do essentially the same thing but take different parameters (such as the printAudit() method), that's a fine choice. But be sure to limit the list of methods to what you need, and no more.

myroot:
@dongsug-song  @hyunjushin Please give your opinion

hyunjushin:
We can consider using partial class for only too big files(more than 1000 line?), if they can be split meaningfully.

myroot:
> 
> 
> We can consider using partial class for only too big files(more than 1000 line?), if they can be split meaningfully.


Partial class not resolve CLOC issue, it is just cheating to get more higher point on SAM-TOOL,

hyunjushin:
@myroot In my point of view, it is not for SAM score. I can't agree 500 line criteria for framework class. I don't care the score. But too big file size is not good readability. In this patch, bindable properties are separated. they are additional part and it looks good  to enhance maintenability. Can't we use partial class at all?

myroot:
@hyunjushin Ok, I also agree of your view to control size of a file, if it is not intent to improve SAM score

PR #1741, opened 2020-06-23
[Bluetooth][Non-ACR] Fix GattConnection event not occured issue

Wootak:
Signed-off-by: Wootak Jung <wootak.jung@samsung.com>


Wootak:
#1739

PR #1747, opened 2020-06-24
[NUI] Make Notification APIs public and fix CA warnings.

rabbitfor:

### API Changes ###

 - ACR: https://code.sec.samsung.net/jira/browse/TCSACR-331

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Notification.cs
+++ b/src/Tizen.NUI.Components/Controls/Notification.cs
@@ -24,9 +24,8 @@ namespace Tizen.NUI.Components
     /// <summary>
     /// Notification helps to raise a notification window with a content View.
     /// </summary>
-    /// <privilege>http://tizen.org/privilege/window.priority.set</privilege>
-    [EditorBrowsable(EditorBrowsableState.Never)]
-    public class Notification
+    /// <since_tizen> 8 </since_tizen>
+    public class Notification : Disposable
     {
         private static HashSet<Notification> instanceSet;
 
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI.Components/Controls/Notification.cs
```diff
@@ -283,6 +274,43 @@ public void ForceQuit()
...
+            }
+
+            if (onDismissAnimation != null)
+            {
+                onDismissAnimation.Dispose();    
+            }
+
+            base.Dispose(type);
+        }
+
```
idkiller, line 313: When calling only one method immediately after null check, it is better to use null conditional operator.

```C#
notificationWindow?.Dispose();
timer?.Dispose();
positionSize?.Dispose();
onPostAnimation?.Dispose();
onDismissAnimation?.Dispose();
```

This improves code readability, reduces code, and actually reduces instructions in IL.

rabbitfor: I tried to use "?" here, but when I built it the Code Analysis created [CA2213](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2213?view=vs-2019) warnings on it. (I don't know why...)

I wanted to keep no Code Analysis warnings in the Notification class so I had to use `if` for null checking.

rabbitfor: I think my code anaylzer is an old one.

rabbitfor: Done fixing!

PR #1749, opened 2020-06-24
[NUI] Hide NUI.Component.StyleManager

rabbitfor:

### API Changes ###

 - ACR: https://code.sec.samsung.net/jira/browse/TCSACR-330

### DIFF
--- a/src/Tizen.NUI.Components/Utils/StyleManager.cs
+++ b/src/Tizen.NUI.Components/Utils/StyleManager.cs
@@ -15,6 +15,7 @@
  *
  */
 using System;
+using System.ComponentModel;
 using System.Collections.Generic;
 using Tizen.NUI.BaseComponents;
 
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI.Components/Utils/StyleManager.cs
```diff
@@ -23,7 +24,8 @@ namespace Tizen.NUI.Components
     /// <summary>
     /// StyleManager is a class to manager all style.
     /// </summary>
-    /// <since_tizen> 8 </since_tizen>
+    /// <since_tizen> none </since_tizen>
```
m-macias, line 27: Please don't use `<since_tizen> none </since_tizen>` tag, because it is not an appropriate tag value. If you hide this API by using EditorBrowsableState.Never you can remove `<since_tizen>` tag completely or just leave it as it is.
```suggestion
    /// <since_tizen> 8 </since_tizen>
```
```suggestion
```

rabbitfor: Done!

PR #1752, opened 2020-06-25
[NUI] Add Capture

dongsug-song:
### Description of Change ###
Capture binding
- This will be provided as hidden API first
- internal CaptureSignal class will be moved to a single file
- CaptureFinishStateEventArgs, this name will be changed (following c# naming rule)
- try to apply TDD in CaptureTest.cs, but it make runtime Exception which seems to be related with Nunit.Framework
- paired with https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/238500/

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/public/Capture.cs
+++ b/src/Tizen.NUI/src/public/Capture.cs
@@ -1,0 +1,298 @@
+using System.Diagnostics;
+using System;
+using System.Drawing;
+/*
+ * Copyright(c) 2020 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+namespace Tizen.NUI
+{
+    using global::System;
+    using global::System.ComponentModel;
+    using global::System.Runtime.InteropServices;
+    using global::System.ComponentModel;
+    using Tizen.NUI.BaseComponents;
+
+    /// <summary>
+    /// Capture snapshots the current scene and save as a file.
+    /// Applications should follow the example below to create capture :
+    /// <code>
+    /// Capture capture = new Capture();
+    /// </code>
+    /// If required, you can also subscribe Finished event :
+    /// <code>
+    /// capture.Finished += onCaptureFinished;
+    /// </code>
+    /// At the subcribed event handler, user can know whether capture finish succeeded state.
+    /// <code>
+    /// private void onCaptureFinished(object sender, CaptureFinishedEventArgs e)
+    /// {
+    ///   if(e.Success) { //capture success, do something. }
+    ///   else { //capture failure, do something. }
+    /// }
+    /// </code>
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class Capture : BaseHandle
+    {
+        /// <summary>
+        /// Create an Capture.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Capture() : this(Interop.Capture.New(), true)
+        {
+            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+        }
+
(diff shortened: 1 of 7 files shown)


src/Tizen.NUI/src/public/Capture.cs
```diff
@@ -0,0 +1,252 @@
...
+                {
+                    finishedCallback = onFinished;
+                    finishedSignal.Disconnect(finishedCallback);
+                }
+            }
+        }
+
+        /// <summary>
+        /// CaptureFinishedState
+        /// </summary>
```
bshsqa, line 166: Would we make this enum private?
We may can change Dali::Capture::FinishedState to the simple bool value in NUI.

src/Tizen.NUI/src/public/Capture.cs
```diff
@@ -0,0 +1,252 @@
...
+                if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+            }
+        }
+
+        private void onFinished(IntPtr data, int state)
+        {
+            if (data != IntPtr.Zero)
+            {
+                var arg = new CaptureFinishStateEventArgs();
+                arg.State = (CaptureFinishedState)state;
```
bshsqa, line 127: Here, if the (CaptureFinishedState)state is Succeeded then arg is true.

src/Tizen.NUI/src/public/Capture.cs
```diff
@@ -0,0 +1,252 @@
...
+        /// <returns></returns>
+        public NativeImageSource GetNativeImageSource()
+        {
+            Tizen.Log.Debug("NUI", $"GetNativeImageSource()");
+            return new NativeImageSource(Interop.Capture.GetNativeImageSourcePtr(swigCPtr), true);
+        }
+
+    }
+
+    /// <summary>
```
bshsqa, line 191: Because we will not open CaptureFinishState, this could be removed.

src/Tizen.NUI/src/public/Capture.cs
```diff
@@ -0,0 +1,252 @@
...
+        public void Disconnect(Delegate func)
+        {
+            IntPtr ip = Marshal.GetFunctionPointerForDelegate<Delegate>(func);
+            {
+                Interop.Capture.SignalDisconnect(swigCPtr, new HandleRef(this, ip));
+                if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+            }
+        }
+
+        public void Emit(Capture src, Capture.CaptureFinishedState state)
```
bshsqa, line 246: The parameter can be like bool CaptureSuccess, and this bool value is needed to changed to the enum to be used in the dali SignalEmit.

src/Tizen.NUI/src/public/Capture.cs
```diff
@@ -0,0 +1,252 @@
...
+    using Tizen.NUI.BaseComponents;
+
+    /// <summary>
+    /// Capture
+    /// </summary>
+    public class Capture : BaseHandle
+    {
+        /// <summary>
+        /// Constructor
+        /// </summary>
```
hyunjushin, line 32: If you don't submit ACR, please use    [EditorBrowsable(EditorBrowsableState.Never)] for all public & protected APIs.

dongsug-song:
@hyunjushin 
@bshsqa 
Hello, Thank you for your reviews.
I changed all according to your comments.
please review again.
thank you~ ^^;

src/Tizen.NUI/src/public/Capture.cs
```diff
@@ -0,0 +1,294 @@
...
+            Interop.Capture.Delete(swigCPtr);
+        }
+
+        /// <summary>
+        /// Generate copy constructor.
+        /// </summary>
+        /// <param name="source">A reference to the copied handle</param>
+        /// <returns></returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static Capture Assign(Capture source)
```
hyunjushin, line 98: I can't find "assign" function in native Capture api. I think "assign" is not common name for public api.

src/Tizen.NUI/src/public/Capture.cs
```diff
@@ -0,0 +1,318 @@
...
+            return url;
+        }
+
+    }
+
+    /// <summary>
+    /// CaptureFinishStateEventArgs
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class CaptureFinishStateEventArgs : EventArgs
```
hyunjushin, line 254: CaptureFinishedEventArgs is commonly used style. please do double check.

src/Tizen.NUI/src/public/Capture.cs
dongsug-song: yes, I was wrong. it was not copy constructor. it was '=' operator override in dali. this is not used in nui so I removed this method. thank you.

src/Tizen.NUI/src/public/Capture.cs
dongsug-song: yes. you are right. I changed it as CaptureFinishedEventArgs.

PR #1757, opened 2020-06-25
[NUI] Implement IDisposable to class containing disposable class

jaehyun0cho:
IDisposable is implemented to class containing disposable class instance
in Tizen.NUI.Components and Tizen.NUI.Wearable to resolve CA1001 of
StyleCop.

Extension classes in Tizen.NUI.Components are not included to this patch
because it has OnDispose() method similar to Dispose().

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/FlexibleView/FlexibleView.Helper.cs
+++ b/src/Tizen.NUI.Components/Controls/FlexibleView/FlexibleView.Helper.cs
@@ -73,7 +73,7 @@ namespace Tizen.NUI.Components
             }
         }
 
-        internal class ChildHelper
+        internal class ChildHelper : Disposable
         {
             private FlexibleView mFlexibleView;
 
@@ -179,6 +179,27 @@ namespace Tizen.NUI.Components
                     return null;
                 }
                 return mViewList[index];
+            }
+
+            protected override void Dispose(DisposeTypes type)
+            {
+                if (disposed)
+                {
+                    return;
+                }
+
+                if (type == DisposeTypes.Explicit)
+                {
+                    Clear();
+
+                    if (mTapGestureDetector != null)
+                    {
+                        mTapGestureDetector.Detected -= OnTapGestureDetected;
+                        mTapGestureDetector.Dispose();
+                        mTapGestureDetector = null;
+                    }
+                }
+                base.Dispose(type);
             }
 
             private void OnTapGestureDetected(object source, TapGestureDetector.DetectedEventArgs e)
--- a/src/Tizen.NUI.Components/Controls/FlexibleView/FlexibleViewLayoutManager.cs
+++ b/src/Tizen.NUI.Components/Controls/FlexibleView/FlexibleViewLayoutManager.cs
@@ -11,7 +11,7 @@ namespace Tizen.NUI.Components
     /// </summary>
     /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
     [EditorBrowsable(EditorBrowsableState.Never)]
-    public abstract class FlexibleViewLayoutManager
+    public abstract class FlexibleViewLayoutManager : Disposable
     {
         /// <summary>
         /// Direction
@@ -706,6 +706,26 @@ namespace Tizen.NUI.Components
         protected virtual FlexibleViewViewHolder FindLastVisibleItemView()
         {
             return null;
+        }
+
+        /// <summary>
+        /// Dispose FlexibleView and all children on it.
+        /// </summary>
+        /// <param name="type">Dispose type.</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected override void Dispose(DisposeTypes type)
+        {
+            if (disposed)
+            {
+                return;
+            }
+
+            if (type == DisposeTypes.Explicit)
+            {
+                mScrollAni?.Dispose();
+            }
+
+            base.Dispose(type);
         }
 
         internal virtual FlexibleViewViewHolder OnFocusSearchFailed(FlexibleViewViewHolder focused, FlexibleViewLayoutManager.Direction direction, FlexibleViewRecycler recycler)
--- a/src/Tizen.NUI.Components/Style/ControlStyle.cs
+++ b/src/Tizen.NUI.Components/Style/ControlStyle.cs
@@ -27,8 +27,16 @@ namespace Tizen.NUI.Components
     /// <since_tizen> 6 </since_tizen>
     /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
     [EditorBrowsable(EditorBrowsableState.Never)]
-    public class ControlStyle : ViewStyle
+    public class ControlStyle : ViewStyle, global::System.IDisposable
     {
+        /// <summary>
+        /// A Flag to check if it is already disposed.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected bool disposed = false;
+
+        private bool isDisposeQueued = false;
+
         static ControlStyle () { }
 
         /// <summary>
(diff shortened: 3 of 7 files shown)


src/Tizen.NUI.Components/Controls/FlexibleView/FlexibleView.Helper.cs
```diff
@@ -181,6 +181,32 @@ public FlexibleViewViewHolder GetChildAt(int index)
...
+                    Clear();
+
+                    if (mTapGestureDetector != null)
+                    {
+                        mTapGestureDetector.Detected -= OnTapGestureDetected;
+                        mTapGestureDetector.Dispose();
+                        mTapGestureDetector = null;
+                    }
+
+                    if (mFlexibleView != null)
```
rabbitfor, line 202: In my opinion,
A FlexibleView owns a ChildHelper instance and ChildHelper.mFlexibleView is just a pointer to its owner. So that the dispose ownership of mFlexibleView is not held by ChildHelper.

[CA1001](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1001?view=vs-2019)
> It is OK to suppress the warning when the dispose ownership of the field(s) is not held by the containing type.

I think instead of disposing mFlexibleView here, it needs to dispose FlexibleView.mChildHelper in FlexibleView.Dispose().

        {
            ......
                if (mChildHelper != null)
                {
                    mChildHelper.Clear();
                    mChildHelper.Dispose();
                    mChildHelper = null;
                }
            }
            base.Dispose(type);
        }


What do you think?

src/Tizen.NUI.Components/Controls/FlexibleView/FlexibleViewLayoutManager.cs
```diff
@@ -708,6 +708,31 @@ protected virtual FlexibleViewViewHolder FindLastVisibleItemView()
...
+            if (disposed)
+            {
+                return;
+            }
+
+            if (type == DisposeTypes.Explicit)
+            {
+                mScrollAni?.Dispose();
+
+                if (mFlexibleView != null)
```
rabbitfor, line 727: ditto

src/Tizen.NUI.Components/Controls/Notification.cs
```diff
@@ -26,7 +26,7 @@ namespace Tizen.NUI.Components
     /// </summary>
     /// <privilege>http://tizen.org/privilege/window.priority.set</privilege>
     [EditorBrowsable(EditorBrowsableState.Never)]
-    public class Notification
+    public class Notification : Disposable
```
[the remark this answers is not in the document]
jaehyun0cho: I see. I will revert the Notification part.

src/Tizen.NUI.Components/Controls/FlexibleView/FlexibleView.Helper.cs
jaehyun0cho: I think you are right! I will update the code.

src/Tizen.NUI.Components/Controls/FlexibleView/FlexibleViewLayoutManager.cs
jaehyun0cho: Thank you! I will also remove the codes related to mFlexibleView here.

rabbitfor:
I think it needs code that dispose both mLayout and mChildHelper in FlexibleView,

    // FlexibleView.cs
    {
        ......
             if (mLayout != null)
             {
                 mLayout.StopScroll(false);
                 mLayout.Dispose();
             }
       .......
            if (mChildHelper != null)
            {
                mChildHelper.Clear();
                mChildHelper.Dispose();
            }
        }
        base.Dispose(type);
    }

src/Tizen.NUI.Components/Style/ControlStyle.cs
```diff
@@ -56,6 +64,63 @@ public ControlStyle(ControlStyle style) : base(style)
...
+            //Throw excpetion if Dispose() is called in separate thread.
+            if (!Window.IsInstalled())
+            {
+                throw new global::System.InvalidOperationException("This API called from separate thread. This API must be called from MainThread.");
+            }
+
+            if (isDisposeQueued)
+            {
+                Dispose(DisposeTypes.Implicit);
+            }
```
idkiller, line 95: How will this code be called?
The code that makes isDisposeQueued true is only in the destructor,
Dispose method cannot be called after Destructor.

jaehyun0cho:
> LGTM

Thank you for your comments!
I have one thing to ask your opinion about mLayout.Dispose() in FlexibleView.cs

mChildHelper is obvious to be disposed because it is constructed in FlexibleView constructor.
But mLayout is set in SetLayoutManager() so if it is designed to keep mLayout alive and use it again after FlexibleView is destructed, then I think that mLayout should not be disposed.

What do you think?

src/Tizen.NUI.Components/Style/ControlStyle.cs
rabbitfor: As I understand,
when the instance is added to the DisposeQueue in destructor,
```
// ControlStyle.cs
~ControlStyle()
{
    if (!isDisposeQueued)
    {
        isDisposeQueued = true;
        DisposeQueue.Instance.Add(this);
    }
}
```
The queue executes callback that calls `dispoable.Dispose()`
```
// DisposeQueue.cs
public void Add(IDisposable disposable)
{
    lock (_listLock)
    {
        _disposables.Add(disposable);
    }

    if (_eventThreadCallback != null)
    {
        _eventThreadCallback.Trigger(); // -> this calls ProcessDisposables() in main thread;
    }
}

public void ProcessDisposables()
{
    lock (_listLock)
    {
        foreach (IDisposable disposable in _disposables)
        {
            disposable.Dispose();
        }
        _disposables.Clear();
    }
}
```

rabbitfor:
> > LGTM
> 
> Thank you for your comments!
> I have one thing to ask your opinion about mLayout.Dispose() in FlexibleView.cs
> 
> mChildHelper is obvious to be disposed because it is constructed in FlexibleView constructor.
> But mLayout is set in SetLayoutManager() so if it is designed to keep mLayout alive and use it again after FlexibleView is destructed, then I think that mLayout should not be disposed.
> 
> What do you think?

You are right.
It needs to call something like mLayout.ClearRecyclerView() instead of calling mLayout.Dispose();

```
// FlexibleView.cs
{
    ......

        if (mLayout != null)
        {
            mLayout.StopScroll(false);
            mLayout.ClearRecyclerView();
            mLayout = null;
        }
    .......
}
```
```
// FlexibleViewLayoutManager.cs

// This clears dependancy of FlexibleView created by SetRecyclerView()
internal void SetRecyclerView(FlexibleView recyclerView) 
{
     mFlexibleView = null;
     mChildHelper = null;
}
```

jaehyun0cho:
> > > LGTM
> > 
> > 
> > Thank you for your comments!
> > I have one thing to ask your opinion about mLayout.Dispose() in FlexibleView.cs
> > mChildHelper is obvious to be disposed because it is constructed in FlexibleView constructor.
> > But mLayout is set in SetLayoutManager() so if it is designed to keep mLayout alive and use it again after FlexibleView is destructed, then I think that mLayout should not be disposed.
> > What do you think?
> 
> You are right.
> It needs to call something like mLayout.ClearRecyclerView() instead of calling mLayout.Dispose();
> 
> ```
> // FlexibleView.cs
> protected override void Dispose(DisposeTypes type)
> {
>     ......
> 
>         if (mLayout != null)
>         {
>             ...
>             mLayout.ClearRecyclerView();
>             mLayout = null;
>         }
>     .......
> }
> ```
> 
> ```
> // FlexibleViewLayoutManager.cs
> 
> // This clears dependancy of FlexibleView created by SetRecyclerView()
> internal void ClearRecyclerView() 
> {
>      mFlexibleView = null;
>      mChildHelper = null;
> }
> ```

Thank you for your reply! I've made your codes into a pull request https://github.com/Samsung/TizenFX/pull/1763

PR #1762, opened 2020-06-26
[NUI] add ControlState class to support combined and custom state

dasg34:
### Custom State

You can define your own control states. first, declare ControlState variable
and initialize using  `ControlState.Create`.
```
public class MyButton : Button {
   ...
   ControlState MyState = ControlState.Create("MyState");
   ...
}
```
And assign to `View.ControlState`.
```
void OnStateChanged(...) {
    ControlState = MyState;
}
```

If you want to assign the value of custom state to `Selector`, use
`Add(ControlState state, object value)`.
```
Selector<string> textSelector = new Selector<string>();
textSelector.Add(ControlState.Pressed, "Pressed!");
textSelector.Add(MyState, "MyText");
```
or
```
Selector<string> textSelector = new Selector<string>()
{
    { ControlState.Pressed, "Pressed!" },
    { MyState, "MyText" }
};
```

 ### Combined State
To implement your own combined control states, you can use the `+` or `+=`
operators.
For example, `MyCombinedStateX` are all the same.
```
ControlState MyCombinedStateA =
   ControlState.Create(ControlState.Pressed, ControlState.Focused);

ControlState MyCombinedStateB = ControlState.Pressed + ControlState.Focused;
```

Note that `Normal` and `All` state cannot be combined with other states.
`Normal` state is ignored. however, `All` state will ignore other states.
```
ControlState.Create(ControlState.Pressed, ControlState.Focused, ControlState.Normal) ==
   ControlState.Create(ControlState.Pressed, ControlState.Focused)

ControlState.Create(ControlState.All, ControlState.Pressed, ControlState.Focused) ==
   ControlState.Create(ControlState.All)
```

 ### ControlState in Xaml

It will support new initialization syntax.
Legacy:
```
<c:TextLabelStyle.Text>
    <c:Selector x:TypeArguments="x:String" Normal="button" Pressed="pressed!!"/>
</c:TextLabelStyle.Text>
```

New:
```
<c:TextLabelStyle.Text>
    <c:StateValuePair x:TypeArguments="x:String" State="Normal" Value="button"/>
    <c:StateValuePair x:TypeArguments="x:String" State="Pressed" Value="pressed!!" />
    <c:StateValuePair x:TypeArguments="x:String" State="MyState" Value="my state!" />
    <c:StateValuePair x:TypeArguments="x:String" State="MyState,Normal,Focused" Value="my combined state!" />
</c:TextLabelStyle.Text>
```

However, it won't work because there is no converter of  `ControlState`.
The next step would be to implement `ControlState` converter.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
@@ -1,0 +1,315 @@
+/*
+ * Copyright(c) 2020 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+
+namespace Tizen.NUI.BaseComponents
+{
+    /// <summary>
+    /// Class for describing the states of the view.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class ControlState : IEquatable<ControlState>
+    {
+        private static readonly Dictionary<string, ControlState> stateDictionary = new Dictionary<string, ControlState>();
+        //Default States
+        /// <summary>
+        /// All State.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly ControlState All = Create("All");
+        /// <summary>
+        /// Normal State.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly ControlState Normal = Create("Normal");
+        /// <summary>
+        /// Focused State.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly ControlState Focused = Create("Focused");
+        /// <summary>
+        /// Pressed State.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly ControlState Pressed = Create("Pressed");
+        /// <summary>
+        /// Disabled State.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly ControlState Disabled = Create("Disabled");
+        /// <summary>
+        /// Selected State.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly ControlState Selected = Create("Selected");
+        /// <summary>
+        /// DisabledSelected State.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly ControlState DisabledSelected = Disabled + Selected;
+        /// <summary>
+        /// DisabledFocused State.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly ControlState DisabledFocused = Disabled + Focused;
+        /// <summary>
+        /// SelectedFocused State.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly ControlState SelectedFocused = Selected + Focused;
+        /// <summary>
+        /// Other State.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly ControlState Other = Create("Other");
+
+        private List<ControlState> stateList = new List<ControlState>();
+        private readonly string name = "";
+
+        /// <summary>
+        /// Gets or sets a value indicating whether it has combined states.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool IsCombined => stateList.Count > 1;
+
+        /// <summary>
+        /// Default Contructor. Please use <see cref="Create(string)"/> or <see cref="Create(ControlState[])"/> instead.
+        /// </summary>
+        // Do not open this constructor. This is only for xaml support.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public ControlState() { }
+
+        private ControlState(string name) : this() => this.name = name;
+
+        /// <summary>
+        /// Create an instance of the <see cref="ControlState"/> with state name.
+        /// </summary>
+        /// <param name="name">The state name.</param>
+        /// <returns>The <see cref="ControlState"/> instance which has single state.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static ControlState Create(string name)
+        {
+            if (name == null)
+                throw new ArgumentNullException(nameof(name));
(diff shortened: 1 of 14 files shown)


rabbitfor:
:thumbsup:

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -0,0 +1,280 @@
...
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public ControlState() => stateSet = new List<ControlState>();
+
+        private ControlState(string name) : this() => this.name = name;
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static ControlState Create(string name)
+        {
+            if (stateDictionary.TryGetValue(name, out ControlState state))
```
rabbitfor, line 67: Is state name case sensitive?

`All` and `all`
`Normal` and `normal`
and so on..

rabbitfor:
Can you add a description of public APIs (including hidden)  to help to understand what the APIs for?
```
/// <summary></summary>
```

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -0,0 +1,280 @@
...
+        {
+            if (states.Length == 1)
+                return states[0];
+
+            ControlState newState = new ControlState();
+            for (int i = 0; i < states.Length; i++)
+            {
+                if (states[i] == Normal)
+                    continue;
+                newState.stateSet.AddRange(states[i].stateSet);
```
rabbitfor, line 131: In my opinion, duplicates should be handled.
```
ControlState.CreateCombinedState(DisableFocused, SelectedFocused) -> Disable, Focused, Selected
```

rabbitfor:
Is it OK not to fix setter/getter of Selector APIs?
-> All, Normal, Pressed, Focused, Selected, Disabled, DisabledFocused, SelectedFocused, DisabledSelected, Other

It doesn't seem to be an item in `List<StateValuePair>`.

dasg34:
> I think `ControlState class` seems to have implemented a flag.
> Is there any reason not to just use an enum with the Flags attribute?
> 
> ```cs
> [Flags]
> public enum ControlState
> {
>   Normal = 1,
>   Focused = 2,
>   Pressed = 4,
>   Disabled = 8,
>   Selected = 16,
>   DisabledSelected = 32,
>   DisabledFocused = 64,
>   SelectedFocused = 128,
>   All = Normal | Focused | Pressed | Disabled | Selected | DisabledSelected | DisabledFocused | SelectedFocused
> }
> ```

Because  it is not possible to add custom enum value to `ControlStates  enum` in runtime.

dasg34:
Selector.cs

   1. Fix setter/getter of Selector APIs.
   2. Remove `public override void Add(ControlState state, object value)`.

ControlState.cs

   1. Rename from `CreateCombinedState` to `Create`.
   2. Rename from `ControlState.HasState` to `ControlState.Contains`.
   3. Rename from `ControlState.stateSet` to `ControlState.stateList`.
   4. Make `ControlState.stateList` distinct list and cleanup `Equals`.
   5. Remove needless `IComparable<ControlState>`.
   6. Remove case-insensitive string comparison of "normal", "all".
   7. Merge `GetState` into `Create`.

Documentation : WIP

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
dasg34: Yes, state name is case sensitive.

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -0,0 +1,280 @@
...
+        private ControlState(string name) : this() => this.name = name;
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static ControlState Create(string name)
+        {
+            if (stateDictionary.TryGetValue(name, out ControlState state))
+                return state;
+
+            state = new ControlState(name);
+            state.stateSet.Add(state);
```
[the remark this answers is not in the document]
dasg34: I guess this is a mistake. `Getstate` is merged into `Create`.

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -0,0 +1,315 @@
...
+            return Contains(other);
+        }
+
+        ///  <inheritdoc/>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override bool Equals(object obj) => this.Equals(obj as ControlState);
+
+        ///  <inheritdoc/>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override int GetHashCode() => stateList.GetHashCode();
```
rabbitfor, line 203: Is it safe? I think List.GetHashCode() may use stateList[i].GetHashCode() which can raises infinite loop.

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -0,0 +1,315 @@
...
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public struct StateValuePair : IEquatable<StateValuePair>
+    {
+        /// <summary>
+        /// The constructor with the specified state and value.
+        /// </summary>
+        /// <param name="state">The state</param>
+        /// <param name="value">The value associated with state.</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public StateValuePair(ControlState state, object value)
```
rabbitfor, line 257: I am wondering if the `Value` is `object` type, what will xaml code would look like,

This is example of Selector<string>, I think it seems not have any problems,
```
<c:TextLabelStyle.Text>
    <c:StateValuePair State="Normal" Value="button"/>
    <c:StateValuePair State="Pressed" Value="pressed!!" />
</c:TextLabelStyle.Text>
```

But how about Selector<Color>?
The Color can be expressed as a string (by Converter), e.g. "Yellow".
```
<c:TextLabelStyle.TextColor>
    <c:StateValuePair State="Normal" Value="Yellow"/>
</c:TextLabelStyle.Text>
```
Is it possible? In my opinion, the StateValuePair would take it as a string (no conversion to Color) and that will cause a problem when getter is called. (e.g. Color a = selector.Normal; -> Can not cast to Color.)

src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
```diff
@@ -158,42 +158,49 @@ public T DisabledSelected
...
         /// <summary>
         /// Get value by State.
         /// </summary>
         /// <since_tizen> 6 </since_tizen>
         /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public T GetValue(ControlStates state)
+        public T GetValue(ControlState state)
         {
             if(All != null)
```
rabbitfor, line 172: In the future, we may think of optimizing speed by adding private field `StateValuePair all;`

Because `All` is the most popular state.
- Selector calls `All` frequently in `GetValue` (which is called everytime when the View's control state has changed).
- The `All` may used a lot because when a user does not need a selector with multiple state value, `All` is used.
Note that the below examples are the same.
```
TextLabelStyle a = new TextLabelStyle
{
    TextColor = "Yellow",
}
```
```
TextLabelStyle a = new TextLabelStyle
{
    TextColor = new Selector<Color>{ All = "Yellow" };
}
```

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
dasg34: Template type T seems much better for `StateValuePair`. it will also support xaml intellisense of `StateValuePair.Value`.

PR #1769, opened 2020-06-30
[NUI] Hide classes that will be move to specific tizen profile

rabbitfor:

### API Changes ###

 - ACR: https://code.sec.samsung.net/jira/browse/TCSACR-335

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Toast.cs
+++ b/src/Tizen.NUI.Components/Controls/Toast.cs
@@ -27,6 +27,7 @@ namespace Tizen.NUI.Components
     /// A toast will automatically disappear after a certain time.
     /// </summary>
     /// <since_tizen> 6 </since_tizen>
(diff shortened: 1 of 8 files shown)


src/Tizen.NUI.Components/Controls/Toast.cs
```diff
@@ -283,7 +294,7 @@ public uint Duration
         /// Apply style to toast.
         /// </summary>
         /// <param name="viewStyle">The style to apply.</param>
-		/// <since_tizen> 8 </since_tizen>
+		[EditorBrowsable(EditorBrowsableState.Never)]
```
dongsug-song, line 297: could you check indentation? you can use "ctrl + k + d" in VS. ^^;

rabbitfor: Done! Thank you :)

PR #1771, opened 2020-07-01
[NUI] Fix to not reset IsSelected on pressed

woohyun:
### Description of Change ###
IsSelected in Button was reset to "false", since "target" state is only set with Pressed without any consideration on current IsSelected value.
(You can easily check the issue by clicking twice CheckBox control. It's never be unselected.)

I just reverted back some logics in UpdateState. If other modifications are needed, please give me feedback.


woohyun:
This would be fixed by another way.

PR #1777, opened 2020-07-01
[NUI] Make Pagination class public and fix CA warnings

Seoyeon2Kim:

### Description of Change ###
- Change Pagination APIs  to public ones and fix CA warnings.

### API Changes ###
 - ACR: TCSACR-338
 (https://code.sec.samsung.net/jira/browse/TCSACR-338)

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Pagination.cs
+++ b/src/Tizen.NUI.Components/Controls/Pagination.cs
@@ -1,5 +1,5 @@
 ﻿/*
- * Copyright(c) 2019 Samsung Electronics Co., Ltd.
+ * Copyright(c) 2020 Samsung Electronics Co., Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,9 +24,7 @@ namespace Tizen.NUI.Components
     /// <summary>
     /// Pagination shows the number of pages available and the currently active page.
     /// </summary>
-    /// <since_tizen> 6 </since_tizen>
-    /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
-    [EditorBrowsable(EditorBrowsableState.Never)]
+    /// <since_tizen> 8 </since_tizen>
     public class Pagination: Control
     {
         private VisualView container;
@@ -41,9 +39,7 @@ namespace Tizen.NUI.Components
         /// <summary>
         /// Creates a new instance of a Pagination.
         /// </summary>
-        /// <since_tizen> 6 </since_tizen>
-	    /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
-	    [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <since_tizen> 8 </since_tizen>
         public Pagination() : base()
         {
             Initialize();
@@ -52,9 +48,8 @@ namespace Tizen.NUI.Components
         /// <summary>
         /// Creates a new instance of a Pagination using style.
         /// </summary>
-        /// <since_tizen> 6 </since_tizen>
-        /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <param name="style">The string to initialize the Pagination</param>
+        /// <since_tizen> 8 </since_tizen>
         public Pagination(string style) : base(style)
         {
             Initialize();
@@ -63,9 +58,8 @@ namespace Tizen.NUI.Components
         /// <summary>
         /// Creates a new instance of a Pagination using style.
         /// </summary>
-        /// <since_tizen> 6 </since_tizen>
-        /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <param name="paginationStyle">The style object to initialize the Pagination</param>
+        /// <since_tizen> 8 </since_tizen>
         public Pagination(PaginationStyle paginationStyle) : base(paginationStyle)
         {
             Initialize();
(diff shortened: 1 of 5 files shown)


src/Tizen.NUI.Components/Controls/Pagination.cs
```diff
@@ -330,7 +301,7 @@ private void CreateIndicator()
                 Size = new Size2D((int)Style.IndicatorSize.Width, (int)Style.IndicatorSize.Height),
                 Opacity = 0.5f,
             };
-            indicator.Position = new Position2D((int)(Style.IndicatorSize.Width + Style.IndicatorSpacing) * indicatorList.Count, 0);
+            indicator.Position = new Vector2((int)(Style.IndicatorSize.Width + Style.IndicatorSpacing) * indicatorList.Count, 0);
```
dasg34, line 304: Note that, CA2000 warning is raised on implicit operator of `Position2D`. when you assign `Position2D` to `Position` Property, implicit operator is called. (https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/public/Position2D.cs#L313)
In this implicit operator, the `position2D` object passed as parameter is not disposed.
CA2000 will be fixed with following code. however, since it is related to dali life-cycle, i'm not sure this is proper fix.
```diff
diff --git a/src/Tizen.NUI/src/public/Position2D.cs b/src/Tizen.NUI/src/public/Position2D.cs
index 48f2595f7..65f696754 100755
--- a/src/Tizen.NUI/src/public/Position2D.cs
+++ b/src/Tizen.NUI/src/public/Position2D.cs
@@ -312,7 +312,9 @@ namespace Tizen.NUI
         /// <since_tizen> 3 </since_tizen>
         public static implicit operator Vector2(Position2D position2d)
         {
-            return new Vector2((float)position2d.X, (float)position2d.Y);
+            var ret = new Vector2((float)position2d.X, (float)position2d.Y);
+            position2d.Dispose();
+            return ret;
         }

         /// <summary>
```

So, I think this kind of CA2000 should be handled in `Position2D`.

rabbitfor: The position2d is from method parameter, so I think the disposal should be handled by the user that passed the argument.

rabbitfor:
```
Position2D p = new Position2D(10, 10);
Vector2 v1 = p;
Vector2 v2 = p;
Vector2 v3 = p;
```

Seoyeon2Kim: Thank you for reviewing it.
I wondered if I would change the operator in `Position2D`.
However, `Position` type of `Visual` is `Vector2`, so I changed the same here.

This patch goes like this,
and I will consider again where the similar warning occurs next time.

src/Tizen.NUI.Components/Controls/Pagination.cs
```diff
@@ -104,9 +95,7 @@ public Size IndicatorSize
         /// <summary>
         /// Gets or sets the background resource of indicator.
         /// </summary>
-        /// <since_tizen> 6 </since_tizen>
-        /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <since_tizen> 8 </since_tizen>
         public Selector<string> IndicatorImageURL
```
m-macias, line 99: Please use the PascalCase naming convention for the acronyms. Also why to Uri return type?
```suggestion
        public Selector<string> IndicatorImageUrl
```

src/Tizen.NUI.Components/Style/PaginationStyle.cs
```diff
@@ -54,33 +49,26 @@ public PaginationStyle(PaginationStyle style) : base(style)
...
         public Size IndicatorSize { get; set; }
 
         /// <summary>
         /// Gets or sets the resource of indicator.
         /// </summary>
-        /// <since_tizen> 6 </since_tizen>
-        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <since_tizen> 8 </since_tizen>
         public Selector<string> IndicatorImageURL { get; set; } = new Selector<string>();
```
m-macias, line 59:
```suggestion
        public Selector<string> IndicatorImageUrl { get; set; } = new Selector<string>();
```

src/Tizen.NUI.Components/Controls/Pagination.cs
Seoyeon2Kim: Changed the name as you suggested.
But, in this sentence "why to Uri return type?", what do you mean?

src/Tizen.NUI.Components/Style/PaginationStyle.cs
Seoyeon2Kim: Done.

src/Tizen.NUI.Components/Controls/Pagination.cs
m-macias: Sorry, I missed the point in the sentence. I meant, Why do you return Selector<string> instead of Uri return type, if your property is returning ImageUrl?

Seoyeon2Kim: Oh, okay. The reason why the return type is 'Selector<>' is that user can set various image paths according to control states, such as Normal, Focused, and Pressed.
For example, in an application :
                IndicatorImageURL = new Selector<string>
                {
                    Normal = Tizen.Applications.Application.Current.DirectoryInfo.Resource + "normalImage.png",
                    Selected = Tizen.Applications.Application.Current.DirectoryInfo.Resource + "focusedImage.png"
                }

Seoyeon2Kim: @m-macias , could you check my comment please?
I'd like to get an accept the ACR and merge this patch by this week.

PR #1774, opened 2020-07-01
[MediaTool] Add new audio/video mimetype

hsgwon:
### Description of Change ###

Add new audio, video mimetypes

### API Changes ###

 - TCSACR-337

PR #1792, opened 2020-07-02
[MediaCodec] Add new codec type - Opus

hsgwon:
### Description of Change ###

Add new codec type - Opus

### API Changes ###

 - TCSACR-344


hsgwon:
https://github.com/Samsung/TizenFX/pull/1774 should be merged first to avoid build break.

PR #1793, opened 2020-07-03
[AudioIO] Add new volume property to set/get recording volume

hsgwon:
### Description of Change ###

Add new volume property to set/get recording volume

### API Changes ###

 - TCSACR-336

### DIFF
--- a/src/Tizen.Multimedia.AudioIO/AudioIO/AudioCapture.cs
+++ b/src/Tizen.Multimedia.AudioIO/AudioIO/AudioCapture.cs
@@ -24,6 +24,7 @@ namespace Tizen.Multimedia
     /// Provides the ability to directly manage the system audio input devices.
     /// </summary>
     /// <privilege>http://tizen.org/privilege/recorder</privilege>
+    /// <feature>http://tizen.org/feature/microphone</feature>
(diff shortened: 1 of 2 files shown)


src/Tizen.Multimedia.AudioIO/AudioIO/AudioCapture.cs
```diff
@@ -176,9 +177,45 @@ internal void ValidateState(params AudioIOState[] desiredStates)
         /// <since_tizen> 3 </since_tizen>
         public AudioSampleType SampleType { get; }
 
+        /// <summary>
+        /// Gets or sets the volume of the audio input data stream.
+        /// </summary>
+        /// <value>
+        /// The default value is 1.0.<br/>
+        /// The valid range is greater than or equal to 0.0 and less than or equal to 2.0.<br/>
+        /// Note that if the value is less than 0.0, it will be set 0.0 and if the value is greater than 2.0, it will be set 2.0.
```
journey2w, line 186:
```suggestion
        /// Note that if the value is less than 0.0, it will be set as 0.0 and if the value is greater than 2.0, it will be set as 2.0.
```

PR #1794, opened 2020-07-03
[NUI] Fix VisualMap.OutputVisualMap

rabbitfor:
* Previously, VisualMap.OutputVisualMap returns a visual map without transform data.
  It causes misunderstanding frequently.
  Now it returns full visual map including transform data.
* Fix VisualMap.OutputVisualMap to create new PropertyMap everytime to
  prevent increasing map size.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/VisualMaps.cs
+++ b/src/Tizen.NUI/src/public/VisualMaps.cs
@@ -661,10 +661,7 @@ namespace Tizen.NUI
         /// <since_tizen> 3 </since_tizen>
         protected virtual void ComposingPropertyMap()
         {
-            if (null == _outputVisualMap)
-            {
-                _outputVisualMap = new PropertyMap();
-            }
+            _outputVisualMap = new PropertyMap();
 
             if (_shader != null) { _outputVisualMap.Add(Visual.Property.Shader, new PropertyValue(_shader)); }
             if (_premultipliedAlpha != null) { _outputVisualMap.Add(Visual.Property.PremultipliedAlpha, new PropertyValue((bool)_premultipliedAlpha)); }
@@ -672,6 +669,13 @@ namespace Tizen.NUI
             if (_opacity != null) { _outputVisualMap.Add(Visual.Property.Opacity, new PropertyValue((float)_opacity)); }
             if (_visualFittingMode != null) { _outputVisualMap.Add(Visual.Property.VisualFittingMode, new PropertyValue((int)_visualFittingMode)); }
             if (_cornerRadius != null) { _outputVisualMap.Add(Visual.Property.CornerRadius, new PropertyValue((int)_cornerRadius)); }
+
+            var transformMap = OutputTransformMap;
+
+            if (!transformMap.Empty())
+            {
+                _outputVisualMap.Add(Visual.Property.Transform, new PropertyValue(transformMap));
+            }
         }
 
         private void ComposingTransformMap()


src/Tizen.NUI/src/public/VisualMaps.cs
```diff
@@ -661,17 +661,21 @@ internal void UpdateVisual(bool force = false)
         /// <since_tizen> 3 </since_tizen>
         protected virtual void ComposingPropertyMap()
         {
-            if (null == _outputVisualMap)
-            {
-                _outputVisualMap = new PropertyMap();
-            }
+            _outputVisualMap = new PropertyMap();
```
dongsug-song, line 664: it looks like _outputVisualMap is created at every time. it seems different from the previous behavior. (previously, it is checked as null) how do think about? ^^;

src/Tizen.NUI/src/public/VisualMaps.cs
```diff
@@ -661,17 +661,21 @@ internal void UpdateVisual(bool force = false)
...
             if (_shader != null) { _outputVisualMap.Add(Visual.Property.Shader, new PropertyValue(_shader)); }
             if (_premultipliedAlpha != null) { _outputVisualMap.Add(Visual.Property.PremultipliedAlpha, new PropertyValue((bool)_premultipliedAlpha)); }
             if (_mixColor != null) { _outputVisualMap.Add(Visual.Property.MixColor, new PropertyValue(_mixColor)); }
             if (_opacity != null) { _outputVisualMap.Add(Visual.Property.Opacity, new PropertyValue((float)_opacity)); }
             if (_visualFittingMode != null) { _outputVisualMap.Add(Visual.Property.VisualFittingMode, new PropertyValue((int)_visualFittingMode)); }
             if (_cornerRadius != null) { _outputVisualMap.Add(Visual.Property.CornerRadius, new PropertyValue((int)_cornerRadius)); }
+
+            var transformMap = OutputTransformMap;
+
+            if (!transformMap.Empty())
```
dongsug-song, line 675: very good to empty check! (b^_^)b

woohyun: One question.
Is there any reason that we cannot use "OutputTransformMap" directly instead of creating  "transformMap" ?

src/Tizen.NUI/src/public/VisualMaps.cs
rabbitfor: If the _outputVisualMap is not created everytime, the map can have multiple values for the same key (MIX_COLOR, OPACITY..)
Because `PropertyMap.Add(key, value)` does not check duplicates. (DALi says it's better to create new one instead reusing)

dongsug-song: OK. I see. thank you for explaining.

src/Tizen.NUI/src/public/VisualMaps.cs
rabbitfor: In the next line, transformMap is re-used.
```C#
var transformMap = OutputTransformMap;

if (!transformMap.Empty())
{
    _outputVisualMap.Add(Visual.Property.Transform, new PropertyValue(transformMap));
}
```
If it use `OutputTransformMap` property directly, it calls property getter twice and that leads creating transform map twice redundantly.

woohyun: I understand. Thanks for answering.

PR #1797, opened 2020-07-03
[NUI] Fix Loading size property not work

zg2nets:
### Description of Change ###
Fix Loading size property not work

### API Changes ###

 - None


rabbitfor:
Please add prefix `[NUI]` to the commit message

PR #1798, opened 2020-07-03
[NUI] Fix Switch selection bug.

rabbitfor:
Previously, when a Switch is selected by code, the thumb is not moved to left/right.
```
var switchButton = new Switch()
{
  IsSelected = true,
}
```
Thumb is moved only if the user touches the switch area or presses return key.

This patch fixes this problem and to do that, it needed to change belows,

* Switch does not listen Touch/Key event anymore to detect selection changed
  Instead it uses ControlState changed event which is clearer.
* Button produces more complicate combined state such as "SelectedPressed".
* Clean up the code that is redundant in Selector.GetValue().

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Extension/SlidingSwitchExtension.cs
+++ b/src/Tizen.NUI.Components/Controls/Extension/SlidingSwitchExtension.cs
@@ -36,16 +36,11 @@ namespace Tizen.NUI.Components.Extension
             slidingAnimation = new Animation(100);
         }
 
-        public override void OnClick(Button button, Button.ClickEventArgs eventArgs)
+        /// <inheritdoc/>
+        public override void OnSelectedChanged(Switch switchButton)
         {
-            if (button as Switch == null)
-            {
-                throw new Exception("SlidingSwitchExtension must be used within a SwitchStyle or derived class.");
-            }
-
-            var switchButton = (Switch)button;
-            var track = switchButton.GetCurrentTrack(this);
-            var thumb = switchButton.GetCurrentThumb(this);
+            var track = switchButton.Track;
+            var thumb = switchButton.Thumb;
 
             if (track == null || thumb == null || null == slidingAnimation)
             {
--- a/src/Tizen.NUI.Components/Controls/Switch.cs
+++ b/src/Tizen.NUI.Components/Controls/Switch.cs
@@ -273,18 +273,7 @@ namespace Tizen.NUI.Components
         /// <since_tizen> 8 </since_tizen>
         public override bool OnKey(Key key)
         {
-            if (!IsEnabled || null == key) return false;
-
-            bool ret = base.OnKey(key);
-            if (key.State == Key.StateType.Up)
-            {
-                if (key.KeyPressedName == "Return")
-                {
-                    OnSelect();
-                }
-            }
-
-            return ret;
+            return base.OnKey(key);
         }
 
         /// <summary>
@@ -296,19 +285,7 @@ namespace Tizen.NUI.Components
         /// <since_tizen> 8 </since_tizen>
         public override bool OnTouch(Touch touch)
         {
-            if(!IsEnabled || null == touch) return false;
-
-            PointStateType state = touch.GetState(0);
-            bool ret = base.OnTouch(touch);
-            switch (state)
-            {
-                case PointStateType.Up:
-                    OnSelect();
-                    break;
-                default:
-                    break;
-            }
-            return ret;
+            return base.OnTouch(touch);
         }
 
         /// <summary>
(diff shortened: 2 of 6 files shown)


src/Tizen.NUI.Components/Controls/Switch.cs
```diff
@@ -343,6 +339,8 @@ protected override void OnThemeChangedEvent(object sender, StyleManager.ThemeCha
 
         private void OnSelect()
         {
+            ((SwitchExtension)Extension)?.OnSelectedChanged(this);
```
dongsug-song, line 342: great! (b^_^)b

https://docs.microsoft.com/ko-kr/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-

src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
```diff
@@ -169,42 +169,29 @@ public T Other
...
+
+            StateValuePair<T> value = Find(x => x.State == state);
+            if (value.Value != null)
+            {
+                return value.Value;
+            }
+
+            if (state.IsCombined)
             {
-                case null:
```
dongsug-song, line 178: looks like this switch-case part has been removed. I couldn't understand the whole logic so I just ask if this is intentionally modified, and if there is replaced logic.

src/Tizen.NUI.Components/Controls/Extension/SlidingSwitchExtension.cs
```diff
@@ -36,16 +36,11 @@ public SlidingSwitchExtension() : base()
             slidingAnimation = new Animation(100);
         }
 
-        public override void OnClick(Button button, Button.ClickEventArgs eventArgs)
+        /// <inheritdoc/>
+        public override void OnSelectedChanged(Switch switchButton)
         {
-            if (button as Switch == null)
-            {
```
dongsug-song, line 42: looks like "throw new Exception(...)" has been removed. isn't it?
is it intentional removal?

src/Tizen.NUI.Components/Controls/Switch.cs
```diff
@@ -361,33 +359,5 @@ public class SelectEventArgs : EventArgs
...
-
-        /// <summary>
-        /// Get current track part to the attached SwitchExtension.
-        /// </summary>
-        /// <remarks>
-        /// It returns null if the passed extension is invaild.
-        /// </remarks>
-        /// <param name="extension">The extension instance that is currently attached to this Switch.</param>
-        /// <return>The switch's track part.</return>
-        [EditorBrowsable(EditorBrowsableState.Never)]
```
dongsug-song, line 373: looks like this is removed. is it OK? ^^;

src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
rabbitfor: I removed switch-case statement intentionally becuase it can be covered by following code,

```C#
// This covers Normal ,Focused, Pressed, Disabled, Selected cases
StateValuePair<T> value = Find(x => x.State == state); 
if (value.Value != null)
{
    return value.Value;
}

// This covers DisabledFocused, DisabledSelected, SelectedFocused cases
if (state.IsCombined)
{
    value = Find(x => state.Contains(x.State));
    if (value.Value != null)
    {
        return value.Value;
    }
}
```

src/Tizen.NUI.Components/Controls/Extension/SlidingSwitchExtension.cs
rabbitfor: Yes :) Previous code, it was `OnClick(Button button)`. I moved body of `OnClick` to `OnSelectedChanged(Switch)`.
Unlike `OnSelectedChanged(Switch)`, the parameter type of `OnClick(Button button)` is Button so that it is not guaranteed the button is type of `Switch`. That was the reason why throwing exception was needed.

src/Tizen.NUI.Components/Controls/Switch.cs
rabbitfor: Since NUI opened some APIs that enabled access to sub-components to a component,
now Switch has API : Thumb, Track.
So the legacy API `GetCurrentTrack` and `GetCurrentThumb` are not needed anymore.

src/Tizen.NUI.Components/Controls/Extension/SlidingSwitchExtension.cs
dongsug-song: OK. I see. Thank you for explanation~

src/Tizen.NUI.Components/Controls/Switch.cs
dongsug-song: OK. I see. Thank you for explanation~

src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
dongsug-song: OK. I see. Thank you for explanation~

PR #1799, opened 2020-07-06
[NUI] Fix Position type converter error when Position get 2 parameters in XAML

elishateng:
### Description of Change ###

1. Fix Position type converter error when Position get 2 parameters in XAML

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/XamlBinding/PositionTypeConverter.cs
+++ b/src/Tizen.NUI/src/internal/XamlBinding/PositionTypeConverter.cs
@@ -59,6 +59,12 @@ namespace Tizen.NUI.Binding
                     int z = (int)GraphicsTypeManager.Instance.ConvertScriptToPixel(parts[2].Trim());
                     return new Position(x, y, z);
                 }
+                else if (parts.Length == 2)
+                {
+                    int x = (int)GraphicsTypeManager.Instance.ConvertScriptToPixel(parts[0].Trim());
+                    int y = (int)GraphicsTypeManager.Instance.ConvertScriptToPixel(parts[1].Trim());
+                    return new Position(x, y);
+                }
             }
 
             throw new InvalidOperationException($"Cannot convert \"{value}\" into {typeof(Position)}");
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/internal/XamlBinding/PositionTypeConverter.cs
```diff
@@ -59,6 +59,12 @@ public override object ConvertFromInvariantString(string value)
                     int z = (int)GraphicsTypeManager.Instance.ConvertScriptToPixel(parts[2].Trim());
                     return new Position(x, y, z);
                 }
+                else if (parts.Length == 2)
```
dongsug-song, line 62: Position and Size gets float typed parameter. and as I know xaml support float type. (maybe we can use Single class)
but this looks like converted and int type. 
how do you think about?

src/Tizen.NUI/src/public/XamlBinding/SizeTypeConverter.cs
```diff
@@ -28,6 +28,12 @@ public override object ConvertFromInvariantString(string value)
                     int z = (int)GraphicsTypeManager.Instance.ConvertScriptToPixel(parts[2].Trim());
                     return new Size(x, y, z);
                 }
+                else if (parts.Length == 2)
```
dongsug-song, line 31: same as above

src/Tizen.NUI/src/internal/XamlBinding/PositionTypeConverter.cs
elishateng: Hello, Mr. SSong, You're right. Thank you for your remind.

src/Tizen.NUI/src/public/XamlBinding/SizeTypeConverter.cs
elishateng: Hello, Mr. SSong, You're right. Thank you for your remind.

PR #1801, opened 2020-07-06
[NUI] Fix the issue that theme could not be changed

huayongxu:
### API Changes ###

 - ACR:


huayongxu:
1. Originally theme is to be lower when theme is set while theme is not set when RegisterStyle is called.
this causes theme could not be found when theme is changed.

2. This patch also fixes a warning [1] because of string.toLower.  String.ToUpper is used instead.

[1]  https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1308?view=vs-2019

PR #1824, opened 2020-07-14
[NUI] BackKeyEvent Proptotype

neostom432:
### Description of Change ###
View can subscribe back key event using BackKeyEvent event handler.

When user adds BackKeyEvent callback to View, view is listed as Subscriber of BackKeyManager.

After that, if back key is comming,
BackKeyManager will find top view among back key subscribers which has Visibility and IsOnWidnow property are true.

The top view will emit its BackKeyEvent.

*** Notice ***

It is based on Window KeyEvent so make sure you don't use Window KeyEvent and View KeyEvent.

After binding backkey event from DALi, user can use Window/View KeyEvent
with BackKeyEvent.

### API Changes ###
NONE

### DIFF
--- a/src/Tizen.NUI/src/internal/BackKeyManager.cs
+++ b/src/Tizen.NUI/src/internal/BackKeyManager.cs
@@ -1,0 +1,180 @@
+/*
+ * Copyright(c) 2020 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+using System;
+using System.ComponentModel;
+using System.Collections.Generic;
+using Tizen.NUI.BaseComponents;
+
+namespace Tizen.NUI
+{
+    /// <summary>
+    /// BackKeyManager is a class to manager back key event.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public sealed class BackKeyManager
+    {
+        private static readonly BackKeyManager instance = new BackKeyManager();
+
+        /// <summary>
+        /// BackKeyManager construct.
+        /// </summary>
+        private BackKeyManager()
+        {
+            NUIApplication.GetDefaultWindow().KeyEvent += OnWindowKeyEvent;
+        }
+
+        /// <summary>
+        /// Subscriber.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public List<View> Subscriber{get; set;} = new List<View>();
+
+        /// <summary>
+        /// BackKeyManager static instance.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static BackKeyManager Instance { get { return instance; } }
+
+        private Container FindParent(Container child, int step)
+        {
+            Container result = child;
+
+            for (int i = 0; i < step; i++)
+            {
+                result = result?.GetParent();
+            }
+
+            return result;
+        }
+
+        // If false, comparison1 is on top and if true, comparison2 is on top.
+        private bool CompareElevation(Container comparison1, Container comparison2)
+        {
+            bool result = false;
+
+            Container parent1 = comparison1?.GetParent();
+            Container parent2 = comparison2?.GetParent();
+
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/internal/BackKeyManager.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// Subscriber.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public List<View> Subscriber{get; set;} = new List<View>();
+
+        /// <summary>
+        /// BackKeyManager static instance.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static BackKeyManager Instance { get; } = new BackKeyManager();
```
dongsug-song, line 49: It seems it could get multiple instances when BackKeyManager.Instance is called multiple times. I am not sure so could you check the FocusManager singleton pattern? ^^;

https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/public/FocusManager.cs#L276

src/Tizen.NUI/src/internal/BackKeyManager.cs
```diff
@@ -0,0 +1,179 @@
...
+            Container parent1 = comparison1?.GetParent();
+            Container parent2 = comparison2?.GetParent();
+
+            if((parent1 == null && parent2 != null) || (parent2 == null && parent1 != null))
+            {
+                // One is in DefaultLayer but the other is not
+                if(parent2 == null)
+                {
+                    // parent1 is in DefaultLayer.
+                    Layer comparison2AsLayer = comparison2 as Layer;
```
dongsug-song, line 77: is there any possibility that comparison2AsLayer is null?

src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
```diff
@@ -462,6 +462,31 @@ private void SendViewAddedEventToWindow(IntPtr data)
             }
         }
 
+        private EventHandler _backKeyPressed;
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public event EventHandler BackKeyPressed
```
dongsug-song, line 468: I think it would be better to add some comments even though it is hidden API. ^^;

src/Tizen.NUI/src/internal/BackKeyManager.cs
```diff
@@ -0,0 +1,179 @@
+/*
+ * Copyright(c) 2019 Samsung Electronics Co., Ltd.
```
dongsug-song, line 2: 2019 to 2020.
and could you add some simple sample code or sample app to figure out how this can be used? ^^;

src/Tizen.NUI/src/internal/BackKeyManager.cs
neostom432: In this case, we have problem that relation between DefaultLayer and other Layer.
They looks like siblings but in fact, they are parent-child.

Because of this situation, when using new Layer, the depth of view can be differ like below.
DefaultLayer - View1 - View2 -> View2.HierarchyDepth is 2
Other Layer - View1 - View2 -> View2.HierarchyDepth is 3.

So, this statement just for one of comparison is in other Layer and the comparison2 is always Layer never null.

src/Tizen.NUI/src/internal/BackKeyManager.cs
neostom432: I used Lazy<T> pattern but it works when using .NET4 or higher.

https://csharpindepth.com/articles/singleton

Do you think we can use this?

src/Tizen.NUI/src/internal/BackKeyManager.cs
neostom432: DONE

src/Tizen.NUI/src/internal/BackKeyManager.cs
neostom432: DONE

dongsug-song: great! (b^_^)b
sorry but I couldn't see Lazy<T> and I think FocusManager uses "forth version" of the article above.
if you use Instance by new BackKeyManager, null checking need be added, I think.
in my private opinion, following FocusManger pattern would be good.
How do you think about?

src/Tizen.NUI/src/internal/BackKeyManager.cs
dongsug-song: OK, I see. 
I didn't know that Layer's depth issue... I think we need to fix that in nui side anyway...

src/Tizen.NUI/src/internal/BackKeyManager.cs
neostom432: Follow FocusManager.
Please check changes.

src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
neostom432: DONE

PR #1811, opened 2020-07-08
[NUI] Add FrameRenderedCallback, FramePresentedCallback as hidden API

dongsug-song:
### Description of Change ###
Add FrameRenderedCallback, FramePresentedCallback as hidden API
- csharp binding for DALi DevelWindow's AddFrameRenderedCallback and AddFramePresentedCallback
- TestCase will not be added in tizen CSharp-TCT, but they were added in test sample to do TDD
- paired with https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/238104/

### API Changes ###
none

PR #1839, opened 2020-07-20
[NUI] Change for AddFrameRenderedCallback to maintain the reference

dongsug-song:
### Description of Change ###
Change for AddFrameRenderedCallback to maintain the reference of user callback
- Fix for the already merged patch : https://github.com/Samsung/TizenFX/pull/1811
- http://wiki.vd.sec.samsung.net/pages/viewpage.action?pageId=4213716
- Using Dictionary to maintain native callback reference so that it is not to be garbage collected
- Small fix added : CaptureTest code's build error is fixed

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/public/Window.cs
+++ b/src/Tizen.NUI/src/public/Window.cs
@@ -1311,6 +1311,60 @@ namespace Tizen.NUI
             Interop.Window.delete_Window(swigCPtr);
         }
 
+        private static Dictionary<int, internalHookCallbackType> frameCallbackList = new Dictionary<int, internalHookCallbackType>();
+
+        private static readonly object locker = new object();
+
+        private static int key = 0;
+
+        private static FrameCallbackType internalHookFrameCallback = OnInternalHookFrameCallback;
+
+        private struct internalHookCallbackType
+        {
+            public FrameCallbackType userCallback;
+            public int frameId;
+        }
+
+        private static void OnInternalHookFrameCallback(int id)
+        {
+            lock (locker)
+            {
+                if (frameCallbackList.ContainsKey(id))
+                {
+                    if (frameCallbackList[id].userCallback != null)
+                    {
+                        frameCallbackList[id].userCallback.Invoke(frameCallbackList[id].frameId);
+                        frameCallbackList.Remove(id);
+                    }
+                    else
+                    {
+                        NUILog.Error($"found userCallback is NULL");
+                        frameCallbackList.Remove(id);
+                    }
+                }
+            }
+        }
+
+        private int AddInterHookCallback(FrameCallbackType callback, int frameId)
+        {
+            if (null == callback)
+            {
+                throw new ArgumentNullException(nameof(callback), "FrameCallbackType should not be null");
+            }
+            var assignedKey = 0;
+            lock (locker)
+            {
+                key++;
+                assignedKey = key;
+                frameCallbackList.Add(assignedKey, new internalHookCallbackType()
+                {
+                    userCallback = callback,
+                    frameId = frameId,
+                });
+            }
+            return assignedKey;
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/Window.cs
```diff
@@ -1311,6 +1311,58 @@ protected override void ReleaseSwigCPtr(System.Runtime.InteropServices.HandleRef
...
+
+        private static FrameCallbackType internalHookFrameCallback = onInternalHookFrameCallback;
+
+        private struct internalHookCallbackType
+        {
+            public FrameCallbackType userCallback;
+            public int frameId;
+        }
+
+        private static void onInternalHookFrameCallback(int id)
```
rabbitfor, line 1328: I think it is better to start with an upper-case letter.

woohyun: I agree with @rabbitfor, and I hope other private methods would also be started with upper-case.

src/Tizen.NUI/src/public/Window.cs
```diff
@@ -1364,14 +1410,8 @@ public void AddFrameRenderedCallback(FrameCallbackType callback, int frameId)
...
-            IntPtr ip = Marshal.GetFunctionPointerForDelegate<Delegate>(callback);
-
-            if (IntPtr.Zero == ip)
-            {
-                throw new ArgumentNullException(nameof(callback), "FrameCallbackType should not be null");
-            }
-
-            Interop.WindowInternal.AddFramePresentedCallback(swigCPtr, new HandleRef(this, ip), frameId);
+            addInterHookCallback(callback, frameId);
+            Interop.WindowInternal.AddFramePresentedCallback(swigCPtr, new HandleRef(this, Marshal.GetFunctionPointerForDelegate<Delegate>(internalHookFrameCallback)), key);
```
woohyun, line 1414: Previously, "frameId" was passed as the last parameter. It's ok to change it to "key" ?
I don't know well about the usage of this parameter, but "key" seems to be existed as a counter of addInterHookCallback.

src/Tizen.NUI/src/public/Window.cs
dongsug-song: Thank you for your reviews. 
I changed only methods as pascal case. 
how do you think the following article about c# coding convention?
https://www.c-sharpcorner.com/UploadFile/8a67c0/C-Sharp-coding-standards-and-naming-conventions/
shall we make this as our internal rule? or is there any rule already proposed internally? ^^;
I think we need to make some our ground rule for NUI coding convention soon.

src/Tizen.NUI/src/public/Window.cs
dongsug-song: could you check the native sample code : https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-demo/+/237623/2/examples/hello-world/hello-world-example.cpp

the frameId in AddFrameRenderedCallback is the value that user induce and it is passed and shown in user defined callback (FrameCallbackType). 

I changed it a little bit and in AddInterHookCallback() user's frameId is saved in internalHookCallbackType instance. The assignedKey is increased and is unique value so it is assigned in AddFrameRenderedCallback() to find saved user callback and frameId in the Dictionary of frameCallbackList.

It could be overhead but it is required to avoid null reference exception occured by garbage collected/unreachable delegate problem.

woohyun: Thanks for explanation. I can understand clearly :)

src/Tizen.NUI/src/public/Window.cs
woohyun: I also hope to set up NUI coding convention asap. Anyway thanks for modification.

PR #1842, opened 2020-07-20
[NUI] remove unused class

dasg34:
since commit 4389bdc8eb5a4ad3852956870d0dfe119d8e04ef, those classes and
methods are no longer used.

### API Changes ###

 - ACR:


dongsug-song:
@elishateng 
Hello, 
this patch would remove xaml related API.
please check if this would be OK.
thank you.

PR #1845, opened 2020-07-21
[NUI] Add ScrollPosition and ScrollCurrentPosition to ScrollableBase

jaehyun0cho:
To get the scroll position given to ScrollTo, ScrollPosition is added.
ScrollPosition is the position in the opposite direction to the position
of ContentContainer.

To get the current scroll position, ScrollCurrentPosition is added.
ScrollCurrentPosition is the position in the opposite direction to the
current position of ContentContainer.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
+++ b/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
@@ -1073,6 +1073,31 @@ namespace Tizen.NUI.Components
             return position;
         }
 
+        /// <summary>
+        /// Scroll position given to ScrollTo.
+        /// This is the position in the opposite direction to the position of ContentContainer.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        public Position ScrollPosition
+        {
+            get
+            {
+                return new Position(-ContentContainer.Position);
+            }
+        }
+
+        /// <summary>
+        /// Current scroll position in the middle of ScrollTo animation.
+        /// This is the position in the opposite direction to the current position of ContentContainer.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        public Position ScrollCurrentPosition
+        {
+            get
+            {
+                return new Position(-ContentContainer.CurrentPosition);
+            }
+        }
     }
 
 } // namespace


jaehyun0cho:
ACR request link is https://code.sec.samsung.net/jira/browse/TCSACR-354

src/Tizen.NUI.Components/Controls/ScrollableBase.cs
```diff
@@ -1073,6 +1073,20 @@ protected virtual float AdjustTargetPositionOfScrollAnimation(float position)
             return position;
         }
 
+        /// <summary>
+        /// Current scroll position given to ScrollTo.
```
neostom432, line 1077: CurrentPosition is not exactly same as "position given to ScrollTo".
It is just current position so when user use SrollTo Api to 4000 and get CurrentPosition at the middle of animation, the value can be "current position", not "position given to ScrollTo".

If you really want to return "position given to ScrollTo", I think Position property is much suitable because it changed to destination value when starting ScrollTo Animation.

jaehyun0cho: I understand your point.

How about supporting ScrollPosition and ScrollCurrentPosition? (same as ContentContainer.Position and ContentContainer.CurrentPosition)
- ScrollableBase.ScrollPosition = -ScrollableBase.ContentContainer.Position
- ScrollableBase.ScrollCurrentPosition = -ScrollableBase.ContentContainer.CurrentPosition

What do you think about the above?

neostom432: I think user always wants to get CurrentPosition instead of Position.

So, just update comment will be fine.
ScrollPosition = -ScrollableBase.ContentContainer.CurrentPosition.

What do you think?

jaehyun0cho: I am not sure if it is sufficient.
Because View class already supports "CurrentPosition" and "Position" to support position information separately. (i.e. visible position and property position)
So I think it would be better that ScrollPosition should be supported in the same way. Otherwise, user cannot get the property position for scrolling.

neostom432: As you mentioned before, please add two APIs

- ScrollableBase.ScrollPosition = -ScrollableBase.ContentContainer.Position
- ScrollableBase.ScrollCurrentPosition = -ScrollableBase.ContentContainer.CurrentPosition

jaehyun0cho: I have updated the patch.
However, for ScrollPosition, I could not assign -ContentContainer.Position because it was not the same with the parameter given to ScrollTo(). (It seems there is a bug to be fixed about ContentContainer.Position)
So I used private variable scrollToPosition to track the parameter of ScrollTo.

jaehyun0cho:
> There are other functions which play ScrollAnimation.
> If you cache ScrollPosition only here, developer can be confused.
> For example,
> 
> 1. Use ScrollTo
> 2. Scroll using finger
> 
> In this case, ScrollPosition is target position of first step.
> 
> Please make clear that you only want to change ScrollPosition property when uses ScrollTo API

Thank you for your comment. I understand your point.
Then, I will modify ScrollPosition as you requested before as follows.
- ScrollableBase.ScrollPosition = -ScrollableBase.ContentContainer.Position

neostom432:
> LGTM

Patch is good but you need to ACR these APIs if you want to make them public.
If you don't, please add [EditorBrowsable(EditorBrowsableState.Never)]

jaehyun0cho:
> Patch is good but you need to ACR these APIs if you want to make them public.
> If you don't, please add [EditorBrowsable(EditorBrowsableState.Never)]

Thank you! The ACR process is ongoing as follows https://code.sec.samsung.net/jira/browse/TCSACR-354

PR #1851, opened 2020-07-21
[NUI] Enable selectors work in View : EnableControlState

rabbitfor:
* Previously, the View did not support ControlState so that selector values are not applied to the View.
* Introduce a new property EnableControlState.
  If this property is set to true, the View can have a touch related ControlState (such as Pressed) when touch.
  By default, it is false in View, true in Control.
  Note that if the value is true, the View will be a touch receptor.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
@@ -245,6 +245,37 @@ namespace Tizen.NUI.BaseComponents
         /// <returns>The <see cref="ControlState"/> containing the result of the addition.</returns>
         [EditorBrowsable(EditorBrowsableState.Never)]
         public static ControlState operator +(ControlState lhs, ControlState rhs) => Create(lhs, rhs);
+
+        /// <summary>
+        /// The substraction operator.
+        /// </summary>
+        /// <param name="lhs">A <see cref="ControlState"/> on the left hand side.</param>
+        /// <param name="rhs">A <see cref="ControlState"/> on the right hand side.</param>
+        /// <returns>The <see cref="ControlState"/> containing the result of the substraction.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static ControlState operator -(ControlState lhs, ControlState rhs)
+        {
+            if (!lhs.IsCombined)
+            {
+                return ReferenceEquals(lhs, rhs) ? Normal : lhs;
+            }
+            
+            var rest = lhs.stateList.Except(rhs.stateList);
+
+            if (rest.Count() == 0)
+            {
+                return Normal;
+            }
+
+            if (rest.Count() == 1)
+            {
+                return rest.First();
+            }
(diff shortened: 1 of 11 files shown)


src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -245,6 +245,37 @@ public override string ToString()
...
         public static ControlState operator +(ControlState lhs, ControlState rhs) => Create(lhs, rhs);
+
+        /// <summary>
+        /// The substraction operator.
+        /// </summary>
+        /// <param name="lhs">A <see cref="ControlState"/> on the left hand side.</param>
+        /// <param name="rhs">A <see cref="ControlState"/> on the right hand side.</param>
+        /// <returns>The <see cref="ControlState"/> containing the result of the substraction.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static ControlState operator -(ControlState lhs, ControlState rhs)
```
dongsug-song, line 256: Hello, could you explain why this - operator overriding is added and what this is for?

dasg34: I think this is useful to remove specific ControlState from combined ControlState. for example,
Pressed State:
```cs
ControlState += ControlState.Pressed;
```
UnPressed State:
```cs
ControlState -= ControlState.Pressed;
```
Also, it is used at line 1125 in ViewInternal.cs

dongsug-song: Thank you for nice explanation~

rabbitfor:
> I have tested with following sample code.
> 
> ```cs
> protected override void OnCreate()
> {
>     base.OnCreate();
>     Window.Instance.BackgroundColor = new Color(227 / 255f, 255 / 255f, 227 / 255f, 1.0f);
>     Window.Instance.KeyEvent += OnKeyEvent;
>             
>     View root = new View();
>     Window.Instance.GetDefaultLayer().Add(root);
> 
>     ViewStyle viewStyle = new ViewStyle()
>     {
>         BackgroundColor = new Selector<Color>(){
>             { ControlState.Pressed, new Color(227 / 255f, 227 / 255f, 255 / 255f, 1.0f) },
>             { ControlState.Normal, new Color(255 / 255f, 227 / 255f, 227 / 255f, 1.0f) },
>         }
>                 
>     };
> 
>     View view = new View(viewStyle)
>     {
>         WidthSpecification = 100,
>         HeightSpecification = 100,
>         EnableControlState = true,
>     };
>     root.Add(view);
> }
> ```
> 
> It works well..

Thank you :)

rabbitfor:
@AdunFang Hi Mr Fang. The patch is merged but can you check if this breaks NUI.DA build

AdunFang:
@rabbitfor 
Hi, Mr. Yang
We have verified this patch, the FHUB.NUI is OK.

PR #1857, opened 2020-07-24
[NUI] Fix View.BackgroundImage getter and View.ImageSyncronousLoading…

rabbitfor:
… to support all possible visuals.

Previous code, it didn't work for other visuals except ImageVisual.

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -2205,15 +2205,11 @@ public bool BackgroundImageSynchronosLoading
...
-                int visualType = 0;
-                Background.Find(Visual.Property.Type)?.Get(out visualType);
-                if (visualType == (int)Visual.Type.Image)
-                {
-                    Background.Find(ImageVisualProperty.URL)?.Get(out bgUrl);
-                }
 
-                if (bgUrl.Length != 0)
+                string bgUrl = null;
+                Background.Find(ImageVisualProperty.URL)?.Get(out bgUrl);
```
dongsug-song, line 2210: is this the idea that if the Visual.Property.Type is Image then it only has URL?
I think it is quite reasonable but I want to be very careful, because this change could give an effect to existing public API's behavior.

rabbitfor: URL can be exist in Visual.Property.Type `Image`, `SVG` and `NPatch`.
Previous code only support `Image` so I fixed it to support all types that includes URL property.

PR #1863, opened 2020-07-27
[NUI] Add IndicatorColor and SelectedIndicatorColor to Pagination

jaehyun0cho:
To support setting indicator color of Pagination, IndicatorColor and
SelectedIndicatorColor properties are added.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Pagination.cs
+++ b/src/Tizen.NUI.Components/Controls/Pagination.cs
@@ -35,6 +35,9 @@ namespace Tizen.NUI.Components
         private int selectedIndex = -1;
         private PaginationStyle paginationStyle => ViewStyle as PaginationStyle;
 
+        private Color indicatorColor;
+        private Color selectedIndicatorColor;
+
         static Pagination() { }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


jaehyun0cho:
ACR link is https://code.sec.samsung.net/jira/browse/TCSACR-355

src/Tizen.NUI.Components/Controls/Pagination.cs
```diff
@@ -185,6 +188,108 @@ public int IndicatorCount
...
+        private void OnIndicatorColorChanged(float r, float g, float b, float a)
+        {
+            IndicatorColor = new Color(r, g, b, a);
+        }
+
+        /// <summary>
+        /// Color of the indicator.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        public Color IndicatorColor
```
idkiller, line 200: This property has a problem that set and get objects are different.
for Example.
```C#
var color = new Color("#ffffff");
pagination.IndicatorColor = color;
var color2 = pagination.IndicatorColor;

color.B = 0x00; // it does not apply to pagination.IndicatorColor
color2.B = 0x00; // it is applies to pagination.IndicatorColor

var isSame = color == color2; // it is true. because Color equal does not compare that has same callback.
```

In general, If property is same type, it is assumed that the get will be the same as the set one.
However, as you can see in the example above, this object actually different and behavior is different also because that does not has callback.

I think these properties must be value types like struct. (But since the legacy Color is already a class... I think it would be difficult without major changes.)
If not, I think it should be made in the form of a method rather than a property... (but I don't come up with a way to make it look good...)

src/Tizen.NUI.Components/Controls/Pagination.cs
```diff
@@ -185,6 +188,108 @@ public int IndicatorCount
...
+        private void OnSelectedIndicatorColorChanged(float r, float g, float b, float a)
+        {
+            SelectedIndicatorColor = new Color(r, g, b, a);
+        }
+
+        /// <summary>
+        /// Color of the selected indicator.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        public Color SelectedIndicatorColor
```
idkiller, line 255: yes... this property has the same problem...

jaehyun0cho: Thank you for your comment. It helps a lot!
By discussing with NUI members, the Setter of Color class properties is going to be deprecated.
i.e. R, G, B, A can be assigned in Color constructor.

PR #1869, opened 2020-07-29
[NUI] Support default focus logic for RecyclerView

neostom432:
### Description of Change ###
Support default focus chain.

Focus logic can vary depending on the type of LayoutManager.

### API Changes ###
NONE


huiyueun:
LGTM

PR #1892, opened 2020-08-05
[NUI] Add MinLineSize in RendererParameters

JoogabYun:
### API Changes ###

 - ACR: NONE


src/Tizen.NUI/src/internal/Interop/Interop.TextUtils.cs
```diff
@@ -126,6 +126,12 @@ internal static partial class RendererParameters
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_RendererParameters_isTextColorSet_get")]
             public static extern bool RendererParameters_isTextColorSet_get(global::System.Runtime.InteropServices.HandleRef jarg1);
 
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_RendererParameters_minLineSize_set")]
```
dongsug-song, line 129: Are these native functions merged in csharp-binder? otherwise, this could cause exception when app launching.

JoogabYun: csharp-binder is merged.

PR #1897, opened 2020-08-05
[MediaVision] Add new barcode types

tyoungroy:
Signed-off-by: Tae-Young Chung <ty83.chung@samsung.com>

### Description of Change ###
Add new barcode types (Native ACR:  https://code.sec.samsung.net/jira/browse/ACR-1564)

### API Changes ###

 - ACR: [TCSACR-361](https://code.sec.samsung.net/jira/browse/TCSACR-361)

### DIFF
--- a/src/Tizen.Multimedia.Vision/MediaVision/BarcodeType.cs
+++ b/src/Tizen.Multimedia.Vision/MediaVision/BarcodeType.cs
@@ -66,6 +66,36 @@ namespace Tizen.Multimedia.Vision
         /// 1D barcode - Interleaved Two of Five.
         /// </summary>
         /// <since_tizen> 3 </since_tizen>
-        I25
+        I25,
+        /// <summary>
+        /// 1D barcode - International Article Number with 2-digit.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        Ean2 = 9,
+        /// <summary>
+        /// 1D barcode - International Article Number with 5-digit.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        Ean5,
+        /// <summary>
+        /// 1D barcode - Code 93.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


src/Tizen.Multimedia.Vision/MediaVision/BarcodeType.cs
```diff
@@ -66,6 +66,36 @@ public enum BarcodeType
         /// 1D barcode - Interleaved Two of Five.
         /// </summary>
         /// <since_tizen> 3 </since_tizen>
-        I25
+        I25,
+        /// <summary>
+        /// 1D barcode - International Article Number with 2-digit.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        Ean2,
```
hsgwon, line 74: Why don't you define simply this as `IEan2 =  9`?
If so, you don't need to convert at all.

tyoungroy: In native code,  the enum MV_BARCODE_UNDEFINE (value is 8) is deprecated and will be removed. 
If i define Ean2 = 9, i have to change the enum value after deprecation.
but as i know, the change of enum value is prohibited.

hsgwon: Will you change the value of native enum `MV_BARCODE_EAN_2`(9) to 8, after removing `MV_BARCODE_UNDEFINED`?
I think it's not possible, because it's ABI break.
If you don't have plan to change the value of `MV_BARCODE_EAN_2`, I think it's better to define `Ean2` as 9 in C# side.

tyoungroy: I fixed it. I totally misunderstood the "Deprecation and remove after 2 version release". 
Thank you for comments!

PR #1898, opened 2020-08-06
[RPCPort][ACR] Support synchronous connection requests and private file sharing

hjhun:
### Description of Change ###

Support synchronous connection requests and private file sharing
To support synchronous connection requests, ConnectSync() is added.
To support private file sharing, ShareFile() and UnshareFile() are added.
When the application calls ShareFile() method, platform will grant a temporary permission to the target application for files and revoke it when the target application is terminated or UnshareFile() method is called.
The target application doesn't have read permission of the directory that is obtained by caller's data path(Tizen.Applications.DirectoryInfo.Data). The developer should open the file path with read only mode directly. Because, platform grants read permission to the passed file path.

### API Changes ###

 - ACR: TCSACR-362

<!--
Added:
- void ConnectSync(string appId, string port);
- void ShareFile(string[] paths);
- void ShareFile(string path);
- void UnshareFile();

### DIFF
--- a/src/Tizen.Applications.Common/Tizen.Applications.RPCPort/Port.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications.RPCPort/Port.cs
@@ -57,7 +57,7 @@ namespace Tizen.Applications.RPCPort
         /// <param name="paths">The file paths to be shared.</param>
         /// <exception cref="InvalidIOException">Thrown when an internal IO error occurrs.</exception>
         /// <since_tizen> 8 </since_tizen>
-        public void ShareFile(IEnumerable<string > paths)
+        public void ShareFile(IEnumerable<string> paths)
         {
             if (paths == null)
                 throw new InvalidIOException();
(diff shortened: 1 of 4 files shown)


src/Tizen.Applications.Common/Tizen.Applications.RPCPort/Port.cs
```diff
@@ -42,5 +42,56 @@ public enum Type
...
+        /// <remarks>
+        /// If all added paths are under the caller application's data path which can be obtained, those will be shared to the target application.
+        /// Platform will grant a temporary permission to the target application for those files and revoke it when the target application is terminated or UnshareFile() is called.
+        /// Paths should be regular files. The target application can just read them.
+        /// Note that the target application doesn't have read permission of the directory that is obtained by caller's data path. You should open the file path with read only mode directly.
+        /// </remarks>
+        /// <param name="paths">The file paths to be shared.</param>
+        /// <exception cref="InvalidIOException">Thrown when an internal IO error occurrs.</exception>
+        /// <since_tizen> 8 </since_tizen>
+        public void ShareFile(string[] paths)
```
myroot, line 58: ✔️ DO prefer using collections over arrays in public APIs. The Collections section provides details about how to choose between collections and arrays.
```suggestion
        public void ShareFile(IEnumerable<string> paths)
```

PR #1917, opened 2020-08-13
[NUI] Provide binding after HoverEvent class pimpling

hrdisaac:
### Description of Change ###
[NUI] Provide binding after HoverEvent class pimpling

### API Changes ###
 - ACR: N/A

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.Hover.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.Hover.cs
@@ -12,28 +12,37 @@ namespace Tizen.NUI
             public static extern global::System.IntPtr new_Hover__SWIG_0();
 
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_new_Hover__SWIG_1")]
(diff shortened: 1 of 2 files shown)


dongsug-song:
@hrdisaac 
Hello,
parent class has been changed => this requires ACR. 
basically, this patch can not be merged until the ACR is accepted.
please check again.
thank you.

dongsug-song:
@hrdisaac 
Hello, Mr. Richard
It will take long time to complete the ACR.
so I recommend for you to add new named class such as Hover2, HoverAlternative... (just an example)
and you can add Hover2 as hidden API by attaching [EditorBrowsable(EditorBrowsableState.Never)]. 
At the same time, you need proceed deprecate-ACR for Hover and add-ACR for Hover2.
it would take times but seems no other way existed. (it is process we have followed so far)
please check again.
thank you.

hrdisaac:
I updated this patch so it doesn't change the base class to Basehandle, therefore no ACR is required, which will allow this patch to be merged with the latest DALi native code.

I will push another patch to change the base class, and I will update the original ACR to link with that new pull request later on.

AdunFang (approved):
Looks good.

adeelkazmi (approved):

PR #1918, opened 2020-08-14
[NUI] restore TypeConverter support for Properties

dasg34:
Since commit 97c4baea0ef21936602db814fd2a9177ddf6e673, TypeConverter of
Properties has not worked. this patch will fix this issue.

Test Code:
```cs
public class TestClass
{
    [TypeConverter(typeof(MyTypeConverter))]
    public int TestProperty { get; set; }

	public class MyTypeConverter : TypeConverter, IExtendedTypeConverter
	{
		object IExtendedTypeConverter.ConvertFromInvariantString(string value, IServiceProvider serviceProvider)
		{
			...
			...
			return value;
		}
		...
	}
}
```

### API Changes ###

 - ACR:


AdunFang (approved):
Looks good.

PR #1925, opened 2020-08-19
[NUI] Sample app for Frame callback.

bshsqa:
- Touch the screen then launcher icons are appeared
 - The icons can be moved by using finger gesture.
 - If a icon is moved by gesture, others follow it with small delay(0.03s)
 - the delayed animation is computed in the frame callback


dongsug-song (approved):
looks good to me

PR #1926, opened 2020-08-19
[NUI] Restore Size2D's minus value as zero when Layout is null

dongsug-song:
### Description of Change ###
[NUI] Restore Size2D's minus value as zero when Layout is null
- to back out this patch : https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-core/+/226385/  
- Size2D is also used to set Layout size specification so it can be minus value.
- TV FLUX discards minus value of Size2D.Width or Size2D.Height so Size2D's values need be restored as zero when NUI.Layout is NULL
- TV FLUX also has it's own Layout, so NUI needs to maintain its behaviour.

### API Changes ###
none


Seoyeon2Kim (approved):
If there's no Layout in a View, then the value of Size should be over 0.
The patch solves the issue above. Looks good to me.

PR #1927, opened 2020-08-19
[NUI] Provide binding after WheelEvent class pimpling

hrdisaac:
### Description of Change ###
[NUI] Provide binding after WheelEvent class pimpling

### API Changes ###
 - ACR: N/A


adeelkazmi (approved):
Looks good but I guess we have to wait till the devel/master branch changes have been ported over to the tizen branch before we submit this.

PR #1928, opened 2020-08-20
[NUI] Fix NUI Layout issue

huiyueun:
The LayoutController is not created when created window at the PreLoad time,
because the Stage does not exist yet.
This operation causes a problem that the layout does not work.
https://github.com/Samsung/TizenFX/blob/9094590d0debe5bd4d9806eee79849132efbaf6a/src/Tizen.NUI/src/public/Window.cs#L51-L57

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/NUIApplication.cs
+++ b/src/Tizen.NUI/src/public/NUIApplication.cs
@@ -383,7 +383,6 @@ namespace Tizen.NUI
         {
             Interop.Application.Application_PreInitialize();
             Application.NewApplication("", Application.WindowMode.Opaque);
-            NUIApplication.GetDefaultWindow();
         }
     }
 


dongsug-song:
@TizenAPI-Bot rerun all

src/Tizen.NUI/src/public/NUIApplication.cs
```diff
@@ -383,7 +383,6 @@ static public void PreLoad()
         {
             Interop.Application.Application_PreInitialize();
             Application.NewApplication("", Application.WindowMode.Opaque);
-            NUIApplication.GetDefaultWindow();
```
dongsug-song, line 386: only this code removed.... ^^;
is this OK? doesn't this cause some behavior change?
I couldn't understand this patch, so I am asking. ^^;

dongsug-song (commented):
could you please ask to rerun auto checker?

src/Tizen.NUI/src/public/NUIApplication.cs
neostom432: If Window is created in NUI, it will try to create LayoutController instance.
LayoutController makes Layout work so it should be created.

But in Preload time, LayoutController cannot be created because stage is not installed as huiyu mentioned in commit message.

So, if we remove this window creation attempt here, there will be no issue.

neostom432 (changes requested):
Please fix build error.

src/Tizen.NUI/src/public/NUIApplication.cs
dongsug-song: I got it. Thank you for explanation~

AdunFang (approved):
It is OK.

neostom432 (approved):
LGTM

PR #1929, opened 2020-08-21
[NUI] Add color constructor

neostom432:
### Description of Change ###
Add color constructor which is initialized by System.Drawing.Color.

When we test something, we always feel our Color class has not enough
color list.

In System.Drawing.Color, there is a ton of predefined colors.
So, added new constructor which can initialized by System.Drawing.Color.

### API Changes ###
NONE

### DIFF
--- a/src/Tizen.NUI/src/public/Color.cs
+++ b/src/Tizen.NUI/src/public/Color.cs
@@ -142,6 +142,19 @@ namespace Tizen.NUI
         }
 
         /// <summary>
+        /// The conversion constructor from an System.Drawing.Color instance.
+        /// </summary>
+        /// <param name="color">System.Drawing.Color instance</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Color(global::System.Drawing.Color color) : this(Interop.Vector4.new_Vector4__SWIG_0(), true)
+        {
+            R = color.R / 255.0f;
+            G = color.G / 255.0f;
+            B = color.B / 255.0f;
+            A = color.A / 255.0f;
+        }
+
+        /// <summary>
         /// The copy constructor.
         /// </summary>
         /// <param name="other">The copy target.</param>


src/Tizen.NUI/src/public/Color.cs
```diff
@@ -141,6 +141,28 @@ public Color(string hexColor) : this(Interop.Vector4.new_Vector4__SWIG_0(), true
             }
         }
 
+        /// <summary>
+        /// The conversion constructor from an System.Drawing.Color instance.
+        /// </summary>
+        /// <param name="color">System.Drawing.Color instance</param>
+        /// <since_tizen> 6 </since_tizen>
```
dongsug-song, line 148: please remove <since_tizen> tag if this is hidden api.

src/Tizen.NUI/src/public/Color.cs
```diff
@@ -141,6 +141,28 @@ public Color(string hexColor) : this(Interop.Vector4.new_Vector4__SWIG_0(), true
...
+        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
```
dongsug-song, line 149: please change "tizen_6.0" as "tizen_next" if this API's open schedule is not confirmed.

src/Tizen.NUI/src/public/Color.cs
```diff
@@ -141,6 +141,28 @@ public Color(string hexColor) : this(Interop.Vector4.new_Vector4__SWIG_0(), true
...
+            try
+            {
+                R = color.R / 255.0f;
+                G = color.G / 255.0f;
+                B = color.B / 255.0f;
+                A = color.A / 255.0f;
+            }
+            catch
+            {
+                throw new global::System.ArgumentException("Please check your System.Drawing.Color instance");
```
dongsug-song, line 162: could you check the following link?
https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1065?view=vs-2019
even though this is not static constructor, if it throws exception in constructor, user could get some big confusion, so is there any other option to handle error case? (for example, giving a empty object or null ...)
I am not sure about this, so I am asking for your idea... ^^;

src/Tizen.NUI/src/public/Color.cs
```diff
@@ -141,6 +141,28 @@ public Color(string hexColor) : this(Interop.Vector4.new_Vector4__SWIG_0(), true
             }
         }
 
+        /// <summary>
+        /// The conversion constructor from an System.Drawing.Color instance.
+        /// </summary>
+        /// <param name="color">System.Drawing.Color instance</param>
+        /// <since_tizen> 6 </since_tizen>
+        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
dongsug-song, line 150: if the api has exception, the comment needs show the description of the exception. 
example) https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/public/Window.cs#L69
could you add some exception comment?

src/Tizen.NUI/src/public/Color.cs
neostom432: DONE

src/Tizen.NUI/src/public/Color.cs
neostom432: DONE

src/Tizen.NUI/src/public/Color.cs
neostom432: Remove exception because if color instance has problem, crash should be happened before.

dongsug-song (approved):
This is very good idea to use .NET's Color class...
How about adding "implicit conversion" for NUI.Color class and Drawing.Color class?
(for example, currently in NUI, Size and Size2D are converting each other implicitly.)
I will figure it out if this is possible...

PR #1933, opened 2020-08-21
[Internal] Add Tizen.Inspection internal API

kkuchciak:
### Description of Change ###
This commit adds new internal Inspection C# API.


src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,270 @@
...
+
+using System;
+using NativeApi = Interop.Diagnostics;
+
+namespace Tizen.Diagnostics
+{
+    /// <summary>
+    /// Enumeration for the context types.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
```
WonyoungChoi, line 25: It should be API Level 8. Please update all since_tizen to 8.
```suggestion
    /// <since_tizen> 8 </since_tizen>
```

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,270 @@
...
+    public enum CtxType
```
WonyoungChoi, line 26:
```suggestion
    public enum ContextType
```

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,270 @@
...
+    /// The Diagnostics class is used for requesting client data and receiving diagnostics report events.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class Diagnostics
+    {
+        /// <summary>
+        /// Event to be invoked when new report context arrives.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public event EventHandler<ReportArrivedEventArgs> ReportArrived;
```
WonyoungChoi, line 63: How about `Received` instead of `Arrived`?
```suggestion
        public event EventHandler<ReportArrivedEventArgs> ReportReceived;
```

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,270 @@
...
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        CrashInfoJson,
+    }
+
+    /// <summary>
+    /// The Diagnostics class is used for requesting client data and receiving diagnostics report events.
+    /// </summary>
+    /// <since_tizen> 6 </since_tizen>
+    public class Diagnostics
```
WonyoungChoi, line 57: Implement IDisposable to unset the notification callback.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        /// <summary>
+        /// Requests other apps to dump diagnostics data.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <feature>
+        /// http://tizen.org/privilege/internal/default/platform
+        /// </feature>
+        public DiagnosticsData RequestClientData(string clientId, string[] parameters)
```
WonyoungChoi, line 81: Could you provide more user-friendly API? User does not have any information about parameters to use this method.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,270 @@
...
+                throw ExceptionFactory.CreateException(ret);
+
+            return new DiagnosticsData(data);
+        }
+
+        /// <summary>
+        /// Stores context type.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        public CtxType Type;
```
WonyoungChoi, line 160: Make as a property instead of `public field`.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,270 @@
...
+    {
+        private IntPtr _handle = IntPtr.Zero;
+        private bool _disposed = false;
+
+        /// <summary>
+        /// The DiagnosticsData class constructor.
+        /// </summary>
+        /// <since_tizen> 6 </since_tizen>
+        /// <param name="handle">Diagnostics data handle.</param>
+        public DiagnosticsData(IntPtr handle)
```
WonyoungChoi, line 204: Do not open API using IntPtr. If you need IntPtr in the public API, consider using SafeHandle instead.
```suggestion
        internal DiagnosticsData(IntPtr handle)
```

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,270 @@
...
+* distributed under the License is distributed on an AS IS BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using NativeApi = Interop.Diagnostics;
+
+namespace Tizen.Diagnostics
```
WonyoungChoi, line 20: How about use `Tizen.System` namespace.
```suggestion
namespace Tizen.System
```

WonyoungChoi (changes requested):

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
kkuchciak: The problem is that the parameters are just passed as an array to the client and it depends on the client how it interprets it. It's just like calling an application with some argv. The Diagnostics API does not know anything about the parameters.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        /// <summary>
+        /// Requests other apps to dump diagnostics data.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        /// <feature>
+        /// http://tizen.org/privilege/internal/default/platform
+        /// </feature>
+        public DiagnosticsData RequestClientData(string clientId, string[] parameters)
```
m-macias, line 91: Parameters of methods must have valid documentation. Method has 2 parameters, 0 are documented. 
Method which returns non-void value must have exactly one `returns` section.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,303 @@
...
+
+        /// <summary>
+        /// Gets report of specific type from the context.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        /// <param name="report">Report type to retrieve from the context.</param>
+        /// <feature>
+        /// http://tizen.org/privilege/internal/default/platform
+        /// </feature>
+        public DiagnosticsData GetReport(ReportType report)
```
m-macias, line 180: Method which returns non-void value must have exactly one `returns` section.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,303 @@
...
+            Dispose(false);
+        }
+
+        /// <summary>
+        /// Reads given number of bytes from diagnostics data.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        /// <param name="size">Number of bytes to read.</param>
+        /// <param name="timeout">Timeout [ms] for reading requested number of bytes.</param>
+        public unsafe byte[] Read(int size, int timeout)
```
m-macias, line 256: Method which returns non-void value must have exactly one `returns` section.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,303 @@
...
+using System;
+using NativeApi = Interop.Diagnostics;
+
+namespace Tizen.System
+{
+    /// <summary>
+    /// Enumeration for the context types.
+    /// </summary>
+    /// <since_tizen> 8 </since_tizen>
+    public enum ContextType
```
WonyoungChoi, line 26: Move public enumerations to the separated file like `DiagnosticsEnumerations.cs`.

src/Tizen.Diagnostics/Interop/Interop.Diagnostics.cs
```diff
@@ -0,0 +1,78 @@
...
+        public enum CtxType
+        {
+            Crash = 0,
+        }
+
+        public enum ReportType
+        {
+            CrashFull = 0,
+            CrashInfoJson,
+        }
```
WonyoungChoi, line 49: Don't use duplicated types. Use public enumerations type `Tizen.Diagnostics.ContextType`

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,314 @@
...
+        /// <summary>
+        /// Reads given number of bytes from diagnostics data.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        /// <param name="size">Number of bytes to read.</param>
+        /// <param name="timeout">Timeout [ms] for reading requested number of bytes.</param>
+        /// <returns>
+        /// An array of bytes read.
+        /// </returns>
+        public unsafe byte[] Read(int size, int timeout)
```
WonyoungChoi, line 267: Avoid to open `unsafe` method as public. Use unsafe block in the method body or make and wrap another unsafe private method.

WonyoungChoi, line 267: what value of timeout if user does not want to set `timeout`? how about providing the overloading method to pass default timeout value?

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,304 @@
...
+        /// <param name="timeout">Timeout [ms] for reading requested number of bytes.</param>
+        /// <returns>
+        /// An array of bytes read.
+        /// </returns>
+        public byte[] Read(int size, int timeout)
+        {
+            return _read(size, timeout);
+        }
+
+        private unsafe byte[] _read(int size, int timeout)
```
WonyoungChoi, line 257: Don't use underscore for private methods. 
```suggestion
        private unsafe byte[] ReadUnsafe(int size, int timeout)
```

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,304 @@
...
+
+        /// <summary>
+        /// Reads given number of bytes from diagnostics data.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        /// <param name="size">Number of bytes to read.</param>
+        /// <returns>
+        /// An array of bytes read.
+        /// </returns>
+        public byte[] Read(int size)
```
WonyoungChoi, line 238: How do users know the size to read?

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,304 @@
...
+using System;
+using NativeApi = Interop.Diagnostics;
+
+namespace Tizen.Diagnostics
+{
+    /// <summary>
+    /// The Diagnostics class is used for requesting client data and receiving diagnostics report events.
+    /// </summary>
+    /// <since_tizen> 8 </since_tizen>
+    public class Diagnostics : IDisposable
```
myroot, line 26: It should be singleton object or static class because native API is a handless so it is working as global

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,304 @@
...
+
+        private void SetNotificationCallback()
+        {
+            _notificationCallback = OnNotification;
+            var ret = NativeApi.SetNotificationCb(_notificationCallback, IntPtr.Zero);
+            if (ret != NativeApi.DiagnosticsError.None)
+                throw ExceptionFactory.CreateException(ret);
+        }
+
+        private void OnNotification(IntPtr ctx, IntPtr data)
```
myroot, line 87: Do you know lifetime of `IntPtr data`?
In Native API, Usually it only valid during on callback, but in C#, it should be valid out scope of callback

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
kkuchciak: This function should be used in loop reading pieces of data until EOF (returns null). The size may (crash report) or may not (client's data) be known, so there is no API to check it.

WonyoungChoi: In that case, It is not good solution to return a byte array allocated in the method. It will make many allocation and copy in the real use-case. 
Please change the signature to
```c#
int Read(Byte[] buffer, int count)
```
if `offset` is possible, 
```c#
int Read(Byte[] buffer, int offset, int count)
```
See: https://docs.microsoft.com/ko-kr/dotnet/api/system.io.stream.read?view=netcore-3.1
And please add more user-kindly description to these methods.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
kkuchciak: Done. Created static class and callback is set when user subscribe to ReportReceived event.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
kkuchciak: It's valid until destroyed in Dispose function. Native API is not releasing this object, after returning from callback. It should be destroyed explicitly by (Native API) user.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
kkuchciak: Done. Unfortunately, offset is not possible.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,285 @@
...
+                SetNotificationCallback();
+                _reportReceived += value;
+            }
+            remove
+            {
+                if (_reportReceived != null)
+                {
+                    UnsetNotificationCallback();
+                    _reportReceived -= value;
+                };
```
myroot, line 52:
```suggestion
                _reportReceived -= value;
                if (_reportReceived == null)
                {
                    UnsetNotificationCallback();
                };
```

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,285 @@
...
+        /// Event to be invoked when new report context arrives.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        public static event EventHandler<ReportReceivedEventArgs> ReportReceived
+        {
+            add
+            {
+                if (_reportReceived != null|| value.GetInvocationList().Length > 1)
+                {
+                    throw new InvalidOperationException("Only one handler allowed");
```
myroot, line 41: You must support multiple event handler

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,285 @@
...
+                }
+                SetNotificationCallback();
+                _reportReceived += value;
```
myroot, line 44:
```suggestion
                if (_reportReceived == null)
                        SetNotificationCallback();
                _reportReceived += value;
```

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
myroot: offsetInt32
The zero-based byte offset in buffer at which to begin storing the data read from the current stream.

Why not support? it is not related with native api

src/Tizen.Diagnostics/Tizen.Diagnostics/DiagnosticsEnumerations.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using System;
+
+namespace Tizen.Diagnostics
+{
+    /// <summary>
+    /// Enumeration for the context types.
+    /// </summary>
+    /// <since_tizen> 8 </since_tizen>
+    public enum ContextType
```
m-macias, line 25: What is the point of the enum with only a single value? I see that it is used in `internal DiagnosticsContext(IntPtr handle)`, but there it will assign only ContextType.Crash value.

kkuchciak: ContextType is going to be extended in the future.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
kkuchciak: I have not much experience with dotnet, but I understand that it will take care of memory management of `IntPtr Ctx` and sharing it across all handlers? Native API will create just one instance of it in the callback and it will be destroyed with `Dispose()` after returning from the first handler.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
kkuchciak: Ok, I thought about offset in reading from source, not in writing to buffer.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
myroot: When `ReportReceived` was raised, DiagnosticsContext will be created and it will be shared to all event handler. 
if anyone dispose context, all hander can't use it

Anyway, we should support multiple handler,
And app developer be careful using of disposable object.

src/Tizen.Diagnostics/Tizen.Diagnostics/ReportReceivedEventArgs.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Tizen.Diagnostics
+{
+    /// <summary>
+    /// ReportReceivedEventArgs is an extended EventArgs class.
+    /// This class contains event arguments for the ReportReceived event from the Diagnostics class.
+    /// </summary>
+    /// <since_tizen> 8 </since_tizen>
+    public class ReportReceivedEventArgs : EventArgs
+    {
+        internal ReportReceivedEventArgs(object sender, IntPtr ctx)
```
myroot, line 29: sender is not used
```suggestion
        internal ReportReceivedEventArgs(IntPtr ctx)
```

myroot (changes requested):

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
kkuchciak: @myroot It follows that below functions are needed for reading:
        public int Read(byte[] buffer, int count);
        public int Read(byte[] buffer, int count, int timeout);
        public int Read(byte[] buffer, int offset, int count);
        public int Read(byte[] buffer, int offset, int count, int timeout);
There is a conflict in signatures. How you prefer to resolve it?

myroot:
``` c#
int Read(byte[] buffer, int offset, int count);
int Read(byte[] buffer, int offset, int count, int timeout);
```

kkuchciak: ok, sure.

kkuchciak: Done.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,298 @@
...
+        /// <summary>
+        /// Requests other apps to dump diagnostics data.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        /// <param name="clientId">An id of app or service to request.</param>
+        /// <param name="parameters">Array of parameters passed to the client.</param>
+        /// <returns>
+        /// The instance of DiagnosticsData.
+        /// </returns>
+        /// <privilege>
```
m-macias, line 60: Empty documentation tag.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,298 @@
...
+
+        /// <summary>
+        /// Gets report of specific type from the context.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        /// <param name="report">Report type to retrieve from the context.</param>
+        /// <returns>
+        /// The instance of DiagnosticsData.
+        /// </returns>
+        /// <privilege>
```
m-macias, line 136: Empty documentation tag.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
kkuchciak: According to the comment: https://code.sec.samsung.net/jira/browse/ACR-1581?focusedCommentId=1137103&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1137103
It's prohibited to mention internal privilege in public API, so Mr. @kd0228-kim decided to leave it blank and add remarks tag with proper info.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
kkuchciak: As above.

WonyoungChoi:
@kkuchciak I think we need to think about the names of the namespace and classes one more time before merging.
Can you explain why we need this API and where it is used (use cases)? What will be added to ContextType? Is the crash information available only?

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,298 @@
...
+        /// <param name="parameters">Array of parameters passed to the client.</param>
+        /// <returns>
+        /// The instance of DiagnosticsData.
+        /// </returns>
+        /// <privilege>
+        /// </privilege>
+        /// <remarks>
+        /// This function is permitted only to an app signed by platform level certificates.
+        /// </remarks>
+        public static DiagnosticsData RequestClientData(string clientId, string[] parameters)
```
WonyoungChoi, line 65: Can't users choose the report type of `DiagnosticsData`? Is the different kind of data with the data from Received event?

kkuchciak:
@WonyoungChoi There are two use cases:
1. Receiving report when some abnormality happen on platform (for now only crash reports are supported).

    `Diagnostics.ReportReceived` event is generated and `DiagnosticsContext` is created which contains only information about context type. To get actual report data, `DiagnosticsContext.GetReport()` should be used. Available reports depend on context provider. For now, crash-worker is the only provider, which can give full report in .tar.gz format (`ReportType.CrashFull`) or JSON report (`ReportType.CrashInfoJson`).

    Report data can be read with `DiagnosticsData.Read()`.

2. Requesting other components (apps, services that support the functionality) to dump own diagnostics logs.

    `Diagnostics.RequestClientData()` is used for this and again report data can be read with `DiagnosticsData.Read()`.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
kkuchciak: `DiagnosticsData` is just a stream, it is not related to any type of report. It can be data from any report requested with `DiagnosticsContext.GetReport()` or data from report requested from other apps with `Diagnostics.RequestClientData()`

kkuchciak: Actually `ReportReceived` event does not provide any specific report data yet. It just provides report context, from where user can get specific type of the report. So user must call `GetReport` in the event handler to get the report. Generated event means something like this: "Hi, I'm crash worker, some app has just crashed and I have some reports to offer. What kind of report do you want?"

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
m-macias: I agree that "internal" privilege should not be included in the "public" API reference, but leaving an empty documentation tag is also not acceptable, because it means nothing for the API user (it won't be even displayed by the IDE, I am not sure about docFX later). Comment of @kd0228-kim applies for the native API documentation, but in C# API it would be better to remove this tag entirely and leave the additional info in the `remarks` tag (as you have already done).

kkuchciak: Ok, done.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,294 @@
...
+                throw ExceptionFactory.CreateException(ret);
+
+            return new DiagnosticsData(data);
+        }
+
+        /// <summary>
+        /// Stores context type.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        public ContextType Type { get; internal set; }
```
myroot, line 150: `Type` is ambiguous with `Object.GetType`
```suggestion
        public ContextType ContextType { get; internal set; }
```

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,294 @@
...
+        private unsafe int ReadUnsafe(byte[] buffer, int offset, int count, int timeout)
+        {
+            var length = Convert.ToUInt32(count);
+            uint bytes_read = 0;
+
+            if (offset + count > buffer.Length)
+                throw new ArgumentException("The sum of offset and count is larger than the buffer length");
+
+            if (buffer == null)
+                throw new ArgumentNullException(nameof(buffer));
```
myroot, line 248: buffer should be checked before use buffer.Length
```suggestion
            if (buffer == null)
                throw new ArgumentNullException(nameof(buffer));

            if (offset + count > buffer.Length)
                throw new ArgumentException("The sum of offset and count is larger than the buffer length");
```

src/Tizen.Diagnostics/Tizen.Diagnostics/ReportReceivedEventArgs.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// ReportReceivedEventArgs is an extended EventArgs class.
+    /// This class contains event arguments for the ReportReceived event from the Diagnostics class.
+    /// </summary>
+    /// <since_tizen> 8 </since_tizen>
+    public class ReportReceivedEventArgs : EventArgs
+    {
+        internal ReportReceivedEventArgs(IntPtr ctx)
+        {
+            Context = new DiagnosticsContext(ctx);
+        }
```
myroot, line 32: I think, it is better reasonable, ReportReceivedEventArgs class does not need to know creation of context.
```suggestion
        internal ReportReceivedEventArgs(DiagnosticsContext ctx)
        {
            Context = ctx;
        }
```

myroot (changes requested):

src/Tizen.Diagnostics/Tizen.Diagnostics/ExceptionFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+            switch (err)
+            {
+                case Interop.Diagnostics.DiagnosticsError.InvalidParameter:
+                    return new ArgumentException("Invalid parameters provided");
+
+                case Interop.Diagnostics.DiagnosticsError.IoError:
+                    return new InvalidOperationException("I/O Error occured");
+
+                case Interop.Diagnostics.DiagnosticsError.OutOfMemory:
+                    return new OutOfMemoryException("Out of memory");
```
myroot, line 36: ❌ DO NOT explicitly throw OutOfMemoryException. This exception is to be thrown only by the CLR infrastructure.

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types#outofmemoryexception

kkuchciak: Should I just remove it or replace with other error?
```
case Interop.Diagnostics.DiagnosticsError.OutOfMemory:
    return new InvalidOperationException("I/O Error occured");
```

kkuchciak: Actually there are quite a few other APIs that are throwing OutOfMemoryException.

myroot: Yes, should be replaced with InvalidOperationException,

kkuchciak: Done. However, I'm not sure about the exception message.

myroot:
Is it a public API or internal API?

kkuchciak:
@myroot I'm not sure if this API is public or internal, but only apps signed with internal/platform certificate can use it.

As I've explained in other comment, there are 2 main use cases:
1. Receiving notification when some app crash on the platform.

    When some crash happen on the platform, **crash-worker** is preparing crash report, the notification is sent via dbus and callback is invoked in Diagnostics API. In this callback user can call `DiagnosticsContext.GetReport()` to get specific report related to the crash. Internally, it is calling `get_file()` method call provided **crash-service** to receive the report.

    The report may be uploaded to external server or processed offline by the client app (an app using Diagnostics API).

    For now, only crash notifications are supported, but it will be extended in the future to support other abnormalities.

2. Requesting status data from other apps.

    The client app can request other app to dump it's own diagnostics data. The 'other app' (sender) must provide this functionality by using **dumpsys** API (internal). The communication is being estabilished by **dumpsys-service** via dbus and data is being sent by pipe.

kkuchciak:
This API is not being used by any application yet.

myroot:
It looks internal API, and should not be exposed in public API documents

myroot:
This API User, should know about this, 
 1. The reporting format that generated by crash-worker 
 2. Which apps are implementing dumpsys API (internal)
But it's only provide as internal and not specfication. 
So, It can't be public API.

kkuchciak:
>     1. The reporting format that generated by crash-worker
I don't understand it. You mean the format of specific reports? It depends on the crash-worker version and is changing quite often.

>     2. Which apps are implementing dumpsys API (internal)
At this moment, no app is utilizing dumpsys API as well.

kkuchciak:
> So, It can't be public API.
I just got response from my HQ PL (KunHoon Baik) and he confirmed that this API is 'public platform privileged'.

src/Tizen.Diagnostics/Tizen.Diagnostics/Diagnostics.cs
```diff
@@ -0,0 +1,294 @@
...
+                _disposed = true;
+            }
+        }
+    }
+
+    /// <summary>
+    /// The DiagnosticsData class is used for reading data (either from report or client).
+    /// </summary>
+    /// <since_tizen> 8 </since_tizen>
+    public class DiagnosticsData : IDisposable
```
idkiller, line 184: If this acts like a ReadOnly Stream, I suggest implementing System.IO.Stream.
Many open sources already implement Stream with ReadOnly, and the implementation can come out almost the same as this class just by throwing a NotSupport exception in the Write method.
Also, by implementing Stream, you can use many classes for Stream.

kkuchciak: > Many open sources already implement Stream with ReadOnly
Can you show some examples?

myroot:
> I don't understand it. You mean the format of specific reports? It depends on the crash-worker version and is changing quite often.
> 

According to your comment, it quite depends on crash-worker, and crash-worker behavior is not opened to app developer, So, it also should be internal.
If you want to make as public API, you should make a specification the behavior of crash-worker and it open to the app developer

myroot:
Native Diagnostics API require a platform level privilege and APIs that requiring platform level privilege are not added into TizenFX, 
if you want add as internal API, it can placed into https://github.com/Samsung/TizenFX/tree/master/internals/src, and it does not expose API documents

internals/src/Tizen.Inspection/Tizen.Inspection/EventReceivedEventArgs.cs
```diff
@@ -0,0 +1,40 @@
...
+using System;
+using NativeApi = Interop.Inspection;
+
+namespace Tizen.Inspection
+{
+    /// <summary>
+    /// EventReceivedEventArgs is an extended EventArgs class.
+    /// This class contains event arguments for the EventReceived event from the Inspection class.
+    /// </summary>
+    /// <since_tizen> 8 </since_tizen>
```
WonyoungChoi, line 26: Remove `since_tizen` tags for internal APIs

internals/src/Tizen.Inspection/Tizen.Inspection/Inspection.cs
```diff
@@ -0,0 +1,296 @@
...
+using System;
+using NativeApi = Interop.Inspection;
+
+namespace Tizen.Inspection
+{
+    /// <summary>
+    /// The Inspection class is used for requesting client data and receiving inspection events.
+    /// </summary>
+    /// <since_tizen> 8 </since_tizen>
+    public static class Inspection
```
WonyoungChoi, line 26: DO NOT use the same name for a namespace and a type in that namespace.
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-namespaces

kkuchciak: Would you suggest to change namespace to Tizen.System?

WonyoungChoi (approved):

myroot (approved):

PR #1937, opened 2020-08-24
[NUI] rename from styleSheet to styleName

dasg34:
Since `StyleSheet` is used in Dali with the JSON file format, this rename will
avoid name confusion.

Co-authored-by: Jiyun Yang <ji.yang@samsung.com>

### API Changes ###

 - ACR:


rabbitfor (approved):
LGTM

PR #1952, opened 2020-08-27
[NUI] Add Seamless transition

huiyueun:
Add NUIFrameBroker & NUIFrameProvider

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.Libraries.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.Libraries.cs
@@ -1,0 +1,27 @@
+﻿/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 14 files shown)


src/Tizen.NUI/src/internal/Interop/Interop.Libraries.cs
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+namespace Tizen.NUI
+{
+    internal static partial class Interop
+    {
+        internal static partial class Libraries
+        {
+            public const string FrameBroker = "libframe-broker.so.1";
```
flyofsky, line 23: When the class is internal, does the property need to be public?

src/Tizen.NUI/src/public/FrameBroker/FrameBrokerBase.cs
```diff
@@ -0,0 +1,281 @@
...
+    /// <summary>
+    /// Represents the Frame Broker.
+    /// </summary>
+    /// <since_tizen> 8 </since_tizen>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public abstract class FrameBrokerBase : IDisposable
+    {
+        private string LogTag = "Tizen.NUI";
+        private readonly SafeFrameBrokerHandle _handle;
+        private static Dictionary<int, Interop.FrameBroker.AppControlResultCallback> _resultCallbackMaps = new Dictionary<int, Interop.FrameBroker.AppControlResultCallback>();
```
flyofsky, line 35: Is there any reason for being private but static?

huiyueun: fixed

src/Tizen.NUI/src/internal/Interop/Interop.Libraries.cs
huiyueun: fixed

PR #1965, opened 2020-09-01
[NUI] replace NUI.ISelectorItem with System.ICloneable

dasg34:
`Tizen.NUI.ISelectorItem` and `System.ICloneable` are exactly same
interface.

### API Changes ###

 - ACR:


rabbitfor (approved):
LGTM

PR #1969, opened 2020-09-02
[NUI] Add KeyInputFocus property to View

Seoyeon2Kim:

### Description of Change ###
- KeyInputFocus is to receive key events to View.
- This property is originally public on DALi package,
 so to use it in NUI app, it should be supplied even as hidden API.

### API Changes ###
- N/A


JoogabYun (approved):

src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -243,6 +243,23 @@ public string StyleName
             }
         }
 
+        /// <summary>
+        /// The KeyInputFocus, type bool.
```
dasg34, line 247: hmm.. `KeyInputFoucs` type is string...

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
```diff
@@ -46,6 +46,26 @@ public partial class View
...
+            var view = (View)bindable;
+            if (newValue != null)
+            {
+                Tizen.NUI.Object.SetProperty(view.swigCPtr, View.Property.KEY_INPUT_FOCUS, new Tizen.NUI.PropertyValue((bool)newValue));
+            }
+        },
+        defaultValueCreator: (bindable) =>
+        {
+            var view = (View)bindable;
+            string temp;
```
dasg34, line 64: It may be `bool`.

dasg34 (changes requested):

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
Seoyeon2Kim: Oh You're right.
I did the wrong type here. Fixed it.
Thank you for reviewing it!!

src/Tizen.NUI/src/public/BaseComponents/View.cs
Seoyeon2Kim: Changed to 'bool' type of `KeyInputFocus`

dasg34 (approved):

AdunFang (approved):
It is OK.

PR #1975, opened 2020-09-03
[NUI] remove StateValueCollection from Selector

dasg34:
State-Value pair is now added to Selector.StateValueList without
duplicate check.

You need to use `Selector.StateValueList` when custom state-value pair is added
to Selector. and it is now able to add custom state and pre-defined
state in the same initializer.

Before:
```
Selector<string> textSelector = new Selector<string>()
{
    Normal = "Defalut",
    { ControlState.Pressed, "Pressed!" }, // build error
    { ControlState.Focused, "Focused!" }  // build error
};
```

After:
```
Selector<string> textSelector = new Selector<string>()
{
    Normal = "Default!",
    StateValueList =
    {
        { ControlState.Pressed, "Pressed!" },
        { ControlState.Focused, "Focused!" }
    }
};
```

Also, this patch fixes a bunch of CA2227(Collection properties should be read
only) warnings.

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
```diff
@@ -27,10 +28,31 @@ namespace Tizen.NUI.BaseComponents
...
-    public class Selector<T> : StateValueCollection<T>
+    public class Selector<T>
     {
         private readonly bool cloneable = typeof(T).IsAssignableFrom(typeof(ICloneable));
 
+        /// <summary>
+        /// The list for adding state-value pair.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public IList<StateValuePair<T>> StateValueList { get; private set; } = new List<StateValuePair<T>>();
```
rabbitfor, line 39: Why don't we just use List instead of IList?

rabbitfor:
Can you show me the xaml example about the custom control state in the selector?

src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
dasg34: MS design-guidelines not recommend `List<T>`.
```
❌ DO NOT use ArrayList or List<T> in public APIs.

These types are data structures designed to be used in internal implementation, not in public APIs. 
List<T> is optimized for performance and power at the cost of cleanness of the APIs and flexibility. 
For example, if you return List<T>, you will not ever be able to receive notifications when client code modifies the collection. 
Also, List<T> exposes many members, such as BinarySearch, that are not useful or applicable in many scenarios. 
The following two sections describe types (abstractions) designed specifically for use in public APIs.
```
(https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections)

dasg34:
> Can you show me the xaml example about the custom control state in the selector?

In xaml, custom control state does not work now. i will add control state converter later.

AdunFang (approved):
It is OK.

dasg34:
> Can you show me the xaml example about the custom control state in the selector?

Here is the custom state xaml example.
https://github.com/Samsung/TizenFX/pull/2002

woohyun:
@rabbitfor recommended to merge this commit a week later (after Sep 21).

dongsug-song:
@dasg34 
Hello,
could you solve conflicts?

PR #1978, opened 2020-09-03
[NUI] Fix exception error when sub Window is disposed

dongsug-song:
### Description of Change ###
[NUI] Fix exception error when sub Window is disposed

- https://code.sec.samsung.net/jira/browse/GRE-2009
- Add null check for Layers and LayoutContainer
- Add sample and test case for sub Window's Dispose()

### API Changes ###
none


AdunFang (approved):
It is OK.

huiyueun (approved):

PR #1982, opened 2020-09-04
[NUI] Fix Gaussian Blur View Size setting issue.

bshsqa:
- Currently, new Size2D is created when one of the WidthSpecification or HeightSpecification is set.
 - Then, the Size2D has incomplete value until the latter one is also entered.
 - e.g. if the size is (500, 500), after just WidthSpecification is entered the Size2D is (500, -2) until HeightSpecification will be entered.
 - (-2 is default negative value that notifing the value is still not set).

 - But, in the property setter of Size2D in View, the value is transfered to the Native API.
 - And the incomplete value make crash in Native API.
 - Because when the latter is entered, we just set Size2D.Width or Size2D.Height not to create Size2D itself. And, exact Size2D is not transfered to the Native.

 - So, this patch postpone to create Size2D utill both WidthSpecification and HeightSpecification are all entered.
 - This create new Size2D whenever new WidthSpecification or HeightSpecification is entered, and it transfers them to the Native well.

Signed-off-by: seungho <seungho@seungho.tn.corp.samsungelectronics.net>


dongsug-song (approved):
looks good to me

PR #1987, opened 2020-09-04
Use a platform agnostic name for native libraries

walac:
This patch makes the DALi C# binder and tizenfx-stub library names
platform agnostic. This guarantees they can be loaded in Windows and
Linux environments.

### DIFF
--- a/src/Tizen.Applications.Common/Interop/Interop.Libraries.cs
+++ b/src/Tizen.Applications.Common/Interop/Interop.Libraries.cs
@@ -18,10 +18,10 @@ internal static partial class Interop
 {
     internal static partial class Libraries
     {
-        public const string AppCommon = "libcapi-appfw-app-common.so.0";
+        public const string AppCommon = "capi-appfw-app-common";
         public const string AppControl = "libcapi-appfw-app-control.so.0";
         public const string AppEvent = "libcapi-appfw-event.so.0";
-        public const string AppManager = "libcapi-appfw-app-manager.so.0";
+        public const string AppManager = "capi-appfw-app-manager";
         public const string Bundle = "libbundle.so.0";
         public const string Rua = "librua.so.0";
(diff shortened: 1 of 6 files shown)


WonyoungChoi:
@HJLeee  Is this available on the dotnet runtime in Tizen?

HJLeee:
@walac I wonder if you have tested it before the submittion as there is no devel pkg in the target and in such case there is no, for example, capi-xxx.so, which links to capi-xxx.so.0.

walac:
@HJLeee I tested it on Windows and Linux environments. Is there another test I should perform?

WonyoungChoi:
@walac You should test on the tizen platform also, The xxxx.so symbolic link files are usually not installed in the platform binaries.

dongsug-song:
@walac 
Hello, this can not be merged, it will definitely make crash on tizen devices.
you can refer to the following links to get some guide for NUI windows/ubuntu backend.

https://code.sec.samsung.net/confluence/display/GFX/NUI+Developer+Getting+started
https://code.sec.samsung.net/confluence/display/GFX/How+to+use+VD+TV+emulator
https://code.sec.samsung.net/confluence/display/GFX/NUI+running+on+Ubuntu+backend
https://code.sec.samsung.net/confluence/display/GFX/NUI+Ubuntu+backend+Debugging
https://github.sec.samsung.net/NUI/windows-backend

thank you.

walac:
Hi @dongsug-song , I can't access the links, are they behind some sort of a VPN or Firewall? I tried to take another approach to the problem following [this dotnet documentation](https://github.com/dotnet/runtime/blob/master/docs/design/features/dllmap.md). The problem is that netstandard2.0 doesn't contain the referred APIs and the `dllmap` configuration is only implemented for [mono](https://www.mono-project.com).

walac:
Closing in favor of #2091

PR #1993, opened 2020-09-07
[WIP][NUI] Subscribe ThemeChanged when the view is added to the window

rabbitfor:
And it also unsubscribe ThemeChanged when the view is removed from the Window.

### API Changes ###

 - ACR:


AdunFang (approved):
It is OK.

PR #1995, opened 2020-09-08
[MachineLearning.Inference] Fix the bug for double type of GetValue<T>

again4you:
### Description of Change ###

This patch fixes the bug for double type of GetValue<T> method.

### API Changes ###
* No ACR since it is bugfix.


myroot (approved):

gichan-jang (approved):
Verified with TM1 device.

PR #1997, opened 2020-09-09
[NUI] Deprecate Adaptor class

dongsug-song:
### Description of Change ###
[NUI] Deprecate Adaptor class, ACR-373

### API Changes ###
https://code.sec.samsung.net/jira/browse/TCSACR-373


src/Tizen.NUI/src/public/Adaptor.cs
```diff
@@ -38,6 +38,8 @@ namespace Tizen.NUI
     ///
     /// </summary>
     /// <since_tizen> 4 </since_tizen>
+    [Obsolete("Deprecated in API8, will be removed in API10. This is not used anymore, please do not use.")]
+    [EditorBrowsable(EditorBrowsableState.Never)]
```
m-macias, line 42: Please DO NOT hide the API that you want to deprecate. It should remain visible until its removal.

AdunFang (approved):
It is OK.

adeelkazmi (approved):

PR #2014, opened 2020-09-15
[ElmSharp] Fix Transit.Go to start transition immediately

jaehyun0cho:
Transit.Go calls elm_transit_go_in with interval.
elm_transit_go_in uses timer so it does not start transition immediately
although interval is 0.

To start transition immediately without interval, Transit.Go is fixed to
call elm_transit_go.

### API Changes ###

 - ACR:


src/ElmSharp/ElmSharp/Transit.cs
```diff
@@ -246,7 +246,18 @@ public double EndAccelerationFactor
         /// <since_tizen> preview </since_tizen>
         public void Go(double interval = 0)
         {
-            Interop.Elementary.elm_transit_go_in(_handle, interval);
+            Double timeInterval = interval;
+
+            if (timeInterval.Equals(0.0))
```
myroot, line 251:
```suggestion
            if (timeInterval == 0)
```

src/ElmSharp/ElmSharp/Transit.cs
```diff
@@ -246,7 +246,18 @@ public double EndAccelerationFactor
         /// <since_tizen> preview </since_tizen>
         public void Go(double interval = 0)
         {
-            Interop.Elementary.elm_transit_go_in(_handle, interval);
+            Double timeInterval = interval;
```
myroot, line 249:
```suggestion
```

src/ElmSharp/ElmSharp/Transit.cs
jaehyun0cho: Thank you for review. I have updated the patch.

src/ElmSharp/ElmSharp/Transit.cs
jaehyun0cho: Thank you for review. I have updated the patch.

myroot (approved):

rookiejava:
One simple question. I couldn't find any information about that on [API reference](https://docs.tizen.org/application/native/api/mobile/5.5/group__Elm__Transit.html#ga5f39f8f0c030eacd601c9d9c6c89ebee).

>elm_transit_go_in uses timer so it does not start transition immediately although interval is 0.

Is this behavior a change from Tizen 6.0 M2? I'm curious because the existing version (4.0, 5.0, 5.5 and even 6.0 M1) works well with the current code.

rookiejava (approved):

PR #2029, opened 2020-09-18
[NUI] Make Scrollable finish when Gesture.StateType is Cancelled

bshsqa:
- Current implementation, the scroll of ScrollableBase will be finished only when Gesture.StateType is Finished.
 - However, by the hw, it can be Cancelled.

Signed-off-by: seungho <seungho@seungho.tn.corp.samsungelectronics.net>


huiyueun (approved):

rabbitfor:
LGTM

PR #2032, opened 2020-09-18
[NUI] Switch do not animate on selection when it is not on Window

rabbitfor:

### API Changes ###

 - ACR:


woohyun (approved):
I think this looks good to me. Thanks.

PR #2033, opened 2020-09-21
Tizen.Peripheral move all classes to it's namespaces.

xerrni:

### Description of Change ###
Move all classes to it's own namespaces.

### API Changes ###
Rename classes:
Adc -> AdcChannel
Gpio -> GpioPin
I2c -> I2cDevice
Spi -> SpiDevice
Pwm -> PwmPin
Uart -> SerialPort

Move class:
1.
```csharp
  public class Tizen.Peripheral.I2cDevice (class):
        public I2cDevice(int bus, int address) (constructor);
        public void Close() (method);
        public void Read(byte[] dataOut) (method);
        public void Write(byte[] data) (method);
        public byte ReadRegisterByte(byte register) (method);
        public void WriteRegisterByte(byte register, byte data) (method);
        public ushort ReadRegisterWord(byte register) (method);
        public void WriteRegisterWord(byte register, ushort data) (method);
        public void Dispose() (method);
```
to `Tizen.Peripheral.I2c` namespace
2. 
```csharp
    public class Tizen.Peripheral.AdcChannel : IDisposable (class):
        public AdcChannel(int device, int channel) (constructor);
        public uint ReadValue() (method);
        public void Close() (method);
        public void Dispose() (method);
```
to `Tizen.Peripheral.Adc` namespace

### DIFF
--- a/internals/src/Tizen.Peripheral/Tizen.Peripheral/Adc.cs
+++ b/internals/src/Tizen.Peripheral/Tizen.Peripheral/Adc.cs
@@ -17,13 +17,13 @@
 using System;
 using NativeAdc = Interop.Peripheral.Adc;
 
-namespace Tizen.Peripheral
+namespace Tizen.Peripheral.Adc
 {
     /// <summary>
     /// The class allows applications to use the platform ADC peripheral.
     /// </summary>
     /// <privilege>http://tizen.org/privilege/peripheralio</privilege>
-    public class Adc : IDisposable
+    public class AdcChannel : IDisposable
     {
         /// <summary>
         /// Native handle to ADC.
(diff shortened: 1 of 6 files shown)


myroot (changes requested):
❌ DO NOT use the same name for a namespace and a type in that namespace.
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-namespaces

I recommend to change class name like below (it comes from MS API)
Adc -> AdcChannel 
Gpio -> GpioPin 
I2c -> I2cDevice 
Spi -> SpiDevice 
Pwm -> PwmPin 
Uart -> SerialPort

WonyoungChoi:
How about moving all classes under `Tizen.Peripherals` namespace?

xerrni:
> 
> 
> ❌ DO NOT use the same name for a namespace and a type in that namespace.
> https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-namespaces
> 
> I recommend to change class name like below (it comes from MS API)
> Adc -> AdcChannel
> Gpio -> GpioPin
> I2c -> I2cDevice
> Spi -> SpiDevice
> Pwm -> PwmPin
> Uart -> SerialPort

done

xerrni:
> 
> 
> How about moving all classes to under `Tizen.Peripherals` namespace?

I would like to keep Tizen.Peripheral name so it will be consistent with native API.

myroot (approved):

xerrni:
I have rebased this commit.

PR #2036, opened 2020-09-22
[SystemSettings] Add lock for EventHandler

jinwangan:
Signed-off-by: JinWang An <jinwang.an@samsung.com>


src/Tizen.System.SystemSettings/Tizen.System.SystemSettings/SystemSettings.cs
```diff
@@ -28,6 +28,11 @@ namespace Tizen.System
     /// <since_tizen> 3 </since_tizen>
     public static class SystemSettings
     {
+        /// <summary>
+        /// Lock for EventHandlers.
+        /// </summary>
+		private static readonly object LockObj = new object();
```
myroot, line 34: replace to space
```suggestion
        private static readonly object LockObj = new object();
```

myroot (approved):

PR #2038, opened 2020-09-22
[Build] Change nitool to dotnettool in build script

HJLeee:
### Description of Change ###
Change `nitool` to `dotnettool` in build script


HJLeee:
S**t.

PR #1988, opened 2020-09-04
[NUI] Update MeasureCallback interface for latest Dali::Toolkit

walac:
Making MeasureCallback return a struct is causing the .NET runtime on
Windows 10 messes up with argument values. We changed the signature of
MeasureCallback on Dali::Toolkit to make the return value be an out
parameter.

The detailed explanation of what's the problem can be found at [1].

Moreover, we declare MeasureCallback as Cdecl, as it is declared this
way in the C++ header file.

[1] https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-toolkit/+/242665/

PR #2043, opened 2020-09-23
[NUI] Update MeasureCallback interface for latest Dali::Toolkit

dongsug-song:
### Description of Change ###
Making MeasureCallback return a struct is causing the .NET runtime on
Windows 10 messes up with argument values. We changed the signature of
MeasureCallback on Dali::Toolkit to make the return value be an out
parameter.

The detailed explanation of what's the problem can be found at [1].

Moreover, we declare MeasureCallback as Cdecl, as it is declared this
way in the C++ header file.

[1] https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-toolkit/+/242665/

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/FlexLayout.cs
+++ b/src/Tizen.NUI/src/public/Layouting/FlexLayout.cs
@@ -94,8 +94,8 @@ namespace Tizen.NUI
                 width = x;
                 height = y;
             }
-            float width;
-            float height;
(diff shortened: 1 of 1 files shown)


dongsug-song:
This is pulled again because https://github.com/Samsung/TizenFX/pull/1988/ has been out dated.

huiyueun (approved):

PR #2069, opened 2020-10-05
[NUI] Add defense code for BaseHandle

huiyueun:
vd kona - DF200925-00594
Check if the swigCptr value is correct after the dispose.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseHandle.cs
+++ b/src/Tizen.NUI/src/public/BaseHandle.cs
@@ -403,7 +403,7 @@ namespace Tizen.NUI
             if (swigCPtrCopy.Handle == IntPtr.Zero)
             {
                 global::System.Diagnostics.StackTrace trace = new global::System.Diagnostics.StackTrace();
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/BaseHandle.cs
```diff
@@ -405,6 +405,11 @@ public bool HasBody()
                 return false;
             }
 
+            if (swigCPtrCopy.Handle == IntPtr.Zero)
```
dongsug-song, line 408: I think it would be better that this codes be moved line 403 where is above "if(disposed == true)"
also could you add some debugging log like this?

            global::System.Diagnostics.StackTrace t = new global::System.Diagnostics.StackTrace();
            Tizen.Log.Error("NUI", t);
            Tizen.Log.Error("NUI", "[ERROR] the native handle is invalid!");

huiyueun: fixed

dongsug-song (approved):

PR #2079, opened 2020-10-06
[NUI] add null checking for equality operator of ControlState

dasg34:
This is patch to prevent a null reference exception in the following
code.

```cs
bool b = null == ControlState.Pressed;
```

### API Changes ###

 - ACR:


huiyueun (approved):

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -235,7 +235,7 @@ public override string ToString()
         /// <param name="rhs">A <see cref="ControlState"/> on the right hand side.</param>
         /// <returns>true if the ControlStates are not equal; otherwise, false.</returns>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public static bool operator !=(ControlState lhs, ControlState rhs) => !lhs.Equals(rhs);
+        public static bool operator !=(ControlState lhs, ControlState rhs) => lhs == null || !lhs.Equals(rhs);
```
rabbitfor, line 238: In case that both `lhs` and `rhs` are null, this will return true. Is it OK?

dasg34: When lhs and rhs are null, `operator !=` should return false.

rabbitfor (approved):
LGTM

PR #2091, opened 2020-10-08
[NUI] Load Tizen.NUI dependencies libs on Windows

Coquinho:
### Description of Change ###
This use `MSBuild` `-p:DefineConstants=NOTIZEN` directive to define a
Windows build, which impact the names of `Tizen.NUI` to search for.

This impacts the load of:
- `dali-csharp-binder`
- `capi-appfw-app-manager`
- `capi-appfw-app-common`
- `dlog`

We first tried to define those names at run-time, but
`DllImportAtribbuts` are `C#` `const strings`, which need to be defined
at compile-time, so it raised the need of the preprocessor directive to
select the correct name.

### API Changes ###
Now, to compile/run on Windows you shoud pass `-p:DefineConstants=NOTIZEN` to
`build.sh` while running it on `Git bash` shell:
```sh
$ ./build.sh full -p:DefineConstants=NOTIZEN
```
While maintaining the original use on Linux:
```sh
$ ./build.sh full
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,9 +43,17 @@ git clone https://github.com/Samsung/TizenFX.git
 cd TizenFX
 ```
 ### How to build
+#### Linux
 ```bash
 ./build.sh full
 ./build.sh pack
+```
+
+#### Windows
+On a `Git bash` shell with `dali-env` seted, run:
+```bash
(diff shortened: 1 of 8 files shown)


Coquinho:
Closing in favor of #2118

PR #2098, opened 2020-10-14
[NUI] Add InterceptTouchEvent

JoogabYun:
The Touch event calls the TouchEvent callback by going back from the last child actor to the parent via hitTest.
InterceptTouchEvent checks the touch event in the parent first.
Returning false from interceptTouchEvent allows child actors to receive TouchEvents.
If it returns true, the actor will receive a TouchEvent.

for example

   View parent = new View();
   View child = new View();
   parent.Add(child);
   child.TouchEvent += childFunctor;
   parent.TouchEvent += parentFunctor;

The callbacks are called in the order childFunctor -> parentFunctor.

If you connect interceptTouchSignal to parentActor.

   parent.InterceptTouchEvent += interceptFunctor;

When interceptFunctor returns false, it is called in the same order childFunctor -> parentFunctor.
If intereptFunctor returns true, it means that the TouchEvent was intercepted.
So the child actor will not be able to receive touch events.
Only the parentFunctor is called.

https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/245321/
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-core/+/245258/

### API Changes ###

 - ACR:


huiyueun:
I think it will be easy to check when releasing if you mention the relevant dali patch in the comment.

AdunFang (approved):
It is OK.

JoogabYun:
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/245321/
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-core/+/245258/

PR #2099, opened 2020-10-16
[NUI] Fix Registry's memory leak

dongsug-song:
### Description of Change ###
[NUI] Fix Registry's memory leak
- new RefObject(xxx, false), this causes unhandled RefObject in dali.
- "cMemoryOwn = false" means that this doesn't care about dali native's object and will not delete native object when nui dispose.
- currently "ViewAddedEventToWindow" is not used and this will be fixed by using Event subscribing pattern soon.

### API Changes ###
none


src/Tizen.NUI/src/internal/Registry.cs
```diff
@@ -77,8 +76,7 @@ internal static void Register(BaseHandle baseHandle)
         /// <param name="baseHandle"> The instance of BaseHandle (C# base class)</param>
         internal static void Unregister(BaseHandle baseHandle)
         {
-            RefObject refObj = baseHandle.GetObjectPtr();
```
huiyueun, line 80: How about deleting 'GetObjectPtr' that has a memory leak issue?
If this api is an internal api that is not used.
If the developer does not know and uses it, the memory leak issue may occur, so if it is difficult to delete it, it would be better to leave a comment.

src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -133,8 +133,9 @@ internal View(global::System.IntPtr cPtr, bool cMemoryOwn, bool shown = true) :
                 PositionUsesPivotPoint = false;
             }
 
-            _onWindowSendEventCallback = SendViewAddedEventToWindow;
-            this.OnWindowSignal().Connect(_onWindowSendEventCallback);
+            //ToDo: this has memory leak and this is not used currently. will be fixed soon by using Event subscribing pattern.
```
huiyueun, line 136: I remember that I made this event addition request to support automatic test in appium. Is there any problem using appium even if I delete it?

src/Tizen.NUI/src/internal/Registry.cs
dongsug-song: it is removed.

src/Tizen.NUI/src/public/BaseComponents/View.cs
dongsug-song: You are right, there would be a problem. 
I will upload patch again soon.

huiyueun (approved):

src/Tizen.NUI/src/public/BaseComponents/View.cs
dasg34: Is there any update for this?

Can i remove `View._onWindowSendEventCallback`?
There is build warning.
`/home/yohoho/tizen/TizenFX/src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs(61,43): warning CS0649: Field 'View._onWindowSendEventCallback' is never assigned to, and will always have its default value null [/home/yohoho/tizen/TizenFX/src/Tizen.NUI/Tizen.NUI.csproj]`

PR #2100, opened 2020-10-16
[NUI] Fix recursion when the view sets KeyboardNavigationSupport to true

rabbitfor:
In Dali code, the method GetNextKeyboardFocuableActor should return
an empty actor (= the base handle without a referenced object)
if there is no actor can be focused.

The corresponding method in NUI is GetNextFocusableView.
It returned new View() (that is not an empty actor) which leads to the wrong result.
So instead returning new View, this patch returns a null IntPtr.

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/BaseComponents/CustomView.cs
```diff
@@ -434,7 +434,7 @@ public virtual void OnFocusLost()
         /// <since_tizen> 3 </since_tizen>
         public virtual View GetNextFocusableView(View currentFocusedView, View.FocusDirection direction, bool loopEnabled)
```
huiyueun, line 435: Do not you need to modify the description of the return value in the comment?
empty handle -> null

rabbitfor:
Please note that, the related dali-csharp-binder patch should be merged first. 
(https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/245938/)

src/Tizen.NUI/src/public/BaseComponents/CustomView.cs
rabbitfor: Done!

huiyueun (approved):

hyunjushin (approved):

rabbitfor:
I will merge this patch after the intergration of dali-csharp-binder is done.

hyunjushin (dismissed):
Please re-push this patch into DevelNUI branch. DevelNUI will be integrated with master branch on every week.

PR #2109, opened 2020-10-21
[NUI] Add DisallowInterceptTouchEvent

JoogabYun:
If child view doesn't want the parent's view to intercept the touch, you can set it to true.
 for example :
     parent.Add(child);
     parent.InterceptTouchEvent += OnInterceptTouchEvent;
     View view = child.GetParent() as View;
     view.DisallowInterceptTouchEvent = true;
 This prevents the parent from interceping touch.

### API Changes ###

 - ACR: NONE


src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
```diff
@@ -812,6 +825,12 @@ private bool OnInterceptTouch(IntPtr view, IntPtr touchData)
                 return true;
             }
 
+            // RequestDisallowInterceptTouchEvent prevents the parent from intercepting touch.
+            if(RequestDisallowInterceptTouchEvent)
```
rabbitfor, line 829: Please add a space between `if` and `(`

src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
```diff
@@ -823,11 +842,6 @@ private bool OnInterceptTouch(IntPtr view, IntPtr touchData)
                 consumed = _interceptTouchDataEventHandler(this, e);
             }
 
-            if (enableControlState && !consumed)
```
rabbitfor, line 826: This code is needed when the View.EnableControlState is true.

rabbitfor (changes requested):

rabbitfor (approved):
LGTM

PR #2111, opened 2020-10-22
[NUI] Bug fixed about TextPageUtil

JoogabYun:
add check remainLength and init totalPageCnt value

### API Changes ###

 - ACR: NONE


dongsug-song (approved):
looks good to me

PR #2114, opened 2020-10-22
[NUI] Fix Size2D to work correctly in Loading

Seoyeon2Kim:

### Description of Change ###
- Currently, Size2D is not working well because of duplicated size properties.
- Fixed both Size and Size2D to work well in Loading

### API Changes ###
- N/A

### DIFF
--- a/src/Tizen.NUI.Components/Style/LoadingStyle.cs
+++ b/src/Tizen.NUI.Components/Style/LoadingStyle.cs
@@ -1,5 +1,5 @@
 ﻿/*
- * Copyright(c) 2019 Samsung Electronics Co., Ltd.
+ * Copyright(c) 2020 Samsung Electronics Co., Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


huiyueun (approved):

src/Tizen.NUI.Components/Style/LoadingStyle.cs
```diff
@@ -58,7 +58,21 @@ public LoadingStyle(LoadingStyle style) : base(style)
...
-        public Size LoadingSize { get; set; }
+        public Size LoadingSize
+        {
+            get
+            {
+                return this.Size;
+            }
+            set
+            {
+                if (value == null)
```
rabbitfor, line 69: I think it is better to allow null value for the style property.

Seoyeon2Kim: Oh I see. Thanks for your review. It's because it's a style property.

rabbitfor (approved):
LGTM

PR #2116, opened 2020-10-22
[NUI] Add EnableMultiSelected to SelectGroup

jaehyun0cho:
To support selecting multiple SelectButtons in SelectGroup,
bool EnableMultiSelected is added to SelectGroup.

By adding EnableMultiSelected in SelectGroup, RadioButtonGroup does not
need to implement its logic to prevent selecting multiple RadioButtons.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/SelectGroup.cs
+++ b/src/Tizen.NUI.Components/Controls/SelectGroup.cs
@@ -70,6 +70,12 @@ namespace Tizen.NUI.Components
         public int SelectedIndex => selectedIndex;
 
         /// <summary>
+        /// EnableMultiSelection is used to indicate if SelectGroup can select multiple SelectButtons.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool EnableMultiSelection { get; set; } = true;
+
+        /// <summary>
         /// Construct SelectionGroup
         /// </summary>
         /// <since_tizen> 6 </since_tizen>
@@ -142,7 +148,7 @@ namespace Tizen.NUI.Components
         }
 
         /// <summary>
-        /// Overrides this method if want to handle behavior after pressing return key by user.
+        /// Called when the state of Selected is changed.
         /// </summary>
         /// <param name="selection">The selection selected by user</param>
         /// <since_tizen> 6 </since_tizen>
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI.Components/Controls/SelectGroup.cs
```diff
@@ -69,6 +69,13 @@ public abstract class SelectGroup
         [EditorBrowsable(EditorBrowsableState.Never)]
         public int SelectedIndex => selectedIndex;
 
+        /// <summary>
+        /// EnableMultiSelected is used to indicate if SelectGroup can select multiple SelectButtons.
+        /// </summary>
+        /// <since_tizen> 9 </since_tizen>
```
m-macias, line 75: Please don't use `<since_tizen>` tag for EditorBrowsable items.

jaehyun0cho: Thank you for your  comment.
I thought that EditorBrowsable items are also required to have since_tizen tags because other EditorBrowsable items have since_tizen tags.
But now I see they are incorrect due to your comment.

I have removed the since_tizen tag.
Thank you!

rabbitfor (approved):
LGTM

src/Tizen.NUI.Components/Controls/SelectGroup.cs
```diff
@@ -69,6 +69,12 @@ public abstract class SelectGroup
         [EditorBrowsable(EditorBrowsableState.Never)]
         public int SelectedIndex => selectedIndex;
 
+        /// <summary>
+        /// EnableMultiSelected is used to indicate if SelectGroup can select multiple SelectButtons.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool EnableMultiSelected { get; set; } = true;
```
rabbitfor, line 76: How about EnableMultiSelection?

jaehyun0cho: Thank you for your opinion!
I thought that "Selection" is used for Text selection so it may confuse users.
However, MultiSelection sounds much more natural.

BTW, based on your opinion, I think SelectedChanged and SelectedChangedEventArgs may be required to be changed their names to SelectionChanged and SelectionChangedEventArgs.
This can be done in API9 by deprecating the SelectedChanged and SelectedChangedEventArgs.

I have updated the code.
Thank you.

hyunjushin (changes requested):
Please re-push this patch into DevelNUI branch. DevelNUI will be integrated with master branch on every week.

PR #2119, opened 2020-10-26
[NUI] Add try-catch block in New function

huiyueun:
- VD Kona : DF200922-00146
- Analyze the log when the problem is reproduced after inserting the try-catch block with log.
- Modify to throw ApplicationException, not NullException in Retrieve().

https://github.sec.samsung.net/dotnet/home/issues/386

### API Changes ###

 - ACR:


src/Tizen.NUI/src/internal/Application.cs
```diff
@@ -1194,12 +1194,21 @@ public static Application New(int argc, string stylesheet, Application.WindowMod
...
+                int argc = args.Length;
+                string argvStr = string.Join(" ", args);
 
-            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+                ret = new Application(NDalicPINVOKE.Application_New__MANUAL_4(argc, argvStr, stylesheet, (int)windowMode), true);
+                if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+            }
+            catch (Exception exception)
+            {
+                Tizen.Log.Fatal("NUI", "message : " + exception.Message);
```
dongsug-song, line 1208: how about to add some message more? 
ex: Tizen.Log.Fatal("NUI", "[Error] got exception during Application New(), this should not occur, msg : " + exception.Message);

src/Tizen.NUI/src/internal/Application.cs
```diff
@@ -1213,11 +1222,22 @@ public static Application New(int argc, string stylesheet, Application.WindowMod
...
+                int argc = args.Length;
+                string argvStr = string.Join(" ", args);
+
+                ret = new Application(NDalicPINVOKE.Application_New_WithWindowSizePosition(argc, argvStr, stylesheet, (int)windowMode, Rectangle.getCPtr(positionSize)), true);
+                if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+            }
+            catch (Exception exception)
+            {
+                Tizen.Log.Fatal("NUI", "message : " + exception.Message);
+                Tizen.Log.Fatal("NUI", "Stack Trace : " + exception.StackTrace);
```
dongsug-song, line 1237: as far as I know, if exception occurs, BT(back trace) call stack is shown by DOTNET_LAUNCHER. isn't it same? I don't know well, so I am just asking.

src/Tizen.NUI/src/internal/Application.cs
huiyueun: Fixed

src/Tizen.NUI/src/internal/Application.cs
huiyueun: Fixed

dongsug-song (approved):
looks good to me

src/Tizen.NUI/src/internal/Application.cs
```diff
@@ -1194,12 +1195,22 @@ public static Application New(int argc, string stylesheet, Application.WindowMod
...
+                string argvStr = string.Join(" ", args);
 
-            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+                ret = new Application(NDalicPINVOKE.Application_New__MANUAL_4(argc, argvStr, stylesheet, (int)windowMode), true);
+                if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+            }
+            catch (Exception exception)
+            {
+                Tizen.Log.Fatal("NUI", "[Error] got exception during Application New(), this should not occur, msg : " + exception.Message);
+                Tizen.Log.Fatal("NUI", "[Error] error line number : " + new StackTrace(exception, true).GetFrame(0).GetFileLineNumber());
```
HJLeee, line 1210: This won't work if there is no 'pdb' file. (I believe this is the case for `tizenfx`.)
Please break down the current try-catch block into pieces until you find the exact point of exception.

src/Tizen.NUI/src/internal/Application.cs
```diff
@@ -1194,12 +1195,22 @@ public static Application New(int argc, string stylesheet, Application.WindowMod
...
-            Application ret = new Application(NDalicPINVOKE.Application_New__MANUAL_4(argc, argvStr, stylesheet, (int)windowMode), true);
+            Application ret = null;
+            try
+            {
+                int argc = args.Length;
+                string argvStr = string.Join(" ", args);
 
-            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+                ret = new Application(NDalicPINVOKE.Application_New__MANUAL_4(argc, argvStr, stylesheet, (int)windowMode), true);
+                if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
```
HJLeee, line 1205: I think `NDalicPINVOKE.SWIGPendingException.Pending` and `NDalicPINVOKE.SWIGPendingException.Retrieve()` should be an atomic operation.

huiyueun: https://github.com/Samsung/TizenFX/blob/d937dc176cc23d88355b9133867a6210d121dc76/src/Tizen.NUI/src/internal/Interop/NDalicPINVOKE.cs#L181

Currently, Retrieve and Set which access numExceptionsPending value are synchronized to Lock.
Is your comment that you need lock even when you get it from Pending?

src/Tizen.NUI/src/internal/Application.cs
huiyueun: Do you mean to separate the try-catch for each line?

src/Tizen.NUI/src/internal/Application.cs
HJLeee: Simply, imagine two threads calling like so;
```
Thread1: `pending`------------------------------->'Retrieve()'
Thread2: -------------'pending'->'Retrieve()'---------------->
```

HJLeee: I figured https://github.com/Samsung/TizenFX/blob/d937dc176cc23d88355b9133867a6210d121dc76/src/Tizen.NUI/src/internal/Interop/NDalicPINVOKE.cs#L151 so it might not be the case I depicted above.

Yet, it's hard to figure out the use of `numExceptionPending` nor the meaning of  API separation of for checking and retrieving exception object.

HJLeee:
This will include debuginfo into dll. No pdb need on the target. Use it for test and make sure it doesn't applied to the final product.

```diff
diff --git a/src/Tizen.NUI/Tizen.NUI.csproj b/src/Tizen.NUI/Tizen.NUI.csproj
index bcffc9095..4774b1139 100644
--- a/src/Tizen.NUI/Tizen.NUI.csproj
+++ b/src/Tizen.NUI/Tizen.NUI.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <NoWarn>$(NoWarn);CS0618;CS0809;CS1591</NoWarn>
+    <DebugType>embedded</DebugType>
   </PropertyGroup>
 
   <ItemGroup>
@@ -34,4 +35,4 @@
     </ItemGroup>
   </Target>
 
-</Project>
\ No newline at end of file
+</Project>
```

src/Tizen.NUI/src/internal/Application.cs
huiyueun: Fixed.

PR #2130, opened 2020-10-27
[NUI] Fix Dispose warning error[CA1001]

huiyueun:
CA100: Types that own disposable fields should be disposable
https://docs.microsoft.com/ko-kr/dotnet/fundamentals/code-analysis/quality-rules/ca1001

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
```diff
@@ -30,20 +30,16 @@ namespace Tizen.NUI.Accessibility
...
         {
             dummy = new View();
             dummy.Name = "dali-atspi-singleton";
         }
-        /// <summary>
-        /// destructor. This is HiddenAPI. recommended not to use in public.
-        /// </summary>
-        ~Accessibility()
-        {
-        }
```
WonyoungChoi, line 46: Need destructor in dispose pattern
```csharp
~Accessibility() {
  Dispose(false);
}
```

huiyueun: I've fixed patch.

WonyoungChoi:
introducing new base class does not make ABI break. you can ignore the `ACR Required` label in this case.

dongsug-song (approved):
looks good to me

PR #2140, opened 2020-10-29
[NUI] Add AttributeUsageAttribute for ScriptableProperty(CA1018)

dasg34:
### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/CustomViewRegistry.cs
```diff
@@ -72,6 +72,7 @@ namespace Tizen.NUI
     ///
     ///
     /// <since_tizen> 3 </since_tizen>
+    [AttributeUsage(AttributeTargets.Property)]
```
dongsug-song, line 75: I am not sure that by this change, another warning has been shown as like as below? 

`src/Tizen.NUI/src/public/CustomViewRegistry.cs#L76
Rename Tizen.NUI.ScriptableProperty to end in 'Attribute'.`

it asks to rename the class name as it should have the "Attribute" suffix.
is this another type of warning? or is this related to this?

dasg34: CA1710(`Rename Tizen.NUI.ScriptableProperty to end in 'Attribute'.`)  is not related with this patch.
This patch is only for fixing CA1018(`Specify AttributeUsage on ScriptableProperty.`).

dasg34:
ping~

dongsug-song (approved):
looks good to me

PR #2142, opened 2020-10-29
[NUI] Make Selector.StateValueList public temporarily

dasg34:
This will fix DA-NUI build error.
Please revert this patch after the DA build error is fixed.

### API Changes ###

 - ACR:


rabbitfor (approved):
LGTM

PR #2154, opened 2020-11-02
[NUI] Fix ScrollableBase Size Issue

bshsqa:
- Fix OnMeasure callback in LinearLayout.(handling margin and padding)
 - Fix to handle only layout of ContentContainer to measure and layout of scrollableBase
 - Fix LayoutGroup to do not assert when the child view has no layout in RemoveChildFromLayoutGroup method


dongsug-song (approved):
looks good to me

PR #2158, opened 2020-11-02
Make sure scrollable shadow is shown only when scrolled vertically

huayongxu:
When the scrollable is scrolled horizontally, a background image
is shown accidentally.
This patch adds a check that scrollable shadow is shown only when
scrolled vertically, and adds a test case for ScrollableBase.

### API Changes ###

 - ACR:


huayongxu:
the branch is not correct. so I close this PR.

AchoWang (approved):
Looks good to me.

dongsug-song (approved):
looks good to me

AdunFang (commented):
It is OK.

PR #2170, opened 2020-11-04
[NUI] Add try-catch block in New function (#2119)

huiyueun:
- VD Kona : DF200922-00146
- Anlayze the log when the problem is reproduced,
 after inserting the try-catch block with log.

Co-authored-by: dongsug-song <35130733+dongsug-song@users.noreply.github.com>

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #2185, opened 2020-11-05
[NUI] Add ControlStateTypeConverter for xaml

dasg34:
Custom ConstrolState is now available in xaml.

 ### Sample
XamlPage.xaml
```xaml
<View x:Class="NUIXamlTemplate1.XamlPage"
      xmlns:c="clr-namespace:Tizen.NUI.Components;assembly=Tizen.NUI.Components"
      xmlns:local="clr-namespace:NUIXamlTemplate1">
    <local:MyButton x:Name="MyButton">
        <x:Arguments>
            <c:ButtonStyle Size="100, 100">
                <c:ButtonStyle.BackgroundColor>
                    <Selector x:TypeArguments="Color" Normal="1, 1, 0, 1" Pressed="0.77, 0.88, 1, 1" Disabled="0.88, 0.88, 0.88, 1">
                        <SelectorItem x:TypeArguments="Color" State="MyState" Value="0,0,1,1" />
                        <SelectorItem x:TypeArguments="Color" State="MyState,Focused" Value="0,1,1,1" />
                    </Selector>
                </c:ButtonStyle.BackgroundColor>
            </c:ButtonStyle>
        </x:Arguments>
    </local:MyButton>
</View>
```

MyButton.cs
```cs
public class MyButton : Button
{
    public MyButton(ButtonStyle buttonStyle) : base(buttonStyle) {}
    public void SetControlState(ControlState s) => ControlState = s;
}
```
Main.cs
```cs
protected override void OnCreate()
{
    base.OnCreate();
    Window.Instance.BackgroundColor = new Color(227 / 255f, 255 / 255f, 227 / 255f, 1.0f);
    Window.Instance.KeyEvent += OnKeyEvent;

    View root = new View();
    root.WidthSpecification = LayoutParamPolicies.MatchParent;
    root.HeightSpecification = LayoutParamPolicies.MatchParent;
    root.Layout = new AbsoluteLayout();
    Window.Instance.GetDefaultLayer().Add(root);

    XamlPage xamlPage = new XamlPage();
    root.Add(xamlPage);

    ControlState MyState = ControlState.Create("MyState");

    Button button = new Button() { Position2D = new Position2D(100, 100) };
    bool flag = false;
    button.Clicked += (object sender, ClickedEventArgs e) =>
    {
        if (!flag)
        {
            xamlPage.MyButton.SetControlState(MyState);
            flag = true;
        }
        else
        {
            xamlPage.MyButton.SetControlState(ControlState.Focused + MyState);
            flag = false;
        }
    };
    root.Add(button);
}
```

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -292,76 +288,26 @@ public override string ToString()
...
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        public override bool Equals(object obj)
-        {
-            if (!(obj is StateValuePair<T>))
-                return false;
+                    ControlState convertedState = new ControlState();
+                    string[] parts = value.Split(',');
+                    foreach (string part in parts)
+                    {
+                        convertedState += Create(part);
```
rabbitfor, line 304: How about doing `Trim` in `Create(string)`?
I think users may think `Selected,Disabled` and `Selected, Disabled` are same.

src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
```diff
@@ -442,4 +441,64 @@ private void OnViewControlState(object obj, View.ControlStateChangedEventArgs co
...
+        {
+            State = state;
+            Value = value;
+        }
+
+        /// <summary>
+        /// The state
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public ControlState State { get; set; }
```
rabbitfor, line 473: How about just providing getter. The setter seems not necessary.

src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
```diff
@@ -442,4 +441,64 @@ private void OnViewControlState(object obj, View.ControlStateChangedEventArgs co
...
+
+        /// <summary>
+        /// The value associated with state.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public T Value { get; set; }
```
rabbitfor, line 479: Ditto.

src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
dasg34: If setter is removed, you can't set `State` in xaml. parser will throw exception.

src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
dasg34: If setter is removed, you can't set `Value` in xaml. parser will throw exception.

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
dasg34: I added `Trim` in `Create(string)`.
Thank you :)

src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
rabbitfor: I see :)
Thank you for the explanation.
I'll approve this code but please think about the issue below.

What I was worrying about is that user may change value in SelectorItem after it applied to the View.
```
var item = new SelectorItem<Color>(ControlState.Pressed, Color.Black);
var colorSelector = new Selector<Color>();
colorSelector.Add(item);

var style = new ViewStyle() { BackgroundColor = colorSelector };

....

item.Value = Color.Red;
```
It is not likely to happen, but I want to avoid these complicate situation as possible.
Maybe providing setter as an internal API is the solution but current xaml parser won't allow internal API of the class that is not a root element. (I think it is a bug)

src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
rabbitfor: ditto

rabbitfor:
LGTM

rabbitfor (approved):
LGTM

dongsug-song:
@dasg34 
Hello, could you resolve the conflict?
then I will merge this.
thank you.

PR #2196, opened 2020-11-06
[NUI] Add ExcludeLayouting to View

bshsqa:
- If ExcludeLayouting property is false, the view is not affected by parent layout
 - Even the property is false, the children Views are affected by layout of current View
 - This ExcludeLayouting property is changed synchronously with SetPositionByLayout of its layout
 - Children of AbsoluteLayout can freely be animated.

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -1,5 +1,5 @@
 /*
- * Copyright(c) 2019 Samsung Electronics Co., Ltd.
+ * Copyright(c) 2020 Samsung Electronics Co., Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -205,6 +206,24 @@ protected set
             }
         }
 
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool TransformedByLayout
```
huiyueun, line 210: Please leave a description comment in the case of hidden API.

bshsqa:
Currently on changing property name.
Please do not merge this.

src/Tizen.NUI/src/public/BaseComponents/View.cs
bshsqa: done. thank you

AdunFang (approved):
It is OK.

hyunjushin (approved):

PR #2199, opened 2020-11-09
[NUI] change orientation angle in GLWindow

sparrow74:
Change orientation angle in GLWindow. It is same with NUI.Window.

### Description of Change ###
GLWindow has window orientation angle enum type. It is different to NUI.Window.
This patch is for convienent both GLWindow and NUI.Window. 
In addition, To be adjusted buffer commit by GLWindow application, the renderFrame callback function is changed.

This interface is not public, so, need not care for ACR.


dongsug-song (approved):
looks good to me

PR #2206, opened 2020-11-09
[NUI] Introduce CornerRadiusPolicy to View

rabbitfor:
The CornerRadiusPolicy indicates whether the CornerRadius value is relative or absolute.
The default policy is absolute, which means the CornerRadius is an absolute float length.
If the policy is relative, the CornerRadius can have percentage value between 0.0f to 1.0f,
that is relative to the smaller of view's width and height.

### API Changes ###

 - ACR:


dongsug-song:
@rabbitfor 
Hello, 
could you resolve conflicts? 
then I will merge.
thank you.

Seoyeon2Kim (approved):
Looks good to me, but someone else must approve this PR.

PR #2209, opened 2020-11-10
Revert "[NUI] change orientation angle in GLWindow"

sparrow74:
### Description of Change ###
Revert "[NUI] change orientation angle in GLWindow"
Because this patch should be only submitted to API9.

### API Changes ###
It is only internal APIs


dongsug-song (approved):
this is for tizen_6.5, so reverted now.

PR #2215, opened 2020-11-10
[NUI] fix ThemeResourceSample build error

dasg34:
This patch will fix build error caused by conflict between global::System.XXX
and Tizen.System.XXX namespace.

### API Changes ###

 - ACR:


rabbitfor (approved):
LGTM

PR #2234, opened 2020-11-13
[NUI] Fix build warning CA1062

Seoyeon2Kim:

### Description of Change ###
- Need to verify whether the argument is null
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062

### API Changes ###
- N/A

### DIFF
--- a/src/Tizen.NUI/src/internal/ItemContainer.cs
+++ b/src/Tizen.NUI/src/internal/ItemContainer.cs
@@ -197,7 +197,7 @@ namespace Tizen.NUI
                 collectionRef = collection;
                 currentIndex = -1;
                 currentObject = null;
-                currentSize = collectionRef.Count;
+                currentSize = collectionRef?.Count ?? 0;
             }
 
             // Type-safe iterator Current
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -1684,6 +1684,7 @@ namespace Tizen.NUI.BaseComponents
         /// <summary>
         /// Gets or sets the minimum size the view can be assigned in size negotiation.
         /// </summary>
+        /// <exception cref="ArgumentNullException"> Thrown when value is null. </exception>
         /// <remarks>
         /// The property cascade chaining set is possible. For example, this (view.MinimumSize.Width = 1;) is possible.
         /// </remarks>
@@ -1697,7 +1698,11 @@ namespace Tizen.NUI.BaseComponents
             }
             set
             {
-                if (_layout != null && value != null)
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(value));
+                }
+                if (_layout != null)
                 {
                     // Note: it only works if minimum size is >= than natural size.
                     // To force the size it should be done through the width&height spec or Size2D.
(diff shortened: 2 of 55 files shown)


src/Tizen.NUI/src/internal/ItemContainer.cs
```diff
@@ -197,7 +197,14 @@ public ItemContainerEnumerator(ItemContainer collection)
...
                 currentObject = null;
-                currentSize = collectionRef.Count;
+                if (collectionRef != null)
+                {
+                    currentSize = collectionRef.Count;
+                }
+                else
+                {
+                    currentSize = 0;
+                }
```
huiyueun, line 207: How about using a '??' ?
currentSize = collectionRef?.Count ?? 0;

src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -2032,9 +2032,12 @@ public LayoutTransition LayoutTransition
                 {
                     _layoutTransitions = new Dictionary<TransitionCondition, TransitionList>();
                 }
-                LayoutTransitionsHelper.AddTransitionForCondition(_layoutTransitions,value.Condition,value, true);
+                if (value != null)
+                {
+                    LayoutTransitionsHelper.AddTransitionForCondition(_layoutTransitions,value.Condition,value, true);
```
huiyueun, line 2037: follow C# code foramt. (Ctrl + K,F in VS)

src/Tizen.NUI/src/public/GLWindow.cs
```diff
@@ -120,13 +120,16 @@ public Size2D WindowSize
...
+                if (value != null)
+                {
+                    global::System.IntPtr intPtr = Interop.GLWindow.GlWindow_GetPositionSize(swigCPtr);
+                    Rectangle val = new Rectangle(intPtr, true);
+                    Rectangle ret = new Rectangle(val.X, val.Y, value.Width, value.Height);
+                    val.Dispose();
+
+                    Interop.GLWindow.GlWindow_SetPositionSize(swigCPtr, Rectangle.getCPtr(ret));
+                    ret.Dispose();
+                }
```
huiyueun, line 132: If the value is null, do not we need a log message or an exception?

src/Tizen.NUI/src/public/KeyValue.cs
```diff
@@ -77,22 +77,25 @@ public object OriginalKey
             }
             set
             {
-                _originalKey = value;
-                if (value is int || value is Int32)
+                if (value != null)
```
huiyueun, line 80: If the value is null, do not we need a log message or an exception?

src/Tizen.NUI/src/public/XamlBinding/ContentPage.cs
```diff
@@ -109,7 +109,14 @@ public ContentPage(Window win)
...
             Root.HeightResizePolicy = ResizePolicyType.FillToParent;
 
-            win.Add(Root);
+            if (win != null)
+            {
+                win.Add(Root);
+            }
+            else
+            {
+                NUIApplication.GetDefaultWindow().Add(Root);
```
huiyueun, line 118: If the parameter window is null, would not it be better not to act than to attach it to the main window?

Seoyeon2Kim: Accepted your opinion! :)

Seoyeon2Kim (commented):
I updated the files as Huiyu's comments.
But, I'm wondering that it would be better to change null check conditionals to the similar ways.

AdunFang (approved):
It is OK.

huiyueun (approved):

bshsqa (approved):
ImageView, View, Capture checked.

PR #2246, opened 2020-11-17
[NUI] Add Animation Finished callback debug log temporarily

dongsug-song:
### Description of Change ###
[NUI] Add Animation Finished callback debug log temporarily

- To do debugging for the rare case when Finished event handler is not called.

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/public/Animation.cs
+++ b/src/Tizen.NUI/src/public/Animation.cs
@@ -1318,11 +1318,14 @@ namespace Tizen.NUI
 
         private void OnFinished(IntPtr data)
         {
+            Tizen.Log.Error("NUI", $"[TEMP]OnFinished() START");
             if (_animationFinishedEventHandler != null)
             {
                 //here we send all data to user event handlers
                 _animationFinishedEventHandler(this, null);
+                Tizen.Log.Error("NUI", $"[TEMP]OnFinished() handler is invoked! should be shown!");
             }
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Animation.cs
```diff
@@ -1318,11 +1318,14 @@ protected override void ReleaseSwigCPtr(System.Runtime.InteropServices.HandleRef
...
         private void OnFinished(IntPtr data)
         {
+            Tizen.Log.Error("NUI", $"[TEMP]OnFinished() START");
             if (_animationFinishedEventHandler != null)
             {
                 //here we send all data to user event handlers
                 _animationFinishedEventHandler(this, null);
+                Tizen.Log.Error("NUI", $"[TEMP]OnFinished() handler is invoked! should be shown!");
             }
+            Tizen.Log.Error("NUI", $"[TEMP]OnFinished() END");
```
huiyueun, line 1328: I am worried about performance when this patch is merged. The log is output in all animations. Even without using the Finish callback.

dongsug-song: @huiyueun 
Yes, You are absolutely right, so this is temporary patch. 
I will remove it when the defect is reproduced and resolved.
if the defect is not reproduced, I will back out this patch by 4th of December.
thank you.

huiyueun (approved):

PR #2250, opened 2020-11-18
[NUI] Unsubscribe theme changed event when the view is not on window

rabbitfor:

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #2252, opened 2020-11-18
[NUI] fix CA1801 build warnings

id213sin:
### Description of Change ###

fix CA1801 build warnings except auto-generated code by SWIG.

### API Changes ###

 - ACR:


src/Tizen.NUI/src/internal/XamlBinding/Device.cs
```diff
@@ -111,10 +111,10 @@ public static T OnPlatform<T>(T iOS, T Android, T WinPhone)
             return iOS;
         }
 
-        public static void OpenUri(Uri uri)
-        {
-            // PlatformServices?.OpenUriAction(uri);
-        }
+        // public static void OpenUri(Uri uri)
```
huiyueun, line 114: What do you think about deleting APIs that are not used?

id213sin: I've commented out this code. Because, it seems that a declaration of "OpenUriAction(Uri uri)" in "IPlatformServices" and the usage of the method ("PlatformServices?.OpenUriAction(uri);") had been commented out in the same time.
(I couldn't find exact history about OpenUriAction and OpenUri.)
I thought a author of these comments wanted to keep the history of OpenUriAction.
So, I've commented out it to keep its history.

If you still think it is better to remove, I'll remove it and update the patch.

huiyueun (approved):

id213sin:
I'll re-upload the patch and re-create pull request for removing merge commits.

id213sin:
Please, review https://github.com/Samsung/TizenFX/pull/2258 instead of this patch.

PR #2265, opened 2020-11-20
Revert "[NUI] Unsubscribe theme changed event when the view is not on…

rabbitfor:
… window"

This revert previous patch and make sure the theme changed event handler is disconnected when dispose.

This reverts commit 592900fdc380ce27df8c8fdb6de8084caffacc15.

### API Changes ###

 - ACR:


dongsug-song (approved):
Looks good to me

PR #2268, opened 2020-11-20
[NUI] Use weak reference for theme changed event

rabbitfor:

### Description of Change ###
Previous theme changed event was holding reference of the view for the theme manager's lifetime
unless the view is disposed by calling Dispose() explicitly.
Since the theme manager is a static class, so there might be a change that the view can't get free by GC.
To solve this problem, this patch apply a weak event that uses weak reference for the handler
so that the views are no longer be caught by theme manager.

### API Changes ###

 - ACR:


src/Tizen.NUI/src/internal/WeakEvent.cs
```diff
@@ -0,0 +1,91 @@
...
+                {
+                    item.Invoke(sender, args);
+                    continue;
+                }
+                handlers.Remove(item);
+            }
+        }
+
+        internal class WeakHandler<T>
+        {
```
huiyueun, line 53: I think warning of new file should be removed.

huiyueun:
Please add detailed comment about this patch.

src/Tizen.NUI/src/internal/WeakEvent.cs
rabbitfor: Done :)

rabbitfor:
> Please add detailed comment about this patch.
Done :)

dongsug-song (approved):
(b^_^)b
I think this would be very good if this is applied previous existing events.
do you think this is possible?
almost all events are connected to native dali callbacks and I am not sure this Weak-Reference-Event is feasible for native binding events.
could you figure this out?
thank you.

dongsug-song:
@rabbitfor 
Hello, could you check my comment above? ^^;
thank you.

PR #2272, opened 2020-11-23
[NUI] remove unnecessary code that flex margin is assigned to flexlayout

dasg34:
Sicne the margin of FlexLayout will be handled by the parent layout,
we don't need to re-assign margin in `OnMeasure`.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
Approved this PR because margin variable of FlexLayout is not used.

Seoyeon2Kim:
After updating the branch, it can be merged :)

PR #2276, opened 2020-11-23
[NUI] separate layouting logic of the independent children

dasg34:
This patch introduces `IterateLayoutChildren()` method that return
an enumerable collection of the child layouts that affected by parent
layout.

Independent children will be measured in `OnMeasureIndependentChildren()`.

By using `IterateLayoutChildren()`, we can remove the codes that check
whether child layout exclude from layout flow in `OnMeasure`. and api
users don't need to check `View.ExcludeLayouting` property in custom layouts.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/AbsoluteLayout.cs
+++ b/src/Tizen.NUI/src/public/Layouting/AbsoluteLayout.cs
@@ -46,39 +46,30 @@ namespace Tizen.NUI
             MeasuredSize.StateType childWidthState = MeasuredSize.StateType.MeasuredSizeOK;
             MeasuredSize.StateType childHeightState = MeasuredSize.StateType.MeasuredSizeOK;
 
-            for (int i = 0; i < LayoutChildren.Count; i++)
+            foreach (LayoutItem childLayout in IterateLayoutChildren())
             {
-                LayoutItem childLayout = LayoutChildren[i];
-                if (childLayout != null)
+                // Get size of child with no padding, no margin. we won't support margin, padding for AbsolutLayout.
+                MeasureChildWithoutPadding(childLayout, widthMeasureSpec, heightMeasureSpec);
+
+                // Determine the width and height needed by the children using their given position and size.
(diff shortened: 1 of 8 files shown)


Seoyeon2Kim:
I think this PR really needs to get Seungho's review because the inheritance of `ScrollableBase` would be changed. (`LayoutGroup` -> `AbsoluteLayout`)
@bshsqa  Could you check this PR please?

bshsqa:
I think this patch looks good to go.
But, currently, we have some issue for scrollableBase and excludeLayouting.
So, If you OK, could we merge this patch after I fix the issue?
Because it can be difficult to find the way to reproduce the issue after the code is changed.

bshsqa (approved):
This patch looks good to go.
If well tested, this is good to be merged.

src/Tizen.NUI/src/public/Layouting/AbsoluteLayout.cs
```diff
@@ -99,21 +90,17 @@ protected override void OnLayout(bool changed, LayoutLength left, LayoutLength t
...
+                Position2D childPosition = childLayout.Owner.Position2D;
 
-                    LayoutLength childLeft = new LayoutLength(childPosition.X);
-                    LayoutLength childTop = new LayoutLength(childPosition.Y);
+                LayoutLength childLeft = new LayoutLength(childPosition.X);
+                LayoutLength childTop = new LayoutLength(childPosition.Y);
 
-                    childLayout.Layout(childLeft, childTop, childLeft + childWidth, childTop + childHeight, true);
-                }
+                childLayout.Layout(childLeft, childTop, childLeft + childWidth, childTop + childHeight);
```
bshsqa, line 103: We suppose that every child of AbsoluteLayout is independent. So, this needs true at the end of the Layout method.

bshsqa:
Thanks. I think it is good to be merged

bshsqa (approved):
Thanks. I think it is good to be merged

PR #2282, opened 2020-11-24
[NUI] Change setter of property

huiyueun:
### Description of Change ###
- Setter is deprecated in API8.(Size/Scale/Color/Position)
- An unnecessary log comes out and causes performance degradation.

### API Changes ###
none


dasg34 (approved):
LGTM d(^_^d)

src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyleBindableProperty.cs
```diff
@@ -346,9 +345,8 @@ public partial class ViewStyle
                 if (viewStyle.size == null)
                 {
                     if ((float)newValue == 0) return;
-                    viewStyle.size = new Size();
                 }
-                viewStyle.size.Height = (float)newValue;
+                viewStyle.size = new Size(viewStyle.size?.Width ?? 0, (float)newValue);
```
dongsug-song, line 349: very good to use ?? operator
(https://docs.microsoft.com/ko-kr/dotnet/csharp/language-reference/operators/null-coalescing-operator)

src/Tizen.NUI.Wearable/src/public/CircularPagination.cs
```diff
@@ -487,6 +487,10 @@ public Position GetIndicatorPosition(int index)
         [EditorBrowsable(EditorBrowsableState.Never)]
         public virtual void SetIndicatorPosition(int index, Position position)
         {
+            if (position == null)
```
dongsug-song, line 490: could you add \<exception cref= ...\> here?
(example) https://github.com/Samsung/TizenFX/blob/d8c72a9533502bd3382ec1e1db12e57614c2a710/src/Tizen.NUI/src/public/Window.cs#L1414

src/Tizen.NUI.Components/Controls/Tab.cs
```diff
@@ -554,8 +554,8 @@ protected virtual void LayoutChild()
                     for (int i = 0; i < totalNum; i++)
                     {
                         preW = (itemList[i].NaturalSize2D != null ? itemList[i].NaturalSize2D.Width : 0);
-                        itemList[i].Position2D.X = w - preW - preX;
-                        itemList[i].Size2D.Width = preW;
+                        itemList[i].PositionX = w - preW - preX;
```
dongsug-song, line 557: I think this would be fine but just for double checking, could you add some TC or sample app to verify whether this works?
I mean set values with Position(xxx, xxx) and check if the PositionX and PositionY has same values and vice versa.

huiyueun: Modify ViewSizeWidthPropertySetTest sample

src/Tizen.NUI.Wearable/src/public/CircularPagination.cs
huiyueun: done

src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyleBindableProperty.cs
huiyueun: Thank you.

PR #2298, opened 2020-11-27
[NUI] Change setter of property

huiyueun:
### Description of Change ###
Setter is deprecated in API8.(Size/Scale/Color/Position)
An unnecessary log comes out and causes performance degradation.

### API Changes ###

 - ACR:


dongsug-song (approved):
This is already reviewed in DevelNUI~

PR #2300, opened 2020-11-27
[NUI] Add direction & displacement into ScrollOutOfBound event.

huayongxu:
### Description of Change ###
This patch is to trigger ScrollOutOfBound event as pan gesture state,
and add pan direction & displacement into ScrollOutOfBound event.
Bound of ScrollOutOfBoundEventArgs is removed, because displacement
can be used for checking Top/Bottom.

### API Changes ###
none


zhouleonlei (approved):
LGTM

AdunFang (commented):
It is OK.

PR #2301, opened 2020-11-27
[NUI] Delete unnecessary log

huiyueun:
### Description of Change ###
- Delete the log commented
- Displays only one log when the value is invalid of color.
- View.Parent is deprecated, modify to use View.GetParent()

### API Changes ###
none


dongsug-song (approved):
looks good to me

PR #2305, opened 2020-11-30
[NUI] Enable/Disable overshooting effect for ScrollableBase.

huayongxu:
A property EnableOverShootingEffect is added. default is false.
And shadow would be raised to top when effect appears.

### API Changes ###

 - ACR:


dongsug-song (approved):
this is reviewd in DevelNUI

AdunFang (commented):
It is OK.

PR #2310, opened 2020-12-01
[NUI] Add Window extension method GetDefaultNavigator

jaehyun0cho:
To provide default navigator of window, Window extension method
GetDefaultNavigator is added.

### API Changes ###

 - ACR:


AdunFang (approved):
It is OK.

PR #2314, opened 2020-12-02
[NUI] Fix a build warning CA1827

Seoyeon2Kim:

### Description of Change ###
- To fix the build warning, replaced `Count()` method with `Any()` method
 https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1827

### API Changes ###
- N/A

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
@@ -285,13 +285,14 @@ namespace Tizen.NUI.BaseComponents
             }
 
             var rest = lhs.stateList.Except(rhs.stateList);
+            var count = rest.Count();
 
-            if (rest.Count() == 0)
+            if (count == 0)
             {
                 return Normal;
             }
 
-            if (rest.Count() == 1)
+            if (count == 1)
             {
                 return rest.First();
             }


src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -286,7 +286,7 @@ public override string ToString()
 
             var rest = lhs.stateList.Except(rhs.stateList);
 
-            if (rest.Count() == 0)
+            if (rest.Any())
```
dasg34, line 289: You forgot `!`.
```suggestion
            if (!rest.Any())
```

BTW, what do you think of using count local variable?
Since `rest.Count()` is used in the line below, we can improve performance by using count local variable.
(IEnumerable<T>.Count() is O(n) Operation.(https://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,41ef9e39e54d0d0b))

dasg34 (changes requested):

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
Seoyeon2Kim: Thank you for your suggestion! I like it! and accept your kind guide. :)

dasg34 (approved):
Thank you~

Seoyeon2Kim:
This PR is changed to use `Count()` method once for a local variable.
Then, we do not have to be worried about CA1827.

To fix the build warning, it could replace `Count()` method with `Any()` method
  https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1827
But, to improve performance as the review of dasg32, use `Count()` itself.

PR #2315, opened 2020-12-02
[NUI] fix CA1801 build warnings from auto-generated codes by SWIG

id213sin:
### Description of Change ###
The 'futureUse' parameter is never being used through the whole code.
It removes 'futureUse' parameter from functions.


Seoyeon2Kim (approved):
Looks really good to me :)
Thanks for creating a new PR even though there're tons of files!

PR #2317, opened 2020-12-03
[NUI] fix exception when Theme.Resource is used in xaml file

dasg34:
Since local value `xamlFile` is set after `LoadFromXaml()`,
`xamlFile` is always null in `Reload()`.

To fix this exception, it sets `xamlFile` before `XamlLoader.Load()`.
Test code : 
```xaml
<Theme
  Resource="MyThemeResource.xaml"
  Id="MyTheme"
  BasedOn="Tizen.NUI.Theme.Common">
```
### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #2323, opened 2020-12-03
[NUI] apply theme resource for components

dasg34:
This patch will work with #2317

### API Changes ###

 - ACR:


dasg34:
Please see https://github.com/Samsung/TizenFX/pull/2405 instead.

PR #2328, opened 2020-12-03
[NUI] Add Dialog and AlertDialog classes

jaehyun0cho:
### Description of Change ###

Dialog is used to show a popup content with background scrim.
Navigator provides static method ShowDialog which pushes Page containing
Dialog.

AlertDialog is used to show a dialog with title, message and action
buttons.
Navigator provides static method ShowAlertDialog which pushes Page
containing AlertDialog.

### API Changes ###

 - ACR:


AdunFang (approved):
It is OK.

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -335,5 +335,94 @@ public static Navigator GetDefaultNavigator(Window window)
...
+
+        /// <summary>
+        /// Shows a dialog by pushing a page containing dialog to default navigator.
+        /// </summary>
+        /// <param name="content">The content of Dialog.</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static void ShowDialog(View content = null)
+        {
+            var window = NUIApplication.GetDefaultWindow();
+            var defaultNavigator = window.GetDefaultNavigator();
```
rabbitfor, line 347: I have question that is not directly related with this patch.
Does the method name `GetDefaultNavigator` means there are multiple Navigator in a window?

dongsug-song:
@jaehyun0cho 
Hello,
could you check the @rabbitfor 's reveiw?
thank you.

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
jaehyun0cho: Yes, it is possible if a user creates multiple Navigator instances in a window.
e.g.
```

var navigator = new Navigator();
window.Add(navigator);

var navigator2 = new Navigator();
window.Add(navigator2);
```

If window.GetDefaultNavigator() is called for the first time, then Navigator instance is created and stored in Dictionary with the window.
If window.GetDefaultNavigator() is called again, then the stored Navigator instance is returned.

Seoyeon2Kim (approved):
Looks good to me, but other's approval is needed.

dongsug-song:
@jaehyun0cho 
Hello,
could you check the @rabbitfor 's reveiw?
thank you.

PR #2335, opened 2020-12-07
[NUI] Modify popup dismiss bug

huiyueun:
when dismiss popup, the shadow does not disappear.
-An error occurs after changing the parent to GetParent ().
-layer of converting View is null.

### API Changes ###

 - ACR:


rabbitfor (approved):
LGTM

PR #2353, opened 2020-12-09
[NUI] Fix CA2214

rabbitfor:

### Description of Change ###
CA2214
Do not call overridable methods in constructors.

* View.cs
  * In the current design, it is hard to remove virtual method call in View constructor.
  * So the patch moved method calling to the newly defined private method.
* ControlStyle.cs
  * Removed unnecessary constructors.
* TextFieldStyle.cs
  * Removed unnecessary constructors.
* ImageViewStyle.cs
  * Removed unnecessary constructors.
* TextLabelStyle.cs
  * Removed unnecessary constructors.

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #2354, opened 2020-12-09
[NUI] Change clipping mode of ScrollableBase

bshsqa:
- ClipChildren clipping mode uses stencil buffer.
 - But, scrollableBase is always pretended as rectangular shape so we can change it to ClipToBoundingBox that uses scissor.
 - Usually, Scissor is better than stencil buffer for the performance. 


huiyueun:
Please comment the brief information about the patch.

Seoyeon2Kim (commented):
If there's a brief description about this PR, we can review yours better :)

huiyueun (approved):

Seoyeon2Kim (approved):
Approved this PR :)  for the better performance!

PR #2355, opened 2020-12-09
[NUI] Fix CA1069

rabbitfor:
CA1069
Enums should not have duplicate values.

This patch changed ProperRanges to struct from enum.

### API Changes ###

 - ACR:


rabbitfor:
Need to check VD build.

dongsug-song (approved):
looks good to me

PR #2359, opened 2020-12-09
[NUI] Fix CA1052

rabbitfor:

### Description of Change ###
    
CA1052
Static holder types should be Static or NotInheritable.

This patch fixes classes that holds only static members to static or sealed.
If the class is obsolete, added suppress message attribute to hide warnings.

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #2362, opened 2020-12-09
[NUI] Fix CS0108

dongsug-song:
### Description of Change ###
[NUI] Fix CS0108

https://docs.microsoft.com/ko-kr/dotnet/csharp/language-reference/compiler-messages/cs0108

'member1' hides inherited member 'member2'. Use the new keyword if hiding was intended.
A variable was declared with the same name as a variable in a base class. However, the new keyword was not used. This warning informs you that you should use new; the variable is declared as if new had been used in the declaration.

### API Changes ###
https://code.sec.samsung.net/jira/browse/TCSACR-394

### DIFF
--- a/src/Tizen.NUI/src/public/InputMethodContext.cs
+++ b/src/Tizen.NUI/src/public/InputMethodContext.cs
@@ -944,6 +944,8 @@ namespace Tizen.NUI
             /// The state if it owns memory
             /// </summary>
             /// <since_tizen> 5 </since_tizen>
+            //ToDo : raise ACR as [Obsolete("This has been deprecated in API9 and will be removed in API11. Please use swigCMemOwn which is declared in parent class")]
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/InputMethodContext.cs
```diff
@@ -1102,6 +1104,8 @@ public class CallbackData : Disposable
             /// The state if it owns memory
             /// </summary>
             /// <since_tizen> 5 </since_tizen>
+            [Obsolete("This has been deprecated in API9 and will be removed in API11. Please use SwigCMemOwn which is declared in parent class")]
```
hyunjushin, line 1107: Please comment out this change on DevelNUI branch as ACR is required on Master branch.

rabbitfor (approved):
LGTM

src/Tizen.NUI/src/public/InputMethodContext.cs
dongsug-song: Yes. you are right. 
it has been commented out.
thank you.

PR #2363, opened 2020-12-09
[NUI] Fix CA1044

dongsug-song:
### Description of Change ###
[NUI] Fix CA1044
- https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1044
- Get accessors provide read access to a property and set accessors provide write access. Although it is acceptable and often necessary to have a read-only property, the design guidelines prohibit the use of write-only properties. This is because letting a user set a value and then preventing the user from viewing the value does not provide any security. Also, without read access, the state of shared objects cannot be viewed, which limits their usefulness.

### API Changes ###
if ACR is needed, I will raise new ACR request soon.


dongsug-song:
@AdunFang 
Hello, 
could you check this API => public AnimationBehavior[] Behaviors ?
this seems to be for XAML animation and doesn't look like being used anywhere in NUI.
I changed this property as mothod.
could you please confirm it's OK if this is removed?
thank you.

huiyueun:
@dongsug-song 
AnimationBehavior[] Behaviors is removed by this patch. 
https://github.com/Samsung/TizenFX/pull/2409

huiyueun (approved):

PR #2365, opened 2020-12-10
[NUI] Fix CA1816

dongsug-song:
### Description of Change ###

[NUI] Fix CA1816
- https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1816?view=vs-2019
- The IDisposable.Dispose method lets users release resources at any time before the object becoming available for garbage collection. If the IDisposable.Dispose method is called, it frees resources of the object. This makes finalization unnecessary. IDisposable.Dispose should call GC.SuppressFinalize so the garbage collector doesn't call the finalizer of the object. To prevent derived types with finalizers from having to reimplement IDisposable and to call it, unsealed types without finalizers should still call GC.SuppressFinalize.

### API Changes ###
if ACR is needed, I will raise new ACR request soon.


huiyueun (approved):

dongsug-song:
this is already merged in other patch.
so close now

PR #2366, opened 2020-12-10
[NUI] Fix build warning CA2007

bshsqa:
- https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2007
 - CA2007: Do not directly await a Task
 - When an asynchronous method awaits a Task directly, continuation usually occurs in the same thread that created the task, depending on the async context. This behavior can be costly in terms of performance and can result in a deadlock on the UI thread. Consider calling Task.ConfigureAwait(Boolean) to signal your intention for continuation.


dongsug-song (approved):
looks good to me

PR #2369, opened 2020-12-10
[NUI] Change property name frome Url to Uri

bshsqa:
We may better to use name Uri than Url, when we use System.Uri return type.


dongsug-song (approved):
looks good to me

PR #2372, opened 2020-12-10
[NUI] Prevent unnecessary Layout call, make independent if it has true ExcludeLayouting and root.

bshsqa:
- Make changed flag in Layout method true when the layout is really changed.
 - If a layout has true value of ExcludeLayouting and it also a root layout, currently it couldn't handled as independent.
 - This patch adds to check ExcludeLayouting Property when it call Layout method.


dasg34 (approved):
Nice catch 👍

PR #2376, opened 2020-12-10
[NUI] Fix build warning CA1720

tscholb:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1720
CA1720: Identifiers should not contain type names

### API Changes ###

 - ACR:


dongsug-song:
@tscholb 
Hello, could you add some brief explanation why the warning is suppressed?
even if the reason is acceptable, we need to remain some history for the future fix.
thank you.

dongsug-song:
@tscholb 
Hello, could you add some brief explanation why the warning is suppressed?
even if the reason is acceptable, we need to remain some history for the future fix.
could you please add reason-explaining comments with // ?
for example, `// reason : need ACR and ACR is impossible because other GBM uses same name.`
could you check again?
thank you.

dongsug-song (approved):
suppress warning reason :  Can't fix because it's already used by other GBM.
looks good to me

PR #2385, opened 2020-12-14
[NUI] Fix CA1066

dkdk-ryu:
CA1066:Implement IEquatable when overriding Equals

### API Changes ###

 - ACR:


Seoyeon2Kim:
Dear @AdunFang  and @bshsqa ,
This PR is related to XmlName and LayoutLength. Could you check them?

dongsug-song:
@m-macias 
Hello,
I have question about TCSACR for derived class's structural change.
could you check the class below changed in this PR?

public struct LayoutLength : IEquatable<LayoutLength>

that is public API and it is changed newly as inheriting IEquatable<T>.
in this case, TCSACR is required even though TizenAPI-bot has checked nothing ?
could you answer for us?
thank you.

huiyueun:
looks good to me

m-macias:
@dongsug-song ,
The addition of the interface changes the public API and would require a new ACR, however in this case those methods (i.e. Equals(...)) were already implemented so there is no real value for the TCSACR process. Additionally, currently those changes are targeting the DevelNUI branch which is not supported by the TCSACR process.

Thank you

dongsug-song (approved):
looks good to me

PR #2390, opened 2020-12-14
[NUI] Fix build warning (CA1714)

sparrow74:
### Description of Change ###

https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1714

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/LayoutTransition.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LayoutTransition.cs
@@ -69,6 +69,57 @@ namespace Tizen.NUI
 
         ChangeOnRemove = 16
     }
+    
+    /// <summary>
+    /// The conditions for transitions.
+    /// </summary>
+    /// <since_tizen> 6.5 </since_tizen>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    [FlagsAttribute]
+    public enum TransitionConditions
+    {
+        /// <summary>
+        /// Default when a condition has not been set.
+        /// </summary>
+        /// <since_tizen> 6.5 </since_tizen>
+       [EditorBrowsable(EditorBrowsableState.Never)]
+        Unspecified = 0,
+
+        /// <summary>
+        /// Animate changing layout to another layout.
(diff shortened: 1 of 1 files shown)


huiyueun:
If you add the changed enum name, it would be better to change it for the existing use case as well.

sparrow74:
@huiyueun 
Thank you for your review. I sent you the mail about this patch.
If this patch is added, we should change more public APIs.

src/Tizen.NUI/src/public/Layouting/LayoutTransition.cs
```diff
@@ -69,6 +69,57 @@ public enum TransitionCondition
 
         ChangeOnRemove = 16
     }
+    
+    /// <summary>
+    /// The conditions for transitions.
+    /// </summary>
+    /// <since_tizen> 6.5 </since_tizen>
```
dongsug-song, line 76: could you remove <since_tizen> tags?
this tag is supposed to be added when "Add ACR" is accepted.
this looks hidden-API, so I don't think versioning tag is needed.
thank you.

src/Tizen.NUI/src/public/Layouting/LayoutTransition.cs
```diff
@@ -69,6 +69,57 @@ public enum TransitionCondition
...
+        /// Animation when an item changes due to a  being added.
+        /// </summary>
+        /// <since_tizen> 6.5 </since_tizen>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        ChangeOnAdd = 8,
+
+        /// <summary>
+        /// Animation when an item changes due to a  being removed.
+        /// </summary>
+        /// <since_tizen> 6.5 </since_tizen>
```
dongsug-song, line 119: same as above

sparrow74:
@dongsug-song 
I think we have one decisions about this patch is alive or not. 
Because if the new enum is added, the public APIs should be changed and need the ACRs.
In addition, this enum and the related public APIs were added in Tizen 6.0. 
So, we will see this build warning

PR #2391, opened 2020-12-14
[NUI] Fix build warning (CA1821)

sparrow74:
### Description of Change ###
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1821


huiyueun:
~DisposeQueue, ~Accessibility -> These are hidden APIs, how about deleting the destructor?

sparrow74:
@huiyueun 
Actually I considered to remove the empty destructors.
If the empty destructors should be removed, i think all they should be removed both public and internal.
I want to avoid the ACR by removing public destructors.

dongsug-song:
@sparrow74 
Hello, 
could you check whether the API-Checker gives error when the empty destructor is removed?
if the API-Checker gives nothing, then we don't need any ACR. 
if the API-Checker gives an API-Break violation then your adding log patch would be OK.
thank you.

sparrow74:
@dongsug-song 
I see, I will try it and share the result.
Thank you :)

hyunjushin (approved):

dongsug-song:
@sparrow74 
Hello, 
sorry but I can not merge your PR. 
image

"Rebase and merge" button has been inactivated.
could you try to merge in your side? if the button is activated, please try to merge.
otherwise, could you try to rebase and do force-push with your git console?
thank you.

sparrow74:
Rebase and merge function is not activated. 
I submit new patch instead of this.
https://github.com/Samsung/TizenFX/pull/2433

PR #2398, opened 2020-12-15
[NUI] Rename EventHandler WithReturnType to end in EventHandler

dasg34:
Fixes CA1710

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/DaliEventHandler.cs
+++ b/src/Tizen.NUI/src/internal/DaliEventHandler.cs
@@ -15,6 +15,7 @@
  *
  */
 using System;
+using System.ComponentModel;
 using System.Runtime.InteropServices;
 
 namespace Tizen.NUI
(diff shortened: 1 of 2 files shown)


huiyueun (approved):

src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
```diff
@@ -231,7 +231,7 @@ private void SendViewAddedEventToWindow(IntPtr data)
         /// If it returns false, the child can receive the touch event. If it returns true, the touch event is intercepted. So child cannot receive touch event.<br />
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public event EventHandlerWithReturnType<object, TouchEventArgs, bool> InterceptTouchEvent
+        public event ReturnTypeEventHandler<object, TouchEventArgs, bool> InterceptTouchEvent
```
JoogabYun, line 234: public event ReturnTypeEventHandler<object, TouchEventArgs, bool> InterceptTouchEvent

public event EventHandlerWithReturnType<object, TouchEventArgs, bool> TouchEvent

 using a mix of EventHandlerWithReturnType and ReturnTypeEventHandler.

It seems to be inconsistent.

@dongsug-song 
what do you think about this?

dongsug-song: @JoogabYun 
@dasg34 
Hello, 
Yes, I think so, but 276 line's change would require ACR. 
I am not sure but Mr. @dasg34  seems to have changed only hidden-APIs which need no ACR.
as you know, doing ACR for method is easier because method can be overridden, but property and event are very difficult to be changed because new named APIs should be added to deprecate the previous property or event.
I am sorry that I can't say surely but it is up to the owner whether proceeding ACR or not. ㅠ_ㅠ

dasg34: I will remove this changes for consistency.

PR #2399, opened 2020-12-15
suppress CA1710 code analysis violation for deprecased APIs

dasg34:
We cannot change public APIs.

### API Changes ###

 - ACR:


dongsug-song (commented):
Hello, could you add some brief explanation why the warning is suppressed?
even if the reason is acceptable, we need to remain some history for the future fix.
could you please add reason-explaining comments with "//" ?
for example, `// reason : need ACR and ACR is impossible because other GBM uses same name.`
could you check again?
I understand the deprecated-API's suppressing but other APIs' suppressing reasons are required.
thank you.

dongsug-song (approved):
suppress warnings for deprecated APIs
looks good to me

PR #2400, opened 2020-12-15
[NUI] fix svace issue for texture

zhouleonlei:
### Description of Change ###
fix svace issue for texture

### DIFF
--- a/src/Tizen.NUI/src/internal/FrameBroker/FrameBrokerBase.cs
+++ b/src/Tizen.NUI/src/internal/FrameBroker/FrameBrokerBase.cs
@@ -364,7 +364,7 @@ namespace Tizen.NUI
                     break;
             }
 
-            texture.Dispose();
+            texture?.Dispose();
             geometry.Dispose();
             shader.Dispose();
             return renderer;


huayongxu (approved):
lgtm.

src/Tizen.NUI/src/internal/FrameBroker/FrameBrokerBase.cs
```diff
@@ -364,7 +364,7 @@ internal Renderer GetRenderer(FrameData data)
                     break;
             }
 
-            texture.Dispose();
+            texture?.Dispose();
```
dongsug-song, line 367: How about geometry and shader ? also don't they need null check?

zhouleonlei: > How about geometry and shader ? also don't they need null check?

They have been New in 
            Geometry geometry = CreateQuadGeometry();
            Shader shader = CreateShader();

And in the Switch case, function return Null directly:
            switch (data.Type)
            {
                case FrameData.FrameType.RemoteSurfaceTbmSurface:
                    if (data.TbmSurface == null)
                    {
                        geometry.Dispose();
                        shader.Dispose();
                        return null;
                    }

AdunFang (commented):
It is OK.

elishateng (commented):
LGTM

PR #2411, opened 2020-12-15
[NUI] deprecate unused FocusRequestArgs class

dasg34:
`FocusChangeREquested` event was removed in
https://github.com/Samsung/TizenFX/pull/903.

### API Changes ###

 - ACR: https://code.sec.samsung.net/jira/browse/TCSACR-393


rabbitfor (approved):
LGTM

src/Tizen.NUI/src/public/BaseHandle.cs
```diff
@@ -504,6 +504,7 @@ protected virtual void ReleaseSwigCPtr(System.Runtime.InteropServices.HandleRef
         /// <summary>
         /// Contains event arguments for the FocusChangeRequested event.
         /// </summary>
+        [Obsolete("Deprecated in API9; Will be removed in API12.")]
```
m-macias, line 507: According to our deprecation policy it should be "Deprecated in API9; Will be removed in API11." instead. Also you should suggest the alternative if there is any.

dasg34: Fixed! thank you~

PR #2416, opened 2020-12-16
[NUI] Null check of Owner property of LayoutItem for Independent View.

bshsqa:
- Current implementation, we didn't check whether the Owner Property is null or not in the Layout method.
 - This patch add null check of it to avoid null exception


Seoyeon2Kim (approved):
Approve this PR for null check of Layout Owner.

PR #2417, opened 2020-12-16
[NUI] Fix build warning CA1820, CA2213

wonrst:
**CA1820: Test for empty strings using string length**
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1820

**CA2213: Disposable fields should be disposed**
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2213


dongsug-song (approved):
looks good to me

PR #2420, opened 2020-12-17
[NUI] Fix build warning[CA1001]

huiyueun:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1001
CA1001: Types that own disposable fields should be disposable


Seoyeon2Kim (commented):
Just my curiosity :
Some Dispose() APIs are added with [EditorBrowsable(EditorBrowsableState.Never)] tags, but some are not.
Is there any difference between the two above?

src/Tizen.NUI/src/internal/DisposeQueue.cs
```diff
@@ -12,7 +12,7 @@
 
 namespace Tizen.NUI
 {
-    internal class DisposeQueue
+    internal class DisposeQueue : IDisposable
```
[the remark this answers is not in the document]
huiyueun: done

huiyueun:
@Seoyeon2Kim  internal class is not added, but public class is added. but some are missing. I've modifed it. thank you.

src/Tizen.NUI/src/internal/DisposeQueue.cs
```diff
@@ -9,9 +9,11 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Tizen.NUI
 {
+    [SuppressMessage("Microsoft.Design", "CA1001:Types that own disposable fields should be disposable")]
```
rabbitfor, line 16: How about adding `Justification` for the future developers?
For example,
```C#
[SuppressMessage("Microsoft.Design", "CA1001:Types that own disposable fields should be disposable", Justification="This is a singleton class and is not disposed")]
```

src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyle.cs
```diff
@@ -26,7 +26,7 @@ namespace Tizen.NUI.BaseComponents
     /// </summary>
     /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
     [EditorBrowsable(EditorBrowsableState.Never)]
-    public partial class ViewStyle : BindableObject
+    public partial class ViewStyle : BindableObject, IDisposable
```
rabbitfor, line 29: The `ControlStyle` class inherits `ViewStyle` and it implements IDisposable already.
Please clean up IDisposable code from the `ControlStyle`.

src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
```diff
@@ -32,7 +32,7 @@ namespace Tizen.NUI.Accessibility
     // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
     [SuppressMessage("Microsoft.Design", "CA1724: Type names should not match namespaces")]
     [EditorBrowsable(EditorBrowsableState.Never)]
-    public class Accessibility
+    public class Accessibility : IDisposable
```
rabbitfor, line 35: Accessibility is also a singleton class like DisposableQueue.

src/Tizen.NUI/src/internal/DisposeQueue.cs
huiyueun: Done

src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyle.cs
huiyueun: The parent inheritance has been changed to implement Disposable.

src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
huiyueun: Done

src/Tizen.NUI/src/public/XamlBinding/BindableObject.cs
```diff
@@ -12,7 +12,7 @@ namespace Tizen.NUI.Binding
     /// Provides a mechanism by which application developers can propagate changes that are made to data in one object to another, by enabling validation, type coercion, and an event system.
     /// </summary>
     [EditorBrowsable(EditorBrowsableState.Never)]
-    public abstract class BindableObject : INotifyPropertyChanged, IDynamicResourceHandler
+    public abstract class BindableObject : Disposable, INotifyPropertyChanged, IDynamicResourceHandler
```
rabbitfor, line 15: I am not sure this is a good idea. View inherits BindableObject but it already implements IDisposable in BaseHandle.
In my opinion, duplicate implements would cause additional build warnings.

rabbitfor (commented):
`ButtonExtension`, for example, is not connected DALi object which means it does not need to be disposed in the main thread using `dispose queue`.

The Disposable class is handy because it implements basic disposable APIs but I think it can be heavy because of the `dispose queue`.

rabbitfor:
CA1001: Types that own disposable fields should be disposable

The main purpose of this warning is to call member.Disposable in the container class's Disposable method.
For example,
```
class abstract ButtonExtension()
{

        {
            if (Disposed)
            {
                return;
            }

            if (type == DisposeTypes.Explicit)
            {
                //Called by User
                //Release your own managed resources here.
                //You should release all of your own disposable objects here.
                TouchInfo?.Dispose();
            }

            base.Dispose();
        }
}
```

rabbitfor (approved):
LGTM

PR #2431, opened 2020-12-18
[NUI] Fix CA2002(Do not lock on objects with weak identity)

dkdk-ryu:
Do not lock on objects with weak identity
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2002


hyunjushin:
Don't we nee lock() at all?

src/Tizen.NUI/src/internal/Application.cs
```diff
@@ -605,32 +605,26 @@ public enum MemoryStatus
         {
             add
             {
-                lock (this)
+                // Restricted to only one listener
+                if (_applicationInitEventHandler == null)
```
dongsug-song, line 609: I agree to remove `lock(this)` in event pattern. but I just want to point out the name of handlers. 
after this patch is merged, could you upload new PR to remove these `_`  prefix?
in c#, pascal case is supposed to be used and for private member, it would be OK to use `_` prefix, but in NUI, we have decided not to use prefix such as `_`, `m`, `p`, and etc.

- NUI ground rule : https://code.sec.samsung.net/confluence/display/GFX/NUI+Ground+Rule
- NUI coding rule : https://code.sec.samsung.net/confluence/pages/viewpage.action?pageId=191373542

src/Tizen.NUI/src/internal/Interop/NDalicPINVOKE.cs
```diff
@@ -171,10 +171,7 @@ public static void Set(global::System.Exception e)
                 if (pendingException != null)
                     throw new global::System.ApplicationException("FATAL: An earlier pending exception from unmanaged code was missed and thus not thrown (" + pendingException.ToString() + ")", e);
                 pendingException = e;
-                lock (typeof(NDalicPINVOKE))
-                {
-                    numExceptionsPending++;
-                }
+                numExceptionsPending++;
```
dongsug-song, line 174: I agree to remove `lock(this)` in event pattern. 
but this is not event pattern. please consider to add new dummy object to check the Mutex.

dongsug-song (commented):
could you check my comments?

src/Tizen.NUI/src/internal/Interop/NDalicPINVOKE.cs
dkdk-ryu: I've added a dummy object.

src/Tizen.NUI/src/internal/Application.cs
dkdk-ryu: Okay, I will do it in new PR.

AdunFang (approved):
It is OK.

PR #2434, opened 2020-12-21
[NUI] Add an exception when deleting a view that is not child

huiyueun:

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewPublicMethods.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewPublicMethods.cs
@@ -165,6 +165,11 @@ namespace Tizen.NUI.BaseComponents
             if (child == null || child.GetParent() == null) // Early out if child null.
                 return;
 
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/BaseComponents/ViewPublicMethods.cs
```diff
@@ -165,6 +165,11 @@ public override void Remove(View child)
             if (child == null || child.GetParent() == null) // Early out if child null.
                 return;
 
+            if (child.Parent != this)
```
dongsug-song, line 168: please add exception noti in document comment. =>
<exception cref=xxx> xxx </exception>
(example) https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/public/Window.cs#L69

huiyueun: done

dongsug-song (approved):
looks good to me

PR #2448, opened 2020-12-22
[NUI] Fix the crash issue when IndicatorSize is null(Pagination).

AchoWang:
### Description of Change ###
If IndicationSize is null,  CreateIndicator() will crash.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Pagination.cs
+++ b/src/Tizen.NUI.Components/Controls/Pagination.cs
@@ -103,6 +103,7 @@ namespace Tizen.NUI.Components
                 }
                 paginationStyle.IndicatorSize = value;
                 UpdateVisual();
+                UpdateContainer();
             }
         }
 
@@ -470,6 +471,10 @@ namespace Tizen.NUI.Components
             if (paginationStyle == null)
             {
                 return;
+            }
+            if (paginationStyle.IndicatorSize == null)
+            {
+                paginationStyle.IndicatorSize = new Size(0, 0);
             }
             ImageVisual indicator = new ImageVisual
             {


AdunFang (approved):
It is OK.

elishateng (approved):
LGTM

zhouleonlei (approved):
LGTM

zg2nets (approved):
looks good

rabbitfor:
In my opinion, most properties in Style class is not by default. So it is normal that the PaginationStyle.IndicatorSize is null.

How about fixing Pagination.CreateIndicator?

dongsug-song:
@AchoWang 
Hello, could you check the @rabbitfor 's comment?

AchoWang:
> In my opinion, most properties in Style class are null by default. So it is normal that the PaginationStyle.IndicatorSize is null. (And it is better to keep null when the user is not set the property in Style for the future)
> 
> How about fixing Pagination.CreateIndicator?

AchoWang:
> In my opinion, most properties in Style class are null by default. So it is normal that the PaginationStyle.IndicatorSize is null. (And it is better to keep null when the user is not set the property in Style for the future)
> 
> How about fixing Pagination.CreateIndicator?

Done, plz check again, thanks.

AchoWang:
> @AchoWang
> Hello, could you check the @rabbitfor 's comment?

Done.

src/Tizen.NUI.Components/Controls/Pagination.cs
```diff
@@ -337,6 +337,10 @@ public int SelectedIndex
             }
             set
             {
+                if (paginationStyle.IndicatorSize == null)
```
rabbitfor, line 340: I think the following case won't work correctly.
```
new Pagination()
{
    SelectedIndex = 2,
    IndicatorCount = 3,
    IndicatorSize = new Size(10, 10)
}
```

rabbitfor (approved):
LGTM

PR #2449, opened 2020-12-22
[NUI] Add ImageList property

huiyueun:

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Loading.cs
+++ b/src/Tizen.NUI.Components/Controls/Loading.cs
@@ -29,7 +29,7 @@ namespace Tizen.NUI.Components
     /// <since_tizen> 6 </since_tizen>
     public class Loading : Control
     {
-        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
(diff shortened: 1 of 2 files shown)


dasg34 (changes requested):
`Framework Design Guidelines` don't recommend using `List<T>` in public APIs.

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections
>❌ DO NOT use ArrayList or List<T> in public APIs.
These types are data structures designed to be used in internal implementation, not in public APIs. List<T> is optimized for performance and power at the cost of cleanness of the APIs and flexibility. For example, if you return List<T>, you will not ever be able to receive notifications when client code modifies the collection. Also, List<T> exposes many members, such as BinarySearch, that are not useful or applicable in many scenarios. The following two sections describe types (abstractions) designed specifically for use in public APIs.

I think you can use `IList<T>` instead of `List<T>`.

dongsug-song:
@dasg34 
Hello, this patch is fixed by your review.
I think it's good to go.
could you review again?
thank you.

hyunjushin (approved):
API reviewed

dasg34 (approved):
👍

PR #2451, opened 2020-12-22
[NUI] Add internal and hidden APIs which are currently used

dongsug-song:
### Description of Change ###
[NUI] Add internal and hidden APIs which are currently used
- Add internal and hidden APIs which are currently used. otherwise removing them causes build errors.
- Let users to use proper APIs by deleting these deprecated APIs
- User can be informed that these are deprecated and should not be used by warning messeges during build time.

### API Changes ###
none


JoogabYun (approved):

AdunFang (approved):
It is OK.

PR #2452, opened 2020-12-22
[NUI] Override Dispose() in Navigator

jaehyun0cho:
Pages pushed into Navigator are disposed if Navigator is disposed.

To override Dispose() in Navigator, two dictionaries are required to
remove disposed navigator from dictionary.

TODO: Needs to consider how to remove disposed window from dictionary.

### API Changes ###

 - ACR:


AdunFang (approved):
It is OK.

PR #2458, opened 2020-12-24
[NUI] Fix the behavior as for View.BackgroundImage to be cleared when null set

dongsug-song:
### Description of Change ###
[NUI] Fix the behavior as for View.BackgroundImage to be cleared when null set
- Change the behavior of View.BackgroundImage property from API9
- https://code.sec.samsung.net/jira/browse/GRE-2078

### API Changes ###
none


rabbitfor (approved):
LGTM

PR #2455, opened 2020-12-23
[NUI] Modify CA2000 warnings for Visual

zhouleonlei:
### Description of Change ###
Modify CA2000 warnings for Visual

CA2000: Dispose objects before losing scope
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2000

### API Changes ###

 - ACR:

PR #2459, opened 2020-12-24
Revert "[NUI] Modify CA2000 warnings for Visual"

rabbitfor:
Reverts Samsung/TizenFX#2455

From the dali-csharp-binder code, if the value is null, calling new PropertyValue throws an excpetion.


dongsug-song (approved):
@zhouleonlei 
Hello, please check this PR again
thank you.

PR #2460, opened 2020-12-24
[NUI] Modify CA2000 Warnings for VisualMap

zhouleonlei:
### Description of Change ###
Modify CA2000 Warnings for VisualMap

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/VisualMaps.cs
+++ b/src/Tizen.NUI/src/public/VisualMaps.cs
@@ -76,16 +76,16 @@ namespace Tizen.NUI
         /// <since_tizen> 3 </since_tizen>
         protected PropertyMap _commonlyUsedMap = null;
 
-        private Vector2 _visualSize = null;
-        private Vector2 _visualOffset = null;
-        private Vector2 _visualOffsetPolicy = null;
-        private Vector2 _visualSizePolicy = null;
-        private Visual.AlignType? _visualOrigin = null;
-        private Visual.AlignType? _visualAnchorPoint = null;
+        private Vector2 visualSize = null;
+        private Vector2 visualOffset = null;
+        private Vector2 visualOffsetPolicy = null;
+        private Vector2 visualSizePolicy = null;
+        private Visual.AlignType? visualOrigin = null;
+        private Visual.AlignType? visualAnchorPoint = null;
 
-        private PropertyMap _visualTransformMap = null;
+        private PropertyMap visualTransformMap = null;
 
-        private int? _depthIndex = null;
+        private int? depthIndex = null;
 
         /// <summary>
         /// Constructor.
@@ -106,14 +106,14 @@ namespace Tizen.NUI
         {
             get
             {
-                return _visualSize ?? (new Size2D(1, 1));
+                return visualSize ?? (new Size2D(1, 1));
             }
             set
             {
-                _visualSize = value;
-                if (_visualSizePolicy == null)
+                visualSize = value;
+                if (visualSizePolicy == null)
                 {
-                    _visualSizePolicy = new Vector2(1.0f, 1.0f);
+                    visualSizePolicy = new Vector2(1.0f, 1.0f);
                 }
                 UpdateVisual();
             }
@@ -130,14 +130,14 @@ namespace Tizen.NUI
         {
             get
             {
-                return _visualOffset ?? (new Vector2(0.0f, 0.0f));
+                return visualOffset ?? (new Vector2(0.0f, 0.0f));
             }
             set
             {
-                _visualOffset = value;
-                if (_visualOffsetPolicy == null)
+                visualOffset = value;
+                if (visualOffsetPolicy == null)
                 {
-                    _visualOffsetPolicy = new Vector2(1.0f, 1.0f);
+                    visualOffsetPolicy = new Vector2(1.0f, 1.0f);
                 }
                 UpdateVisual();
             }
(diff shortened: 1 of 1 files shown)


elishateng (approved):
LGTM

AchoWang (approved):
Looks good to me.

rabbitfor (approved):
LGTM

src/Tizen.NUI/src/public/VisualMaps.cs
```diff
@@ -660,27 +660,85 @@ protected virtual void ComposingPropertyMap()
...
+            if (visualSize != null)
+            {
+                PropertyValue temp = new PropertyValue(visualSize);
+                visualTransformMap.Add((int)VisualTransformPropertyType.Size, temp);
+                temp.Dispose();
+            }
+            if (visualOffset != null)
+            {
+                PropertyValue temp = new PropertyValue(visualOffset);
+                visualTransformMap.Add((int)VisualTransformPropertyType.Offset, temp);
```
dongsug-song, line 713: could you check if the `temp` is copied when `.Add()` is called?
I am concerned about that `visualTransfromMap` is used in other place and if `temp` is disposed right after `Add()`, then the `visualTransformMap` could have null `ProperyValue` inside. 
how do you think about?

src/Tizen.NUI/src/public/VisualMaps.cs
```diff
@@ -660,27 +660,85 @@ protected virtual void ComposingPropertyMap()
...
+                temp.Dispose();
+            }
+            if (visualOffsetPolicy != null)
+            {
+                PropertyValue temp = new PropertyValue(visualOffsetPolicy);
+                visualTransformMap.Add((int)VisualTransformPropertyType.OffsetPolicy, temp);
+                temp.Dispose();
```
dongsug-song, line 720: same as above

src/Tizen.NUI/src/public/VisualMaps.cs
```diff
@@ -660,27 +660,85 @@ protected virtual void ComposingPropertyMap()
...
+            }
+            if (visualSizePolicy != null)
+            {
+                PropertyValue temp = new PropertyValue(visualSizePolicy);
+                visualTransformMap.Add((int)VisualTransformPropertyType.SizePolicy, temp);
+                temp.Dispose();
```
dongsug-song, line 726: same as above

src/Tizen.NUI/src/public/VisualMaps.cs
```diff
@@ -660,27 +660,85 @@ protected virtual void ComposingPropertyMap()
...
+            {
+                PropertyValue temp = new PropertyValue((int)visualOrigin);
+                visualTransformMap.Add((int)VisualTransformPropertyType.Origin, temp);
+                temp.Dispose();
+            }
+            if (visualAnchorPoint != null)
+            {
+                PropertyValue temp = new PropertyValue((int)visualAnchorPoint);
+                visualTransformMap.Add((int)VisualTransformPropertyType.AnchorPoint, temp);
+                temp.Dispose();
```
dongsug-song, line 738: same as above

huayongxu (commented):
lgtm

rabbitfor:
@dongsug-song ,
I've checked it.
When NUI creates PropetyValue, the corresponding dali object is created.
And then when adding it to the PropertyMap, it copies the dali object (not reference copy) in the dali code.
So I think disposing it after `Add` is safe.

src/Tizen.NUI/src/public/VisualMaps.cs
zhouleonlei: > could you check if the `temp` is copied when `.Add()` is called?
> I am concerned about that `visualTransfromMap` is used in other place and if `temp` is disposed right after `Add()`, then the `visualTransformMap` could have null `ProperyValue` inside.
> how do you think about?
I have checked the Dali codes:
 Dali::Property::Value *arg3 = 0 ;
result = (Dali::Property::Map *) &(arg1)->Add(arg2,(Dali::Property::Value const &)*arg3);
The arg3 is the pointer of ProperyValue, it did value copied when Adding. So it is safe as Ms.Yang said.

zhouleonlei:
> @dongsug-song ,
> I've checked it.
> When NUI creates PropetyValue, the corresponding dali object is created.
> And then when adding it to the PropertyMap, it copies the dali object (not reference copy) in the dali code.
> So I think disposing it after `Add` is safe.

Thanks for Ms.Yang careful review~

rabbitfor:
> > @dongsug-song ,
> > I've checked it.
> > When NUI creates PropetyValue, the corresponding dali object is created.
> > And then when adding it to the PropertyMap, it copies the dali object (not reference copy) in the dali code.
> > So I think disposing it after `Add` is safe.
> 
> Thanks for Ms.Yang careful review~


Mr. Lei, could you double check?
I checked it too long ago. It is necessary to check again with the latest code.

zhouleonlei:
> > > @dongsug-song ,
> > > I've checked it.
> > > When NUI creates PropetyValue, the corresponding dali object is created.
> > > And then when adding it to the PropertyMap, it copies the dali object (not reference copy) in the dali code.
> > > So I think disposing it after `Add` is safe.
> > 
> > 
> > Thanks for Ms.Yang careful review~
> 
> Mr. Lei, could you double check?
> I checked it too long ago. It is necessary to check again with the latest code.

I have checked Dail codes and make a Sample below:
            PropertyMap _outputVisualMap = new PropertyMap();
            PropertyValue temp = new PropertyValue("123");
            _outputVisualMap.Add(Visual.Property.Shader, temp);
            temp.Dispose();
            temp = new PropertyValue("abc");
            _outputVisualMap.Add(Visual.Property.PremultipliedAlpha, temp);
            temp.Dispose();
            for (uint i = 0; i < _outputVisualMap.Count(); i++)          
            {
                string str;
                PropertyValue v = _outputVisualMap.GetValue(i);
                v.Get(out str);
                Console.WriteLine("~~index{0} = {1}~~", i, str);
            }

It works fine. Thank you

PR #2312, opened 2020-12-01
[NUI] Use correct calling convention

walac:
Some callback and functions from DALi expected the calling convention to
be cdecl, this commit makes it right. The amd64 platforms this is rarely
a problem since arguments are passed through CPU registers, but in x86
this leads to stack corruption.

PR #2462, opened 2020-12-24
[NUI] Use correct calling convention

dongsug-song:
### Description of Change ###
Some callback and functions from DALi expected the calling convention to
be cdecl, this commit makes it right. The amd64 platforms this is rarely
a problem since arguments are passed through CPU registers, but in x86
this leads to stack corruption.

### API Changes ###
none


dongsug-song:
@walac 
Hello, 
this is same as https://github.com/Samsung/TizenFX/pull/2312
please check and please close https://github.com/Samsung/TizenFX/pull/2312
thank you.

hyunjushin (approved):

huiyueun (approved):

PR #2464, opened 2020-12-24
[Device] Add internal api for power off

lokilee73:
Signed-off-by: lokilee73 <changjoo.lee@samsung.com>

### Description of Change ###
Add a internal api for power off

Changed:
 -  Add DevicePowerPowerOff


lokilee73:
Close to update patch again

PR #2468, opened 2020-12-28
[NUI] make GridLayout.ColumnSpan/RowSpan work properly in xaml

dasg34:
Since xaml parser find property setter by property name, the parser will find
`View.ColumnSpan` even if you use "GridLayout.ColumnSpan" in xaml.
GridLayout.ColumnSpan/RowSpan didn't work for the above reason.

To fix this, SetValue of `GridLayout.ColumnSpan/RowSpan` is added in
`View.ColumnSpan/RowSpan`.

Here is xaml code for the test.
```xaml
<View x:Class="Demo.XamlPage"
      xmlns:c="clr-namespace:Tizen.NUI.Components;assembly=Tizen.NUI.Components"
      >
    <View.Layout>
        <GridLayout Columns="2"/>
    </View.Layout>

    <c:Button x:Name="left" Text="left" GridLayout.ColumnSpan="2" GridLayout.HorizontalStretch="ExpandAndFill" BackgroundColor="1,0.96,0.85,1"/>
    <c:Button x:Name="left2" Text="left22" GridLayout.HorizontalStretch="ExpandAndFill" BackgroundColor="0,0.96,0.85,1"/>
    <c:Button x:Name="left3" Text="left33" GridLayout.HorizontalStretch="ExpandAndFill" BackgroundColor="0,0.96,1,1"/>

</View>
```

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -637,6 +637,7 @@ namespace Tizen.NUI.BaseComponents
             set
             {
                 SetValue(RowSpanProperty, value);
+                SetValue(GridLayout.RowSpanProperty, (int)value);
                 NotifyPropertyChanged();
             }
         }
@@ -654,6 +655,7 @@ namespace Tizen.NUI.BaseComponents
             set
             {
                 SetValue(ColumnSpanProperty, value);
+                SetValue(GridLayout.ColumnSpanProperty, (int)value);
                 NotifyPropertyChanged();
             }
         }


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -637,6 +637,7 @@ public float RowSpan
             set
             {
                 SetValue(RowSpanProperty, value);
+                SetValue(GridLayout.RowSpanProperty, (int)value);
```
dongsug-song, line 640: I am not sure whether this is OK because I don't understand Layout very well,
but with quick glance, it seems to set same value for two times, even when GridLayout is not used.
I don't know this is right or not so I think other member's review is required.

src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -654,6 +655,7 @@ public float ColumnSpan
             set
             {
                 SetValue(ColumnSpanProperty, value);
+                SetValue(GridLayout.ColumnSpanProperty, (int)value);
```
dongsug-song, line 658: same as above

src/Tizen.NUI/src/public/BaseComponents/View.cs
rabbitfor: I have question, too. 
As fas as I understand,
`SetValue(GridLayout.RowSpanProperty, (int)value)` will eventually call `view?.Layout?.RequestLayout();`.
If this code was all what you wanted, how about just adding `Layout?.RequestLayout()` here, instead `SetValue`?

dasg34: `Layout?.RequestLayout()` is not related with this issue.
GridLayout will try to get `GridLayout.RowSpanProperty` value of children on `GridLayout.OnMeasure()`. howerver, this value is always 1(default value) because of the reason in the commit message.

rabbitfor: Thank you for the explanation.
I am not familiar with attached bindable properties, so I have search for it,
https://docs.microsoft.com/en-us/xamarin/xamarin-forms/xaml/attached-properties#consume-an-attached-property
```
<Label Text="Label Shadow Effect" local:ShadowEffect.HasShadow="true" />
```
In Xamarin, above example works.

In the description you wrote,
> Since xaml parser find property setter by property name, the parser will find
View.ColumnSpan even if you use "GridLayout.ColumnSpan" in xaml.

I wonder why NUI is different from Xamarin despite it uses same xaml parser.

dasg34: > I wonder why NUI is different from Xamarin despite it uses same xaml parser.

The order of setting property is different.
Please check NUI `XamlG`(https://github.com/Samsung/TizenFX/pull/1639) and Xamarin.Forms `XmalG` (https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Xamarin.Forms.Xaml/ApplyPropertiesVisitor.cs#L405) 

Xamarin.Forms `XamlG` try to set bindable property first, but NUI `XamlG` try to set normal property.
If you test without https://github.com/Samsung/TizenFX/pull/1639 patch, `GridLayout.ColumnSpan` will work properly.


BTW, NUI `XamlC` is the same to Xamarin.Forms `XamlC`

NUI `XamlC` :
- https://github.com/nui-dali/Tizen.NUI.XamlBuild/blob/master/src/internal/Xaml/ApplyPropertiesVisitor.cs#L346

Xamarin.Forms `XamlC` :
-  https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Xamarin.Forms.Build.Tasks/SetPropertiesVisitor.cs#L943

So, if you add below code in `YourDemoApp.csproj`(that means using XamlC instead of XamlG), `GridLayout.ColumnSpan` will work properly.
```xaml
  <PropertyGroup>
    <NeedInjection>True</NeedInjection>
  </PropertyGroup>
```

> Is it because there are duplicate property names between View and GridLayout?

Yes, if property names are different, `TrySetProperty` will return false and `TrySetValue` try to assing attached property.

rabbitfor (approved):
LGTM

But there's one thing that I am not sure about -> TableView
From the code, View's `RowSpan` and `ColumnSpan` are for the TableView.
When the view is a child of a TableView and it sets those properties, GridLayout's attached properties also changed, and then it triggers `Layout?.RequestLayout()` in case it is not created by xaml.
Do you think this patch does not affect TableView's behavior?

dasg34:
> LGTM
> 
> But there's one thing that I am not sure about -> TableView
> From the code, View's `RowSpan` and `ColumnSpan` are for the TableView.
> Do you think this patch does not affect TableView's behavior?
> 
> test/Tizen.NUI.Samples uses TableView, so please check this still works fine before merging this patch.
> Thank you.

If `GridLayout` is child of `TableView`, It can be problem.
We cannot apply this changes :(

PR #2210, opened 2020-11-10
[NUI] Fixed Svace issues

zhouleonlei:
### Description of Change ###
Fixed Svace issues

PR #2187, opened 2020-11-05
[NUI] Add ExcludeLayouting to View

bshsqa:
- If ExcludeLayouting property is false, the view is not affected by parent layout.
 - Even the property is false, the children Views are affected by layout of current View.
 - This ExcludeLayouting property is changed synchronously with SetPositionByLayout of its Layout.
 - Children of AbsoluteLayout can freely be animated.

PR #2469, opened 2020-12-28
[NUI] remove needless null validation in LayoutItem.SetFrame

dasg34:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
@@ -600,22 +600,19 @@ namespace Tizen.NUI
                 }
                 else
                 {
-                    if (Owner.Position != null)
+                    if (independent)
                     {
-                        if(independent)
-                        {
-                            // If height or width specification is not explicitly defined,
-                            // the size of the owner view must be reset even the ExcludeLayouting is true.
-                            if (Owner.HeightSpecification < 0 || Owner.WidthSpecification < 0)
-                            {
-                                Owner.SetSize(right - left, bottom - top);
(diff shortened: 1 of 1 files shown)


dongsug-song (commented):
looks good to me, but other member's review is required

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
```diff
@@ -600,23 +600,20 @@ private bool SetFrame(float left, float top, float right, float bottom, bool ind
                 }
                 else
                 {
-                    if (Owner.Position != null)
+                    if (independent)
```
huiyueun, line 603: It is added by this patch : https://github.com/Samsung/TizenFX/pull/2210/files#diff-7f6a6b6c3dd860842f5d89bcf00c8d5746b0bbd1554dd6611534cd6461738bd5R561
Is there any svace problem with this fix?

dasg34: Svace warning is reported because of `Owner.Position.Z`(https://github.com/Samsung/TizenFX/pull/2210/files#diff-7f6a6b6c3dd860842f5d89bcf00c8d5746b0bbd1554dd6611534cd6461738bd5R563).
However, `Owner.Position.Z` was removed here(https://github.com/Samsung/TizenFX/pull/2187/files#diff-7f6a6b6c3dd860842f5d89bcf00c8d5746b0bbd1554dd6611534cd6461738bd5L563).
So, i think we don't need to consider Svace problem anymore.

huiyueun (approved):

PR #2474, opened 2020-12-29
[NUI] Fix CA1716 build warning

MoonheeChoi:
### Description of Change ###
CA1716: Identifiers should not match keywords
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1716

- Change one hidden API. no ACR
- Since two enums are public APIs, Added suppress messages for CA1716 build warning, not fix them.

### API Changes ###

 - ACR:


dongsug-song (approved):
can not change the names. it's public API.

PR #2477, opened 2020-12-30
[Packagemanager] Fix registering internal event callback

jeremy-jang:
### Description of Change ###
Do not reset internal event callback. This can cause unexpected result
when user add event handler during package installation.
(user may not be able to listen the package event completely.)

### API Changes ###
N/A


pjh9216 (approved):
LGTM

PR #2478, opened 2020-12-30
[NUI] Fix null exception in Loading and improve code

rabbitfor:
This fixes null pointer exception in Loading when running following sample,

```C#
var loading1 = new Loading(new LoadingStyle
{
    Size = new Size(100, 100),
    Position = new Position(0, 100),
    Images  = imageArray
});
```

Plus, this patch removes redundant behaviors when applying a style to a Loading component.
All components copies a default style to the component properties when it is created.
When copying a style, all bindable properties defined in the style class that is not null value are copied to the component properties,
so it is important to keep null value for the property that is not set by user.
Hence this patch removes code that allocates empty list object in the getter of ImageList bindable property.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Loading.cs
+++ b/src/Tizen.NUI.Components/Controls/Loading.cs
@@ -54,7 +54,7 @@ namespace Tizen.NUI.Components
             {
                 var newValueList = newValue as List<string>;
                 instance.loadingStyle.ImageList = newValueList;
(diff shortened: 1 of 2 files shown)


Seoyeon2Kim (approved):
Looks good to me, but someone else must approve. :)

hyunjushin (approved):

PR #2488, opened 2021-01-04
[NUI] Fix build warning CA1712

wonrst:
CA1712: Do not prefix enum values with type name

The name can not be changed becuase this enum is publicly open.
Suppress warning messages.

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #2494, opened 2021-01-06
[NUI] Fix CA2000 Warnings

zhouleonlei:
### Description of Change ###
Fix CA2000 Warnings

### API Changes ###

 - ACR:

PR #2504, opened 2021-01-07
Revert "[NUI] Fix CA2000 Warnings"

rabbitfor:
Reverts Samsung/TizenFX#2494


rabbitfor:
@zhouleonlei 
I am sorry to revert this.
In my opinion, storing signal as a field just for removing warnings is redundant.
For now, I think it'd better to suppress warnings for the signals.
We need to find better way after then.

zhouleonlei:
> @zhouleonlei
> I am sorry to revert this.
> In my opinion, storing signal as a field just for removing warnings is redundant.
> For now, I think it'd better to suppress warnings for the signals.
> We need to find better way after then.

OK Got it

dongsug-song (approved):
I also agree. approve

PR #2515, opened 2021-01-12
[NUI] use Size2DProperty for GetNaturalSize() instead of local field

dasg34:
Since we can take the size value using `Size2DProperty`, we don't need to
save Size2D value in local value.
It will save a bit of memory.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
Looks good to me.
It's an old code before Size2Dproperty is created.
To remove this, the memory will be reduced a little.

PR #2521, opened 2021-01-13
[NUI] Add API for delete widgetView

tscholb:
WidgetViewManager provide API for delete widgetView instead TerminateWidget().
TerminateWidget() will be deprecated.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/WidgetViewManager.cs
+++ b/src/Tizen.NUI/src/public/WidgetViewManager.cs
@@ -25,6 +25,7 @@ namespace Tizen.NUI
     /// <since_tizen> 3 </since_tizen>
     public class WidgetViewManager : BaseHandle
     {
+        private static WidgetViewManager instance;
 
         /// <summary>
         /// Creates a new widgetView manager object.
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/WidgetViewManager.cs
```diff
@@ -55,6 +55,18 @@ public WidgetView AddWidget(string widgetId, string contentInfo, int width, int
             return ret;
         }
 
+        /// <summary>
+        /// Remove a widget view object.
+        /// </summary>
+        /// <param name="widgetView">widgetView to remove</param>
+        /// <returns> True on success, false otherwise.</returns>
+        public bool RemoveWidget(WidgetView widgetView)
```
m-macias, line 63: The public API changes should be processed in master branch only.
So, make API as hidden by adding [EditorBrowsable(EditorBrowsableState.Never)] attribute, then in DevelNUI branch API-Checker will pass.
After the patch is merged into master branch by regular integration work, you can make new PR in master branch by removing [EditorBrowsable(EditorBrowsableState.Never)] attribute to proceed ACR.
if the ACR is urgent, you can upload PR on master branch directly. that PR will be merged on DevelNUI branch by rebasing during the regular integration work.

rabbitfor:
ACR need to be done in master branch.

* DevelNUI : Add new API with `[EditorBrowsable(EditorBrowsableState.Never)]` (hidden)
-> after integrate it to master,
* master : Remove hidden tag and go with ACR

rabbitfor (approved):
LGTM

PR #2522, opened 2021-01-13
[NUI] remove internal child properly when child view of FlexLayout is removed

dasg34:
In `FlexLayout`, when we add new view, the existing view shrink.
but when removed they are in shrinked state. they should change back.

To fix this issue, this patch remove internal child(child of yoga
layout) when child view of `FlexLayout` is removed.

For now, `OnChildRemove` API will work correctly.

Thanks Aman Jeph for the report.

### API Changes ###

 - ACR:


rabbitfor (approved):
Looks good to me. Thank you!

PR #2533, opened 2021-01-19
[NUI] Add Value Indicator of Slider

Seoyeon2Kim:

### Description of Change ###
- Value Indicator is a tooltip of slider.
- Add some properties for value indicator :
  IsValueShown, ValueIndicatorText, ValueIndicatorSize, and ValueIndicatorUrl

### API Changes ###
- N/A


src/Tizen.NUI.Components/Controls/Slider.cs
```diff
@@ -66,7 +66,9 @@ public class SliderSlidingFinishedEventArgs : EventArgs
     /// <since_tizen> 6 </since_tizen>
     public partial class Slider : Control
     {
-        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
+        #region Fields
+
+        /// This will be public opened later after ACR done. Before ACR, need to be hidden as inhouse API.
```
huiyueun, line 71: We may not open bindable properties. Please remove wrong comment.

src/Tizen.NUI.Components/Controls/Slider.cs
```diff
@@ -66,7 +66,9 @@ public class SliderSlidingFinishedEventArgs : EventArgs
     /// <since_tizen> 6 </since_tizen>
     public partial class Slider : Control
     {
-        /// This will be public opened in tizen_6.0 after ACR done. Before ACR, need to be hidden as inhouse API.
+        #region Fields
```
m-macias, line 69: I am not sure what is our policy regarding the '#region' directive, but there is a rule: 
StyleCop SA1124: You should not use regions.

src/Tizen.NUI.Components/Controls/Slider.cs
Seoyeon2Kim: I've just changed the sentences from 'tizen_6.0' to 'later'. Then, can I leave just [EditorBrowsable] tags?

src/Tizen.NUI.Components/Controls/Slider.cs
Seoyeon2Kim: Hmm Okay. Thank you for your review.
I've just used the '#region' to divide sections in a little long code in Slider.cs.
I'll remove these regions now. :)

huiyueun (approved):
Looks good to me

PR #2535, opened 2021-01-19
[NUI] Change Feedback as working on TouchEvent

dongsug-song:
### Description of Change ###
[NUI] Change Feedback as working on TouchEvent
- https://code.sec.samsung.net/jira/browse/TDAF-1366

### API Changes ###
none


zhouleonlei (approved):
LGTM

huiyueun (approved):
Looks good to me

src/Tizen.NUI.Components/Controls/Control.cs
```diff
@@ -320,6 +303,8 @@ private void Initialize()
             StateFocusableOnTouchMode = false;
 
             EnableControlState = true;
+
+            Feedback = true;
```
rabbitfor, line 307: After set it true as default, the control won't work in Linux backend.
We need to implement dummy feedback library for Linux or set it true only for fhub.

rabbitfor: If you want to enables it only in fhub, I think this will work.
```
#if PROFILE_MOBILE
Feedback = true;
#endif

```

PR #2547, opened 2021-01-21
[MachineLearning.Inference][Non-ACR] Fix the potential bug

again4you:
### Description of Change ###

This patch fixes the potential bug of MachineLearning API. Detailed items are as below.

* Update the version number of the shared library
* Remove the name comparision in TensorsInfo.Equals()
* Add Clone() to TensorsInfo class for deep copy

### API Changes ###
 - ACR: None

### Related patches ###
* CSharp TC: https://review.tizen.org/gerrit/#/c/test/tct/csharp/api/+/251938/

#### v3 ####
* Remove `ICloneable` interface and use `internal` Clone() method

#### v2 ####
* Explicitly dispose the `_tensorsInfo` when TensorsData is disposed.
* Update the comments about TensorsInfo

#### v1 ####
* First draft


myroot (changes requested):
1. Now, `_tensorsInfo` is totally owned by TensersData, So you need to dispose it on Dispose(bool)
2. /// <param name="info">The handle of tensors info. (Default: null)</param> 
 it should be update, info can't be null

again4you:
### Submit v2
* Explicitly dispose the `_tensorsInfo` when TensorsData is disposed.
* Update the comments about TensorsInfo

### Question
To @myroot 

Since `Clone()` method is an **internal** access modifier, it is not exposed to the public. I generated the Class diagram after modifying the code but I can't find any update for `Clone()` method. If then, should I make an ACR for Clone() method?


* New class diagram
Tizen MachineLearning Inference_internal

myroot:
> 
> 
> Since `Clone()` method is an **internal** access modifier, it is not exposed to the public. I generated the Class diagram after m

Explicit implementation is not need access modifier, it is always `public`
So, you are adding a new API `Tizen.MachineLearning.Inference.TensorsInfo::System.ICloneable.Clone()` as public

myroot (approved):

myroot (approved):

again4you:
@myroot 
Ping!. To merge this PR, should I do something more? If then, let me know it.

myroot:
> @myroot 
> Ping!. To merge this PR, should I do something more? If then, let me know it.
> 

Nothing, just request merge

PR #2559, opened 2021-01-26
[NUI] Add FontSizeScale property in TextLabel, TextField and TextEditor

id213sin:
### Description of Change ###
The value of FontSizeScale will be used for multiplying the given PointSize.
If FontSizeScale.UseSystemSetting, will use the SystemSettings.FontSize internally.

 - Added (hidden):
  float Tizen.NUI.BaseComponents.TextLabel.FontSizeScale { set; get; }
  float Tizen.NUI.BaseComponents.TextField.FontSizeScale { set; get; }
  float Tizen.NUI.BaseComponents.TextEditor.FontSizeScale { set; get; }
  struct Tizen.NUI.FontSizeScale { public const float UseSystemSetting = -1.0; }

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/TextEditor.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/TextEditor.cs
@@ -34,6 +34,8 @@ namespace Tizen.NUI.BaseComponents
         private string textEditorPlaceHolderTextSid = null;
         private bool systemlangTextFlag = false;
         private InputMethodContext inputMethodContext = null;
+        private float fontSizeScale = 1.0f;
+        private bool hasFontSizeChangedCallback = false;
 
         /// <summary>
         /// Creates the TextEditor control.
@@ -1268,6 +1270,54 @@ namespace Tizen.NUI.BaseComponents
         }
 
         /// <summary>
+        /// The FontSizeScale property. <br />
+        /// The default value is 1.0. <br />
+        /// If FontSizeScale.UseSystemSetting, will use the SystemSettings.FontSize internally. <br />
+        /// </summary>
+        /// <since_tizen> 9 </since_tizen>
+        /// This will be public opened in tizen_6.5 after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public float FontSizeScale
+        {
+            get
+            {
+                return fontSizeScale;
+            }
+            set
+            {
+                float newFontSizeScale;
+
+                if (fontSizeScale == value) return;
+
+                fontSizeScale = value;
+                if (fontSizeScale == Tizen.NUI.FontSizeScale.UseSystemSetting)
+                {
+                    SystemSettingsFontSize systemSettingsFontSize;
+
+                    try
+                    {
+                        systemSettingsFontSize = SystemSettings.FontSize;
+                    }
+                    catch (Exception e)
+                    {
+                        Console.WriteLine("{0} Exception caught.", e);
+                        systemSettingsFontSize = SystemSettingsFontSize.Normal;
+                    }
+                    newFontSizeScale = TextUtils.GetFontSizeScale(systemSettingsFontSize);
+                    addFontSizeChangedCallback();
+                }
+                else
+                {
+                    newFontSizeScale = fontSizeScale;
+                    removeFontSizeChangedCallback();
+                }
+
+                SetValue(FontSizeScaleProperty, newFontSizeScale);
+                NotifyPropertyChanged();
+            }
+        }
+
+        /// <summary>
         /// Scroll the text control by specific amount..
         /// </summary>
         /// <param name="scroll">The amount (in pixels) of scrolling in horizontal &amp; vertical directions.</param>
@@ -1342,6 +1392,8 @@ namespace Tizen.NUI.BaseComponents
                 SystemSettings.LocaleLanguageChanged -= SystemSettings_LocaleLanguageChanged;
             }
 
+            removeFontSizeChangedCallback();
+
             //Release your own unmanaged resources here.
             //You should not access any managed member here except static instance.
             //because the execution order of Finalizes is non-deterministic.
@@ -1400,6 +1452,47 @@ namespace Tizen.NUI.BaseComponents
             if (textEditorPlaceHolderTextSid != null)
             {
                 PlaceholderText = NUIApplication.MultilingualResourceManager?.GetString(textEditorPlaceHolderTextSid, new CultureInfo(e.Value.Replace("_", "-")));
+            }
+        }
+
+        private void SystemSettingsFontSizeChanged(object sender, FontSizeChangedEventArgs e)
+        {
+            float newFontSizeScale = TextUtils.GetFontSizeScale(e.Value);
+            SetValue(FontSizeScaleProperty, newFontSizeScale);
+            NotifyPropertyChanged();
+        }
+
+        private void addFontSizeChangedCallback()
+        {
+            if (hasFontSizeChangedCallback != true)
+            {
+                try
+                {
+                    SystemSettings.FontSizeChanged += SystemSettingsFontSizeChanged;
+                    hasFontSizeChangedCallback = true;
+                }
+                catch (Exception e)
+                {
+                    Console.WriteLine("{0} Exception caught.", e);
+                    hasFontSizeChangedCallback = false;
+                }
+            }
+        }
+
+        private void removeFontSizeChangedCallback()
+        {
+            if (hasFontSizeChangedCallback == true)
+            {
+                try
+                {
+                    SystemSettings.FontSizeChanged -= SystemSettingsFontSizeChanged;
+                    hasFontSizeChangedCallback = false;
+                }
+                catch (Exception e)
+                {
+                    Console.WriteLine("{0} Exception caught.", e);
+                    hasFontSizeChangedCallback = true;
+                }
             }
         }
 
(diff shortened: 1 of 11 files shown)


src/Tizen.NUI/src/public/BaseComponents/TextEditor.cs
```diff
@@ -42,6 +48,7 @@ public partial class TextEditor : View
         public TextEditor() : this(Interop.TextEditor.New(), true)
         {
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+            initializeTextEditor();
```
rabbitfor, line 51: This is duplicate of line:86

src/Tizen.NUI/src/public/BaseComponents/TextEditor.cs
```diff
@@ -54,6 +61,7 @@ public TextEditor(bool shown) : this(Interop.TextEditor.New(), true)
         {
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
             SetVisible(shown);
+            initializeTextEditor();
```
rabbitfor, line 64: This is duplicate of line:86

src/Tizen.NUI/src/public/BaseComponents/TextEditor.cs
```diff
@@ -64,6 +72,8 @@ internal TextEditor(TextEditor handle, bool shown = true) : this(Interop.TextEdi
             {
                 SetVisible(false);
             }
+
+            initializeTextEditor();
```
rabbitfor, line 76: This is duplicate of line:86

src/Tizen.NUI/src/public/BaseComponents/TextField.cs
```diff
@@ -1475,6 +1548,48 @@ private void SystemSettings_LocaleLanguageChanged(object sender, LocaleLanguageC
             }
         }
 
+        private void SystemSettings_FontSizeChanged(object sender, FontSizeChangedEventArgs e)
```
rabbitfor, line 1551: Please avoid to use `_` for the method name.
NUI is not using `_` for the new code since last version.

src/Tizen.NUI/src/public/BaseComponents/TextEditorBindableProperty.cs
```diff
@@ -935,5 +935,22 @@ public partial class TextEditor
...
             return temp;
         }));
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly BindableProperty FontSizeScaleProperty = BindableProperty.Create(nameof(FontSizeScale), typeof(float), typeof(TextEditor), default(float), propertyChanged: (BindableProperty.BindingPropertyChangedDelegate)((bindable, oldValue, newValue) =>
+        {
+            var textEditor = (TextEditor)bindable;
+            if (newValue != null)
+            {
+                Tizen.NUI.Object.SetProperty((System.Runtime.InteropServices.HandleRef)textEditor.SwigCPtr, TextEditor.Property.FontSizeScale, new Tizen.NUI.PropertyValue((float)newValue));
```
rabbitfor, line 945: According to rule [CA2000](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2000), disposable object need to be disposed before losing scope.
How about dispose a `PropertyValue` using `using statement`?
It is safe to dispose a created property after set it to the TextEditor.
```C#
using (var property = new Tizen.NUI.PropertyValue((float)newValue))
{
    Tizen.NUI.Object.SetProperty((System.Runtime.InteropServices.HandleRef)textEditor.SwigCPtr, TextEditor.Property.FontSizeScale, property);
}
```

rabbitfor:
In my opinion, using `null` as a special value that means "follow system settings" has a tricky part.

There is style classes for each component that can decorate its component,
for example,
```C#
TextLabelStyle normalPointSizeStyle = new TextLabelStyle()
{
  PointSize = 10.0f,  // TextLabelStyle.PointSize is "float?" type, while TextLabel.PointSize is float type.
}

// Create a TextLabel with predefined style.
var label = new TextLabel(normalPointSizeStyle);
```

In style, all properties are nullable types and having `null` value means it do not copy the value to its component.
So when we implement `FontSizeScale` in `TextLabelStyle` and `TextFieldStyle`, it can not express "follow system settings" as a `null` value.

So I suggest to define a constant which is a very large value to indicate "follow system settings" instead `null`.

src/Tizen.NUI/src/public/BaseComponents/TextEditor.cs
id213sin: The whole initializeText****() functions are not necessary if we decide to do not use null-able property.
I will remove it.

src/Tizen.NUI/src/public/BaseComponents/TextEditor.cs
id213sin: The whole initializeText****() functions are not necessary if we decide to do not use null-able property.
I will remove it.

src/Tizen.NUI/src/public/BaseComponents/TextEditor.cs
id213sin: The whole initializeText****() functions are not necessary if we decide to do not use null-able property.
I will remove it.

src/Tizen.NUI/src/public/BaseComponents/TextField.cs
id213sin: I'll remove "_" from the method name.

src/Tizen.NUI/src/public/BaseComponents/TextEditorBindableProperty.cs
id213sin: Thank you for suggestion. I'll fix it with the "using" statement.

id213sin:
> In my opinion, using `null` as a special value that means "follow system settings" has a tricky part.
> 
> There is style classes for each component that can decorate its component,
> for example,
> 
> ```cs
> TextLabelStyle normalPointSizeStyle = new TextLabelStyle()
> {
>   PointSize = 10.0f,  // TextLabelStyle.PointSize is "float?" type, while TextLabel.PointSize is float type.
>   TextColor = Color.Red,
> }
> 
> // Create a TextLabel with predefined style.
> var label = new TextLabel(normalPointSizeStyle);
> ```
> 
> In style, all properties are nullable types and having `null` value means it do not copy the value to its component.
> So when we implement `FontSizeScale` in `TextLabelStyle`, it can not express "follow system settings" with a `null` value.
> 
> So I suggest to use other way to indicate "follow system settings" instead `null`.

I've never thought about supporting FontSizeScale in TextLabelStyle.
If we need to enable the feature as default in some NUI Components, the feature needs to be separated from the value.
I will add a new API for the feature. The API would be like "public bool EnableSystemSettingFontSizeScale".
If you have any good idea for the name, please let me know.

id213sin:
@rabbitfor
I updated the patch. Please, review the latest patch set.

rabbitfor:
> > In my opinion, using `null` as a special value that means "follow system settings" has a tricky part.
> > There is style classes for each component that can decorate its component,
> > for example,
> > ```cs
> > TextLabelStyle normalPointSizeStyle = new TextLabelStyle()
> > {
> >   PointSize = 10.0f,  // TextLabelStyle.PointSize is "float?" type, while TextLabel.PointSize is float type.
> >   TextColor = Color.Red,
> > }
> > 
> > // Create a TextLabel with predefined style.
> > var label = new TextLabel(normalPointSizeStyle);
> > ```
> > 
> > 
> > In style, all properties are nullable types and having `null` value means it do not copy the value to its component.
> > So when we implement `FontSizeScale` in `TextLabelStyle`, it can not express "follow system settings" with a `null` value.
> > So I suggest to use other way to indicate "follow system settings" instead `null`.
> 
> I've never thought about supporting FontSizeScale in TextLabelStyle.
> If we need to enable the feature as default in some NUI Components, the feature needs to be separated from the value.
> I will add a new API for the feature. The API would be like "public bool EnableSystemSettingFontSizeScale".
> If you have any good idea for the name, please let me know.

Sorry for the late response.
I checked updated code and your proposal to use `EnableSystemSettingFontSizeScale`  is good,
but I have another idea.
I think it will be simple to define a constant indicating "follow system settings".
In NUIConstants.cs, you can find constants values that is commonly used in NUI apps, for example,
```C#
// The View's ParentOrigin is Position type.
// And Position(0, 0, 0.5f) indicates top-left position but it is hard to type and remember,
// so NUI provide a constant "TopLeft". It is defined in NUIConstants.cs
view.ParentOrigin = ParentOrigin.TopLeft; 
```
How about providing such thing for FontSizeScale?
```C#
// In NUIConstants.cs
struct FontSizeScale
{
  const float UseSystemSetting = -1.0f;
}
```
```C#
// In the application,
textLabel.FontSizeScale = FontSizeScale.UseSystemSetting;
```
What do you think?

dongsug-song:
@id213sin 
Hello,
please resolve the conflicts. otherwise please upload new PR and close this PR.
thank you.

id213sin:
@rabbitfor 
That is a good idea!
I'll check your suggestion and update the patch soon. Thank you!

id213sin:
@rabbitfor 
I updated the patch as your suggestion.
Please, check the latest patch set.

rabbitfor (approved):
LGTM

PR #2566, opened 2021-01-26

dongsug-song:
### Description of Change ###

### API Changes ###
none


huayongxu (approved):
lgtm.

rabbitfor (commented):
As I commented https://github.com/Samsung/TizenFX/pull/2535,
After set it true as default, the control won't work in Linux backend.
We need to implement dummy feedback library for Linux.
Until it's ready, how about enabling it only for fhub.
```
#if PROFILE_MOBILE
    Feedback = true;
#endif
```

PR #2419, opened 2020-12-16
[NUI] Add some new APIs like Context, Settings, etc.

huayongxu:
Originally only WebView is used to add new APIs, it will become
bigger, and be hard to maintain.
This patch is to move some related APIs into a class like WebContext,
WebSettings, etc.
And it also adds some new APIs like WebBackForwardList, etc.

### API Changes ###

 - ACR:

PR #2567, opened 2021-01-26
[NUI] Add some new APIs like Context, Settings, etc. (#2419)

rabbitfor:
Originally only WebView is used to add new APIs, it will become
bigger, and be hard to maintain.
This patch is to move some related APIs into a class like WebContext,
WebSettings, etc.
And it also adds some new APIs like WebBackForwardList, etc.

### API Changes ###

 - ACR:


hinohie (approved):

hinohie (approved):

dongsug-song (approved):
looks good to me

PR #2569, opened 2021-01-27
Add some new APIs like Context, Settings, etc.

huayongxu:
Originally only WebView is used to add new APIs, it will become
bigger, and be hard to maintain.
This patch is to move some related APIs into a class like WebContext,
WebSettings, etc.
And it also adds some new APIs like WebBackForwardList, etc.

### API Changes ###

 - ACR:


rabbitfor:
Hi Mr. Xu, I already have merge it. Please check https://github.com/Samsung/TizenFX/pull/2567

PR #2578, opened 2021-01-28
[NUI] Add some APIs into WebView.

huayongxu:
Title, Favicon, ClearAllTilesResources are added.
Meanwhile WebView can be created with arguments argc & args
that will be delivered into web engine.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/WebView/WebView.cs
+++ b/src/Tizen.NUI/src/internal/WebView/WebView.cs
@@ -168,6 +168,7 @@ namespace Tizen.NUI
             internal static readonly int ScrollPosition = Interop.WebView.ScrollPositionGet();
             internal static readonly int ScrollSize = Interop.WebView.ScrollSizeGet();
             internal static readonly int ContentSize = Interop.WebView.ContentSizeGet();
+            internal static readonly int Title = Interop.WebView.TitleGet();
         }
 
         private static readonly BindableProperty UrlProperty = BindableProperty.Create(nameof(Url), typeof(string), typeof(WebView), string.Empty, propertyChanged: (BindableProperty.BindingPropertyChangedDelegate)((bindable, oldValue, newValue) =>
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/internal/WebView.cs
```diff
@@ -572,6 +591,31 @@ public Size ContentSize
...
+
+        /// <summary>
+        /// Gets fav icon.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public ImageView Favicon
+        {
+            get
+            {
+                global::System.IntPtr imageView = Interop.WebView.GetFavicon(SwigCPtr);
```
rabbitfor, line 614: You need to catch exception with
```C#
```
Since I`nterop.WebView.GetFavicon` is connected with `CSharp_Dali_WebView_GetFavicon` in c++ and the method can produce an error.

huayongxu: Done.
A little difference is that there is a VS warning if throw an exception, null is returned instead to avoid warning.

src/Tizen.NUI/src/internal/WebView/WebView.cs
```diff
@@ -254,6 +263,16 @@ public WebView(string locale, string timezoneId) : this(Interop.WebView.New2(loc
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
         }
 
+        /// <summary>
+        /// Creates an uninitialized WebView.
+        /// <param name="args">args array</param>
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public WebView(string[] args) : this(Interop.WebView.New3(args.Length, args), true)
```
rabbitfor, line 271: Can you check the build warning at line 271?
It seems that you need to check nullity of args before referencing it.

huayongxu: Done.

rabbitfor (approved):
LGTM

PR #2579, opened 2021-01-28

dongsug-song:
### Description of Change ###

- Add #if PROFILE_MOBILE to make NUI.Components run on Ubuntu

### API Changes ###
none


rabbitfor (approved):
LGTM

PR #2580, opened 2021-01-28
Fix WatchTime.UtcTimestamp.Millisecond bug

gotoextreme:
For now, WatchTime.UtcTimestamp.Millisecond always return 0. This patch will add proper millisecond to the WatchTime.UtcTimestamp so that WatchTime.UtcTimestamp.Millisecond can return proper value.


src/Tizen.Applications.WatchApplication/Tizen.Applications/WatchTime.cs
```diff
@@ -292,7 +292,7 @@ public DateTime UtcTimestamp
                     else
                         Log.Error(LOGTAG, "Failed to get UtcTimestamp err : " + err);
                 }
-                return (new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc)).AddSeconds(ts).ToLocalTime();
+                return (new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc)).AddSeconds(ts).AddMilliseconds(Millisecond).ToLocalTime();
```
myroot, line 295: Why do you convert to local time from UTC?

hjhun (approved):

pjh9216 (approved):

PR #2584, opened 2021-01-28
[NUI] Fix Selector bug and add new clone method

rabbitfor:

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
@@ -1,5 +1,5 @@
 ﻿/*
- * Copyright(c) 2019 Samsung Electronics Co., Ltd.
+ * Copyright(c) 2021 Samsung Electronics Co., Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
```diff
@@ -297,6 +301,22 @@ public Selector<T> Clone()
...
+        {
+            if (null == converter) throw new ArgumentNullException(nameof(converter));
+
+            Selector<TOut> result = new Selector<TOut>();
+            result.All = converter(All);
+            result.SelectorItems = SelectorItems.ConvertAll<SelectorItem<TOut>>(m => new SelectorItem<TOut>(m.State, converter(m.Value)));
+
+            return result;
+        }
+
```
dongsug-song, line 319: could you check the build warning below in line 546?
I think null checking for `list` is reasonable.
Or you can add null argument exception.

when you add exception throwing, please add exception comment also.
=> <exception cref=xxx> xxx </exception>
Example: https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/public/Window.cs#L69

Seoyeon2Kim (approved):
Looks good to me!
This PR is to add new clone method in Selector :)

src/Tizen.NUI/src/public/BaseComponents/Style/Selector.cs
rabbitfor: Done!

PR #2587, opened 2021-01-29
[NUI] Add GetOriginalImageSize() considering rotation

tscholb:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/ImageLoading.cs
+++ b/src/Tizen.NUI/src/public/ImageLoading.cs
@@ -211,7 +211,7 @@ namespace Tizen.NUI
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/ImageLoading.cs
```diff
@@ -220,7 +220,24 @@ public static Size2D GetClosestImageSize(string filename)
         [EditorBrowsable(EditorBrowsableState.Never)]
```
bshsqa, line 220: We need to add comment that this method will not respect any rotation of image

Seoyeon2Kim (approved):
Looks good to me. 
This PR is to add a hidden API.

PR #2602, opened 2021-02-02
[NUI] Integration TextPageUtils

JoogabYun:
### Description of Change ###

Integration TextPageUtils

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/BaseComponents/TextLabel.cs
```diff
@@ -1345,6 +1363,24 @@ public PropertyMap TextFit
             }
         }
 
+        /// <summary>
+        /// The MinLineSize property.<br />
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public float MinLineSize
```
m-macias, line 1371: Why are you adding a new API (although hidden) to a closed branch? Additionally, this branch is targeting API6 and you are adding API8 item. What is the reason of those changes?

JoogabYun: VD wants to apply the features applied in tizen 6.0 to tizen 5.5 as well.
VD is said that the tizen_6.0 TV based App is applied to tizen_5.5 TV without modification.

huiyueun (approved):

PR #2606, opened 2021-02-02
[NUI] Add a VideoHoleEnabled property into WebView.

huayongxu:
Video hole is used for playing videos with high quality in WebView.
This patch is to add a VideoHoleEnabled property.

### API Changes ###

 - ACR:


rabbitfor (approved):
LGTM

PR #2620, opened 2021-02-05
[ComponentBased] Fix ComponentPort

hjhun:
### Description of Change ###

- The unnecessary Interop Parcel methods are removed.
- The descriptions of ComponentPort is changed.
The timeout is milliseconds. If the timeout is '-1', the interval will be changed to the default timeout.
And, the WaitForEvent() runs a main loop until the Cancel() is called. The code in the next line will not run until the Cancel() is called. It's recommended that the method in the sub thread.


src/Tizen.Applications.ComponentBased.ComponentPort/Tizen.Applications.ComponentBased.ComponentPort/ComponentPort.cs
```diff
@@ -89,6 +88,63 @@ public void AddPrivilege(string privilege)
         /// <summary>
         /// Waits for events.
         /// </summary>
+        /// <reamrks>
+        /// This method runs a main loop until Cancel() is called.
+        /// The code in the next line will not run until Cancel() is called.
+        /// It is recommended that this method uses in the sub thread.
+        /// </reamrks>
```
m-macias, line 95: Spelling.
```suggestion
        /// <remarks>
        /// This method runs a main loop until Cancel() is called.
        /// The code in the next line will not run until Cancel() is called.
        /// It is recommended that this method uses in the sub thread.
        /// </remarks>
```

hjhun: Done. Thanks.

src/Tizen.Applications.ComponentBased.ComponentPort/Tizen.Applications.ComponentBased.ComponentPort/ComponentPort.cs
```diff
@@ -89,6 +88,63 @@ public void AddPrivilege(string privilege)
         /// <summary>
         /// Waits for events.
         /// </summary>
+        /// <remarks>
+        /// This method runs a main loop until Cancel() is called.
+        /// The code in the next line will not run until Cancel() is called.
+        /// It is recommended that this method uses in the sub thread.
+        /// </remarks>
+        /// <example>
+        /// <code>
```
m-macias, line 97: What is the point of this `<code>` section? First of all, it is not related to this method. Second of all, it is too long and redundant.

hjhun: I've fixed it.

WonyoungChoi:
@hjhun If the directory name is changed, the assembly (dll) name also changed together. It means that the backward compatibility could be broken. If you really want to change the name, you should provide a facade assembly using type-forwarding ( https://docs.microsoft.com/en-us/dotnet/standard/assembly/type-forwarding)

PR #2621, opened 2021-02-05
[NUI] NUI ThemeManager works with TizenThemeManager

rabbitfor:
### API Changes ###

 - ACR:


huiyueun (approved):

dongsug-song:
"ACR Required" tag is not related to NUI and it's from other module.
so "ACR Required" tag can be ignored.

PR #2623, opened 2021-02-05
[NUI] Remove wrong force casting, Layer as View

hinohie:
Remove force casting at LayoutItem.cs
just change `Owner.GetParent() as View;`
+
tiny typo error fix

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
@@ -168,10 +168,7 @@ namespace Tizen.NUI
             OnAttachedToOwner();
             // Add layout to parent layout if a layout container
             View parent = Owner.GetParent() as View;
-            if (parent != null)
-            {
-                (parent?.Layout as LayoutGroup)?.Add( this );
-            }
+            (parent?.Layout as LayoutGroup)?.Add( this );
 
             // If Add or ChangeOnAdd then do not update condition
             if (ConditionForAnimation.Equals(TransitionCondition.Unspecified))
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
```diff
@@ -168,7 +168,10 @@ internal void AttachToOwner(View owner)
             OnAttachedToOwner();
             // Add layout to parent layout if a layout container
             View parent = Owner.GetParent() as View;
-            (parent?.Layout as LayoutGroup)?.Add( this );
+            if (parent != null)
```
bshsqa, line 171: Do we need this change?

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
```diff
@@ -551,8 +554,7 @@ private bool SetFrame(float left, float top, float right, float bottom)
                                                          " top:" + _layoutPositionData.Top +
                                                          " right:" + _layoutPositionData.Right +
                                                          " bottom:" + _layoutPositionData.Bottom );
-                Container onwerContainer = Owner.GetParent();
-                View onwerView = onwerContainer is Layer ? new View(Layer.getCPtr(onwerContainer).Handle, false) : onwerContainer as View;
+                View onwerView = Owner.GetParent() as View;
```
bshsqa, line 557: Layer is not a View.
If Parent is Layer it is not safe to cast directly from Layer to View.

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
hinohie: Nope. Revert.

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
hinohie: As I know, `as` operation is safe enough when casting failed. If it is not safe,.. We have to fix a lot of codes.

huiyueun: I think it is safe code. If the parent is Layer, there is no additional work required.

huiyueun (approved):

PR #2626, opened 2021-02-08
[NUI] Add all Dali colors to NUI

Ali-Alzyoud:
### Description of Change ###
Add all DAli Colors into NUI

### API Changes ###

 - ACR:


huiyueun (approved):

rabbitfor:
Colors in NDalic only used for internal purpose. (but it seems that there is no use cases)
For public colors, please refer to [this](https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/public/Common/Color.cs#L35).

Ali-Alzyoud:
> Colors in NDalic only used for internal purpose. (but it seems that there is no use cases)
> For public colors, please refer to [this](https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/public/Common/Color.cs#L35).

I see, hmmm but I think redefine them again seems redundant.
The color class seems a wrapper for Vector4, So how about use these defines in src/Tizen.NUI/src/public/Common/Color.cs instead of setting colors values again.

rabbitfor:
> > Colors in NDalic only used for internal purpose. (but it seems that there is no use cases)
> > For public colors, please refer to [this](https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/public/Common/Color.cs#L35).
> 
> I see, hmmm but I think redefine them again seems redundant.
> The color class seems a wrapper for Vector4, So how about use these defines in src/Tizen.NUI/src/public/Common/Color.cs instead of setting colors values again.

I agree :)

PR #2627, opened 2021-02-09
[NUI] Change LayoutController process

huiyueun:
Register a callback when Layout is first used.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
Looks good to me. This PR is to improve the performance if an app does not use Layout at all. :)

PR #2644, opened 2021-02-11
Mark all Tizen.Security.TEEC APIs as deprecated

dmichaluk:
Change-Id: Ic002190023d3f05f9e120c4481e8879deea03cb8

### Description of Change ###

Mark all Tizen.Security.TEEC APIs as deprecated

### API Changes ###

Mark all Tizen.Security.TEEC APIs as deprecated


m-macias:
According to our deprecation policy all of those Obsolete attributes comments should mention the API version and alternative method if there is any. For example
`[Obsolete("Deprecated in API9, Will be removed in API11. Please use XXX instead.")]`
If this API will be removed without any alternatives, please mention it in the comment of the Obsolete attribute.

PR #2645, opened 2021-02-11
[NUI] Removed old dependency in Tizen.NUI.csproj

m-bialota:
* This enables building Tizen.NUI as a reference in Visual Studio
project.

### Description of Change ###
Removed ProjectReference of Tizen.Applications.ThemeManager from Tizen.NUI.csproj

### API Changes ###
none

PR #2652, opened 2021-02-16
Version (nui22011) rebase into master

lwc0917:
### Description of Change ###
Version (nui22011) rebase into master

### API Changes ###
Rebase PR no API Changes


lwc0917:
@m-macias

The 'ACR required' tag is attached, could you check it?

Any idea about it?

lwc0917:
Need to check fork git master history can cause this problem.

PR #2655, opened 2021-02-16
Rebase develNUI 

lwc0917:
### Description of Change ###
Version (nui22011) rebase into master

### API Changes ###
Rebase PR no API Changes


lwc0917:
@m-macias

The 'ACR required' tag is attached, 

The API bot builds both branches and gets the list of APIs and compares them, 
 so I think it's a metadata problem of the forked Git, so I tried this and that, but it still gets ACR Required tag...

https:/samsung.github.io/TizenFX/internals/api/Tizen.Applications.ThemeManager.Theme.html

APIs reported as problems as above are in internal.

Do you think there's a problem?

PR #2659, opened 2021-02-17
Rebase develnui

lwc0917:
### API Changes ###

 - ACR:


lwc0917:
@m-macias

The 'ACR required' tag is attached,

The API bot builds both branches and gets the list of APIs and compares them,
so I think it's a metadata problem of the forked Git, so I tried this and that, but it still gets ACR Required tag...

https:/samsung.github.io/TizenFX/internals/api/Tizen.Applications.ThemeManager.Theme.html

APIs reported as problems as above are in internal.

Do you think there's a problem?

m-macias:
@lwc0917 I am not responsible for the TizenAPI-Bot, but I agree that "ACR Required" tag is unnecessary in this case. It may be related to the fact that you are removing the project reference, but those changes are not tracked in TCSACR process. Nonetheless, I suggest using master branch only for public API changes (as was discussed with @dongsug-song ) and using DevelNUI branch only for in-house and internal API changes, because there are multiple tools that may detect your changes as false positive.

WonyoungChoi:
@lwc0917 The problem is that the `Tizen.NUI` was referring to the `Tizen.Applications.ThemeManager`, which is an internal module. For that reason, the Internal API was promoted to public API. But current rebase changes is removing that wrong reference. Therefore, you can ignore the `ACR Required` tag now.

huiyueun (approved):

lwc0917:
@WonyoungChoi  Thank you for your opinion.

PR #2661, opened 2021-02-18
[MachineLearning.Inference][Non-ACR] Fix the potential bug (#2547)

huiyueun:
* [MachineLearning.Inference] Update the version number of the shared library

This patch updates the version number of the nnstreamer library.
* libcapi-nnstreamer.so.0 -> libcapi-nnstreamer.so.1

* [MachineLearning.Inference] Remove the name comparision in TensorsInfo.Equals()

Since the name of the TensorsInfo is not used when negociating the
pipeline, the name comparison in TensorsInfo.Equals() should be removed.

* [MachineLearning.Inference] Add Clone() to TensorsInfo class for deep copy

It is not a common case but developers can modify the TensorsInfo after
creating the TensorsData. Because of this reason, TensorsData should
have a hard-copied instance of the TensorsInfo object. If not, it shows
the wrong information of TensorsData. This patch fixes this potential
bug.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.MachineLearning.Inference/Interop/Interop.Nnstreamer.cs
+++ b/src/Tizen.MachineLearning.Inference/Interop/Interop.Nnstreamer.cs
@@ -22,7 +22,7 @@ internal static partial class Interop
 {
     internal static partial class Libraries
     {
-        public const string Nnstreamer = "libcapi-nnstreamer.so.0";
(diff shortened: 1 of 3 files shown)


dongsug-song (approved):
sync patch.

PR #2670, opened 2021-02-22
[NUI] Create TextLayout when parent has layout

huiyueun:

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #2680, opened 2021-02-25
[NUI] Implement CameraView

JoogabYun:
### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/BaseComponents/CameraView.cs
```diff
@@ -0,0 +1,102 @@
...
+            Image     //  texture stream
+        };
+
+        /// <summary>
+        /// Creates an initialized CameraView.
+        /// </summary>
+        /// <param name="handle">Multimedia Camera handle</param>
+        /// <param name="type">DisplayType</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public CameraView(global::System.IntPtr handle, DisplayType type = DisplayType.Window) : this(Interop.CameraView.New(), true)
```
dongsug-song, line 43: as you know, raw pointer (IntPtr) is not supposed to be exposed in public API. 
this needs to be changed as using `SafeHandle`. (https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/public/BaseComponents/VideoView.cs#L502)
currently, this is hidden-API so it would be OK. but please fix when this is published by ACR.
thank you.

dongsug-song (approved):
looks good to me,
but please check the comment before merging

PR #2695, opened 2021-02-26
[NUI] Remove rarely used properties from the Style class

rabbitfor:
This can leads application launching time improvement.

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me.
this is for cleaning up unused APIs.
these are all Hidden-API so it can also be added again later if required.

PR #2718, opened 2021-03-09
[NUI] Add AppBarStyle class with the latest AppBar UX

jaehyun0cho:
To apply the latest AppBar UX, AppBarStyle class is added.

AppBarStyle.BackButton is applied to AppBar.DefaultNavigationContent.
AppBarStyle.TitleTextLabel is applied to AppBar.DefaultTitleContent.
AppBarStyle.ActionView and ActionButton are applied to the children of
AppBar.DefaultActionContent added by AppBar.AddActions().

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
```diff
@@ -129,6 +150,27 @@ protected override void Dispose(DisposeTypes type)
             base.Dispose(type);
         }
 
+        /// <summary>
```
rabbitfor, line 153: This does not need. Please remove it.

jaehyun0cho: Thank you for your review!
The property Style has been removed.

Do you mean that the property ViewStyle works instead of this property?

rabbitfor: Yes. NUI decided not to open public `Style` for all component. Existing ones should be removed next ACR.

PR #2656, opened 2021-02-16
[NUI] Introduce NUI Palette APIs

lwc0917:
### Description of Change ###
Introduce NUI Palette APIs

All the detail about it
https://code.sec.samsung.net/confluence/display/ENUIFWC/94.+NUI+Palette

### API Changes ###
No public APIs yet

PR #2723, opened 2021-03-10
[NUI] Introduce NUI Palette APIs

lwc0917:
### Description of Change ###
Add NUI Palette features.

#2656 It's been reverted due to TizenFX not allow to has dependency with System.Drawing
So i changed internal logic and remove other module dependency.

### API Changes ###
No Public API Changes


src/Tizen.NUI/src/internal/Utility/ColorCutQuantizer.cs
```diff
@@ -0,0 +1,646 @@
...
+                minRed = minGreen = minBlue = 0xff;
+                maxRed = maxGreen = maxBlue = 0x0;
+                for (int i = lowerIndex; i <= upperIndex; i++)
+                {
+                    int red = (colors[i] >> 16) & 0xff;
+                    int green = (colors[i] >> 8) & 0xff;
+                    int blue = colors[i] & 0xff;
+                    if (red > maxRed)
+                    {
+                        maxRed = red;
```
JoogabYun, line 446: How about using trinomial operator?
maxRed = red > maxRed ? red : maxRed;

JoogabYun (approved):

src/Tizen.NUI/src/internal/Utility/ColorCutQuantizer.cs
lwc0917: That would be better.
Fix done.

PR #2730, opened 2021-03-12
[NUI] Add TouchArea property.

JoogabYun:
### Description of Change ###

    
    Default touchable area is view's size.
    TouchArea can reset the view's touchable area.
    
    This is usefull when the view is small, but it should have a larger touch area.
    
    for example
    
          View view = new View();
          view.Size = new Size(10, 10);
          view.TouchEvent += OnTouchEvent;
          view.TouchArea = new Size2D(200, 200);
    
    The view is small, If you want to set the touch area to a larger area,
    you can use the TouchArea property.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.ActorProperty.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.ActorProperty.cs
@@ -204,6 +204,9 @@ namespace Tizen.NUI
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Actor_Property_BLEND_EQUATION_get")]
             public static extern int BlendEquationGet();
 
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Actor_Property_TOUCH_AREA_get")]
+            public static extern int TouchAreaGet();
+
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_new_Actor_Property")]
             public static extern global::System.IntPtr NewActorProperty();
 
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
```diff
@@ -1364,5 +1364,25 @@ private void OnWindowWheelEvent(object sender, Window.WheelEventArgs e)
...
+        /// If you set the TouchArea on an view, when you touch the view, the touch area is used rather than the size of the view.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Size2D TouchArea
+        {
+            get
+            {
+                Size2D temp = new Size2D(0, 0);
+                GetProperty(View.Property.TouchArea).Get(temp);
+                return new Size2D(temp.Width, temp.Height);
```
dongsug-song, line 1378: are these different?
`return new Size2D(temp.Width, ...)` and `return temp` ?
if they are same, I think the latter would be better. 
How do you think?

src/Tizen.NUI/src/internal/Interop/Interop.ActorProperty.cs
```diff
@@ -204,6 +204,9 @@ internal static partial class ActorProperty
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Actor_Property_BLEND_EQUATION_get")]
             public static extern int BlendEquationGet();
 
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Actor_Property_TOUCH_AREA_get")]
```
dongsug-song, line 207: could you share the csharp-binder's patch?
sometimes, runtime crash comes caused by mismatch of charp-binder version.

JoogabYun: https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/246392/

dongsug-song (approved):
looks good to me

PR #2733, opened 2021-03-12
[NUI] Add Progress Indeterminate mode

Seoyeon2Kim:

### Description of Change ###
- Indeterminate type means an unspecified amount of wait time.
 If `ProgressState` property sets `ProgressStatusType.Indeterminate` mode,
 then progress is continually repeating until the process in complete.

### API Changes ###
- N/A
-


src/Tizen.NUI.Components/Controls/Progress.cs
```diff
@@ -262,6 +265,38 @@ public string BufferImageURL
...
+        {
+            get
+            {
+                if (indeterminateImage == null)
+                {
+                    return null;
+                }
+                else
+                {
+                    return ((string)indeterminateImage.URL);
```
dongsug-song, line 284: I think this can be shortened like this =>
`return indeterminateImage?.URL;`
( https://docs.microsoft.com/ko-kr/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and- )
how do you think?

src/Tizen.NUI.Components/Controls/Progress.cs
```diff
@@ -472,6 +508,20 @@ private void UpdateValue()
             }
         }
 
+        /// <summary>
+        /// Update Animation for Indeterminate mode.
+        /// </summary>
+        /// This will be public opened later after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        private void UpdateIndeterminateAnimation()
```
hyunjushin, line 516: private method can't be opened. I'm not sure if we need separate Animation method. It might be UX dependent. right?

hyunjushin (commented):
Please check and fix warning messages.

src/Tizen.NUI.Components/Controls/Progress.cs
Seoyeon2Kim: Yes, you're right. I'll update it as you suggested.

src/Tizen.NUI.Components/Controls/Progress.cs
Seoyeon2Kim: Yes, private method can't be used by app developers.
Currently I set the basic animation for indeterminate mode, but I'll think about adding a new public method so that app developer can change other Animation. Then, I'll create another PR separately.

src/Tizen.NUI.Components/Controls/Progress.cs
```diff
@@ -127,10 +128,13 @@ public class Progress : Control
         private ImageView trackImage = null;
         private ImageView progressImage = null;
         private ImageView bufferImage = null;
+        private ImageVisual indeterminateImage = null;
         private float maxValue = 100;
         private float minValue = 0;
         private float currentValue = 0;
         private float bufferValue = 0;
+        private Animation indeterminateAnimation = null;
+        private ProgressStyle progressStyle => ViewStyle as ProgressStyle;
```
rabbitfor, line 137: `progressStyle` is not used in the code. Could you remove the variable?

src/Tizen.NUI.Components/Controls/Progress.cs
```diff
@@ -223,6 +227,38 @@ public string BufferImageURL
             set => bufferImage.ResourceUrl = value;
         }
 
+        /// <summary>
+        /// The property to get/set the indeterminate image.
+        /// </summary>
+        /// <exception cref="NullReferenceException">Thrown when setting null value.</exception>
+        /// This will be public opened later after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public string IndeterminateImage
```
rabbitfor, line 236: According to other Progress's API,
the string path is expressed as `XXXURL`. (TrackImageURL, ProgressImageURL).
For the naming consistency, this API should be `IndeterminateImageURL`.

But the `URL` is against the c# coding convention. (`Url` is the correct.)

Could you please make all APIs to have correct name?
* System.Uri TrackImageUri
* System.Uri ProgressImageUri
* System.Uri IndeterminateImageUri

src/Tizen.NUI.Components/Controls/Progress.cs
Seoyeon2Kim: Okay, I'll remove it.

src/Tizen.NUI.Components/Controls/Progress.cs
Seoyeon2Kim: As we discuss it today, I'll change the name to `IndeterminateImageUrl`.
Do you think we should add `TrackImageUrl` or `ProgressImageUrl` and deprecate the current one?

rabbitfor: I think it'll be better to deprecate existing `XXXURL` and add new `XXXUrl`. Thank you!

rabbitfor (approved):
LGTM

src/Tizen.NUI.Components/Controls/Progress.cs
Seoyeon2Kim: Okay, great. I'll add new ones together when I proceed ACR for other Slider public properties. :)

PR #2736, opened 2021-03-12
[NUI] Fix TV TCT fails

dongsug-song:
### Description of Change ###
[NUI] Fix TV TCT fails

- FeedKey and Accessibility are FHUB(mobile) feature
- TV doesn't use Feedkey nor NUI Accessibility
- Need to be seperated by profile when building

### API Changes ###
none


rabbitfor (approved):
LGTM

PR #2737, opened 2021-03-12
[Peripheral] Fix Dispose pattern in Tizen.Peripheral

xerrni:

### Description of Change ###
Fix Dispose pattern in Tizen.Peripheral API.

### API Changes ###
None, This is private API, so there is no need for ACR.

### DIFF
--- a/internals/src/Tizen.Peripheral/Tizen.Peripheral/Gpio.cs
+++ b/internals/src/Tizen.Peripheral/Tizen.Peripheral/Gpio.cs
@@ -112,32 +112,40 @@ namespace Tizen.Peripheral.Gpio
         public GpioPin(int pinNumber, GpioPinDriveMode mode)
         {
             var ret = NativeGpio.Open(pinNumber, out _handle);
-            if (ret != ErrorCode.None)
-                throw ExceptionFactory.CreateException(ret);
-
-            PinNumber = pinNumber;
-            switch (mode)
+            try
             {
-                case GpioPinDriveMode.Input:
-                    ret = NativeGpio.SetEdgeMode(_handle, NativeGpio.EdgeType.Both);
-                    if (ret != ErrorCode.None)
-                        throw ExceptionFactory.CreateException(ret);
+                if (ret != ErrorCode.None)
+                    throw ExceptionFactory.CreateException(ret);
 
-                    ret = NativeGpio.SetDirection(_handle, NativeGpio.Direction.In);
-                    if (ret != ErrorCode.None)
-                        throw ExceptionFactory.CreateException(ret);
-                    SetIntteruptedCallback();
-                    break;
-                case GpioPinDriveMode.OutputInitiallyLow:
-                    ret = NativeGpio.SetDirection(_handle, NativeGpio.Direction.OutLow);
-                    if (ret != ErrorCode.None)
-                        throw ExceptionFactory.CreateException(ret);
-                    break;
-                case GpioPinDriveMode.OutputInitiallyHigh:
-                    ret = NativeGpio.SetDirection(_handle, NativeGpio.Direction.OutHigh);
-                    if (ret != ErrorCode.None)
-                        throw ExceptionFactory.CreateException(ret);
-                    break;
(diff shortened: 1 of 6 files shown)


xerrni:
pull request is ready for review.

myroot (commented):
Why do you dispose the object when exception occurred?

xerrni:
```csharp
namespace ConsoleApp1
{
    class Tester : IDisposable
    {
        public void T1()
        {
            throw new Exception("exception");
        }
        ~Tester()
        {
            Console.WriteLine("Destructor");
            Dispose(false);
        }
        public void Dispose()
        {
            Console.WriteLine("No parameter Dispose");
            Dispose(true);
            GC.SuppressFinalize(this);
        }
        {
            if(_disposed)
            {
                Console.WriteLine($"Already Disposed: {disposing}");
                return;
            }
            //dispose code...
            _disposed = true;
            Console.WriteLine($"Disposed: {disposing}.");
        }
    }
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello World!");
            Tester t1 = new Tester();
            try
            {
                t1.T1();
            }
            catch(Exception e)
            {
                Console.WriteLine($"Caught: {e.Message}");
            }
            t1.Dispose();
        }
    }
}
```
```csharp
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello World!");
            Tester t1 = new Tester();
            t1.T1();
            // Dispose not called and resource is busy
        }
    }
```
In this example `Dispose()` is called, but when e.g. exception occurs and is not caught it will not be called.
I need to ensure that `Dispose()` is always  called because it free's resources associated with e.g. hardware pin on the board.
Without it, this pin will be in busy state.

xerrni:
I rebased this branch.

myroot:
> 
> In this example `Dispose()` is called, but when e.g. exception occurs and is not caught it will not be called.
> I need to ensure that `Dispose()` is always called because it free's resources associated with e.g. hardware pin on the board.
> Without it, this pin will be in busy state.


Normal C# Programmer....

Programming like this
``` c#
     class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello World!");
            using (Tester t1 = new Tester()) 
            {
                 t1.T1();
            }
        }
    }
```
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement
https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/using-objects

myroot:
The object life cycle should be controlled by programmer, programmer could be decide when object disposed
your code not allow it

xerrni:
Following your suggestion, I removed `Dispose()` calls before throwing exception.

internals/src/Tizen.Peripheral/Tizen.Peripheral/Gpio.cs
```diff
@@ -111,39 +111,33 @@ public class GpioPin : IDisposable
...
+                case GpioPinDriveMode.OutputInitiallyLow:
+                    ret = NativeGpio.SetDirection(_handle, NativeGpio.Direction.OutLow);
+                    if (ret != ErrorCode.None)
+                        throw ExceptionFactory.CreateException(ret);
+                    break;
+                case GpioPinDriveMode.OutputInitiallyHigh:
+                    ret = NativeGpio.SetDirection(_handle, NativeGpio.Direction.OutHigh);
+                    if (ret != ErrorCode.None)
+                        throw ExceptionFactory.CreateException(ret);
+                    break;
```
myroot, line 140: You need try and catch, because it is a constructor, if fail to constructor it does not create object so, developer can't release internal handle

``` c#
 try { 
 ....
 catch (Exception err) {
       Dispose();
       throw;
 }
```

myroot (changes requested):

myroot (approved):
You do better update PR title

xerrni:
I rebased changes. Can this pull request be merged?

xerrni:
@WonyoungChoi 
I have rebased this commit. Can it be merged ?

PR #2739, opened 2021-03-15
[NUI] change type from Size2D to Size at TouchArea api

JoogabYun:
### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #2744, opened 2021-03-16
[NUI] : Add ScrollPosition in ScrollEventArgs for user convenience.

everLEEst:
### Description of Change ###
Add ScrollPosition in ScrollEventArgs.

Current Position property in ScrollEventArgs is position of
ContentContainer which is negative.
user controls scrollableBase via ScrollTo and other API with
positive position of pan not ContentContainer so user must convert
current Position Property to use.

By this patch, we provide positive ScrollPosition which represent
position of scroll pan, so no need to convert value for using it.

### API Changes ###
No API Changes.


rabbitfor (approved):
LGTM

jaehyun0cho (approved):

PR #2756, opened 2021-03-18
[Inputmethod] Fix the crash issue when setting PreEditString's attribute

Inhong:
### Description of Change ###
The crash issue occurred when C library released memory for string attribute.
So I modified to use Malshal class instead of GCHandle.

### API Changes ###
There is no need for ACR.


myroot (approved):

imfine98 (approved):
LGTM

PR #2763, opened 2021-03-18
[NUI] CheckBox, RadioButton and Switch can place text without setting…

rabbitfor:
… text size

### API Changes ###

 - ACR:


dongsug-song (commented):
looks good to me

huiyueun (approved):

huiyueun:
LGTM

jaehyun0cho:
@rabbitfor 
After applying this patch, it seems that ellipsis effect does not happen in button text string when text string is longer than button size.
Is it ok?

PR #2772, opened 2021-03-22
[NUI] Button text ellipsis enabled and etc.

rabbitfor:
* Button text ellipsis when the text is longer than the button width.
* Switch thumb can get control state changed callback from parent.
* Button IconPadding is ignored when the icon is not exist.
* Button TextPadding is ignored when the text is not exist.

### API Changes ###

 - ACR:


wonrst (approved):
Looks good to me.
This patch guarantees the size of Button and TextLabel in several situations.
This should be merged with the patch below.
https://github.com/Samsung/TizenFX/pull/2774

PR #2777, opened 2021-03-23
[NUI] : Scrollbar thumb support image and update styles

everLEEst:
Change thumb View to ImageView to support image and color properly.
- View thumb -> ImageView thumb

Add new property for thumb images on Scrollbar and ScrollbarStyle.
- ThumbVerticalImageUrl
- ThumbHorizontalImageUrl

Add new image resource for thumb image.
- nui_component_default_scroll_vbar.#.png
- nui_component_default_scroll_hbar.#.png

Now color can be color of background for thumb when image is not
existed, or color of image if url is existed.


everLEEst:
need to change thumb size.
resource min size is 14 but GUI size is 12.

rabbitfor (commented):
Please check github build warnings in `File changed` tab.
You may ignore the warning, such as,
```
Change the type of property XXX from string to System.Uri
```
I will suppress them all soon.

everLEEst:
Still resource is not updated, but It seems horitzontal and vertical resources are interchanged....
update the resource and checked it looks fine but wait for UX designer's confirmation.

dongsug-song:
@everLEEst 
Hello, could you resolve build error?

src/Tizen.NUI.Components/Controls/Scrollbar.cs
```diff
@@ -475,6 +527,43 @@ private void UpdateTrackPadding(Extents padding)
             thumbView.Size = calculator.CalculateThumbSize(ThumbThickness, trackView.Size);
             thumbView.Position = calculator.CalculateThumbPaddingPosition(trackView.Size, thumbView.Size, thumbView.Position, trackPadding);
         }
+        private void UpdateThumbColor(Color color)
+        {
+            thumbColor = color;
+            if (String.IsNullOrEmpty(thumbView.ResourceUrl))
+            {
+                thumbView.Color = color;
```
rabbitfor, line 535: Does it work? The property `Color` is a mixed color value that will be multiplied with other visuals color in a View. So I think when there is no resource url, it can be just a backgroundColor.

everLEEst: This is typo after change build error.
the condition must be !String.IsNullOrEmpty, so if thumeView resourceUrl is exist, set color on the view.

rabbitfor (approved):
LGTM

PR #2783, opened 2021-03-24
[NUI] Make possible to capture on the old driver devices.

bshsqa:
Some devices those have old gpu driver cannot use NativeImageSource for Capture.
This patch adds more option to use captured buffer with uri directly


rabbitfor:
Could you add `[NUI]` prefix in the commit message?

dongsug-song (approved):
LGTM

PR #2785, opened 2021-03-24
[NUI] Add apis of webview settings.

dongsug-song:
### API Changes ###

 - ACR:


dongsug-song:
should make a PR on DevelNUI first

PR #2789, opened 2021-03-25
[NUI] Change button theme to show Focused+Pressed state well.

rabbitfor:

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
+++ b/src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
@@ -41,8 +41,8 @@ namespace Tizen.NUI.Components
                 BackgroundColor = new Selector<Color>()
                 {
                     Normal = new Color(0.039f, 0.055f, 0.29f, 1),
-                    Focused = new Color(0, 0.2f, 0.545f, 1),
                     Pressed = new Color(0.106f, 0.412f, 0.792f, 1),
+                    Focused = new Color(0, 0.2f, 0.545f, 1),
                     Disabled = new Color(0.765f, 0.792f, 0.824f, 1),
                 },
                 Text = new TextLabelStyle()


src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
```diff
@@ -41,8 +41,8 @@ public Theme Create()
                 BackgroundColor = new Selector<Color>()
                 {
                     Normal = new Color(0.039f, 0.055f, 0.29f, 1),
-                    Focused = new Color(0, 0.2f, 0.545f, 1),
                     Pressed = new Color(0.106f, 0.412f, 0.792f, 1),
+                    Focused = new Color(0, 0.2f, 0.545f, 1),
```
myroot, line 45: It looks like the same code as before.

myroot: If the order in which the property is set is important, it is really scary.

rabbitfor: Yeah... I agree. NUI `Selector` class is designed to act like a collection but it is not look like.
So there is a plan to refactor `Selector` class in the near future and after then this code will be
```C#
BackgroundColor = new Selector<Color>
{
  new SelectorItem<Color>(ControlState.Normal, new Color(0.039f, 0.055f, 0.29f, 1),
  new SelectorItem<Color>(ControlState.Pressed, new Color(0.106f, 0.412f, 0.792f, 1)),
  new SelectorItem<Color>(ControlState.Focused, new Color(0, 0.2f, 0.545f, 1)),
  new SelectorItem<Color>(ControlState.Disabled, new Color(0.765f, 0.792f, 0.824f, 1)),
}
```

dongsug-song (approved):
LGTM

PR #2792, opened 2021-03-25
[NUI] Fix layout null crash issue

huiyueun:
need to check 'value'

### API Changes ###

 - ACR:


rabbitfor (approved):
LGTM

PR #2795, opened 2021-03-26
[VoiceControl] Support multi handle for event and add thread lock for these handles

stom-hwang:
### Description of Change ###

This patch makes event handler support multi handles and add thread locks for these event handlers

### API Changes ###

 - No ACR


urmain (approved):
LGTM

syeonholic (approved):
It looks good to me.

PR #2797, opened 2021-03-26
[Internal] Extend Tizen.Inspections internal API

kkuchciak:
### Description of Change ###
Extending internal Inspections API after extending native API (diagnostics).

### API Changes ###
Added:

class Inspector:
 - EventHandler<DataRequestReceivedEventArgs> DataRequestReceived;
 - void SubscribeEvent(string eventName, string moduleID);
 - void SendEvent(string eventName, Bundle eventData);
 - void RequestBugreport(int pid);
 - void SetDataRequestCallback();
 - void UnsetDataRequestCallback()

class InspectionContext:
 - string EventName; // property
 - Bundle EventData; // property

class InspectionData:
 - int Write(byte[] buffer, int offset, int count);


internals/src/Tizen.Inspections/Tizen.Inspections/DataRequestReceivedEventArgs.cs
```diff
@@ -0,0 +1,50 @@
+/*
+* Copyright (c) 2020 Samsung Electronics Co., Ltd All Rights Reserved
```
m-macias, line 2: I suggest changing the year if it is a new file.
```suggestion
* Copyright (c) 2021 Samsung Electronics Co., Ltd All Rights Reserved
```

WonyoungChoi (approved):

PR #2798, opened 2021-03-26
[Peripheral] Gpio callback return pin value from structure.

xerrni:
Change implementation so interrupted callback reads pin value from
structure rather than invoking Read() method.

### Description of Change ###
This commit allows to get more precise data about pin value, as it is saved inside CAPI callback request, rather than invoking Read() method in C# which leads to getting PinValue with higher timestamp, which can mean that PinValue has already changed.

This commit depends on:
https://github.com/Samsung/TizenFX/pull/2737
https://review.tizen.org/gerrit/#/c/platform/core/api/peripheral-io/+/255967/ - merged

### API Changes ###
None, this is private API.


internals/src/Tizen.Peripheral/Tizen.Peripheral/Gpio.cs
```diff
@@ -168,7 +168,16 @@ private void SetIntteruptedCallback()
...
         private void OnInterrupted(IntPtr handle, ErrorCode error, IntPtr data)
         {
-            ValueChanged?.Invoke(this, new PinUpdatedEventArgs(PinNumber, Read()));
+            NativeGpio.PeripherialGpio pio = (NativeGpio.PeripherialGpio)
+                                            System.Runtime.InteropServices.Marshal.PtrToStructure(handle,
+                                                typeof(NativeGpio.PeripherialGpio));
+            // check magic values to verify capi structures integrity
+            if (pio.vermagic != 13712 || pio.cbInfo.vermagic != 14469)
+            {
+                throw ExceptionFactory.CreateException(ErrorCode.BadAddress);
```
myroot, line 177: Do not throw exception in Callback, nobody to catch it

myroot: Hm.. I don't know which is better, throw excpetion and terminate app, or just print out log and ignore this callcak.

And I don't know why do you need magic value check, because TizenFX is including on platform binary and it always compatiable with library built in platform binary

myroot: And it is not big issue, so i approve it

myroot (approved):

internals/src/Tizen.Peripheral/Tizen.Peripheral/Gpio.cs
xerrni: As you mentioned callback should not throw exception, so I will change `throw` exception to `TIzen.Log`.
I want to check structure magic number to check if it's parsed correctly rather than using it as it -- read arbitrary data pointed by handle.

PR #2799, opened 2021-03-29
[NUI] Add to check xaml image resource

huiyueun:
- Check IsCreateByXaml of parent because components.
 If we create component by xaml, we can't set image resource.
- Add BackgroundImage checker for xaml

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -38,7 +38,7 @@ public class ImageView : View
             var imageView = (ImageView)bindable;
             string url = (string)newValue;
             url = (url == null ? "" : url);
-            if (imageView.IsCreateByXaml && url.Contains("*Resource*"))
+            if ((imageView.IsCreateByXaml || (imageView.GetParent()?.IsCreateByXaml ?? false)) && url.Contains("*Resource*"))
```
rabbitfor, line 41: Is there a case that the parent is not ceated by xaml but the chlid is from xaml?

huiyueun: If crate button by xaml, icon of button is created code. so button.Icon.IsCreateByXaml is false.
Problem is that IconUrl is not drawing on xaml editor.
```csharp
<Button x:Name="btnDashBoard2" Position="0,75" Text="Dashboard" PointSize="8" TextColor="#007EF9" WidthResizePolicy="FillToParent" IconURL="*Resource*/profile.png" Size="184,26" Margin="0,0,10,10" Opacity="Infinity"/>
```

rabbitfor: If so checking only a parent won't solve all problems, especially if the component is made of 3 depth objects.

rabbitfor: As far as I know, switch's thumb is 3depth object.
Is there a way to set IsCreateByXaml to true for all sub-components?

huiyueun: Right, I also think this issue. but there is no idea about that.
How about deleting check IsCreateByXaml?, I mean just only check the resource name if it is contain '*Resource*'.

rabbitfor: How about override `IsCreateByXaml` of Button to set sub-component's `IsCreateByXaml` to true? (Switch is also a Button)

huiyueun: I've updated my patch. I've added to override 'IsCreateByXaml' in Control.

jaehyun0cho: Is it possible to add child to parent after parent's IsCreateByXaml is set to be true?
(the added child's IsCreateByXaml is false as @rabbitfor mentioned above)

If so, isn't it required to set child's IsCreateByXaml whenever child is added?

huiyueun: IsCreateByXaml is set to true at the time it creates an object by xaml. If you add child after the parent's IsCreateByXaml is set, I think the added object has no problem with false.

elishateng: Hi, Mr. Eun,
I talked with Mr. Fang, we suggest another way to reach it, like the following snippets.

`
            if (url.StartWith("*Resource*"))
            {
                string resource = Tizen.Applications.Application.Current.DirectoryInfo.Resource;
                url = url.Replace("*Resource*", resource);
            }
`

In this way, user can use '*Resource*' way to set the URL of ImageView/BackgoundImage both in c# codes and XAML.
We don't need set child's IsCreateByXaml of NUI Components control when it create by xaml.

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
```diff
@@ -149,6 +149,12 @@ public partial class View
                 return;
             }
 
+            if (view.IsCreateByXaml && url.Contains("*Resource*"))
```
elishateng, line 152: How about change this line into this:
```
if (url.StartWith("*Resource*"))
```

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
huiyueun: @rabbitfor @Jaehyun-Cho could you check teng's opinion? I think it is good way to fix.

src/Tizen.NUI.Components/Controls/Control.cs
```diff
@@ -152,6 +152,25 @@ public object CommandParameter
             set { SetValue(CommandParameterProperty, value); }
         }
 
+        /// Only used by the IL of xaml, will never changed to not hidden.
+        /// If the parent is created by xaml, all children are recognized as objects created with xaml.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override bool IsCreateByXaml
```
elishateng, line 158: We don't need it.

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
jaehyun0cho: I think @elishateng and @AdunFang 's idea is great :)

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
huiyueun: done

src/Tizen.NUI.Components/Controls/Control.cs
huiyueun: done

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
huiyueun: @rabbitfor @Jaehyun-Cho @elishateng @AdunFang 
Please check my new patch.

elishateng (approved):
LGTM

rabbitfor (approved):
LGTM

PR #2806, opened 2021-03-30
[NUI] Fix Svace issue

huiyueun:

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Common/Color.cs
+++ b/src/Tizen.NUI/src/public/Common/Color.cs
@@ -1056,8 +1056,9 @@ namespace Tizen.NUI
         /// The copy constructor.
         /// </summary>
         /// <param name="other">The copy target.</param>
+        /// <exception cref="ArgumentNullException"> Thrown when other is null. </exception>
         [EditorBrowsable(EditorBrowsableState.Never)]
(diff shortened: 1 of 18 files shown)


Seoyeon2Kim (approved):
Looks good to me. This PR is to fix SVACE issues for null check.

src/Tizen.NUI/src/public/Common/Color.cs
```diff
@@ -1232,7 +1233,12 @@ public float A
         /// <since_tizen> 3 </since_tizen>
         public static implicit operator Vector4(Color color)
         {
-            return new Vector4((float)color?.R, (float)color.G, (float)color.B, (float)color.A);
+            if (color == null)
+            {
+                Tizen.Log.Error("NUI", "color is null");
```
rabbitfor, line 1238: I don't think this is an error case to show an error message.
```C#
Color color = null;
Vector4 vector4 = color;
```
This can happen sometimes.

tscholb (approved):
Looks good to me

src/Tizen.NUI/src/public/Common/Color.cs
huiyueun: delete all logs

Seoyeon2Kim (approved):
Agree with Jiyun's review!  Great!! :+1:

PR #2809, opened 2021-03-30
[NUI] Fix button layout when resize policy is FillToParent and etc. 

rabbitfor:
* Add ItemAlignment property to make it "center" for a Button and "begin" for others.
* Add ItemSpacing to set space between icon and text.

### API Changes ###

 - ACR:


src/Tizen.NUI.Components/Controls/Button.Internal.cs
```diff
@@ -413,22 +424,12 @@ private void OnTextOrIconUpdated()
             }
 
             // If the button has fixed width and the text is not empty, the text should not exceed button boundary.
-            if (WidthSpecification >= 0 && !String.IsNullOrEmpty(buttonText.Text))
+            if (WidthSpecification != LayoutParamPolicies.WrapContent && !String.IsNullOrEmpty(buttonText.Text))
```
huiyueun, line 427: Don't we need to check about 'MatchParent' property?

rabbitfor: In case `MatchParent`, the `size` parameter is equals to the parent size so it can calculate text's maximum size.

src/Tizen.NUI.Components/Controls/Button.cs
```diff
@@ -127,8 +127,11 @@ public partial class Button : Control
...
             var instance = (Button)bindable;
-            instance.iconPadding = (Extents)((Extents)newValue).Clone();
-            instance.OnTextOrIconUpdated();
+            if (instance.buttonIcon == null)
+            {
+                return;
+            }
+            instance.buttonIcon.Padding = (Extents)newValue;
         },
         defaultValueCreator: (bindable) => ((Button)bindable).iconPadding);
```
jaehyun0cho, line 136: Do we need to change this as well?

src/Tizen.NUI.Components/Controls/Button.cs
```diff
@@ -137,17 +140,42 @@ public partial class Button : Control
...
             var instance = (Button)bindable;
-            instance.textPadding = (Extents)((Extents)newValue).Clone();
-            instance.OnTextOrIconUpdated();
+            if (instance.buttonText == null)
+            {
+                return;
+            }
+            instance.buttonText.Padding = (Extents)newValue;
         },
         defaultValueCreator: (bindable) => ((Button)bindable).textPadding);
```
jaehyun0cho, line 149: Do we need to change this as well?

src/Tizen.NUI.Components/Controls/Button.cs
```diff
@@ -623,6 +651,28 @@ public Extents TextPadding
...
+            set => SetValue(ItemAlignmentProperty, value);
+        }
+
+        /// <summary>
+        /// The space between icon and text.
+        /// The value is applied when there exist icon and text both.
+        /// The width value is used when the items are arranged horizontally. Otherwise, the height value is used.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Size2D ItemSpacing
```
jaehyun0cho, line 670: How about changing this name to CellPadding like LinearLayout?

src/Tizen.NUI.Components/Controls/Button.cs
rabbitfor: Done! Thank you.

src/Tizen.NUI.Components/Controls/Button.cs
rabbitfor: Done! Thank you.

src/Tizen.NUI.Components/Controls/Button.cs
rabbitfor: In button, I think `Item` is more suitable than `Cell` so I used `Item`
* ItemAlignment
* ItemSpacing

jaehyun0cho (approved):
LGTM

PR #2819, opened 2021-03-31
[NUI] Fix not to check non-nullable value

jaehyun0cho:
NavigatorPages always contains non-nullable values.
Therefore, to resolve svace issue (DEREF_AFTER_NULL), null pointer check
is not done for non-nullable values.

### API Changes ###

 - ACR:


huiyueun (approved):
LGTM

Seoyeon2Kim (approved):
Looks good to me!

PR #2839, opened 2021-04-02
[NUI][AT-SPI] Improve StringToVoid implementation

grzelewski:
This commit must be merged first: https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/256117/

### Description of Change ###
This pull request allows to return result to caller of Dali::Signal<>::Emit() by first parameter.

In this case Bridge implemented inside dali is the caller.

This pull request also remove Emit function from C# api.
Emit function should be used only by Bridge and there is
no reason to allow call of this function from C# api.

### API Changes ###

NONE

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/StringToVoidSignal.cs
+++ b/src/Tizen.NUI/src/internal/Common/StringToVoidSignal.cs
@@ -35,11 +35,18 @@ namespace Tizen.NUI
         }
 
         [EditorBrowsable(EditorBrowsableState.Never)]
-        static internal string ConvertParam1(global::System.IntPtr data)
+        static internal string GetResult(global::System.IntPtr data)
         {
-            string result = Interop.StringToVoidSignal.ConvertParam1(data);
+            string result = Interop.StringToVoidSignal.GetResult(data);
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
             return result;
+        }
+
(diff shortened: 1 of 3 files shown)


lol-github (approved):
LGTM

lol-github (approved):
LGTM

src/Tizen.NUI/src/internal/Common/StringToVoidSignal.cs
```diff
@@ -89,13 +96,6 @@ public void Disconnect(System.Delegate func)
                 if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
             }
         }
-
-        [EditorBrowsable(EditorBrowsableState.Never)]
```
Seoyeon2Kim, line 93: Could you  explain why Emit() is removed in more detail?

grzelewski: Emit function should be used only by Bridge and there is
no reason to allow call of this function from C# api.

Moreover emit functions (as designed in c++) should take one parameter - reference to std::string (c++ reference not c# reference to string) - so current implementation of Emit is broken.

Seoyeon2Kim: I see :+1:

Seoyeon2Kim (approved):
Looks good to me!
csharp-binder patch is  merged now.

PR #2841, opened 2021-04-06
[NUI] Improve Process of LayoutController

huiyueun:
Add to check if layout is needed before OnMeasure
It makes performance improve of layouting.

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
```diff
@@ -202,31 +202,30 @@ internal void AttachToOwner(View owner)
...
-            if (needsLayout)
-            {
-                OnMeasure(widthMeasureSpec, heightMeasureSpec);
-                OnMeasureIndependentChildren(widthMeasureSpec, heightMeasureSpec);
-                flags = flags | LayoutFlags.LayoutRequired;
-                flags &= ~LayoutFlags.ForceLayout;
-            }
-            oldWidthMeasureSpec = widthMeasureSpec;
-            oldHeightMeasureSpec = heightMeasureSpec;
+            return LayoutRequested || needsLayout;
```
jaehyun0cho, line 228: How about comparing separately?

```
if (LayoutRequested) // line 217
{
    return true;
}
```

```
return needsLayout; //last line of NeedsLayout
```

jaehyun0cho (commented):
Originally, PerformLayout() was called although Measure() was not called.
In this patch, PerformLayout() is not called as well if Measure() was not called.

Layout() uses SetFrame() to check if there is a change.
It seems that NeedsLayout() does not work like SetFrame().

Do we need to check if PerformLayout() can be called or not in MeasureAndLayout()?

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
huiyueun: Thank you. I've fixed patch.

I agree with the problem you mentioned. I've modified the logic so that it doesn't change.
When the Width / Height is changed, the SetFrame will also change by measure. 

I think the problem is position. I decided that the change of position is not a factor in the layouting.
However, logic changes may affect the operation, so I will reflect it after more testing.

jaehyun0cho (approved):
LGTM

PR #2844, opened 2021-04-06
[NUI] Keep the previously set FittingMode value

tscholb:
FittingMode should be applied even if Url is set later.
so ImageView need to keep the previsouly set fittingMode value.

### API Changes ###

 - ACR:


bshsqa:
Isn't there any other properties of ImageVisual?

myroot:
You should guarantee that returning previously setted value. 
``` c#
image.FittingMode = FittingModeType.ScaleToFill;
Debug.Assert(image.FittingMode == FittingModeType.ScaleToFill);
```

myroot:
It's horrible situation
``` c#
image.FittingMode = FittingModeType.ScaleToFill;
anotherImage.FittingMode = image.FittingMode;
```

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -889,12 +890,13 @@ public FittingModeType FittingMode
         {
             get
             {
-                int ret = (int)VisualFittingModeType.Fill;
+                int ret = (int)_fittingMode;
```
myroot, line 893:
```suggestion
                return ConvertVisualFittingModetoFittingMode(_fittingMode);
```

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1223,6 +1226,13 @@ private void UpdateImage(int key, PropertyValue value)
                 border?.Dispose();
             }
 
+            if(_fittingMode != VisualFittingModeType.Fill)
```
myroot, line 1229:
```suggestion
            if (key != Visual.Property.VisualFittingMode && _fittingMode != VisualFittingModeType.Fill)
```

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
tscholb: In the current structure, the dali and NUI values may be different, so i obtained the dali value.
i think the property map of ImageView need to refactoring, and i will changed it that time

myroot: According to 
``` c#
if(key != Visual.Property.VisualFittingMode && _fittingMode != VisualFittingModeType.Fill)
```
A default value of fittingMode is `Fill` on Dali native, 
And you set `_fittingMode = VisualFittingModeType.Fill;` as default value
So, _fittingMode value and value from dali must be same. 

Could you tell me when it could be different?

If do not same between `_fittingMode` and value from Dali native
It also horrible things.
``` c#
image.FittingMode = FittingModeType.ScaleToFill;
anotherImage.FittingMode = image.FittingMode;
```
I expect to set, `FittingModeType.ScaleToFill` on image and anotherImage, 
But according to your comment deoes not.

tscholb: To resolve this problem, I am planning a new PR for refactoring PropertyMap in ImageView.
I think we'd better check this problem in that PR.

bshsqa (approved):
LGTM

dongsug-song:
according to Mr. Kim, a new PR will be submitted soon and I merge this now.

@tscholb mentioned =>
To resolve this problem, I am planning a new PR for refactoring PropertyMap in ImageView.
I think we'd better check this problem in that PR.

PR #2845, opened 2021-04-06
[NUI] Fix not to block touch event by Scrollbar

jaehyun0cho:
Now, Scrollbar blocks touch events to ScrollableBase's Contents.

Since Scrollbar does not support Handler mode now, Scrollbar's
EnableControlState is set to be false not to catch TouchEvent.

### API Changes ###

 - ACR:


everLEEst (approved):
Looks Good to me.

rabbitfor (approved):
:)

PR #2846, opened 2021-04-06
[NUI] Add Devel(internal) TCs

dongsug-song:
### Description of Change ###
[NUI] Add Devel(internal) TCs

### API Changes ###
none


AchoWang (approved):
Looks good to me.

Seoyeon2Kim (approved):
I'm afraid of whether Accessibility signals could be working on desktop test cases. But, let's get  in first. We can change and fix it when we complete the base of NUI devel TCT frame.
Approved.

PR #2850, opened 2021-04-06
[NUI] Make bindable properties in Style and View have same name and etc.

rabbitfor:

### Description of Change ###
* Remove TextFieldStyle's Selector.

### API Changes ###

 - ACR:


dongsug-song:
@rabbitfor 
Thank you for big patch.
could you explain about this briefly?

everLEEst (commented):
Thank you for this update.
I was also thinking it is weird that property on View is normally string or color or url and Style is Selector.
It looks good to me!
only one tiny worry. Changing style is not breaking any backward compatibility yet?

rabbitfor:
> Thank you for this update.
> I was also thinking it is weird that property on View is normally string or color or url and Style is Selector.
> It looks good to me!
> only one tiny worry. Changing style is not breaking any backward compatibility yet?

Style is not open yet (Non ACR API). So I think it is OK to change.

rabbitfor:
> @rabbitfor
> Thank you for big patch.
> could you explain about this briefly?

Regarding NUI `View`, there were 2 problems related with bindable properties.
1. Some APIs do not have same logic with its bindable property. For example, `Color`, `BackgroundColor` and etc. This leads to the wrong result when the user sets a value through the bindable property.
2. `Style` class should have same set of bindable properties with corresponding `View`, but some properties have different names. For example, View.BackgroundColorProperty and ViewStyle.BackgroundColorSelectorProperty. It looks weird and make it hard to understand.


This patch tries to fix them.
Note that, regarding problem 1, this patch fix only part of it.

dongsug-song:
@rabbitfor 
Thank you for detailed explanation. 
I will merge this after the conflict is resolved. (please check the below error) 
thank you.

dongsug-song (approved):
looks good to me

PR #2853, opened 2021-04-07
[NUI] Fix Window.Instance null issue

huiyueun:
If call Window.Instance before initializing, Window.Instance is always null

### API Changes ###

 - ACR:


myroot:
I want to re-struct, because, current implemation is using bidirectional reference between `Window` and `Application`

On Window class,
``` c#
 public static Window Instance { get; internal;}
```

On Application class
``` c#
  // on initialize time of application
 Window.Instance = GetWindow();
```

huiyueun:
Thank you. I've modified structure of window and application.

myroot (approved):

PR #2855, opened 2021-04-07
[NUI] Picker fix scroll position moves to an incorrect place after an…

lwc0917:
### Description of Change ###
Fix an issue where the position moves to the wrong place when scrolling occurs while an animation for aligning the item is in progress.

### API Changes ###
none.


rabbitfor (approved):
LGTM. I think it'd be good to have if-statements in the same format as other NUI code in your next patch.
```C#
if (condition)
{ // next line
  // ...
}
```

lwc0917:
@rabbitfor  I missed it i will mind the rule. Thans you for reviewing.

PR #2859, opened 2021-04-08
[NUI] Add null check in LayoutController.Process

huiyueun:
add null check,
It can try to layout the view that has already been released.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
Looks good to me.
Should add null check for `view`

PR #2860, opened 2021-04-08
[NUI] Add Obsolete attribute on deprecated property setter

myroot:
### Description of Change ###

 Remove worng Obsolete attribute
 Some Property setter was obsoleted but, Obsolete attribute make getter obsolete also. it is annoying a developer
 These obsolete attribute was removed but someone remained, So I remove it

##  Updated!! 
I found way to set Obsolete on property setter, it need to use C# version 8
I change it on some properties, but i can't apply all deprecated setter, because it is not my job. 
I hope that someone complete this PR.

I will close it when volunteer is coming

### API Changes ###

 - ACR:


huiyueun (approved):
LGTM

PR #2861, opened 2021-04-08
[NUI] AppBar code refactoring

jaehyun0cho:
To enhance usability, some APIs are added and removed as follows.

Title and Actions are added to set these properties with constructor.

SetTitle(), AddActions() and RemoveActions() are removed because the
above Title and Actions cover theses methods.

Comment's indentation is also fixed.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
+++ b/src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
@@ -14,8 +14,9 @@
  * limitations under the License.
  *
  */
-using System;
+
 using System.ComponentModel;
+using System.Collections.Generic;
 using Tizen.NUI.BaseComponents;
 
 namespace Tizen.NUI.Components
@@ -29,13 +30,16 @@ namespace Tizen.NUI.Components
     {
         private bool autoNavigationContent = true;
 
+        private string title = null;
+
+        private View navigationContent = null;
+        private View titleContent = null;
+        private View actionContent = null;
+        private IEnumerable<View> actionContentViews = null;
+
         private View defaultNavigationContent = null;
         private View defaultTitleContent = null;
         private View defaultActionContent = null;
-
-        private View appBarNavigation = null;
-        private View appBarTitle = null;
-        private View appBarAction = null;
 
         private AppBarStyle appBarStyle => ViewStyle as AppBarStyle;
 
@@ -84,18 +88,18 @@ namespace Tizen.NUI.Components
 
             if (type == DisposeTypes.Explicit)
             {
-                if (appBarNavigation == defaultNavigationContent)
+                if (navigationContent == defaultNavigationContent)
                 {
-                    if (appBarNavigation != null)
+                    if (navigationContent != null)
                     {
-                        Utility.Dispose(appBarNavigation);
+                        Utility.Dispose(navigationContent);
                     }
                 }
                 else
                 {
-                    if (appBarNavigation != null)
+                    if (navigationContent != null)
                     {
-                        Utility.Dispose(appBarNavigation);
+                        Utility.Dispose(navigationContent);
                     }
 
                     if (defaultNavigationContent != null)
@@ -104,18 +108,18 @@ namespace Tizen.NUI.Components
                     }
                 }
 
-                if (appBarTitle == defaultTitleContent)
+                if (titleContent == defaultTitleContent)
                 {
-                    if (appBarTitle != null)
+                    if (titleContent != null)
                     {
-                        Utility.Dispose(appBarTitle);
+                        Utility.Dispose(titleContent);
                     }
                 }
                 else
                 {
-                    if (appBarTitle != null)
+                    if (titleContent != null)
                     {
-                        Utility.Dispose(appBarTitle);
+                        Utility.Dispose(titleContent);
                     }
 
                     if (defaultTitleContent != null)
@@ -124,18 +128,18 @@ namespace Tizen.NUI.Components
                     }
                 }
 
-                if (appBarAction == defaultActionContent)
+                if (actionContent == defaultActionContent)
                 {
-                    if (appBarAction != null)
+                    if (actionContent != null)
                     {
-                        Utility.Dispose(appBarAction);
+                        Utility.Dispose(actionContent);
                     }
                 }
                 else
                 {
-                    if (appBarAction != null)
+                    if (actionContent != null)
                     {
-                        Utility.Dispose(appBarAction);
+                        Utility.Dispose(actionContent);
                     }
 
                     if (defaultActionContent != null)
@@ -158,22 +162,22 @@ namespace Tizen.NUI.Components
         {
             get
             {
-                return appBarNavigation;
+                return navigationContent;
             }
             set
             {
-                if (appBarNavigation == value)
+                if (navigationContent == value)
                 {
                     return;
                 }
 
-                if (appBarNavigation != null)
+                if (navigationContent != null)
                 {
-                    Remove(appBarNavigation);
+                    Remove(navigationContent);
                 }
 
-                appBarNavigation = value;
-                if (appBarNavigation == null)
+                navigationContent = value;
+                if (navigationContent == null)
                 {
                     return;
                 }
@@ -183,32 +187,65 @@ namespace Tizen.NUI.Components
         }
 
         /// <summary>
+        /// Title text of AppBar.
+        /// SetTitle sets title text to the default title content.
+        /// If TitleContent is not TextLabel, then Title does not set title text of the TitleContent.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public string Title
+        {
+            get
+            {
+                return title;
+            }
+            set
+            {
+                if (title == value)
+                {
+                    return;
+                }
+
+                title = value;
+
+                if (TitleContent is TextLabel textLabel)
+                {
+                    textLabel.Text = title;
+                }
+            }
+        }
+
+        /// <summary>
         /// Title content of AppBar. TitleContent is added to Children automatically.
-        /// If TitleContent is set by user, then SetText does not set title text of the replaced TitleContent.
+        /// If TitleContent is not TextLabel, then Title does not set title text of the TitleContent.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
         public View TitleContent
         {
             get
             {
-                return appBarTitle;
+                return titleContent;
             }
             set
             {
-                if (appBarTitle == value)
+                if (titleContent == value)
                 {
                     return;
                 }
 
-                if (appBarTitle != null)
+                if (titleContent != null)
                 {
-                    Remove(appBarTitle);
+                    Remove(titleContent);
                 }
 
-                appBarTitle = value;
-                if (appBarTitle == null)
+                titleContent = value;
+                if (titleContent == null)
                 {
                     return;
+                }
+
+                if (titleContent is TextLabel textLabel)
+                {
+                    textLabel.Text = Title;
                 }
 
                 ResetContent();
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
```diff
@@ -215,12 +252,60 @@ public View TitleContent
...
         }
 
+        /// <summary>
+        /// Action views of AppBar.
+        /// Action views are added to ActionContent of AppBar.
+        /// The Action and ActionButton styles of AppBarStyle are applied to actions only by setting Actions.
+        /// If you do not want to apply Action and ActionButton styles to action views, then please use ActionContent.Add() instead.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public View[] Actions
```
rabbitfor, line 262: Please check build warning : Properties should not return arrays.

jaehyun0cho: Thank you for your comment!
I have removed the getter of Actions.

Seoyeon2Kim: It looks like these two warnings are because Actions have only setter.
Should we suppress the messages of CA1044 and CA1819?

myroot: Do not use Array on public API

myroot: Below code will be help to implement it

 https://github.com/xamarin/Xamarin.Forms/blob/caab66bcf9614aca0c0805d560a34e176d196e17/Xamarin.Forms.Core/Page.cs#L130

https://github.com/xamarin/Xamarin.Forms/blob/caab66bcf9614aca0c0805d560a34e176d196e17/Xamarin.Forms.Core/Page.cs#L62-L64

https://github.com/xamarin/Xamarin.Forms/blob/caab66bcf9614aca0c0805d560a34e176d196e17/Xamarin.Forms.Core/Page.cs#L530-L536

myroot: Normally, ObservableCollection is used to handle adding and removing item by API user

But, it does not expose original type, due to below rule
✔️ DO use the least-specialized type possible as a parameter type. Most members taking collections as parameters use the IEnumerable<T> interface.

myroot:
We made a guidelines for API design and I hope to NUI teams also following this guidelines

https://code.sec.samsung.net/confluence/display/API/Framework+Design+Guidelines

It was based on .net framework desigin guidelines
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
jaehyun0cho: @rabbitfor @Seoyeon2Kim @myroot 
Could you let me know which one would be better between the followings?

**1. Use Property with ReadOnlyCollection**
```
// Property
public ReadOnlyCollection<View> Actions
{
    get
    {
       ...
    }
    set
    {
        ...
    }
}

// Usage
var page = new ContentPage()
{
    AppBar = new AppBar()
    {
        Actions = new ReadOnlyCollection<View>(new List<view> { new Button() { ... }, ... }),
    },
    ...
};
```

**2. Use Method instead of Property**
```
// Method
public void SetActions(params View[] actions)
{
    ...
}

// Usage
var page = new ContentPage()
{
    ...
};

var appBar = new AppBar();
appBar.SetActions(new Button() { }, ...);

page.AppBar = appBar;
```

myroot: I prefer readonly property, because method can't use on XAML directly

for example 
``` xml
   <AppBar>
      <AppBar.Actions>
          <Button Text="....">
          <Button Text="....">
     </AppBar.Actions>
  </AppBar>
```

myroot: But I don't know that you guys are willing to support XAML and MVVM(data binding)

jaehyun0cho:
@myroot 
Thank you for your comments and guides.
I have updated patch to use ReadOnlyCollection<View> type for the Actions property.

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
```diff
@@ -215,12 +253,60 @@ public View TitleContent
...
         }
 
+        /// <summary>
+        /// Action views of AppBar.
+        /// Action views are added to ActionContent of AppBar.
+        /// The Action and ActionButton styles of AppBarStyle are applied to actions only by setting Actions.
+        /// If you do not want to apply Action and ActionButton styles to action views, then please use ActionContent.Add() instead.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public ReadOnlyCollection<View> Actions
```
myroot, line 263: use IEnumerable < View >

jaehyun0cho: I have updated the code.
Thank you!

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
```diff
@@ -215,12 +253,60 @@ public View TitleContent
...
+                {
+                    return;
+                }
+
+                foreach (var action in appBarActionViews)
+                {
+                    // Apply Action and ActionButton styles.
+                    if ((action is Button) && (appBarStyle?.ActionButton != null))
+                    {
+                        ((Button)action)?.ApplyStyle(appBarStyle.ActionButton);
```
myroot, line 294: No need cast and null operator, `ApplyStyle` is ovrride method, and action never null
```suggestion
                        action.ApplyStyle(appBarStyle.ActionButton);
```

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
```diff
@@ -182,9 +184,45 @@ public View NavigationContent
...
+                    return null;
+                }
+
+                if (TitleContent is TextLabel)
+                {
+                    return ((TextLabel)TitleContent).Text;
+                }
+
+                return null;
+            }
```
myroot, line 208:
```suggestion
            get
            {
                if (TitleContent is TextLabel label)
                {
                    return label.Text;
                }
                return null;
            }
```

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
```diff
@@ -182,9 +184,45 @@ public View NavigationContent
             }
         }
 
+        /// <summary>
+        /// Title text of AppBar.
+        /// SetTitle sets title text to the default title content.
+        /// If TitleContent is not TextLabel, then Title does not set title text of the TitleContent.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public string Title
```
myroot, line 193: I think no need this property, 

Because, TitleContent was provide by API user, So, he can control value of Text by himself. 
It just make a confusing

``` c#
 appbar.Title = "aaa";
 assert(appbar.Title == "aaa"); // it will be fail if TitleContent is not TextLabel
```
``` c#
 appbar.TitleContent = new TextLabel();
 appbar.Title = "aaa";
 assert(appbar.Title == "aaa"); // it is pass

 appbar.TitleContent = new TextLabel();
 assert(appbar.Title == "aaa"); // it is fail
```

myroot: If a you want to give a responsibility of set a title on AppBar, 
You need to hold a title value, and you should update when TitleContent was updated
like below
``` c#
  public string Title {get; set;}
 
  public View TitleContent
  {
       set 
       {
             if (value is TextLabel label) 
             {
                   label.Text = Title;
             }
       }
  }

```

myroot: I want to request to you that separate between data and UI, in this case, Title is data, View is UI, 

UI could be change by want, but Data(title) should not be reflected by those changes.
And
Title also could be changed by want, but UI no need to change

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
```diff
@@ -215,12 +253,60 @@ public View TitleContent
...
+                    if ((action is Button) && (appBarStyle?.ActionButton != null))
+                    {
+                        ((Button)action)?.ApplyStyle(appBarStyle.ActionButton);
+                    }
+                    else if (appBarStyle?.ActionView != null)
+                    {
+                        action?.ApplyStyle(appBarStyle.ActionView);
+                    }
+
+                    ActionContent.Add(action);
```
myroot, line 301: ActionContent could be a null
```suggestion
                    ActionContent?.Add(action);
```

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
```diff
@@ -215,12 +253,60 @@ public View TitleContent
...
+                    }
+                }
+
+                appBarActionViews = value;
+
+                if (appBarActionViews == null)
+                {
+                    return;
+                }
+
```
myroot, line 288: You should check ActionContent was not null
```suggestion
    if (ActionContent == null)
         return;
```

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
jaehyun0cho: Thank you for your comments.
I understand your point and I would like to explain why I added "Title" property.

As you said, user can set TextLabel to "TitleContent".
Since NUI supports theme style by its type (e.g. TextLabelStyle is defined for default TitleContent), if "TitleContent" is not TextLabel, then we cannot apply Tizen UX for AppBar Title to the "TitleContent".

In other words, if user sets any Control to "TitleContent" except TitleLabel, then Tizen UX for AppBar Title is not applied.

So I wanted to provide "Title" for most of users who just wants to set text on Title area.
If user wants to set something else, then they can use "TitleContent".

jaehyun0cho: I considered your implementation before, but I had a problem as the following case.

```
appbar.TitleContent = new TextLabel();
appbar.Title = "aaa";
(appbar.TitleContent as TextLabel).Text = "bbb";
```

In the above case, the data "Title" would be still "aaa", then it is not synchronized with the UI text "TitleContent.Text".

I thought that it is not that good so I chose the current implementation.
I would like to change the current implementation to your suggestion if the above case would be fine from application developer's perspective.

myroot: I think, it is not problem, 
Because It is **intent** of user,

This rule was also applied on data binding

If user set a value on property that was binding some value, 
how to work? 
in this case binding was broken because it is intent of user. 

I think it is basically same with our case

What is your opinion on this case?
1. `appbar.Title` should be "bbb"?
2. or `(appbar.TitleContent as TextLabel).Text` should not change to "bbb" so still it's value was "aaa"?

if your think is 1, 
We can track changes of Text on TextLabel by using `PropertyChanged` event and update a value of `Title`
But, I don't think it is need

jaehyun0cho: Thank you very much for your answer!
Your comment is so clear that I think I can apply the way of thinking to other cases as well.

About the preference between 1 and 2, I agree with you that I don't think 1 is needed.
And 2 is not what I want.

So I am going to apply your suggestion about implementing "Title" and "TitleContent".
Thank you :)

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
jaehyun0cho: Thank you for your comment.
The code has been updated.

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
jaehyun0cho: Thank you for your comment.
I have updated Title property as you suggested below.

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
jaehyun0cho: Thank you for your comment.
The code has been updated.

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
jaehyun0cho: Thank you for your comment.
I have updated code to do null check ActionContent in the beginning as you suggested below.

src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
jaehyun0cho: Thank you for your comment.
The code has been updated.

myroot (approved):

PR #2862, opened 2021-04-08
[NUI] Change TouchArea to TouchAreaOffset

JoogabYun:
Change TouchArea operation

You can set offset of the touch area.

for example)
  View view = new View();
  view.Size = new Size(10, 10);
  view.TouchEvent += OnTouch;
  view.SetTouchAreaOffset(-100, 100, 100, -100); // left, right, bottom, top

then touch area is 210x210.
this is view.width + TouchAreaOffset.right - TouchAreaOffset.left
and view.height + TouchAreaOffset.bottom -TouchAreaOffset.top

### API Changes ###

 - ACR:


huiyueun (approved):
LGTM

everLEEst (commented):
Looks good to me but someone else need to approve.

dongsug-song:
@JoogabYun 
could you update the csharp-binder's patch on this PR? and csharp-binder patch should be merged prior to this PR. otherwise there can be a runtime exception which blocks the release of tizenfx by TCT hurdle

JoogabYun:
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/256790/

I will merge this NUI patch after the csharp binder is merged.

src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
```diff
@@ -1290,21 +1290,22 @@ private void OnWindowWheelEvent(object sender, Window.WheelEventArgs e)
         }
 
         /// <summary>
-        /// TouchArea can reset the view's touchable area.<br/>
-        /// If you set the TouchArea on an view, when you touch the view, the touch area is used rather than the size of the view.
+        /// TouchArea can expand the view's touchable area.<br/>
+        /// If you set the TouchAreaOffset on an view, when you touch the view, the touch area is used rather than the size of the view.<br/>
+        /// TouchAreaOffset = new Rectangle(left, right, bottom, top);
```
myroot, line 1295: Is it not possible to extend like a margin?
If I want to extend touch area just right size, could I do?
if it possible `Extents` more proper

JoogabYun: You can only extend one of left, right, bottom and top.
This is based on the top left x, y coordinates.

These coordinates can have negative value. so I can not using 'Extents'.
because 'Extents' is unsigned int value.

myroot: How to use to extend 10 pixel for 4 direction (left, right, bottom, top)?
image

1. new Rectangle(-10, 10, 10, -10)
2. new Rectangle(-10, View.Size.Width + 10, View.Size.Height + 10, -10);

And there is no constructor to accept left, right, bottom, top

JoogabYun: 1. new Rectangle(-10, 10, 10, -10)
This is why I added an 'offset' to the name. 

 there is no constructor to accept left, right, bottom, top
--> sorry,  please explain what you are saying.

myroot: Do not use `Rectangle` class, it not proper to represent offset.
How about think about it?
``` c# 
var is_it_a_width_of_touch_area = view.TouchAreaOffset.Width ; // ???
```
>there is no constructor to accept left, right, bottom, top

See it, no constructor that has arguments as `(int left, int right, int top, int down)`
https://github.com/Samsung/TizenFX/blob/c6147db7131e1b759a1d2da515df3d3af361426a/src/Tizen.NUI/src/public/Common/Rectangle.cs#L34-L74

I recommend to extend `Vector4` class to represent four direction offset

JoogabYun: Oh I see, I will consider it.

JoogabYun:
conflicts 발생으로 https://github.com/Samsung/TizenFX/pull/2948로 패치를 다시 올립니다.

PR #2863, opened 2021-04-08
[NUI] Change CornerRadius type to Vector4 from float

rabbitfor:
Note that, only X value of Vector4 will be applied to the view until dali support it.

### API Changes ###

 - ACR:


huiyueun (approved):

src/Tizen.NUI/src/internal/Common/PropertyHelper.cs
```diff
@@ -209,6 +209,17 @@ private static object ObjectIntToFloat(object value)
...
 
+        private static object ObjectVector4ToFloat(object value)
+        {
+            Type type = value.GetType();
+            if (type.Equals(typeof(Vector4)))
+            {
+                return ((Vector4)value).X;
+            }
+
+            return value;
```
myroot, line 220: It is type casting of normal C# programmer, I wonder what is your intent?
``` c#
 if (value is Vector4 vector) 
 {
    return vector.X;
 }
 return value;
```

rabbitfor: I took your opinion in my next patch https://github.com/Samsung/TizenFX/pull/2866.

PR #2871, opened 2021-04-09
[NUI] Implement some new APIs of web view.

huayongxu:
This patch is to add some new APIs such as PageZoomFactor,
GetScreenshotAsynchronously, etc

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/WebView/WebView.cs
+++ b/src/Tizen.NUI/src/internal/WebView/WebView.cs
@@ -30,6 +30,11 @@ namespace Tizen.NUI
     [EditorBrowsable(EditorBrowsableState.Never)]
     public class WebView : View
     {
+        private Vector4 documentBackgroundColor;
+        private bool tilesClearedWhenHidden;
+        private float tileCoverAreaMultiplier;
+        private bool cursorEnabledByClient;
+
         private readonly WebViewPageLoadSignal pageLoadStartedSignal;
         private EventHandler<WebViewPageLoadEventArgs> pageLoadStartedEventHandler;
         private WebViewPageLoadCallbackDelegate pageLoadStartedCallback;
@@ -180,6 +185,27 @@ namespace Tizen.NUI
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
         [EditorBrowsable(EditorBrowsableState.Never)]
         public delegate void JavaScriptPromptCallback(string message1, string message2);
+
+        /// <summary>
+        /// The callback function that is invoked when screen shot is captured.
+        /// </summary>
+        [UnmanagedFunctionPointer(CallingConvention.StdCall)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public delegate void ScreenshotCapturedCallback(ImageView image);
+
+        /// <summary>
+        /// The callback function that is invoked when video playing is checked.
+        /// </summary>
+        [UnmanagedFunctionPointer(CallingConvention.StdCall)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public delegate void VideoPlayingCallback(bool isPlaying);
+
+        /// <summary>
+        /// The callback function that is invoked when geolocation permission is requested.
+        /// </summary>
+        [UnmanagedFunctionPointer(CallingConvention.StdCall)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public delegate void GeolocationPermissionCallback(string host, string protocol);
 
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
         private delegate void WebViewPageLoadCallbackDelegate(IntPtr data, string pageUrl);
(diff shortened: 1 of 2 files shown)


huayongxu:
This patch is related to patches at csharp-binder side:
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/254454/
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/254552/

src/Tizen.NUI/src/internal/WebView/WebView.cs
```diff
@@ -181,6 +186,27 @@ protected override void Dispose(DisposeTypes type)
         [EditorBrowsable(EditorBrowsableState.Never)]
         public delegate void JavaScriptPromptCallback(string message1, string message2);
 
+        /// <summary>
+        /// The callback function that is invoked when screen shot is captured.
+        /// </summary>
+        [UnmanagedFunctionPointer(CallingConvention.StdCall)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public delegate void ScreenshotCapturedCallback(ImageView image);
```
rabbitfor, line 194: I wonder why they are not events.

src/Tizen.NUI/src/internal/WebView/WebView.cs
```diff
@@ -662,6 +749,90 @@ public bool KeyEventsEnabled
             }
         }
 
+        /// <summary>
+        /// Background color of web page.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Vector4 DocumentBackgroundColor
```
rabbitfor, line 756: When I first commented about the name, I though it indicates [document](https://developer.mozilla.org/en-US/docs/Web/API/Document) of the web page literally.
스크린샷 2021-04-09 오후 4 23 55

Is it correct? The API sets the background color of `document`?

src/Tizen.NUI/src/internal/WebView/WebView.cs
```diff
@@ -662,6 +749,90 @@ public bool KeyEventsEnabled
...
+        {
+            get
+            {
+                return documentBackgroundColor;
+            }
+            set
+            {
+                documentBackgroundColor = value;
+                SetValue(DocumentBackgroundColorProperty, value);
```
rabbitfor, line 765: Why don't you put this line `documentBackgroundColor = value;` into the corresponding bindable property setter?
The logic should be same in both property setter and the bindable property setter.

src/Tizen.NUI/src/internal/WebView/WebView.cs
```diff
@@ -858,6 +1167,36 @@ public void LoadHtmlString(string data)
...
         }
 
+        /// <summary>
+        /// Loads the specified html as the content of the view to override current history entry.
+        /// <param name="html">The html to be loaded</param>
+        /// <param name="basicUri">Basic URL used for relative paths to external objects</param>
+        /// <param name="unreachableUri">URL that could not be reached</param>
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool LoadHtmlStringOverrideCurrentEntry(string html, string basicUri, string unreachableUri)
```
rabbitfor, line 1177: The term `basicUri` and `baseUri` are different? In LoadContents, `baseUri` is used but here it is `basicUri`.
I'd like to know the differences.

src/Tizen.NUI/src/internal/WebView/WebView.cs
huayongxu: This callback is called when getting screen shot asynchronously:
public bool GetScreenshotAsynchronously(Rectangle viewArea, float scaleFactor, ScreenshotAcquiredCallback callback)
you mean,  ScreenshotCapturedCallback is renamed to ScreenshotCapturedEventHandler?
but if that, there would be no difference between callback and real event hanlder, such as PageLoadErrorEventHandler.

src/Tizen.NUI/src/internal/WebView/WebView.cs
huayongxu: after this property is set, background color of ewk view of chromium-efl would be changed.
so it is the same as background color of NUI view.
here this property would be overridden.

src/Tizen.NUI/src/internal/WebView/WebView.cs
huayongxu: Done.

src/Tizen.NUI/src/internal/WebView/WebView.cs
huayongxu: Done.
basicUri is ranamed to baseUri.

src/Tizen.NUI/src/internal/WebView/WebView.cs
```diff
@@ -662,6 +749,86 @@ public bool KeyEventsEnabled
             }
         }
 
+        /// <summary>
+        /// Background color of web view.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public new Color BackgroundColor
```
rabbitfor, line 756: Original `BackgroundColor` is a background visual's color. Background visual can have larger size than a view.  And the property is tightly connect to a `Background` and a `BackgroundImage` properties. So overriding original logic has a lot of potential problem.
Plus, it is not good that a property and a bindable property have different name: BackgroundColor and WebViewBackgroundColor.

I think setting WebView's background should be done in DALi.
(When a user sets the BackgroundColor property, a WebView control should set both actor's background color and ewk's background color).
But here we have no time to update DALi again, I think `ContentBackgroundColor` is enough for now.

src/Tizen.NUI/src/internal/WebView/WebView.cs
rabbitfor: The events we already provide, such as, PageLoadFinished is used like,
```C#
webView.PageLoadeFinished += OnPageLoadFinished;
webView.LoadUrl(url);
// OnPageLoadFinished is called when the page is loaded.
```

So I though the new callbacks have no different with PageLoadFinished.
```C#
webView.ScreenShotAcquired += OnScreenShotAcquired;
webView.TakeScreenshot(viewArea, scaleFactor);
```

src/Tizen.NUI/src/internal/WebView/WebView.cs
huayongxu: Done.

rabbitfor: Please make sure the name is not using `Document` unless it really means the document's background color. I assume `ContentBackgroundColor` is enough as I commented.

huayongxu: Done.

src/Tizen.NUI/src/internal/WebView/WebView.cs
huayongxu: All is ok.
then RegisterJavaScriptAlertCallback/RegisterJavaScriptConfirmCallback/RegisterJavaScriptPromptCallback need be changed too?

huayongxu: Sorry, I checked and have another question.
If RegisterGeolocationPermissionCallback  is changed to event,
many event handlers can be added.
but RegisterGeolocationPermissionCallback had better be called only once.

rabbitfor: If so, when user call `RegisterGeolocationPermissionCallback` multiple times, only last one registered?

huayongxu: yes, only last one is valid.

rabbitfor: OK then I will go with your current code :)

rabbitfor (approved):
LGTM

zg2nets (approved):
looks good to me！

AchoWang (approved):
Looks good for me.

elishateng (approved):
LGTM

zhouleonlei (approved):
LGTM

AdunFang (approved):
It is OK.

PR #2875, opened 2021-04-09
[NUI] Enable TextLabel's BindableProperties to be affected by Layout change

dongsug-song:
### Description of Change ###
- for the first step, I have changed TextLabel but this is not entire fix.
- after get review and if my approach is right, then I will fix all Text related BindableProperty.

### API Changes ###
none


src/Tizen.NUI/src/public/BaseComponents/TextLabel.cs
```diff
@@ -235,8 +235,7 @@ public string Text
             set
             {
                 SetValue(TextProperty, value);
-                selectorData?.Text?.Reset(this);
-                NotifyPropertyChangedAndRequestLayout();
+                NotifyPropertyChanged();
```
myroot, line 238: This method trigger `PropertySet` event, I wonder what is different from `PropertyChanged` event, 
What means of `PropertySet`? is it no need to trigger when value was changed by data binding?

src/Tizen.NUI.Components/Controls/Slider.cs
```diff
@@ -161,6 +161,35 @@ public partial class Slider : Control
...
         });
 
+        /// <summary>
+        /// Bindable property of CurrentValue
+        /// <remark>
+        /// Hidden API, used for NUI XAML data binding
+        /// </remark>
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly BindableProperty CurrentValueProperty = BindableProperty.Create(nameof(CurrentValue), typeof(float), typeof(Slider), 0.0f,
```
myroot, line 171: It need to `TwoWay` binding, CurrentValue could be changed by data source and source also could be changed by UI
```suggestion
        public static readonly BindableProperty CurrentValueProperty = BindableProperty.Create(nameof(CurrentValue), typeof(float), typeof(Slider), 0.0f, BindingMode.TwoWay
```

src/Tizen.NUI/src/public/BaseComponents/TextLabel.cs
dongsug-song: thank you for review. `PropertySet` was added before the NUI XAML added. (NUI XAM has been copied from Xamarin.Forms open source) and it's role is exactly same as `PropertyChanged'. I can move all `NotifyPropertyChanged();` into `SetValue()` but I just want to get your guide whether removal of this `PropertySet' is better for usability? if so, then I will do ACR.

myroot: @dongsug-song Yes, `PropertySet` event should be deprecated, if it is same with `PropertyChanged`, It make a confused and inconsistent

dongsug-song: @myroot 
OK, I see. I will proceed the ACR (https://code.sec.samsung.net/jira/browse/TCSACR-413) and will make it deprecated.

myroot (approved):

PR #2876, opened 2021-04-09
[NUI] Page and Dialog code refactoring

jaehyun0cho:
### Description of Change ###

Page and Dialog code refactoring.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/AlertDialog.cs
+++ b/src/Tizen.NUI.Components/Controls/AlertDialog.cs
@@ -1,5 +1,5 @@
 /*
- * Copyright(c) 2020 Samsung Electronics Co., Ltd.
+ * Copyright(c) 2021 Samsung Electronics Co., Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


jaehyun0cho:
This PR should NOT be merged by "Squash and merge".
**This PR should be merged by "Rebase and merge".**

src/Tizen.NUI.Components/Controls/AlertDialog.cs
```diff
@@ -17,154 +17,95 @@
...
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        Negative
-    }
-
     /// <summary>
     /// AlertDialog class shows a dialog with title, message and action buttons.
     /// </summary>
     [EditorBrowsable(EditorBrowsableState.Never)]
-    public class AlertDialog : Dialog
+    public class AlertDialog : Control
```
dongsug-song, line 29: I am not sure, but is the Dialog class removed?
I think Control is base class and AlertDialog seems to be a kind of Dialog class, so I think it looks better the previous change of AlertDialog inherits from Dialog. 
is there any reason to change the base class?

src/Tizen.NUI.Components/Controls/AlertDialog.cs
```diff
@@ -175,425 +116,369 @@ public View TitleContent
...
         }
 
         /// <summary>
-        /// Title text of AlertDialog.
+        /// Action views of AlertDialog.
+        /// Action views are added to ActionContent of AlertDialog.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public string Title
+        public IEnumerable<View> Actions
```
dongsug-song, line 216: I think it is very good to use IEnumerable as a return type. is it of the purpose not to use `array[]` as a return type?
I am not sure, but if IEnumerable is used, IEnumerator should be defined. isn't it? I don't know very well so I am just asking to figure out together. ^^;
I saw this : https://docs.microsoft.com/en-us/dotnet/api/system.collections.ienumerable?view=net-5.0

src/Tizen.NUI.Components/Controls/AlertDialog.cs
```diff
@@ -175,425 +116,369 @@ public View TitleContent
...
             {
-                return Message;
+                parentSize = NUIApplication.GetDefaultWindow().Size;
             }
+
+            Position2D = new Position2D((parentSize.Width - size.Width) / 2, (parentSize.Height - size.Height) / 2);
         }
     }
-}
+}
```
dongsug-song, line 484: could you add new line at the end of file?

reference: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

src/Tizen.NUI.Components/Controls/Navigation/Page.cs
```diff
@@ -40,166 +39,37 @@ public class PageDisappearingEventArgs : EventArgs
     /// The Page class is a class which is an element of navigation.
     /// </summary>
     [EditorBrowsable(EditorBrowsableState.Never)]
-    public class Page : Control
+    public abstract class Page : Control
```
dongsug-song, line 42: could you explain why this is changed to `abstract` class?

src/Tizen.NUI.Components/Controls/AlertDialog.cs
jaehyun0cho: Originally, Scrim (dimmed background area) was a property of Dialog. So AlertDialog was derived from Dialog.
But now, Scrim was a property of DialogPage and Dialog and AlertDialog only shows the popup content itself.

Dialog has only one content as follows.
- View Content

AlertDialog has three contents as follows.
- View TitleContent
- View Content
- View ActionContent

So to inherit Dialog, we have either 2 ways.
1. The positioning contents logic may be different between Dialog and AlertDialog
   Dialog positions Content at the center of the screen.
   AlertDialog positions 3 contents at the center of the screen.

OR

2. AlertDialog sets its 3 contents into Dialog.Content.
   (i.e. AlertDialog's TitleContent, BodyContent, ActionContent -> Dialog.Content)

I thought that AlertDialog may not be derived from AlertDialog because Scrim is no longer Dialog's property and it would be complicated the above.

Could you give me your opinion about the above?

src/Tizen.NUI.Components/Controls/AlertDialog.cs
jaehyun0cho: I was guided by https://github.com/Samsung/TizenFX/pull/2861#discussion_r609570342 and https://github.com/Samsung/TizenFX/pull/2861#discussion_r609643301

The above comments refer https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections and it says ArrayList or List<T> are not recommended because it does not notice when user modifies the collections.

So IEnumerable can be used instead of ArrayList or List<T>.

src/Tizen.NUI.Components/Controls/Navigation/Page.cs
jaehyun0cho: Because I think ContentPage and DialogPage are used for basic usages.

There would be more usages related to Page. So maybe more Page classes would be added later.

To give clear usage related to Page classes, I changed Page from class to abstract class.
ContentPage can be used for the full screen formatted Page cases (Title with Content layout).

In Xamarin forms, Page is not abstract class but it is hard to find Page class is used instead of other specific Page classes.
https://docs.microsoft.com/ko-kr/dotnet/api/xamarin.forms.page?view=xamarin-forms
https://docs.microsoft.com/ko-kr/xamarin/xamarin-forms/user-interface/controls/pages

src/Tizen.NUI.Components/Controls/AlertDialog.cs
jaehyun0cho: Thank you for your comment.
The code has been updated.

jaehyun0cho:
dialog_page_diagram

src/Tizen.NUI.Components/Controls/Navigation/Page.cs
dongsug-song: OK, I see. thank you for explanation.

src/Tizen.NUI.Components/Controls/AlertDialog.cs
dongsug-song: OK, I see. Thank you for explanation.

src/Tizen.NUI.Components/Controls/AlertDialog.cs
dongsug-song: OK, I see. I think it's good to go. thank you for explanation.

dongsug-song (approved):
looks good to me

PR #2878, opened 2021-04-12
[NUI] Make Palette APIs public

lwc0917:
### Description of Change ###
Palette API ACR

### API Changes ###
https://code.sec.samsung.net/jira/browse/TCSACR-409


src/Tizen.NUI/src/public/Utility/Palette.cs
```diff
@@ -153,7 +154,6 @@ public static void GenerateAsync(PixelBuffer pixelBuffer, Tizen.NUI.Rectangle re
         /// <param name="pixelBuffer">A Target image's pixelBuffer.</param>
         /// <exception cref="ArgumentNullException">Thrown when the argument pixelBuffer is null.</exception>
         /// <returns>the palette instance.</returns>
-        [EditorBrowsable(EditorBrowsableState.Never)]
         public static Palette generate(PixelBuffer pixelBuffer)
```
wiertel, line 157: generate -> Generate, lowercase name for public method is against design guide

src/Tizen.NUI/src/public/Utility/Palette.cs
```diff
@@ -204,7 +203,6 @@ public static Palette Generate(PixelBuffer pixelBuffer, Rectangle region)
         /// Returns all of the swatches which make up the palette.
         /// </summary>
         /// <returns>The swatch list</returns>
-        [EditorBrowsable(EditorBrowsableState.Never)]
         public ReadOnlyCollection<Swatch> GetSwatches()
```
wiertel, line 206: it's recommended to declare returning interfaces rather than specific classes, Please add one letter to signature and keep the implementation as it is
```suggestion
        public IReadOnlyCollection<Swatch> GetSwatches()
```

src/Tizen.NUI/src/public/Utility/Palette.cs
```diff
@@ -127,7 +130,6 @@ private Palette(List<Swatch> swatcheList)
         /// <param name="pixelBuffer">A Target image's pixelBuffer.</param>
         /// <param name="paletteGeneratedEventHandler">A method will be called with the palette when generated.</param>
         /// <exception cref="ArgumentNullException">Thrown when the argument pixelBuffer, PaletteGeneratedEventHandler is null.</exception>
-        [EditorBrowsable(EditorBrowsableState.Never)]
         public static void GenerateAsync(PixelBuffer pixelBuffer, PaletteGeneratedEventHandler paletteGeneratedEventHandler)
```
wiertel, line 133: please use C# features for asynchronous methods. An asynchronous method should return [Task<T>](https://docs.microsoft.com/en-gb/dotnet/api/system.threading.tasks.task-1?view=netcore-3.0) or Task. Optionally be marked with `async` keyword - but you probably will not need that. This way user will call:
```
   var pallete = await Pallete.GenerateAsync(buffer);
```

I've noticed that you've already implemented a Task returning method, so just return that this task and don't expose the callback.

src/Tizen.NUI/src/public/Utility/Palette.cs
```diff
@@ -584,7 +570,6 @@ public Tizen.NUI.Color GetBodyTextColor()
             /// return the number of pixels represented by this swatch.
```
wiertel, line 570: return -> Returns, and I'm not sure about "represented by" - who is representing who? maybe "pixels that belong to this swatch", "pixels that has been assigned to this swatch", "pixels matching this swatch" - it depends on how you would describe the process of deciding if a pixel is related to a swatch.

wiertel (changes requested):

src/Tizen.NUI/src/public/Utility/Palette.cs
```diff
@@ -141,7 +143,6 @@ public static void GenerateAsync(PixelBuffer pixelBuffer, PaletteGeneratedEventH
         /// <param name="region">A rectangle used for region.</param>    
         /// <param name="paletteGeneratedEventHandler">A method will be called with the palette when generated.</param>
         /// <exception cref="ArgumentNullException">Thrown when the argument pixelBuffer, PaletteGeneratedEventHandler is null.</exception>
-        [EditorBrowsable(EditorBrowsableState.Never)]
```
Seoyeon2Kim, line 144: After removing this tag, should we add 'since tizen' tag at the same place?
like -> /// <since_tizen> 8 </since_tizen>

lwc0917:
@wiertel 
Thank you for your review.
I agreed to the reviewed content and modified the code.

src/Tizen.NUI/src/public/Utility/Palette.cs
```diff
@@ -48,7 +48,7 @@ namespace Tizen.NUI
     ///     // Use generated instance
```
wiertel, line 48: the line above is no longer correct, the GenerateAsync method has been changed

also - most of this description should go to \<remarks/\> section - summary should be short, so it would be \<summary\>....\</summary\>\<remarks\>...\</remarks\>
the code examples will not be formatted as in this code, please add \<br/\>in lines 43, 46, 47, 48, 49

src/Tizen.NUI/src/public/Utility/Palette.cs
```diff
@@ -204,18 +208,18 @@ public static Palette Generate(PixelBuffer pixelBuffer, Rectangle region)
         /// Returns all of the swatches which make up the palette.
         /// </summary>
         /// <returns>The swatch list</returns>
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        public ReadOnlyCollection<Swatch> GetSwatches()
+        /// <since_tizen> 9 </since_tizen>
+        public IReadOnlyCollection<Swatch> GetSwatches()
         {
-            return new ReadOnlyCollection<Swatch>(swatches);
+            return swatches as IReadOnlyCollection<Swatch>;
```
wiertel, line 214: the cast (`as`) is probably not necessary

src/Tizen.NUI/src/public/Utility/Palette.cs
lwc0917: Thanks for the guide. The contents will be updated separately on Doc git.

src/Tizen.NUI/src/public/Utility/Palette.cs
lwc0917: Thank you.

jaehyun0cho (approved):
LGTM

PR #2895, opened 2021-04-13
[NUI] Optimizing item positioning on MeasureFirst

everLEEst:
using List<T> Data Structure is redundant in MeasureFirst case,
and it makes worse sometimes for searching indexed item's position.
In MeasureFirst, every item has same size,
thus we can calculate it's position in linear or grouped cases.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #2907, opened 2021-04-14
[NUI] Deprecate BaseHandle.PropertySet and Add BindableProperty.PropertyChanged

dongsug-song:
### Description of Change ###
Deprecate BaseHandle.PropertySet and Add BindableProperty.PropertyChanged
- PropertySet has been added before the NUI XAML is added.
- PropertySet is duplicated of PropertyChanged so it can be deprecated.
- BindableProperty class is parent of BaseHandle class, so PropertyChanged is better to be used.

### API Changes ###
https://code.sec.samsung.net/jira/browse/TCSACR-413


src/Tizen.NUI/src/public/Common/BaseHandle.cs
```diff
@@ -115,6 +115,7 @@ internal BaseHandle(global::System.IntPtr cPtr)
         /// Event when a property is set.
         /// </summary>
         /// <since_tizen> 5 </since_tizen>
```
wiertel, line 117: We could optionally also add a way to quickly jump to the new event.
```suggestion
        /// <since_tizen> 5 </since_tizen>
        /// <seealso cref="BindableObject.PropertyChanged" />
```

dongsug-song: thank you for your review.
it's added.

everLEEst (approved):
LGTM

wiertel (approved):

PR #2912, opened 2021-04-14
[NUI] Add Menu, MenuItem, MenuItemGroup classes

jaehyun0cho:
To support menu displayed at the given anchor position, Menu, MenuItem,
MenuItemGroup classes are added.

Menu adds MenuItems and also adds MenuItems to MenuItemGroup to select
only one item at a time.

Menu is displayed at the AnchorPosition if it is enough space to display
Menu at the AnchorPosition.
Otherwise, Menu is displayed near AnchorPosition to display Menu.

To support Menu easily, static method ShowMenu() is added to DialogPage.

### API Changes ###

 - ACR:


src/Tizen.NUI.Components/Controls/Navigation/DialogPage.cs
```diff
@@ -284,5 +284,65 @@ public static void ShowAlertDialog(string title, string message, params View[] a
...
+            {
+                Content = new Menu()
+                {
+                    Items = items,
+                    AnchorPosition = anchorPosition,
+                },
+                ScrimColor = Color.Transparent,
+            };
+
+            NUIApplication.GetDefaultWindow().GetDefaultNavigator().Push(dialogPage);
```
huiyueun, line 316: Isn't it necessary in a multi-window situation?

jaehyun0cho: Thank you for your comment.
I think it's good point.
For now, static helper methods only supports default window and default navigator for simple form of method.
I think we can overload this method with window and navigator parameters when we meet developers' requests.
```
pubic static void ShowMenu(View anchor, params MenuItem[] items, Window window, Navigator navigator)

huiyueun: Ok

huiyueun (approved):
LGTM

PR #2913, opened 2021-04-14
[NUI] Fix to pass touch event when DialogPage's Scrim is disabled

jaehyun0cho:
To pass touch event when DialogPage's Scrim is disabled, DialogPage's
EnableControlState is set to be false.

### API Changes ###

 - ACR:


huiyueun (approved):
LGTM

PR #2918, opened 2021-04-14
[NUI] Make Dialog APIs public

jaehyun0cho:
This is ACR patch to make Dialog APIs public.

This pull request includes properties and methods of DialogPage, Dialog,
AlertDialog and AlertDialogStyle classes.

### API Changes ###

 - ACR: TCSACR-419

### DIFF
--- a/src/Tizen.NUI.Components/Controls/AlertDialog.cs
+++ b/src/Tizen.NUI.Components/Controls/AlertDialog.cs
@@ -25,7 +25,7 @@ namespace Tizen.NUI.Components
     /// <summary>
     /// AlertDialog class shows a dialog with title, message and action buttons.
     /// </summary>
-    [EditorBrowsable(EditorBrowsableState.Never)]
+    /// <since_tizen> 9 </since_tizen>
     public class AlertDialog : Control
     {
         private string title = null;
@@ -50,16 +50,33 @@ namespace Tizen.NUI.Components
         /// <summary>
         /// Creates a new instance of AlertDialog.
         /// </summary>
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <since_tizen> 9 </since_tizen>
         public AlertDialog() : base()
         {
             Initialize();
         }
 
         /// <summary>
-        /// Dispose AlertDialog and all children on it.
+        /// Creates a new instance of AlertDialog.
         /// </summary>
-        /// <param name="type">Dispose type.</param>
+        /// <param name="style">Creates AlertDialog by special style defined in UX.</param>
+        /// <since_tizen> 9 </since_tizen>
+        public AlertDialog(string style) : base(style)
+        {
+            Initialize();
+        }
+
+        /// <summary>
+        /// Creates a new instance of AlertDialog.
+        /// </summary>
+        /// <param name="alertDialogStyle">Creates AlertDialog by style customized by user.</param>
+        /// <since_tizen> 9 </since_tizen>
+        public AlertDialog(AlertDialogStyle alertDialogStyle) : base(alertDialogStyle)
+        {
+            Initialize();
+        }
+
+        /// <inheritdoc/>
         [EditorBrowsable(EditorBrowsableState.Never)]
         protected override void Dispose(DisposeTypes type)
         {
@@ -100,7 +117,7 @@ namespace Tizen.NUI.Components
         /// Applies style to AlertDialog.
         /// </summary>
         /// <param name="viewStyle">The style to apply.</param>
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <since_tizen> 9 </since_tizen>
         public override void ApplyStyle(ViewStyle viewStyle)
         {
             styleApplied = false;
@@ -133,8 +150,10 @@ namespace Tizen.NUI.Components
 
         /// <summary>
         /// Title text of AlertDialog.
+        /// Title text is set to TitleContent's Text if TitleContent is TextLabel.
+        /// If TitleContent's Text is set manually by user, then it is not guaranteed that TitleContent's Text is the same with Title text.
         /// </summary>
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <since_tizen> 9 </since_tizen>
         public string Title
         {
             get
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI.Components/Controls/AlertDialog.cs
```diff
@@ -160,7 +180,7 @@ public string Title
         /// <summary>
         /// Title content of AlertDialog. TitleContent is added to Children automatically.
         /// </summary>
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <since_tizen> 9 </since_tizen>
```
hyunjushin, line 183: "TitleContent is added as a child of AlertDialog automatically. " is better?

src/Tizen.NUI.Components/Controls/Navigation/DialogPage.cs
```diff
@@ -25,7 +25,7 @@ namespace Tizen.NUI.Components
     /// The DialogPage class is a class which shows a dialog on the page.
     /// DialogPage contains dialog and dimmed scrim behind the dialog.
     /// </summary>
-    [EditorBrowsable(EditorBrowsableState.Never)]
+    /// <since_tizen> 9 </since_tizen>
```
hyunjushin, line 28: DialogPage is a Contol. Why don't you use DialogPage's backgroud color instead of seperate scrim View?

src/Tizen.NUI.Components/Controls/AlertDialog.cs
jaehyun0cho: Thank you for your comment.
I think your description is better for understanding.
The description has been updated.

src/Tizen.NUI.Components/Controls/Navigation/DialogPage.cs
jaehyun0cho: Scrim catches touch event and DialogPage is popped when user touches Scrim region (dimmed background region).

If we make DialogPage do the above action, the touch point should be checked if it is on DialogPage's children.
(Basically, DialogPage has one child which is "Content" but user can add more views to DialogPage.)

Moreover, if DialogPage catches the touch event, DialogPage should check if a child on the touch point is touchable or not. (e.g. if View.Sensitive is false, then View and its children do not get touch events)

I think that catching the touch event by DialogPage requires us to think more cases to implement Scrim's action.
So I think that let the separate scrim view do the Scrim's action would be easier to maintain.

jaehyun0cho:
ACR number is TCSACR-419

dongsug-song:
https://code.sec.samsung.net/jira/browse/TCSACR-419
has not been accepted yet.
and now I put the **\*\*NO MERGE\*\*** tag on this PR.
you can remove **\*\*NO MERGE\*\*** tag for yourself when this PR is ACR accepted and ready to be merged.
(**\*\*NO MERGE\*\*** tag has no restriction to submit PR or to edit this patch but just for notification and marking.
as you know, this PR can not be merged before the ACR is accepted.
thank you.

src/Tizen.NUI.Components/Controls/AlertDialog.cs
```diff
@@ -133,8 +150,10 @@ public override void ApplyStyle(ViewStyle viewStyle)
 
         /// <summary>
         /// Title text of AlertDialog.
+        /// Title text is set to TitleContent's Text if TitleContent is TextLabel.
+        /// If TitleContent's Text is set manually by user, then it is not guaranteed that TitleContent's Text is the same with Title text.
```
wiertel, line 154: do you mean that if it's changed later then they might differ?

src/Tizen.NUI.Components/Controls/Navigation/DialogPage.cs
```diff
@@ -80,9 +80,11 @@ protected override void Dispose(DisposeTypes type)
         }
 
         /// <summary>
-        /// Content of DialogPage. Content is added to Children automatically.
+        /// Content of DialogPage.
+        /// Content is used as dialog so Content is displayed above the dimmed scrim.
```
wiertel, line 84: i would add a coma:    ... as dialog, so ...

src/Tizen.NUI.Components/Style/AlertDialogStyle.cs
```diff
@@ -23,25 +23,26 @@ namespace Tizen.NUI.Components
     /// <summary>
     /// AlertDialogStyle is a class which saves AlertDialog's ux data.
```
wiertel, line 24: ux -> UX   as always with abbreviations in description text

src/Tizen.NUI.Components/Style/AlertDialogStyle.cs
```diff
@@ -50,27 +51,27 @@ public AlertDialogStyle(AlertDialogStyle style) : base(style)
...
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <since_tizen> 9 </since_tizen>
         public ViewStyle ActionContent { get; set; } = new ViewStyle();
 
         /// <summary>
         /// Style's clone function.
         /// </summary>
         /// <param name="bindableObject">The style that need to copy.</param>
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <since_tizen> 9 </since_tizen>
```
wiertel, line 74: if this override doesn't do anything special, then it's probably better to use "\<inheritdoc/\>"

src/Tizen.NUI.Components/Controls/AlertDialog.cs
jaehyun0cho: Yes, we provide a property Text.
But we cannot prohibit user from setting TitleContent's Text manually.
Therefore, they might be differ if user sets TitleContent's Text manually.

src/Tizen.NUI.Components/Controls/Navigation/DialogPage.cs
jaehyun0cho: Thank you for your comment.
It has been updated.

src/Tizen.NUI.Components/Style/AlertDialogStyle.cs
jaehyun0cho: Thank you for your comment.
It has been updated.

src/Tizen.NUI.Components/Style/AlertDialogStyle.cs
jaehyun0cho: Thank you for your comment.
It has been updated.

dongsug-song:
looks good to me but other members' review is required.
@wiertel 
could you review again and approve?
thank you.

everLEEst (approved):
Looks good to me but someone else must approve it.

dongsug-song:
the ACR has been accepted so this is merged now.

PR #2922, opened 2021-04-16
[NUI] Make Menu APIs public

jaehyun0cho:
This is ACR patch to make Menu APIs public.

This pull request includes properties and methods of Menu, MenuItem and
DialogPage classes.

Menu.AnchorPosition is replaced with Menu.Anchor.
Menu.HorizontalAlignment and Menu.VerticalAlignment are added for
Menu.Anchor.

### API Changes ###

 - ACR: TCSACR-422


Seoyeon2Kim (approved):
Looks good to me, but before merging this PR, please add ACR number you added.
(TCSACR-xxx)

jaehyun0cho:
ACR number is TCSACR-422

src/Tizen.NUI.Components/Controls/Menu.cs
```diff
@@ -121,6 +122,7 @@ public IEnumerable<MenuItem> Items
         /// If there is no enough space to display menu at the anchor position,
         /// then menu is displayed at the proper position near anchor position.
         /// </summary>
+        /// <since_tizen> 9 </since_tizen>
```
hyunjushin, line 125: Should developers set position manually?  
If the required functionality is similar to Android "PopupMenu", anchor view concept is more suitable to calculate position of the menu automatically.

src/Tizen.NUI.Components/Controls/Navigation/DialogPage.cs
```diff
@@ -294,7 +294,7 @@ public static void ShowAlertDialog(string title, string message, params View[] a
         /// </summary>
         /// <param name="anchor">The anchor view where menu is displayed.</param>
         /// <param name="items">The menu items.</param>
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <since_tizen> 9 </since_tizen>
```
hyunjushin, line 297: why are you getting menu items instead of menu?

src/Tizen.NUI.Components/Controls/Navigation/DialogPage.cs
```diff
@@ -325,7 +325,7 @@ public static void ShowMenu(View anchor, params MenuItem[] items)
         /// </summary>
         /// <param name="anchorPosition">The anchor position where menu is displayed.</param>
         /// <param name="items">The menu items.</param>
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <since_tizen> 9 </since_tizen>
```
hyunjushin, line 328: anchor position need to be checked again.

hyunjushin (changes requested):

src/Tizen.NUI.Components/Controls/Menu.cs
jaehyun0cho: I think that there are 2 ways of usages.
- Showing menu at a given object (such as drop down button)
  - "Anchor" property may be useful.
- Showing menu at a given point (such as copy paste menu)
  - "Anchor Position" property may be useful.

I think that the properties "Anchor" and "Anchor Position" affect each other.
(For Android case, they use only "Anchor")

I think that "Anchor Position" may cover "Anchor" as well because we can set the proper position of "Anchor" to "Anchor Position". (but I think that "Anchor" cannot cover "Anchor Position")
However, it is not so easy to get and set "Anchor"'s position to "Anchor Position".

So I adds the static methods for menu as follows in DialogPage.
- public static void ShowMenu(View anchor, params MenuItem[] items)
- public static void ShowMenu(Position2D anchorPosition, params MenuItem[] items)
```
DialogPage.ShowMenu(anchor, menuItem, menuItem2, ...);
```

If you think that "Anchor" property is better than "Anchor Position", then I can provide "Anchor" property instead of "Anchor Position".
But I am not sure how to support "Anchor Position" usage with "Anchor" property. I need to think about it more.

src/Tizen.NUI.Components/Controls/Navigation/DialogPage.cs
jaehyun0cho: It is to support both "Anchor" and "Anchor Position".

Since Menu supports "Anchor Position", to support "Anchor" user needs to set something more with their menu.
So I tried to provide the easier way to show menu with both "Anchor" and "Anchor Position".

src/Tizen.NUI.Components/Controls/Navigation/DialogPage.cs
jaehyun0cho: As I mentioned https://github.com/Samsung/TizenFX/pull/2922#discussion_r615518743 , we can think about "Anchor" and "Anchor Position" more.

src/Tizen.NUI.Components/Controls/Menu.cs
jaehyun0cho: I have added "Anchor" property and removed "AnchorPosition" property.

src/Tizen.NUI.Components/Controls/Navigation/DialogPage.cs
jaehyun0cho: I have added "Anchor" property and removed "AnchorPosition" property.

jaehyun0cho:
AnchorPosition is replaced with Anchor.
HorizontalAlignment and VerticalAlignment are added for Anchor.

jaehyun0cho:
Now, Menu is not used by DialogPage. (DialogPage.ShowMenu is removed)

The new usage is as follows.

```
var menu = new Menu()
{
    Anchor = moreNutton,
    HorizontalPositionToAnchor = Menu.HorizontalRelativePosition.Center,
    VerticalPositionToAnchor = Menu.VerticalRelativePosition.Bottom,
    Items = new MenuItem[] { menuItem, menuItem2, menuItem3, ... },
};
menu.Post();
```

src/Tizen.NUI.Components/Controls/Navigation/DialogPage.cs
jaehyun0cho: AnchorPosition is replaced with Anchor.
HorizontalPositionToAnchor and VerticalPositionToAnchor are added for Anchor.

src/Tizen.NUI.Components/Controls/Menu.cs
```diff
@@ -93,9 +131,9 @@ public IEnumerable<MenuItem> Items
                 {
                     foreach (var oldItem in menuItems)
                     {
-                        if (scrollableBase.Children?.Contains(oldItem) == true)
+                        if (content.Children?.Contains(oldItem) == true)
                         {
-                            scrollableBase.Children.Remove(oldItem);
+                            content.Children.Remove(oldItem);
```
myroot, line 136: You should call `content.Remove(oldItem)` instead `content.Children.Remove` because `content.Children` is not Observerable collection. so it does not anything when children was removed

And, you do not need to check `Contains` as my opinions, 
I think, just call `Unparen() `is enough

``` c#
   oldItem.Unparent();
```

src/Tizen.NUI.Components/Controls/Menu.cs
```diff
@@ -55,31 +57,67 @@ protected override void Dispose(DisposeTypes type)
...
+                Window.RemoveLayer(layer);
+                layer.Dispose();
             }
 
             base.Dispose(type);
         }
 
+        /// <summary>The Menu's relative position to Anchor horizontally.</summary>
+        /// <since_tizen> 9 </since_tizen>
+        public enum HorizontalRelativePosition
```
myroot, line 87: I think you can share enum type if you define like below
and you do better to use start/end instead `left,right` if you want to support RTL language
``` c#
public enum RelativePosition
        {
            /// <summary>At the start of the Anchor.</summary>
            Start = 0,
            /// <summary>At the center of the Anchor.</summary>
            Center = 1,
            /// <summary>At the end of the Anchor.</summary>
            End = 2,
        }
```

dongsug-song:
https://code.sec.samsung.net/jira/browse/TCSACR-422
has not been accepted yet.
and now I put the **\*\*NO MERGE\*\*** tag on this PR.
you can remove **\*\*NO MERGE\*\*** tag for yourself when this PR is ACR accepted and ready to be merged.
(**\*\*NO MERGE\*\*** tag has no restriction to submit PR or to edit this patch but just for notification and marking.
as you know, this PR can not be merged before the ACR is accepted.
thank you.

src/Tizen.NUI.Components/Controls/Menu.cs
jaehyun0cho: Thank you for your comment.
I have changed as follows based on your comment.
```
if (content.Children?.Contains(oldItem) == true)
{
    content.Remove(oldItem);
}
```

View.Remove(child) calls functions related to Remove although View does not contain the child.
(It only shows Tizen.Log.Error messages if View does not contain the child.)
So I think that checking if content contains oldItem may prevent unnecessary calls related to Remove.

I am not sure what app developer does to Items.
e.g. app developer may add Items to other View right before setting a new Items to Menu.
So I think that content.Remove() is safer than oldItem.Unparent().

src/Tizen.NUI.Components/Controls/Menu.cs
jaehyun0cho: Thank you for your comment.
I have changed as you commented.

myroot (approved):

hyunjushin (approved):

PR #2928, opened 2021-04-19
[NUI] Propgate BindingContext recursively on RecyclerViewItem.

everLEEst:
propagate BindingContext to chlidren and descendant recursively
for applying binding on DataTemplate.

- previously
RecyclerItemView
|_ Child View - BindingContext Propagated
     |_ GrandChild View BindingContext Not Propagated
     
- from now on
RecyclerViewItem
|_ Child View - BindingContext Propagated
          ....
           |_ Descendant View - BindingContext Propagated.
           
           
Yet user has to set IsCreateByXaml true for using SetBinding() in cs code,    
will be renamed IsBinded property with automatically set by SetBinding().


everLEEst:
need to fix the conflict.

everLEEst:
patch is rebased.

jaehyun0cho (approved):
LGTM

PR #2939, opened 2021-04-20
[NUI] Update some descriptions of View properties

Seoyeon2Kim:

### Description of Change ###
- To avoid confusion of using some `View` properties,
 added remarks to use them for `TableView` class.

### API Changes ###
- N/A

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -2297,7 +2297,7 @@ namespace Tizen.NUI.BaseComponents
         }
 
         ///<summary>
-        /// Set a layout transitions for this View.
+        /// Sets a layout transitions for this View.
         ///</summary>
         /// <exception cref="ArgumentNullException"> Thrown when value is null. </exception>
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -814,6 +814,7 @@ public Vector4 FlexMargin
         /// </summary>
         /// <remarks>
         /// The property cascade chaining set is possible. For example, this (view.CellIndex.X = 0.1f;) is possible.
+        /// Also, this property is for `TableView` class. Please use the property for the child position of TableView.
```
jaehyun0cho, line 817: I am not sure if it is correct, but can we use as follows?

```
/// Also, this property is for <see cref="TableView"/> class. Please use the property for the child position of <see cref="TableView"/>.
```

Seoyeon2Kim: Sure, we can!
Thanks for your review :)

jaehyun0cho (approved):
LGTM

PR #2942, opened 2021-04-21
[NUI] Fix null reference check in CollectionViews.

everLEEst:
This issues are detected by self-diagnosis.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #2947, opened 2021-04-22

kkhos:
### Description of Change ###

### API Changes ###

 - ACR:NONE


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -467,6 +467,21 @@ public Vector4 CornerRadius
         // Accessibility
         //
 
+        /// <summary>
+        /// Gets or sets a value that allows the automation framework to find and interact with this element.
+        /// </summary>
+        /// This will be public opened in tizen_5.0 after ACR done. Before ACR, need to be hidden as inhouse API.
```
dongsug-song, line 473: please remove this comment~ this comment is not appropriate for this API.
_**/// This will be public opened in tizen_5.0 after ACR done. Before ACR, need to be hidden as inhouse API.**_

src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -467,6 +467,21 @@ public Vector4 CornerRadius
...
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override string AutomationId
```
dongsug-song, line 475: How about change as followings
`public override string AutomationId` => `public new string AutomationId` 
`public virtual string AutomationId` => `public string AutomationId` (no change)
(personally, Element is base class and it looks not good to change base class's interface)

src/Tizen.NUI/src/public/XamlBinding/Element.cs
```diff
@@ -57,7 +57,7 @@ public abstract partial class Element : BindableObject, IElement, INameScope, IE
         /// </summary>
         /// This will be public opened in tizen_5.0 after ACR done. Before ACR, need to be hidden as inhouse API.
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public string AutomationId
+        public virtual string AutomationId
```
dongsug-song, line 60: How about change as followings
public override string AutomationId => public new string AutomationId
public virtual string AutomationId => public string AutomationId (no change)
(personally, Element is base class and it looks not good to change base class's interface)

everLEEst:
One question regarding the automationID....
have you checked other services who's using this automation ID?
declairing new and overriding is almost same but calling property in which class may change the actions...

kkhos:
Remove comment and virtual keyword.

dongsug-song:
@kkhos 
Hello, could you check @everLEEst 's comment above?
thank you.

kkhos:
@everLEEst 
AutomationId was created for appium. Tizen Appium is working on wearable and mobile device. 
So I think AutomationId has not been used usually. There isn't usage in codegrok also.
And If user doesn't use up casting, new and overriding is almost same. 
User doesn't use up casting with Element class, so I think there is no problem.

dongsug-song (approved):
looks good to me

dongsug-song:
@kkhos 
could you resolve the conflicts? then this can be merged.~

kkhos:
@dongsug-song 
Request a new pull request because properties of the accessibility have been moved to another file.
https://github.com/Samsung/TizenFX/pull/3028

PR #2949, opened 2021-04-23
[NUI][ATSPI] Alert dialog

a-bartak:
### API Changes ###

 - ACR:


lol-github (approved):
LGTM

Seoyeon2Kim (approved):
Looks good to me, but @Jaehyun-Cho should approve this PR.

jaehyun0cho:
@a-bartak 
This PR sets AccessibilityState.Modal true as follows.
```
states.Set(AccessibilityState.Modal, true);
```
It seems that this causes to show dali window popup.
AlertDialog itself is a dialog (popup).

Is it correct if dali window popup is also displayed when AlertDialog is Show()?

a-bartak:
@jaehyun0cho 
I have added Modal AccessibilityState to AlertDialog as without that  AlertDialog was not modal and while changing highlighted element using flick highlight was jumping out of the dialog.
Could you specify in more detail what problems do you anticipate?

Seoyeon2Kim:
> @jaehyun0cho
> I have added Modal AccessibilityState to AlertDialog as without that AlertDialog was not modal and while changing highlighted element using flick highlight was jumping out of the dialog.
> Could you specify in more detail what problems do you anticipate?

So, `AddPopup()` is not literally to add and show popup window, but it just means that popup, which is `Dialog` in NUI Components, is highlighted and checked by Accessible bridge. Do I understand correctly?

jaehyun0cho:
@a-bartak 
I don't have enough knowledge about accessibility so I was worried about a new window created by Modal setting.
However, based on your comment, it turns out that the new window is only for the highlight and it is not to move highlight outside of the AlertDialog.
If I understand correctly, then your PR is absolutely correct.
Thank you.

lol-github (approved):
LGTM, Yes the AddPopup() doesn't create new instance of widget it only tells the ATSPI bridge that the existing popup has been shown/hiden by application. Bridge is using this information to manage the highlight navigation context, as it has to support navigation in case of many stacked popups. Probably more accurate name for this function would be RegisterPopup.

jaehyun0cho:
@lol-github 
Thank you for the explanation.

jaehyun0cho (approved):
LGTM

PR #2951, opened 2021-04-26
[NUI] Apply Vector4 corner radius internally.

rabbitfor:

### API Changes ###

 - ACR:


src/Tizen.NUI/src/internal/Common/PropertyHelper.cs
```diff
@@ -109,15 +110,14 @@ private static SearchResult GenerateVisualPropertySearchResult(View view, Visual
...
             {
                 property.Dispose();
-                return null;
+                return data.RelatedData == null ? null : GenerateVisualPropertySearchResult(view, data.RelatedData);
             }
 
             SearchResult result = new SearchResult(property, data.ObjectConverter, data.PropertyValueConverter);
 
-            while (data.RelatedData != null)
+            if (data.RelatedData != null)
```
dongsug-song, line 118: this looks to change the behavior of the `while() ... data = data.RelatedData` 
is this OK?

rabbitfor: Yes :) The previous code has a bug. It need to be fixed to remove `while`.

dongsug-song: OK, I got it~

dongsug-song (approved):
looks good to me

PR #2954, opened 2021-04-27
[NUI][ATSPI] remove build warning messages

kimcinoo:
This patch is for removing following build warnings.

(1) Do not nest type SayFinishedEventArgs. Alternatively,
  change its accessibility so that it is not externally visible.

(2) Only FlagsAttribute enums should have plural names

### API Changes ###

 - ACR:


dongsug-song (approved):
LGTM

PR #2957, opened 2021-04-27
[NUI] Implements view update on NotifyCollectionChanged in collectionView.

everLEEst:
When Collection is updated,
View can pick the notify event with updated info.
this patch implement normal collection updates including

- single item inserted
- single item removed
- single item moved
- group insrted
- group removed
- group moved

and apply samples regarding new implemented feature.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #2959, opened 2021-04-27
[NUI] set Internal SizingStrategy to hide MeasureAll option.

everLEEst:
Currently CollectionView only support MeasureFirst,
and MeasureAll need to be implemented.

before this feature is ready,
make setter internal to disallow unimplemented mode.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #2960, opened 2021-04-27
[NUI] fix crash on collecitonView which looks like bug action..

everLEEst:
currently create two collectionView can be aborted while
accessing padding of layouter style.

use "using" keyword as GetViewStyle simply Copy the original
ViewStyle and will be disposed by using keyword,
but it occurs crash as CopyFrom only new the parent and copy
all existing values.

This looks issue on CopyFrom but to fix collectionView issue quickly,
remove using keyward on first step.

Need to discuss more regarding the issue.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #2961, opened 2021-04-27
[NUI] Initialize CollectionView when scrolling direction changed.

everLEEst:
if scrolling direction is changed,
all layout need to be updated and initialized.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #2972, opened 2021-04-29
[NUI] Add GLWindow Rendering mode

sparrow74:
Add GLWindow rendering mode.
It has two mode as continuous and on demand.
If the mode is continuous, the rendering user callback is called continuously.
If the mode is on demand, the rendering user callback is called by call renderOnce().

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window/GLWindow.cs
+++ b/src/Tizen.NUI/src/public/Window/GLWindow.cs
@@ -136,26 +136,6 @@ namespace Tizen.NUI
         }
 
         /// <summary>
-        /// This Enumeration is used the GLES version for EGL configuration.<br />
-        /// If the device can not support GLES version 3.0 over, the version will be chosen with GLES version 2.0<br />
-        /// </summary>
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        public enum GLESVersion
-        {
-            /// <summary>
-            /// GLES version 2.0
-            /// </summary>
-            [EditorBrowsable(EditorBrowsableState.Never)]
-            Version_2_0 = 0,
-
-            /// <summary>
-            /// GLES version 3.0
-            /// </summary>
-            [EditorBrowsable(EditorBrowsableState.Never)]
-            Version_3_0
-        }
-
-        /// <summary>
         /// Sets egl configuration for GLWindow
         /// </summary>
         /// <param name="depth">The flag of depth buffer. If true is set, 24bit depth buffer is enabled.</param>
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/Window/GLWindow.cs
```diff
@@ -494,6 +494,50 @@ public void Destroy()
...
 
+        /// <summary>
+        /// Enumeration for rendering mode
+        /// This Enumeration is used to choose the rendering mode.
+        /// It has two options.
+        /// One of them is continuous mode. It is rendered continuously.
+        /// The other is on demand mode. It is rendered by application.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public enum GLWindowRenderingMode
```
huiyueun, line 505: How about changing 'RenderingModeType'?

sparrow74: Thank you for your kind review.
Actually, DALi has the similar enumeration. The name is "RenderingMode".
I think same name is better.

dongsug-song (commented):
Hello, 
could you check the warning in line 149, 155?

**Remove the underscores from member name Tizen.NUI.GLWindow.GLESVersion.Version_2_0.**

could you change them as PascalCase?

dkdk-ryu:
In OpenTK case, They name the namespaces as OpenTK.Graphics.ES20 and ES30.
https://docs.microsoft.com/en-us/dotnet/api/opentk.graphics.es20?view=xamarin-ios-sdk-12
I think It should be better to follow the naming of OpenTK and don't break the Tizen rule.

sparrow74:
Thank you for detail review.
NUI has the enumeration and OpenTK has the namespace.
I think it is different case.
In addition, the Enumeration's name is "GLESVerion".
So, The prefix's candidates are 'GLES", "Version", "Ver" or "V".
If "ES" is the best choice, the enumeration's name should be changed with "ESVersion".
Could you share your opinion?

dkdk-ryu:
OK, I think Version20 is better.

sparrow74:
I'm appreciated your review.
I will update with your opinion.

dkdk-ryu:
I think GLView would need GLRenderingMode.
So how about moving GLRenderingMode to NUIConstants.cs?

sparrow74:
@dkdk-ryu  
You're right.
GLRenderingMode should be moved to NUIConstants.cs for GLView.
In addition, GLESVersion enum should be moved, too.
Thank you for your review. :)

dongsug-song (approved):
looks good to me

dkdk-ryu:
Looks good to me.

PR #2973, opened 2021-04-29
[NUI] Make LayoutParamPolicies public

jaehyun0cho:
This is ACR patch to make LayoutParamPolicies public.

This pull request includes static class LayoutParamPolicies and its
const int MatchParent and WrapContent which decide the size of View when
the View is laid out in its parent View.

### API Changes ###

 - ACR: TCSACR-423


src/Tizen.NUI/src/public/BaseComponents/ViewEnum.cs
```diff
@@ -38,23 +38,20 @@ public enum ViewLayoutDirectionType
     }
 
     /// <summary>
-    /// [Draft] Available policies for layout parameters
+    /// Layout policies to decide the size of View when the View is laid out in its parent View.
     /// </summary>
-    /// This will be public opened in tizen_5.5 after ACR done. Before ACR, need to be hidden as inhouse API.
-    [EditorBrowsable(EditorBrowsableState.Never)]
+    /// <since_tizen> 9 </since_tizen>
```
dongsug-song, line 43: could you check the enum class in .NET ?

https://docs.microsoft.com/ko-kr/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/enumeration-classes-over-enum-types

I think this looks actually `enum` but it was declared as static class. I don't know why. 
so could you check the difference of `enum` and `enum class` in .NET ? then you may use more appropriate one.

jaehyun0cho: If the static class is changed to enum as follows,
```
     /// Layout policies to decide the size of View when the View is laid out in its parent View.
     /// </summary>
    public enum LayoutParamPolicies
     {
         /// <summary>
         /// Constant which indicates child size should match parent size.
         /// </summary>
        MatchParent = -1,
         /// <summary>
         /// Constant which indicates parent should take the smallest size possible to wrap it's children with their desired size.
         /// </summary>
        WrapContent = -2
     }

```

then build error occurs as follows.
```
TizenFX/src/Tizen.NUI/src/internal/Layouting/LayoutController.cs(235,39): error CS0019: Operator '==' cannot be applied to operands of type 'int' and 'LayoutParamPolicies' [TizenFX/src/Tizen.NUI/Tizen.NUI.csproj]
```

The build error occurs by comparing int value and Enum value.
Although the build error can be resolved by casting like (int)LayoutParamPolicies.MatchParent, I think static class with const int value would be easier to use.

myroot: I think, enum type is much better than static class for API User. 
It is common things to use enum type on this case, 
I want to make API with common sense

myroot: Of course, if you change it to enum, the ABI will be broken and I will get a trouble with my app

jaehyun0cho: @myroot 
I totally agree with your opinion.
However, we have opened ```View.WidthSpecification```/```View.HeightSpecification``` already in API6, so I am not sure making ```LayoutParamPolicies.MatchParent```/```LayoutParamPolicies.WrapContent``` ```enum``` type is better.

As I replied above, app developers need to use type casting if ```LayoutParamPolicies.MatchParent```/```LayoutParamPolicies.WrapContent``` are ```enum``` type.
```
view.WidthSpecification = (int)LayoutParamPolicies.MatchParent;
```

Moreover, ```View.WidthSpecification```/```View.HeightSpecification``` are ```int``` type, so ```LayoutParamPolicies.MatchParent```/```LayoutParamPolicies.WrapContent``` are not auto-completed when app developers write codes.

Based on the current NUI View's implementation, ```const int``` type with ```static class``` for ```LayoutParamPolicies.MatchParent```/```LayoutParamPolicies.WrapContent``` is better although it is not the best solution...

myroot: @jaehyun0cho I also agree to you, if we could not change type of `WidthSpecification` , static int value could be alternative solution

dongsug-song (approved):
looks good to me

PR #2974, opened 2021-04-29
[Multimedia] Add Validation util and rename filename

hsgwon:
Add Validation util and rename filename

### DIFF
--- a/src/Tizen.Multimedia/Common.Internal/ValidationUtil.cs
+++ b/src/Tizen.Multimedia/Common.Internal/ValidationUtil.cs
@@ -15,7 +15,9 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 
 namespace Tizen.Multimedia
 {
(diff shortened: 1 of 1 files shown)


src/Tizen.Multimedia/Common.Internal/ValidationUtil.cs
```diff
@@ -44,5 +46,26 @@ internal static void ValidateFeatureSupported(string featureKey)
                 throw new NotSupportedException($"The feature({featureKey}) is not supported.");
             }
         }
+
+        internal static void ValidateIsNullOrEmpty(string value, string paramName)
```
sangchul1011, line 50: I checked, the caller of this method call it like..
ValidateIsNullOrEmpty(somestring, nameof(somestring));

I think, if the second parameter is used for just getting the name out of the first value, we can remove the second parameter, and get the name inside of this function.

src/Tizen.Multimedia/Common.Internal/ValidationUtil.cs
```diff
@@ -44,5 +46,26 @@ internal static void ValidateFeatureSupported(string featureKey)
...
+        {
+            if (value == null)
+            {
+                throw new ArgumentNullException(paramName);
```
sangchul1011, line 54: nameof(value)

src/Tizen.Multimedia/Common.Internal/ValidationUtil.cs
```diff
@@ -44,5 +46,26 @@ internal static void ValidateFeatureSupported(string featureKey)
...
+            }
+
+            if (string.IsNullOrWhiteSpace(value))
+            {
+                throw new ArgumentException(paramName + " is a zero-length string.", paramName);
```
sangchul1011, line 59: nameof(value)

src/Tizen.Multimedia/Common.Internal/ValidationUtil.cs
sangchul1011: done

sangchul1011 (approved):

PR #2982, opened 2021-04-30
[NUI][AT-SPI] Add Value interface to Pagination

grzelewski:
### Description of Change ###
Add support value interface in pagination widget.

Old implementation in screen-reader reads information by analyzing
widget structures. This causes problem with pagination widget in nui
library as this one may have multiple "visual" elements that are inside
containers. As a result nui pagination widget always have two
children.

This problem was solved by using value interface to inform screen-reader
about number of visual elements and position of selected one.

### API Changes ###
NONE

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Pagination.cs
+++ b/src/Tizen.NUI.Components/Controls/Pagination.cs
@@ -367,7 +367,8 @@ namespace Tizen.NUI.Components
 
                 SelectIn(indicatorList[selectedIndex]);
 
-                if (IsHighlighted) {
+                if (IsHighlighted) 
+                {
                     EmitAccessibilityEvent(ObjectPropertyChangeEvent.Value);
                 }
             }
@@ -421,11 +422,11 @@ namespace Tizen.NUI.Components
         [EditorBrowsable(EditorBrowsableState.Never)]
         protected override bool AccessibilitySetCurrent(double value)
         {
-            int f = (int)value;
+            int integerValue = (int)value;
 
-            if (f >= 0 && f <= IndicatorCount)
+            if (integerValue >= 0 && integerValue <= IndicatorCount)
             {
-                SelectedIndex = f;
+                SelectedIndex = integerValue;
                 return true;
             }
 
--- a/src/Tizen.NUI.Components/Controls/Pagination.cs
+++ b/src/Tizen.NUI.Components/Controls/Pagination.cs
@@ -366,6 +366,10 @@ namespace Tizen.NUI.Components
                 selectedIndex = refinedValue;
 
                 SelectIn(indicatorList[selectedIndex]);
+
+                if (IsHighlighted) {
+                    EmitAccessibilityEvent(ObjectPropertyChangeEvent.Value);
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI.Components/Controls/Pagination.cs
```diff
@@ -366,6 +366,10 @@ public int SelectedIndex
                 selectedIndex = refinedValue;
 
                 SelectIn(indicatorList[selectedIndex]);
+
+                if (IsHighlighted) {
```
Seoyeon2Kim, line 370: Please update it like the below.
```suggestion
                if (IsHighlighted)
                {
```

src/Tizen.NUI.Components/Controls/Pagination.cs
```diff
@@ -384,11 +388,67 @@ public Position GetIndicatorPosition(int index)
...
+            return (double)IndicatorCount;
+        }
+
+        /// <summary>
+        /// Current value.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected override bool AccessibilitySetCurrent(double value)
+        {
+            int f = (int)value;
```
Seoyeon2Kim, line 424: Please reconsider the variable name.
We don't use either ambiguous abbreviations or arbitrary characters that do not convey meaning for the variable.

src/Tizen.NUI.Components/Controls/Pagination.cs
```diff
@@ -384,11 +388,67 @@ public Position GetIndicatorPosition(int index)
...
+        {
+            return 1.0;
+        }
+
         /// <inheritdoc/>
         [EditorBrowsable(EditorBrowsableState.Never)]
         public override void OnInitialize()
         {
             base.OnInitialize();
+            SetAccessibilityConstructor(Role.ScrollBar, AccessibilityInterface.Value);
```
Seoyeon2Kim, line 449: I'm not sure whether there's any proper role for `Pagination` class in ATSPI role.
But, as you know, there's `Scrollbar` class in NUI. And `Pagination` is similar to CSS Pagination.
https://www.w3schools.com/css/tryit.asp?filename=trycss_ex_pagination

lol-github: We have tried to mimic the approach applied in case of elm_index widget. In EFL this widget when combined with "pagecontrol" style behaves and looks identically as Pagination hence as there is no dedicated "pagination" role in the set of ATSPI roles we decided to use the "scrollbar" role combined with attribute "pagecontrolbyvalue". Using a combination of role and attribute gives us the freedom of  distinction on screen-reader side between Scrollbar widget and Pagination widget cases.

Seoyeon2Kim: Okay, I see. When we find the proper role of ATSPI in the future, we can update it. :)

lol-github (approved):
LGTM

src/Tizen.NUI.Components/Controls/Pagination.cs
lol-github: Done

src/Tizen.NUI.Components/Controls/Pagination.cs
lol-github: Done

dongsug-song (approved):
looks good to me

PR #2985, opened 2021-05-03
[NUI] Refactoring itemSelectionMode names and adding SingleAlways

everLEEst:
Change names in ItemSelectionMode

SingleSelection and MultipleSelections overlapped "Selection" keyward
with their enum name.
Removing Unnecessary keyward from enum members,
and adding Suppress condition to avoid CA1720 warnings.
Single is class Identifier name but there are no harm to using name
Single in a member of enum.

Single allow deselection by clicking selected item again,
but we have another demands who want to act single selection like radio,
which always have one single selected item after once selected.

this Singlelways for that demands,
even click again on selected item, item will not deselected,
and will not occurs any SelectionChanged events.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #2986, opened 2021-05-03
[NUI] Fix dispoing collectionView and related classes.

everLEEst:
items and container is not cleared after view disposed.
fix collectionView and item dispose properly.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #2987, opened 2021-05-03
[NUI] Fix Utility.Dispose not to set null to method parameter

jaehyun0cho:
Setting a value to method parameter does not work unless the method
parameter is defined with out keyword.

Not to confuse users, setting null to method parameter is removed.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM

PR #2988, opened 2021-05-03
[ElmSharp] Fix EcoreEvent

JoonghyunCho:
### Description of Change ###
Fix EcoreEvent handler logic which does not remove event handler properly.

When event handler is removed from `On`, it should also be removed from the inner list containing handlers
so it does not try removing at `Dispose()` again.

### API Changes ###
- None


arosis78 (approved):
LGTM

sung-su (approved):

shyunMin (approved):

PR #2992, opened 2021-05-04
[NUI] Replace duplicate documentation of CopyFrom with inheritdoc

jaehyun0cho:
The duplicate documentation of BindableObject.CopyFrom is replaced with
inheritdoc.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM

PR #2993, opened 2021-05-04
[NUI] Replace NavigationPages.Count with PageCount

jaehyun0cho:
Not to provide List property to users, NavigationPages is removed from
Navigator class.
Instead of NavigationPages.Count, PageCount is added to Navigator class.

### API Changes ###

 - ACR:


JSUYA (approved):
LGTM. I have check sample app works fine.

PR #2994, opened 2021-05-04
[NUI] Change property name Caption to Label and related names on DefaultGridItem

everLEEst:
…ultGridItem.

TextLabel property is commonly named Label so the name Caption can makes
confusing.

Change Caption to Label on DefaultGridItem and its style,
and change CaptionRelativeOrientation property to LabelOrientationType.

this patch also @fix the Tizen.NUI.Samples build error.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #3006, opened 2021-05-07
[NUI] Deprecate BaseHandle.PropertySet and Add BindableProperty.PropertyChanged

dongsug-song:
### Description of Change ###
Deprecate BaseHandle.PropertySet and Add BindableProperty.PropertyChanged

PropertySet has been added before the NUI XAML is added.
PropertySet is duplicated of PropertyChanged so it can be deprecated.
BindableProperty class is parent of BaseHandle class, so PropertyChanged is better to be used.

### API Changes ###
https://code.sec.samsung.net/jira/browse/TCSACR-413


Seoyeon2Kim (approved):
Looks good to me for ACR process!

PR #3008, opened 2021-05-07
[NUI] Deprecate ResizePolicyType

jaehyun0cho:
Up to now, both ResizePolicyType and LayoutParamPolicies are supported.
Originally, ResizePolicyType is for a view whose parent view does not
have Layout.
In contrast, LayoutParamPolicies is for a view whose parent view has
Layout.

However, ResizePolicyType can be used as LayoutParamPolicies as well
internally.

This makes users confused which one they need to use in each case.

Not to confuse users, ResizePolicyType is deprecated.

Instead, LayoutParamPolicies can be used to set the size of a view
relative to its parent or children.

Specifically, LayoutParamPolicies.MatchParent and
LayoutParamPolicies.WrapContent work with parent view which does not
have Layout as how ResizePolicyType.FillToParent and
ResizePolicyType.FitToChildren work.

### API Changes ###

 - ACR: TCSACR-425

### DIFF
--- a/src/Tizen.NUI/src/public/Common/NUIConstants.cs
+++ b/src/Tizen.NUI/src/public/Common/NUIConstants.cs
@@ -136,48 +136,57 @@ namespace Tizen.NUI
     /// Enumeration for size negotiation resize policies.
     /// </summary>
     /// <since_tizen> 3 </since_tizen>
+    [Obsolete("Deprecated. Please use Tizen.NUI.BaseComponents.LayoutParamPolicies instead.")]
     public enum ResizePolicyType
     {
         /// <summary>
         /// Size is fixed as set by SetSize.
         /// </summary>
         [Description("FIXED")]
+        [Obsolete("Deprecated. Instead, please set a positive integer value to Tizen.NUI.View.WidthSpecification or Tizen.NUI.View.HeightSpecification.")]
         Fixed,
         /// <summary>
         /// Size is to use the actor's natural size.
         /// </summary>
         /// <see cref="ViewImpl.GetNaturalSize"/>
         [Description("USE_NATURAL_SIZE")]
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI/src/public/Common/NUIConstants.cs
```diff
@@ -136,48 +136,57 @@ public enum DrawModeType
     /// Enumeration for size negotiation resize policies.
     /// </summary>
     /// <since_tizen> 3 </since_tizen>
+    [Obsolete("Deprecated. Please use Tizen.NUI.BaseComponents.LayoutParamPolicies instead.")]
```
huiyueun, line 139: Do you have a plan to add other type to LayoutParameterPolicies ?( UseNaturalSize, SizeRelativeToParent ... )

jaehyun0cho: For now, I don't have plan for that.
I think the current LayoutParamPolicies are sufficient for sizing usages.

jaehyun0cho:
This PR does not support the following previous action directly.

```
public Vector3 SizeModeFactor
This view's size is set to the view's size multiplied by or added to this factor, depending on ResizePolicyType
```

However, IMHO, this method's usecase is very rare and also it can be substituted by other methods.
e.g. LayoutParamPolicies.MatchParent with Weight
e.g. WidthSepcification and HeightSpecification

dongsug-song:
@jaehyun0cho 
I accept this, because this change follows our internal discussion. 
this requires deprecated-ACR, so please update the ACR link.
after ACR is accepted, then this can be merged.
thank you.

Seoyeon2Kim (approved):
Looks good to me!
This PR is totally to clean up the size negotiation of NUI and make user less confused!:)
After posting deprecated ACR regarding this, please leave a message here and merge by yourself.

PR #3011, opened 2021-05-10
[Camera] Add new internal API - extra preview event

hsgwon:
### Description of Change ###

Add new internal API for extra preview event


hsgwon:
There's no need to proceed ACR because public API is not changed, just moved to other file and changed `Camera` class to partial.
And some internal API is added.

aferin (approved):

PR #3012, opened 2021-05-10
[NUI] Fix crash issue when favicon is null.

huayongxu:
### API Changes ###

 - ACR:


src/Tizen.NUI/src/internal/WebView.cs
```diff
@@ -645,7 +645,7 @@ public ImageView Favicon
             get
             {
                 global::System.IntPtr imageView = Interop.WebView.WebView_GetFavicon(swigCPtr);
-                if (NDalicPINVOKE.SWIGPendingException.Pending)
+                if (imageView == null || imageView == IntPtr.Zero)
```
rabbitfor, line 648: I think comparing IntPtr to null is just same as comparing to IntPtr.Zero.

huayongxu: Done.

rabbitfor (approved):
LGTM

zhouleonlei (commented):
LGTM

PR #3013, opened 2021-05-10
[NUI] Fix not to dereference nullable handle ItemsLayouter

jaehyun0cho:
ItemsLayouter is nullable handle so it is checked before it is
dereferenced.

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #3018, opened 2021-05-11
[NUI] Notification throws a proper exception when the device does not support it.

rabbitfor:
… support it.

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #3028, opened 2021-05-12

kkhos:
### Description of Change ###

### API Changes ###

 - ACR:NONE


src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityProperties.cs
```diff
@@ -107,5 +107,15 @@ public bool AccessibilityAnimated
                 NotifyPropertyChanged();
             }
         }
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
dongsug-song, line 110: could you add some comment even it is hidden API?

lwc0917:
@kkhos Please add AutomationId description.

kkhos:
rebase patch and add description 
https://github.com/Samsung/TizenFX/pull/3081

PR #3029, opened 2021-05-12
[NUI] Fix MatchParent to fill remaining space in LinearLayout

jaehyun0cho:
Previously, in LinearLayout, LayoutParamPolicies.MatchParent sets child
size to its parent size.

Now, in LinearLayout, LayoutParamPolicies.MatchParent calculates the
remaining space of parent and fills child in the remaining space of
parent.

There are 2 cases to fill child in the remaining space of parent.
1. MatchParent (both with or without weight)
 - If child has MatchParent specification, then child is filled in the
   remaining space of parent.
2. Specification 0 with weight
 - If child has zero specification and weight is greater than 0, then
   child is filled in the remaining space of parent proportional to its
   weight.

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/Layouting/LinearLayout.cs
```diff
@@ -268,88 +243,68 @@ private void MeasureHorizontal(MeasureSpecification widthMeasureSpec, MeasureSpe
             var widthMode = widthMeasureSpec.Mode;
             var heightMode = heightMeasureSpec.Mode;
             bool isExactly = (widthMode == MeasureSpecification.ModeType.Exactly);
-            bool matchHeight = false;
+            bool isHeightExactly = (heightMode == MeasureSpecification.ModeType.Exactly);
```
dongsug-song, line 246: could you check the VS warning below?

_**src/Tizen.NUI/src/public/Layouting/LinearLayout.cs#L247
The variable 'allFillParent' is assigned but its value is never used**_

jaehyun0cho: Thank you for your comment.
The unused variable has been removed.

dongsug-song (commented):
looks good to me, but other member's review is required.

everLEEst (commented):
Looks good to me except one old use case.

in old use case,
add content with MatchParent param in ScrollableBase
will create N-th Page Scrollable View.

can be guide them proper solution after this patch merged.

jaehyun0cho:
> Looks good to me except one old use case.
> 
> in old use case,
> add content with MatchParent param in ScrollableBase
> will create N-th Page Scrollable View.
> 
> can be guide them proper solution after this patch merged.

As @everLEEst  and I discussed later, this case is not a problem case.
The reason as follows.

1. User defined ScrollableBase case
    - Suppose that user defined ScrollableBase uses LinearLayout.
      i. If ScrollableBase's ContentContainer uses WrapContent, then children cannot use MatchParent. (circular dependency)
      ii. If ScrollableBase's ContentContainer uses MatchParent or specific size and children uses MatchParent, then children's size fills ScrollableBase's ContentContainer, so scrolling cannot work.

2. Default ScrollableBase case
    - Default ScrollableBase's ContentContainer does not use Layout. Therefore, children cannot use MatchParent/WrapContent to set their position.

As a result, this PR does cause the above backward compatibility issue.

dongsug-song (approved):
looks good to me

PR #3034, opened 2021-05-13
fix nullable member access code to prevent future crash.

everLEEst:
nullable members need to be check it is null or not,
and carefully null set after disposing.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #3035, opened 2021-05-13
[NUI] Add Borderline Properties

hinohie:
Add borderline properies for view
 - BorderlineWidth (float, default = 0.0f)
   : Width of the borderline
 - BorderlineColor (Vector4, default = Color.Black)
   : Color of the borderline
 - BorderlineOffset (float, default = 0.0f)
   : Relative position offset from 'Real' borderline of visual.
     == 0.0f then half is inside, and half is outside of visual.
     == -1.0f then all borderline will be rendered inside of visual.
     == 1.0f then all borderline will be rendered outside of visual.

These three properties are animatable

If it is ImageView, this property only allow to ImageView.Property.IMAGE.
Otherwise, allow to View.Property.BACKGROUND.

Also, current version only works when Border is not setup.


src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyle.cs
```diff
@@ -418,6 +421,40 @@ public Vector4 CornerRadius
             set => SetValue(CornerRadiusPolicyProperty, value);
         }
 
+        /// <summary>
+        /// The width for the borderline of the View.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public float? BorderlineWidth
+        {
+            get => (float?)GetValue(BorderlineWidthProperty);
```
myroot, line 430: Why do you use nullable object(`float?`)? 
GetValue method never return null object, because it is defined as `float` not `float?`

``` c#
        public static readonly BindableProperty BorderlineWidthProperty = BindableProperty.Create(nameof(BorderlineWidth), typeof(float), typeof(ViewStyle), default(float), propertyChanged: (bindable, oldValue, newValue) =>
        {
            ((ViewStyle)bindable).borderlineWidth = (float)newValue;
        }, defaultValueCreator: (bindable) => ((ViewStyle)bindable).borderlineWidth);
```

src/Tizen.NUI/src/public/Visuals/VisualMaps.cs
```diff
@@ -82,6 +82,25 @@ public class VisualMap : IDisposable
         [EditorBrowsable(EditorBrowsableState.Never)]
         private Vector4 cornerRadius = null;
 
+        /// <summary>
+        /// The borderline width of the visual.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        private float? borderlineWidth = null;
```
myroot, line 89: Why do you declare it? I think you no need it, because you use `BindableProperty`, it used for property bag.

src/Tizen.NUI/src/public/Visuals/VisualMaps.cs
```diff
@@ -82,6 +82,25 @@ public class VisualMap : IDisposable
...
+
+        /// <summary>
+        /// The borderline Color of the visual.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
myroot, line 94: No need to add `EditorBrowsable(EditorBrowsableState.Never)` attribute on private field

src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyle.cs
hinohie: Oh, I miss that points. I'll make GetValue method retulrn `float?` type.

src/Tizen.NUI/src/public/Visuals/VisualMaps.cs
hinohie: Fix done.

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
```diff
@@ -1589,6 +1589,54 @@ public partial class View
...
+        public static readonly BindableProperty BorderlineColorProperty = BindableProperty.Create(nameof(BorderlineColor), typeof(Color), typeof(View), null, propertyChanged: (bindable, oldValue, newValue) =>
+        {
+            var view = (View)bindable;
+            (view.backgroundExtraData ?? (view.backgroundExtraData = new BackgroundExtraData())).BorderlineColor = (Color)newValue;
+            view.ApplyBorderline();
+        },
+        defaultValueCreator: (bindable) =>
+        {
+            var view = (View)bindable;
+            return view.backgroundExtraData == null ? Color.Black : view.backgroundExtraData.BorderlineColor;
```
myroot, line 1621: it should be null, 
because, backgroundExtraData.BorderlineColor default value is null
``` c#
internal Color BorderlineColor { get; set; }
```

example....
``` c#
 View view1 = ....
 // Do not set any BackgroundExtraData
 Debug.Assert(view1.BorderlineColor == Color.Black); // according to definition of BorderlineColorProperty

View view2 = ......
 // Set BorderlineOffset
 view2.BorderlineOffset = 0.1; // it will create BackgroundExtraData and 

// according to definition of BorderlineColorProperty, default value is Color.Black
// But actually it is null
 Debug.Assert(view2.BorderlineColor == Color.????); 
 
```

src/Tizen.NUI/src/public/Visuals/VisualMaps.cs
myroot: it is my miss understand, this class is not View and ViewStyle.

src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyleBindableProperty.cs
```diff
@@ -385,6 +385,27 @@ public partial class ViewStyle
             defaultValueCreator: (bindable) => ((ViewStyle)bindable).cornerRadiusPolicy
         );
 
+        /// <summary> Bindable property of BorderlineWidth. Please do not open it. </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly BindableProperty BorderlineWidthProperty = BindableProperty.Create(nameof(BorderlineWidth), typeof(float?), typeof(ViewStyle), null, propertyChanged: (bindable, oldValue, newValue) =>
+        {
+            ((ViewStyle)bindable).borderlineWidth = (float?)newValue;
+        }, defaultValueCreator: (bindable) => ((ViewStyle)bindable).borderlineWidth);
```
myroot, line 393:
```suggestion
        public static readonly BindableProperty BorderlineWidthProperty = BindableProperty.Create(nameof(BorderlineWidth), typeof(float?), typeof(ViewStyle), null);
```
It is normal code to use bindinable property, I wonder why do you use private field such as `borderlineWith`?

hinohie: If I change this code without private field, that mean "EVERY" codes in `ViewStyleBindableProperyes.cs` need to fix in this patch. I think It will not works in this PR.

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
hinohie: fix done. now value will return Color.Black as default.

src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyleBindableProperty.cs
myroot: Ok, I understand, you do like to keep a consistent with others code

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
myroot: @hinohie Could you check it?
I think, default value of BorderlineColor should be null because default value of `BackgroundExtraData`.`BorderlineColor` is null
``` c#
return view?.backgroundExtraData.BorderlineColor ?? null;
```

hinohie: @myroot  Yes I check it. Now `BackgroundExtraData.BorderlineColor` default value is `Color.Black` 

I don't set this value as null. In DALi area, It'll work as `Color.BLACK` when we don't setup `BorderlineColor`.
When we set default value as `null`, then user may get confuse. 
"Now I look some black line now. but why `BorderlineColor` doesn't return `Color.Black`??"

myroot (approved):

PR #3047, opened 2021-05-17
[NUI] Fix MatchParent to fill remaining space in LinearLayout

jaehyun0cho:
Previously, in LinearLayout, LayoutParamPolicies.MatchParent sets child
size to its parent size.

Now, in LinearLayout, LayoutParamPolicies.MatchParent calculates the
remaining space of parent and fills child in the remaining space of
parent.

There are 2 cases to fill child in the remaining space of parent.
1. MatchParent (both with or without weight)
 - If child has MatchParent specification, then child is filled in the
   remaining space of parent.
2. Specification 0 with weight
 - If child has zero specification and weight is greater than 0, then
   child is filled in the remaining space of parent proportional to its
   weight.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM. it already reviewed in DevelNUI

PR #3049, opened 2021-05-17
[NUI] Add Borderline Properties

hinohie:
Add borderline properies for view
 - BorderlineWidth (float, default = 0.0f)
   : Width of the borderline
 - BorderlineColor (Vector4, default = Color.Black)
   : Color of the borderline
 - BorderlineOffset (float, default = 0.0f)
   : Relative position offset from 'Real' borderline of visual.
     == 0.0f then half is inside, and half is outside of visual.
     == -1.0f then all borderline will be rendered inside of visual.
     == 1.0f then all borderline will be rendered outside of visual.

These three properties are animatable

If it is ImageView, this property only allow to ImageView.Property.IMAGE.
Otherwise, allow to View.Property.BACKGROUND.

Also, current version only works when Border is not setup.

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me
this is same as https://github.com/Samsung/TizenFX/pull/3035

PR #3052, opened 2021-05-17
[NUI] Calculate correct MatchParent size for ContentPage and TabView

jaehyun0cho:
Currently, WidthSpecification and HeightSpecification are updated
internally.

e.g.
In the beginning, WidthSpecification = LayoutParamPolicies.MatchParent;
After size calculation, WidthSpecification = 1280;

Until the above issue is resolved, ContentPage and TabView changes its
size calculation logic to calculate its MatchParent children size
correctly.

### API Changes ###

 - ACR:


dongsug-song (approved):
please resolve the `FIXME` parts as soon as possible~
looks good to me.

PR #3060, opened 2021-05-20
[NUI] CanvasView: Manage added drawables as a list

JSUYA:
### Description of Change ###

Canvasview manages drawables as a list so that added drawables
used in canvasview are not released by gc.

### API Changes ###

 - ACR:no acr

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/VectorGraphics/CanvasView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/VectorGraphics/CanvasView.cs
@@ -16,6 +16,7 @@
 */
 using System;
 using System.ComponentModel;
+using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 
 namespace Tizen.NUI.BaseComponents.VectorGraphics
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/BaseComponents/VectorGraphics/CanvasView.cs
```diff
@@ -68,18 +72,29 @@ protected override void ReleaseSwigCPtr(System.Runtime.InteropServices.HandleRef
         {
             Interop.CanvasView.DeleteCanvasView(swigCPtr);
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+
+            drawables.Clear();
```
dongsug-song, line 76: Hello, I think this `drawables.Clear();` moves into line 66.
managed resource, `drawables` needs to be freed before native objects be freed.
so it would be =>

        {
            if (disposed)
            {
                return;
            }
            drawables.Clear();
            base.Dispose(type);
        }

myroot: to access managed object in Dispose, we only access when disposing is true,
if not, managed object could be already released

myroot (approved):

JSUYA:
rebase patch. please merge this :)

PR #3062, opened 2021-05-20
[NUI] Fix to calculate LinearLayout's MatchParent children correctly

jaehyun0cho:
The followings have been fixed to calculate LinearLayout's MatchParent
children correctly.

- MatchParent child layout is calculated only if LinearLayout has
  positive Specification or MatchParent.

- MatchParent child layout's margin is not shared with other MatchParent
  child layouts.
  e.g.
  LinearLayout has size 100.
  Child layout1 is MatchParent and its margin is 20.
  Child layout2 is MatchParent and its margin is 0.
  Then, child layout1's size is 30 and child layout2's size is 50.

- Setting Width/HeightSpecification has been removed because
  Width/HeightSpecification should preserve the value set by user.

- Typo has been fixed. e.g. width, height

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/LinearLayout.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LinearLayout.cs
@@ -202,11 +202,11 @@ namespace Tizen.NUI
 
             if (horizontal)
             {
-                childWidthMeasureSpec = new MeasureSpecification(new LayoutLength(childsShare), MeasureSpecification.ModeType.Exactly);
+                childWidthMeasureSpec = new MeasureSpecification(new LayoutLength(childsShare - (childLayout.Margin.Start + childLayout.Margin.End)), MeasureSpecification.ModeType.Exactly);
 
                 childHeightMeasureSpec = GetChildMeasureSpecification(
                                             new MeasureSpecification(
-                                                new LayoutLength(heightMeasureSpec.Size - (childLayout.Owner.Margin.Top + childLayout.Owner.Margin.Bottom)),
+                                                new LayoutLength(heightMeasureSpec.Size - (childLayout.Margin.Top + childLayout.Margin.Bottom)),
                                                 heightMeasureSpec.Mode),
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Layouting/LinearLayout.cs
```diff
@@ -325,70 +326,78 @@ private void MeasureHorizontal(MeasureSpecification widthMeasureSpec, MeasureSpe
...
+            float remainingWidth = widthSize - totalLength;
             float totalWeightLength = 0.0f;
 
             // 2ND PHASE:
             //
             // We measure all children whose width specification policy is MatchParent without weight.
             // After 2nd phase, all children's widths are calculated without considering weight.
             // And the widths of all weighted children are accumulated to calculate weighted width.
-            foreach (LayoutItem childLayout in IterateLayoutChildren())
+            if (isWidthExactly)
```
everLEEst, line 342: Could be better to print error about circular dependency when parent is wrapContent and child is MatchParent

src/Tizen.NUI/src/public/Layouting/LinearLayout.cs
```diff
@@ -495,70 +505,78 @@ private void MeasureVertical(MeasureSpecification widthMeasureSpec, MeasureSpeci
...
+                } // 2ND PHASE foreach
+            }
 
             // 3RD PHASE:
             //
             // We measure all weighted children whose owner has weight greater than 0.
             // After 3rd phase, all weighted children has height which is proportional to their weights
             // in remaining height of parent.
-            if (totalWeight > 0)
+            if ((isHeightExactly) && (totalWeight > 0))
```
everLEEst, line 574: Could be better to print error about circular dependency when parent is wrapContent and child is MatchParent

jaehyun0cho: Thank you for your comment.
The error log has been added.

everLEEst (approved):
LGTM

PR #3065, opened 2021-05-20
[Bluetooth][Non-ACR] Add public constructor for BluetoothDevice

Wootak:

PR #3067, opened 2021-05-21
[Bluetooth][Non-ACR] Add public constructor for BluetoothDevice (#3065)

Wootak:


Wootak:
same #3065

dh79pyun (approved):
Looks good

PR #3071, opened 2021-05-21
[NUI] Fix to calculate LinearLayout's MatchParent children correctly

jaehyun0cho:
The followings have been fixed to calculate LinearLayout's MatchParent
children correctly.

- MatchParent child layout is calculated only if LinearLayout has
  positive Specification or MatchParent.

- MatchParent child layout's margin is not shared with other MatchParent
  child layouts.
  e.g.
  LinearLayout has size 100.
  Child layout1 is MatchParent and its margin is 20.
  Child layout2 is MatchParent and its margin is 0.
  Then, child layout1's size is 30 and child layout2's size is 50.

- Setting Width/HeightSpecification has been removed because
  Width/HeightSpecification should preserve the value set by user.

- Typo has been fixed. e.g. width, height

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/LinearLayout.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LinearLayout.cs
@@ -202,11 +202,11 @@ namespace Tizen.NUI
 
             if (horizontal)
             {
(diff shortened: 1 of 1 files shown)


everLEEst (approved):
LGTM

PR #3078, opened 2021-05-25
[NUI] Fix crash when background color of webview is set.

huayongxu:
### Description of Change ###

This patch is to rename some functions in WebContext too.

### API Changes ###

 - ACR:


zg2nets (approved):
looks good！

rabbitfor (commented):
It seems that this patch is including not only crash fix but also renaming WebContext APIs.
Since I already have reported to VD about out 6.5 API list so I will review it tomorrow if it is ok.

rabbitfor:
Could you rebase the patch? The base branch has been reset.

huayongxu:
> Could you rebase the patch? The base branch has been reset.

Done.

huayongxu:
> It seems that this patch is including not only crash fix but also renaming WebContext APIs.
> Since I already have reported to VD about out 6.5 API list so I will review it tomorrow if it is ok.

Old APIs are reserved still in the latest patch.

rabbitfor (approved):
LGTM

PR #3079, opened 2021-05-25
[NUI] Fix BindableObject.SetValue to allow Selector types flexibly

rabbitfor:
When the flag isCreateByXaml is true, the BindableObject.SetValue is comparing types strictly.
But some bindable properties, such as BackgroundColor, are designed to allow both types: Color and Selector<Color>.
Hence this patch allows BindableObject.SetValue to take both types.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM

PR #3090, opened 2021-05-26
[NUI][ATSPI] Not highlightable for AlertDialog

jaehyun0cho:
Since AlertDialog is a container, it is not needed to be highlightable.

### API Changes ###

 - ACR:


huiyueun (approved):
LGTM

PR #3095, opened 2021-05-27
[NUI] Disable desired size when image loading fail

tscholb:
### Description of Change ###

[NUI] Disable desired size when image loading fail
When ImageLoading is failed, GetOriginalImageSize() returns zero size.
In this case, Calculating for fittingMode is able to make error because it use invalid size.
To avoid this problem, i added this patch.

### API Changes ###

 - ACR: NONE

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
@@ -1305,30 +1305,37 @@ namespace Tizen.NUI.BaseComponents
                 if (_resourceUrl != null)
                 {
                     Size2D imageSize = ImageLoading.GetOriginalImageSize(_resourceUrl, true);
-
-                    int adjustedDesiredWidth, adjustedDesiredHeight;
-                    float aspectOfDesiredSize = (float)_desired_height / (float)_desired_width;
-                    float aspectOfImageSize = (float)imageSize.Height / (float)imageSize.Width;
-                    if (aspectOfImageSize > aspectOfDesiredSize)
+                    if( imageSize.Height > 0 && imageSize.Width > 0 && _desired_width > 0  && _desired_height > 0 )
                     {
-                        adjustedDesiredWidth = _desired_width;
-                        adjustedDesiredHeight = imageSize.Height * _desired_width / imageSize.Width;
+                        int adjustedDesiredWidth, adjustedDesiredHeight;
+                        float aspectOfDesiredSize = (float)_desired_height / (float)_desired_width;
+                        float aspectOfImageSize = (float)imageSize.Height / (float)imageSize.Width;
+                        if (aspectOfImageSize > aspectOfDesiredSize)
+                        {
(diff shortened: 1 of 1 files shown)


hinohie (approved):
LGTM. need to merge

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1305,30 +1305,36 @@ private void UpdateImage(int key, PropertyValue value)
                 if (_resourceUrl != null)
                 {
                     Size2D imageSize = ImageLoading.GetOriginalImageSize(_resourceUrl, true);
-
-                    int adjustedDesiredWidth, adjustedDesiredHeight;
-                    float aspectOfDesiredSize = (float)_desired_height / (float)_desired_width;
-                    float aspectOfImageSize = (float)imageSize.Height / (float)imageSize.Width;
-                    if (aspectOfImageSize > aspectOfDesiredSize)
+                    if( imageSize.Height > 0 && imageSize.Width > 0 )
```
hinohie, line 1308: How about to check here to `_desired_width > 0 && _desired_height > 0`  ?

`_desired_width = 0` also have same potential of error

hinohie (commented):
I think

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1305,30 +1305,37 @@ private void UpdateImage(int key, PropertyValue value)
...
+                            adjustedDesiredHeight = _desired_height;
+                        }
+
+                        PropertyValue returnWidth = new PropertyValue(adjustedDesiredWidth);
+                        imageMap?.Insert(ImageVisualProperty.DesiredWidth, returnWidth);
+                        returnWidth?.Dispose();
+                        PropertyValue returnHeight = new PropertyValue(adjustedDesiredHeight);
+                        imageMap?.Insert(ImageVisualProperty.DesiredHeight, returnHeight);
+                        returnHeight?.Dispose();
+                        PropertyValue scaleToFit = new PropertyValue((int)FittingModeType.ScaleToFill);
```
bshsqa, line 1330: Isn't this required for the failed case?
I think it is better to transfer these 3 properties to the dali as default if image loading is failed.

bshsqa: Hm.. ignore this comment if there is no problem.

dongsug-song:
@tscholb 
Hello, 
could you add some description or even a simple explanation?

tscholb:
When ImageLoading is failed, GetOriginalImageSize() returns zero size.
In this case,  Calculating for fittingMode is able to make error because it use invalid size.
To avoid this problem, i added this patch.

dongsug-song (approved):
looks good to me.
and I think this patch should be submitted and merged in API8 (tizen_6.0) also.
please check. thank you.

PR #3096, opened 2021-05-27
[NUI][ATSPI] Modify Accessbility for picker components

lwc0917:
### Description of Change ###
picker components does not need value interface.
picker works like container so it does not need highlightable.

### API Changes ###
no changes

### DIFF
--- a/src/Tizen.NUI.Components/Controls/DatePicker.cs
+++ b/src/Tizen.NUI.Components/Controls/DatePicker.cs
@@ -211,7 +211,7 @@ namespace Tizen.NUI.Components
         public override void OnInitialize()
         {
             base.OnInitialize();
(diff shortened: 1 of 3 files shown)


kimcinoo (approved):
Looks good to me
It would be good if @lukasz oleksak have a look this PR

dongsug-song:
@kimcinoo 
Hello, when Mr. lukasz gets the reviewer privilege, please add him as a reviewer to process this PR.
thank you.

lol-github (commented):
LGTM, DatePicker and TimePicker are just containers and they do not support any attributes that could be mapped to Value interface, so they shouldn't implement Value interface.

Re. Picker, it is also a container (list) of items but this items  seams to simulate range of discrete   values there are currentValue and max/min value attributes which could map to Value interface.

huiyueun (approved):
LGTM

PR #3097, opened 2021-05-27
[NUI] fix CollecitonView item sizing on MeasureFirst

everLEEst:
item sizing on MeasureFirst need to reuse pre-calculated result.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #3099, opened 2021-05-27
[NUI] Fix ScrollableBase Remove issue

huiyueun:
- [TDAF-1430] ScrollableaBase Child remove issue
 If the ScrollableBase has one child, it does not need to move the page

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
+++ b/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
@@ -729,8 +729,7 @@ namespace Tizen.NUI.Components
             {
                 // Target View is current page and also last child.
                 // CurrentPage should be changed to previous page.
-                CurrentPage = Math.Max(0, CurrentPage - 1);
-                ScrollToIndex(CurrentPage);
+                ScrollToIndex(CurrentPage - 1);
             }
 
             ContentContainer.Remove(view);


huiyueun:
https://code.sec.samsung.net/jira/browse/TDAF-1430

src/Tizen.NUI.Components/Controls/ScrollableBase.cs
```diff
@@ -725,7 +725,7 @@ public override void Add(View view)
         /// <since_tizen> 8 </since_tizen>
         public override void Remove(View view)
         {
-            if (SnapToPage && CurrentPage == Children.IndexOf(view) && CurrentPage == Children.Count - 1)
+            if (SnapToPage && CurrentPage == Children.IndexOf(view) && CurrentPage == Children.Count - 1 && Children.Count > 1)
```
bshsqa, line 728: Actually, ScrollToIndex check the input index is positive or not.
So I think it is enough remove CurrentPage = Math.Max(0, CurrentPage - 1); at the next line.
And use ScrollToIndex(CurrentPage - 1);, because CurrentPage is also set in the ScrollToIndex as a Index value.

It is just my preference that to do not want to use specific number in the code. At this case "1" for comparing.

huiyueun: Ok, I've fixed it. I think it is simple way to fix.

dongsug-song (approved):
looks good to me

PR #3101, opened 2021-05-27
[NUI][ATSPI] Default accessibility support for CollectionView and rel…

everLEEst:
Default accessibility support for CollectionView and related classes.

### DIFF
--- a/test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/internal/Common/TSAlignment.cs
+++ b/test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/internal/Common/TSAlignment.cs
@@ -1,0 +1,434 @@
+﻿using global::System;
+using NUnit.Framework;
+using NUnit.Framework.TUnit;
+using Tizen.NUI.Components;
+using Tizen.NUI.BaseComponents;
+
+namespace Tizen.NUI.Devel.Tests
+{
+    using tlog = Tizen.Log;
+
+    [TestFixture]
+    [Description("internal/Common/Alignment")]
+    public class InternalAlignmentTest
+    {
+        private const string tag = "NUITEST";
+
+        [SetUp]
+        public void Init()
+        {
+            tlog.Info(tag, "Init() is called!");
+        }
+
+        [TearDown]
+        public void Destroy()
+        {
+            tlog.Info(tag, "Destroy() is called!");
+        }
+
+        [Test]
+        [Category("P1")]
+        [Description("Alignment constructor.")]
+        [Property("SPEC", "Tizen.NUI.Alignment.Alignment C")]
+        [Property("SPEC_URL", "-")]
+        [Property("CRITERIA", "CONSTR")]
+        [Property("AUTHOR", "guowei.wang@samsung.com")]
+        public void AlignmentConstructor()
+        {
+            tlog.Debug(tag, $"AlignmentConstructor START");
+
+            var testingTarget = new Alignment();
+            Assert.IsNotNull(testingTarget, "Can't create success object Alignment");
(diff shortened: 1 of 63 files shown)


src/Tizen.NUI.Components/Controls/RecyclerView/CollectionView.cs
```diff
@@ -134,6 +134,7 @@ public CollectionView() : base()
         {
             FocusGroup = true;
             SetKeyboardNavigationSupport(true);
+            SetAccessibilityConstructor(Role.List, AccessibilityInterface.None);
```
Seoyeon2Kim, line 137: I think you don't have to add the second argument when it's `None` value. 
But, someone else may approve it. :)

m-bialota: I agree with Ms. Kim, it is better to remove the second parameter :).

m-bialota: However please consider if the CollectionView should not support Selection interface.

everLEEst: It support items selection not it's own selection.
Count can be readable and selected item count also readable then should it be value?

aswigon: Note: AccessibilityInterface.Selection PR: https://github.com/Samsung/TizenFX/pull/3110

everLEEst: okay I'll wait till it merged :)

dongsug-song:
@everLEEst 
@m-bialota 
Hello, I can assume that this PR depends on other patch which should be merged first.
so I put the **\*\*NO MERGE\*\*** label on this PR now.
**You can remove this label for yourself when this PR is ready to be reviewed and merged.**
(**\*\*NO MERGE\*\*** label has no restriction to submit PR or to edit this patch. Just for notification and marking)
Thank you.

lol-github (changes requested):

src/Tizen.NUI.Components/Controls/RecyclerView/Item/RecyclerViewItem.Internal.cs
```diff
@@ -226,6 +226,8 @@ public override void OnInitialize()
             Layout = new AbsoluteLayout();
             UpdateState();
 
+            SetAccessibilityConstructor(Role.ListItem, AccessibilityInterface.Value);
```
lol-github, line 229: IMHO, RecyclerViewItem doesn't represent a numerical value from the specified range so it shouldn't implement AccessibilityInterface.Value. Removing second parameter should be OK for this class.

However, this class should override the AccessibilityCalculateStates method to support "Selected" and "Selectable" states.

src/Tizen.NUI.Components/Controls/RecyclerView/Item/DefaultGridItem.cs
```diff
@@ -265,7 +265,10 @@ protected virtual TextLabel CreateLabel(TextLabelStyle textStyle)
         [EditorBrowsable(EditorBrowsableState.Never)]
         protected virtual ImageView CreateImage(ImageViewStyle imageStyle)
         {
-            return new ImageView(imageStyle);
+            return new ImageView(imageStyle)
+            {
+                AccessibilityHighlightable = false,
```
lol-github, line 270: this line might not be needed soon, as there is a change which makes ImageView not highlightable by default:
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-toolkit/+/259931/

lol-github (changes requested):

PR #3115, opened 2021-05-28
[NUI] Disable desired size when image loading fail

tscholb:
### Description of Change ###
When ImageLoading is failed, GetOriginalImageSize() returns zero size.
In this case, Calculating for fittingMode is able to make error because it use invalid size.
To avoid this problem, i added this patch.

### API Changes ###
none


dongsug-song (approved):
looks good to me.
this is the same patch of the merged one in master branch which resolves "Divided By Zero" exception.

PR #3120, opened 2021-05-28
[NUI] Cancel pressed state properly on touch events

everLEEst:
fix bug that touch highlight remaining when touch is not stable.


jaehyun0cho (approved):
LGTM

PR #3125, opened 2021-05-31
[NUI] Set Notification Window transparent

jaehyun0cho:
Notification creates its own Window.
Not to cover the existing view behind Notification Window, Notification
Window's BackgroundColor is set to be transparent.

### API Changes ###

 - ACR:


rabbitfor (commented):
Nice catch :)

PR #3127, opened 2021-05-31
[NUI] Fix ABI break for API8 NUI image

dongsug-song:
### Description of Change ###
[NUI] Fix ABI break for API8 NUI image
- GetOriginalImageSize() was Hidden API and it was changed as to use default parameter in API9.
- This caused ABI break for tizen_6.0 application which references API9 NUI dll.
- Revert back to API8's GetOriginalImageSize() method form.

### API Changes ###
none


tscholb (approved):
LGTM :)

Seoyeon2Kim (approved):

PR #3138, opened 2021-06-01
[NUI] Fix Navigator's animation and Page's visible state

jaehyun0cho:
Previously, Navigator's animation did not keep the final state.
e.g. Opacity from 1 to 0 animation did not keep opacity 0 when the
     animation was finished.
Now, Navigator's animation keeps the final state.

Previously, hidden Pages did not have actually hidden visibilities.
e.g. When the peek Page was hidden by newly pushed Page, the hidden
     Page's visible state was true.
Now, hidden Pages have actually hidden visibilities.

### API Changes ###

 - ACR:


huiyueun (approved):
LGTM

PR #3142, opened 2021-06-02
[NUI] Fix button CellPadding policy and default CornerRadius

rabbitfor:

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Button.Internal.cs
+++ b/src/Tizen.NUI.Components/Controls/Button.Internal.cs
@@ -335,13 +335,17 @@ namespace Tizen.NUI.Components
             buttonText.Unparent();
             overlayImage?.Unparent();
 
+#pragma warning disable CA2000
+            Size2D cellPadding = String.IsNullOrEmpty(buttonText.Text) ? new Size2D(0, 0) : itemSpacing;
+#pragma warning restore CA2000
+
             if (IconRelativeOrientation == IconOrientation.Left)
             {
                 Layout = new LinearLayout()
                 {
                     LinearOrientation = LinearLayout.Orientation.Horizontal,
                     LinearAlignment = itemAlignment,
-                    CellPadding = itemSpacing ?? new Size2D(0, 0)
+                    CellPadding = cellPadding
                 };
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI.Components/Controls/Button.Internal.cs
```diff
@@ -335,13 +335,15 @@ protected virtual void LayoutItems()
             buttonText.Unparent();
             overlayImage?.Unparent();
 
+            Size2D cellPadding = String.IsNullOrEmpty(buttonText.Text) ? new Size2D(0, 0) : null;
```
dongsug-song, line 338: could you check the VS warning below? =>
**_src/Tizen.NUI.Components/Controls/Button.Internal.cs#L338
Use recommended dispose pattern to ensure that object created by 'new Size2D(0, 0)' is disposed on all paths. If possible, wrap the creation within a 'using' statement or a 'using' declaration. Otherwise, use a try-finally pattern, with a dedicated local variable declared before the try region and an unconditional Dispose invocation on non-null value in the 'finally' region, say 'x?.Dispose()'. If the object is explicitly disposed within the try region or the dispose ownership is transfered to another object or method, assign 'null' to the local variable just after such an operation to prevent double dispose in 'finally'._**


how about this? ^^;

            Size2D cellPadding = null;
            if (String.IsNullOrEmpty(buttonText.Text))
            {
                 cellPadding = new Size2D(0, 0);
            }
**_after cellPadding is used_** 

            cellPadding.Dispose();

rabbitfor: @dongsug-song 
I disabled warning in the line, because the ownership of disposing `cellPadding` is passed to the Layout object.
So if we call `cellPadding.Dispose()` here, there will be a crash.

dongsug-song (approved):
looks good to me

src/Tizen.NUI.Components/Controls/Button.Internal.cs
dongsug-song: ok. I see

PR #3144, opened 2021-06-02
[NUI] Fix Picker value changed not invoked when align animation ended

lwc0917:
### Description of Change ###
Fix value change is not invoke when animation finished

### API Changes ###
no changes


jaehyun0cho (approved):
LGTM

PR #3145, opened 2021-06-02
[NUI] Fix to require Layout to fit to children or fill to parent

jaehyun0cho:
This patch is necessary to the following patch.
"[NUI] Deprecate ResizePolicyType"

It is impossible that LayoutParamPolicies.WrapContent supports both
ResizePolicyType.UseNaturalSize and FitToChildren.

Therefore, LayoutParamPolicies.WrapContent does not support
ResizePolicyType.UseNaturalSize nor FitToChildren.
Instead, LayoutParamPolicies.WrapContent makes View fit to its children
only if the View sets its Layout.

In the same way, LayoutParamPolicies.MatchParent does not support
ResizePolicyType.FillToParent.
Instead, LayoutParamPolicies.MatchParent makes View fill to its parent
only if the View sets its Layout.

If View uses LayoutParamPolicies.WrapContent or MatchParent while it
does not set its Layout, then the View does not fit to its children nor
fill to its parent.
Instead, ResizePolicyType.UseNaturalSize is supported by default.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyle.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyle.cs
@@ -247,7 +247,7 @@ namespace Tizen.NUI.BaseComponents
         /// Gets or sets the width resize policy to be used.
         /// </summary>
         /// <since_tizen> 9 </since_tizen>
-        [Obsolete("Deprecated. Please use Tizen.NUI.View.WidthSpecification instead.")]
+        [Obsolete("Deprecated. Please set Tizen.NUI.View.Layout and use Tizen.NUI.View.WidthSpecification instead.")]
         public ResizePolicyType? WidthResizePolicy
         {
             get => (ResizePolicyType?)GetValue(WidthResizePolicyProperty);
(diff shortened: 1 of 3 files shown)


Seoyeon2Kim (approved):
Looks good to me.
For a long term, this PR should be merged and it'll be updated more.

bshsqa:
This patch looks prevent NUI Layout to affect View's resize policy.
Then... if we deprecated resize policy and width/height specification not affect the resize policy, how we can control non-layout View's size policy?

jaehyun0cho:
> This patch looks prevent NUI Layout to affect View's resize policy.
> Then... if we deprecated resize policy and width/height specification not affect the resize policy, how we can control non-layout View's size policy?

Based on this PR, View can refer its parent or child's size only if the View's Layout is set and Width/HeightSpecification is used.

Up to now, non-layout View users should use Width/HeightResizePolicy.
(Width/HeightSpecification does not work for non-layout View)

Since Width/HeightResizePolicy are deprecated, they are not used any more.
Users need to use Width/HeightSpecification so they need to set View's Layout.

I think the issue case that you raised should be handled by a new approach.
The issue case
- If a non-layout view has children and the non-layout view is added to layout view, then the non-layout view's children positions refer their Width/HeightSpecification which are not set intentionally by user.

PR #3146, opened 2021-06-03
[Camera] Hide CDM interface from user

hsgwon:
### Description of Change ###

Hide CDM interface from user

### DIFF
--- a/src/Tizen.Multimedia.Camera/Camera/Camera.cs
+++ b/src/Tizen.Multimedia.Camera/Camera/Camera.cs
@@ -58,6 +58,8 @@ namespace Tizen.Multimedia
         /// <feature> http://tizen.org/feature/camera </feature>
         public Camera(CameraDevice device)
         {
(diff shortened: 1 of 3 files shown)


src/Tizen.Multimedia.Camera/Camera/Camera.cs
```diff
@@ -58,56 +58,46 @@ public partial class Camera : IDisposable, IDisplayable<CameraError>
         /// <feature> http://tizen.org/feature/camera </feature>
         public Camera(CameraDevice device)
         {
```
aferin, line 60: How about checking validity here?
```suggestion
        {ValidationUtil.ValidateEnum(typeof(CameraDevice), device, nameof(device));}
```

hsgwon: OK I'll add it.

aferin (approved):

PR #3148, opened 2021-06-03
[NUI] make IsGrouped setter internal as only the GroupHeaderTemplate can enable it

everLEEst:
### Description of Change ###
[NUI] make IsGrouped setter internal as only the GroupHeaderTemplate can enable it

### API Changes ###
none


dongsug-song (approved):
looks good to me

PR #3155, opened 2021-06-04
[NUI] Add Page's Appeared and Disappeared events

jaehyun0cho:
Page's Appeared event is invoked when the Navigator's animation is
finished and the page becomes visible.

Page's Disappeared event is invoked when the Navigator's animation is
finished and the page becomes invisible.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM

PR #3164, opened 2021-06-08
[Camera] Add new CameraPixelFormat - VP8,VP9

hsgwon:
### Description of Change ###

Add new CameraPixelFormat - VP8, VP9

### API Changes ###

 - TCSACR-427


src/Tizen.Multimedia.Camera/Camera/CameraEnums.cs
```diff
@@ -240,7 +240,17 @@ public enum CameraPixelFormat
         /// Encoded pixel format : Motion JPEG
         /// </summary>
         /// <since_tizen> 8 </since_tizen>
-        Mjpeg
+        Mjpeg,
+        /// <summary>
+        /// Encoded pixel format : VP8
```
wiertel, line 245: The change is ok, no need to modify anything. I just want to share comment: when I see an abbreviation that is not cleat to me at a glance I think we could link it to description, e.g. on Wikipedia. On the other hand, everybody who is going to choose this value most probably know what it means.
```suggestion
        /// Encoded pixel format : <a href="https://en.wikipedia.org/wiki/VP8">VP8</a>
```
I you decide to apply this, then please do the same to VP9.

hsgwon: Thank you for your suggestion. 
But, I think that users of these APIs are familiar with these terms. 
Even if not, they can find what it means easily.

aferin (approved):

PR #3167, opened 2021-06-08
[NUI] Fix RelativeLayout's descendant size and position calculations

jaehyun0cho:
Previously, 2 descendant sizes and positions calculation problems
exitsted on RelativeLayout as follows.

1. MatchParent grand children's sizes and positions were not calculated.
 - RelativeLayout set its children's sizes and positions but it did not
   set its children's MeasuredWidth/Height.
 - If children's MeasuredWidth/Height are not set, then the MatchParent
   grand children's sizes and positions are not calculated.

2. MatchParent children's sizes fill to the RelativeLayout by default.
 - MatchParent children's sizes should not fill to the RelativeLayout by
   default because children's sizes and positions should be calculated
   by RelativeLayout's APIs.

Now, the above problems have been fixed as follows.

1. RelativeLayout sets its children's MeasuredWidth/Height, so its
   MatchParent grand children's sizes and positions are calculated
   correctly.

2. MatchParent children's sizes are calculated by RelativeLayout's APIs.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/LinearLayout.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LinearLayout.cs
@@ -354,7 +354,12 @@ namespace Tizen.NUI
                     {
                         needToMeasure = true;
                     }
-                    else
+                    // RelativeLayout's MatchParent children should not fill to the RelativeLayout.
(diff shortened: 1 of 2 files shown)


everLEEst (approved):
LGTM

PR #3169, opened 2021-06-09
[NUI] Fix implicit dispose exception defect

dongsug-song:
### Description of Change ###
[NUI] Fix implicit dispose exception defect

- when implicit dispose is done by DisposeQueue, the view which has empty body of dali gives exception.
- the "in Children" => this.Children give null reference exception.

### API Changes ###
none


rabbitfor (approved):
LGTM

Seoyeon2Kim (approved):

PR #3174, opened 2021-06-09
[NUI] Add source type to TapGesture.

JoogabYun:
This is similar to MouseButton in TouchEvent.

Now, you can see from which input the tap was made.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Events/TapGesture.cs
+++ b/src/Tizen.NUI/src/public/Events/TapGesture.cs
@@ -87,6 +87,20 @@ namespace Tizen.NUI
             }
         }
 
+        /// <summary>
+        /// The gesture source type of touches property (read-only).
+        /// If you tap with a mouse button, this will tell you which mouse input you tapped.
+        /// Primary(Left), Secondary(Right). Tertiary(Wheel).
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public GestureSourceType SourceType
+        {
+            get
+            {
+                return sourceType;
+            }
+        }
+
         private uint numberOfTaps
         {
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Events/TapGesture.cs
```diff
@@ -87,6 +87,18 @@ public Vector2 LocalPoint
             }
         }
 
+        /// <summary>
+        /// The gesture source type of touches property.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public GestureSourceType SourceType
```
rabbitfor, line 94: Could you add  more description about the `SourceType` and its enum values? I am not a expert for a gesture system so I don't get what it means.

JoogabYun: 네 설명을 좀 더 추가했습니다.

src/Tizen.NUI/src/public/Events/TapGesture.cs
```diff
@@ -87,6 +87,20 @@ public Vector2 LocalPoint
             }
         }
 
+        /// <summary>
+        /// The gesture source type of touches property (read-only).
+        /// If you tap with a mouse button, this will tell you which mouse input you tapped.
+        /// Primary(Left), Secondary(Right). Center(Wheel).
```
myroot, line 93: It is not matched with `GestureSourceType` you intent `Tertiary`?
```suggestion
        /// Primary(Left), Secondary(Right). Tertiary(Wheel).
```

dongsug-song (approved):
looks good to me

PR #3175, opened 2021-06-09
[NUI] Add SpaceEvenly to FlexJustification

jaehyun0cho:
SpaceAround distributes the equal space to before, and after each item.
Therefore, the space between adjacent items is twice bigger than the
space between the beginning edge and the first item.

To distribute the equal space to before, between, and after each item,
SpaceEvenly is added to FlexJustification.

Requires:
https://review.tizen.org/gerrit/259554/

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
Looks good to me,.

PR #3187, opened 2021-06-14
[NUI] Fix Slider Thumb shown issue

Seoyeon2Kim:

### Description of Change ###
- After default theme update, Slider Thumb disappears when user presses it.
- The reason is that the default resource is not enough,
 and the code is just to get the empty resource. (null value)
- Fixed to show slider thumb for normal, pressed, and disabled states.

### API Changes ###
- N/A


src/Tizen.NUI.Components/Controls/Slider.cs
```diff
@@ -1460,7 +1460,19 @@ private void UpdateState(bool isFocusedNew, bool isPressedNew)
             if (!isFocused && !isPressed)
             {
                 ControlState = ControlState.Normal;
-                thumbImage.ResourceUrl = thumbImageUrlSelector?.Normal;
+
+                if (thumbImageUrlSelector != null)
+                {
+                    thumbImage.ResourceUrl = thumbImageUrlSelector?.Normal;
```
rabbitfor, line 1466: Once the `thumbImage.ResourceUrlSelector` is set to `thumbImageUrlSelector` then you don't have to set image resource url manually whenever the control state is changed. There is a property to use for this case: `View.EnableControlStatePropagation`.

Please refer Button's case,
https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI.Components/Controls/Button.Internal.cs#L273

Seoyeon2Kim: Removed thumbImageUrlSelector variable and added 'EnableControlState = true' to Thumb ImageView.

rabbitfor (approved):
:)

bshsqa (approved):

hinohie (approved):
LGTM! Finally we use BorderlineWidth!

PR #3189, opened 2021-06-14
[NUI] Add ImageUrl

tscholb:
ImageUrl is a class for wrapping url that is get from external buffer.
it is used for lifecycle management of external buffers.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.ImageUrl.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.ImageUrl.cs
@@ -1,0 +1,34 @@
+/*
+ * Copyright(c) 2021 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+namespace Tizen.NUI
+{
+    using global::System;
+    using global::System.Runtime.InteropServices;
+
+    internal static partial class Interop
+    {
+        internal static partial class ImageUrl
+        {
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_delete_ImageUrl")]
+            public static extern void Delete(HandleRef jarg1);
+
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_ImageUrl_Get")]
+            public static extern string GetUrl(HandleRef jarg1);
+        }
+    }
+}
--- a/src/Tizen.NUI/src/public/Images/ImageUrl.cs
+++ b/src/Tizen.NUI/src/public/Images/ImageUrl.cs
@@ -1,0 +1,76 @@
+/*
+ * Copyright(c) 2021 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+using System;
+using System.ComponentModel;
+
+namespace Tizen.NUI
+{
+    using global::System;
+    using global::System.ComponentModel;
+    using global::System.Runtime.InteropServices;
+    using Tizen.NUI.BaseComponents;
+
+    /// <summary>
+    /// Url for Image
+    /// </summary>
+    /// <remarks>Hidden API: Only for inhouse or developing usage. The behavior and interface can be changed anytime.</remarks>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class ImageUrl : BaseHandle
+    {
+        internal ImageUrl(IntPtr cPtr, bool cMemoryOwn) : base(cPtr, cMemoryOwn)
+        {
+        }
+
+        /// <summary>
+        /// Dispose
+        /// </summary>
+        /// <param name="type"></param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected override void Dispose(DisposeTypes type)
+        {
+            if (disposed)
+            {
+                return;
+            }
+
+            if (type == DisposeTypes.Explicit)
+            {
+                //Called by User
+                //Release your own managed resources here.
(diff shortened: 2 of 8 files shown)


src/Tizen.NUI/src/public/Images/ImageUrl.cs
```diff
@@ -0,0 +1,108 @@
...
+            Interop.ImageUrl.Delete(swigCPtr);
+            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+        }
+
+        /// <summary>
+        /// Set Url to ImageUrl
+        /// </summary>
+        /// <param name="url">The uniform resource identifier to load.</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public void SetUrl (string url)
```
rabbitfor, line 82: Please remove a space between `SetUrl` and `(`.

rabbitfor (commented):
In C# class, if there are of SetXXX and GetXXX, they normally to be a [Property](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties).

rabbitfor:
It seems that the current `ImageUrl` class only provide string `Get/SetUrl` method. Do you have any plan to add more methods or properties? If not, I don't think the API `ImageUrl` type of NativeImageSource.Url` need to be provided. (It can be just a string type.)

src/Tizen.NUI/src/public/Images/ImageUrl.cs
rabbitfor: And in normal case, if a class provide SetXXX and GetXXX, it can be a (Property)[https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties].

src/Tizen.NUI/src/public/Images/ImageUrl.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            Interop.ImageUrl.SetUrl(SwigCPtr, url);
+            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+        }
+
+                /// <summary>
+        /// Set Url to ImageUrl
+        /// </summary>
+        /// <param name="url">The uniform resource identifier to load.</param>
```
rabbitfor, line 96: Invalid XML comment. Please check a build warning.

src/Tizen.NUI/src/internal/Interop/Interop.ImageUrl.cs
```diff
@@ -0,0 +1,40 @@
...
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_new_ImageUrl")]
+            public static extern IntPtr New();
+
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_ImageUrl_New_With_Url")]
+            public static extern IntPtr New(string jarg1);
+
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_delete_ImageUrl")]
+            public static extern void Delete(HandleRef jarg1);
+
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_get_ImageUrl")]
```
bshsqa, line 36: Get

src/Tizen.NUI/src/public/Images/ImageUrl.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// This will not be public opened.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected override void ReleaseSwigCPtr(HandleRef swigCPtr)
+        {
+            Interop.ImageUrl.Delete(swigCPtr);
+            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+        }
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override String ToString()
```
myroot, line 78:
```suggestion
        public override string ToString()
```

myroot: How about provide method to get a url?

```c#
public string Url => ToString();
```

src/Tizen.NUI/src/public/Images/NativeImageSource.cs
```diff
@@ -38,18 +38,14 @@ internal NativeImageSource(IntPtr cPtr, bool cMemoryOwn) : base(Interop.NativeIm
         }
 
         /// <summary>
-        /// Get URI from native image source.
+        /// Generate Url from current native image source.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public string Url
+        public ImageUrl GenerateCurrentUrl()
```
myroot, line 44: You do better keep API, and manage a lifecycle of ImageUrl in `NativeImageSource`


``` c#
ImageUrl _imageUrl = null;
public string Url 
{
    get 
    {
          _imageUrl?.Dispose();
          _imageUrl = new ImageUrl(Interop.NativeImageSource.GenerateUrl(this.SwigCPtr.Handle), true);
            if (NDalicPINVOKE.SWIGPendingException.Pending) throw new InvalidOperationException("FATAL: get Exception", NDalicPINVOKE.SWIGPendingException.Retrieve());
         return _imageUrl.ToString();
    }
}

```

tscholb: GenerateCurrentUrl() changes every time it's called,
so I think the method is better than the property.
What do you think?

myroot: Even though every times called, the content of `ImageUrl` will not changed if `NativeImageSource` was not changed. 
So, I thinks it is a property of `NativeImageSource`

myroot: If I optimize it, I will add a boolean field to check buffer was dirty and if image buffer was dirty, I will generate a new ImageUrl, if not I will return cached ImageUrl, 


``` c#

bool _isDirty = false;
public IntPtr AcquireBuffer(ref int width, ref int height, ref int stride) 
{
   _isDirty = true;
 ...
}


ImageUrl _imageUrl = null;
public string Url 
{
    get 
    {
          if (!_isDirty && _imageUrl != null)
                return _imageUrl;

          _imageUrl?.Dispose();
           _isDirty = false;
          _imageUrl = new ImageUrl(Interop.NativeImageSource.GenerateUrl(this.SwigCPtr.Handle), true);
            if (NDalicPINVOKE.SWIGPendingException.Pending) throw new InvalidOperationException("FATAL: get Exception", NDalicPINVOKE.SWIGPendingException.Retrieve());
         return _imageUrl.ToString();
    }
}
```

tscholb: Unfortunately, ImageUrl seems to be newly created every time even if nativeImageSource is not changed.
For example, texture needs to be created internally when the raw buffer is changed.
so I think the method is better.

myroot: If you want to provide as method, I recommend to rename like `GenerateSnapshot`
I don't know the detail implemention of it, but `ImageUrl` seem to not connect a `NativeImageSource`, it looks just a snapshot if `NativeImageSource`, 
So in this view, Method will be better than property

tscholb: because of this problem, I added Get"Current" Url but it still ambiguous.
I will think more about this.

myroot (approved):

PR #3193, opened 2021-06-14
[NUI] Fix flickering buttons in ActionContent of AlertDialog

jaehyun0cho:
The buttons in the ActionContent of AlertDialog are flickering because
the CellPadding of ActionContent's LinearLayout did not set initially.

To resolve this issue, the CellPadding of ActionContent's LinearLayout
is set initially.

However, this should be updated to apply CellPadding from
AlertDialogStyle.

### API Changes ###

 - ACR:


rabbitfor (approved):
LGTM

PR #3204, opened 2021-06-17
[NUI] fix IsSelected default value true to false

everLEEst:
### Description of Change ###

IsSelected Property defaultValue need to be false not true.


rabbitfor (approved):
LGTM

jaehyun0cho (approved):
LGTM

PR #3206, opened 2021-06-17
[NUI] make Tizen.NUI.Devel.Tests be built in csharp-tct by copying the folder itself

dongsug-song:
### Description of Change ###
[NUI] make Tizen.NUI.Devel.Tests be built in csharp-tct by copying the folder itself
- only related to csharp-tct, no change of NUI
- just change folder's name
- fix csproj file

### API Changes ###
none


dongsug-song:
can not resolve conflict so I will upload again.

PR #3208, opened 2021-06-17
[NUI] Add description for Text component's PropertyMap.

wonrst:

### Description of Change ###

In the case of NUI PropertyMap, it is difficult for users to understand the name, type, and meaning of key and value.
This PR added description to all PropertyMap of Text Components and added some examples.


dongsug-song (approved):
finally, you have found the way to add table in comment! great!
looks good to me and other members' reviews are required.

rabbitfor (approved):

PR #3213, opened 2021-06-18
[NUI] Fix not to make LinearLayout.CellPadding nullable

jaehyun0cho:
LinearLayout.CellPadding is Size2D, so it is not nullable type.
Therefore, if null is assigned to LinearLayout.CellPadding, then it
returns Size2D(0, 0).

### API Changes ###

 - ACR:


lwc0917 (commented):
LGTM

dongsug-song (approved):
looks good to me

PR #3215, opened 2021-06-21
[NUI] Add fade transition

bshsqa:
Add Fade Transition for the appearing/disappearing transition of Page

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Navigation/Page.cs
+++ b/src/Tizen.NUI.Components/Controls/Navigation/Page.cs
@@ -57,7 +57,22 @@ namespace Tizen.NUI.Components
     /// <since_tizen> 9 </since_tizen>
     public abstract class Page : Control
     {
+        private const int DefaultTransitionDuration = 500;
+
         private Navigator navigator = null;
+
+        // Default transition is Fade.
+        private TransitionBase appearingTransition = new Fade()
+        {
+            TimePeriod = new TimePeriod(DefaultTransitionDuration),
+            AlphaFunction = new AlphaFunction(AlphaFunction.BuiltinFunctions.Default),
+        };
+
+        private TransitionBase disappearingTransition = new Fade()
+        {
+            TimePeriod = new TimePeriod(DefaultTransitionDuration),
+            AlphaFunction = new AlphaFunction(AlphaFunction.BuiltinFunctions.Default),
+        };
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


src/Tizen.NUI.Components/Controls/Navigation/Page.cs
```diff
@@ -89,6 +104,38 @@ internal set
             }
         }
 
+        /// <summary>
+        /// Transition properties for the transition of Views in this page during this page is pushed to Navigator.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public TransitionBase AppearingTransition
```
dongsug-song, line 111: could you check the https://code.sec.samsung.net/jira/browse/GRE-2025 ?
there seems to be an issue if the `TransitionBase AppearingTransition` used as like : `page.AppearingTransition.AlphaFuction = xxx or page.AppearingTransition.TimePeriod = xxx`
this is why GRE-2025 deprecated setters and use only parameter constructor.
I think it looks late because TransitionBase is already opened as public.
but anyway, I just want to talk about the old hot issue which is similar to this. thank you.

src/Tizen.NUI.Components/Controls/Navigation/Page.cs
```diff
@@ -57,8 +57,23 @@ public class PageDisappearedEventArgs : EventArgs
     /// <since_tizen> 9 </since_tizen>
     public abstract class Page : Control
     {
+        private static readonly int DefaultTransitionDuration = 500;
```
dongsug-song, line 60: please check the VS warning below.

**src/Tizen.NUI.Components/Controls/Navigation/Page.cs#L60
Field 'DefaultTransitionDuration' is declared as 'readonly' but is initialized with a constant value. Mark this field as 'const' instead.**

bshsqa: Thank you I fixed.

src/Tizen.NUI.Components/Controls/Navigation/Page.cs
bshsqa: HAT!!!!!!!!!!!
Sorry I didn't know the history.
I think it didn't create such a big problem in the ordinary use case of transition fortunately.
But I will remember that......
Thank you.

dongsug-song (approved):
looks good to me

PR #3221, opened 2021-06-22
[NUI] Add SelectedChanged event to the SelectGroup and etc.

rabbitfor:
* Fix default SelectGroup.SelectedIndex value to -1.
* Add event SelectGroup.SelectedChanged.
* Add method SelectGroup.AddAllToView().
* Add method RadioButtonGroup.GetSelectedItem().
* Add attached property RadioButtonGroup.IsGroupHolder. View can make a RadioButtonGroup for its radio button children.
```xaml
<View x:Name="rootView" RadioButtonGroup.IsGroupHolder="true">
  <RadioButton Text="1"/>
  <RadioButton Text="2"/>
</View>
```
```cs
RadioButtonGroup.GetRadioButtonGroup(rootView).SelectedChanged += (s, e) => {
   // TODO
};
```

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/RadioButtonGroup.cs
+++ b/src/Tizen.NUI.Components/Controls/RadioButtonGroup.cs
@@ -14,7 +14,11 @@
  * limitations under the License.
  *
  */
+
+using System;
 using System.ComponentModel;
+using Tizen.NUI.BaseComponents;
+using Tizen.NUI.Binding;
 
 namespace Tizen.NUI.Components
 {
@@ -35,6 +39,14 @@ namespace Tizen.NUI.Components
     public class RadioButtonGroup : SelectGroup
     {
         /// <summary>
+        /// IsGroupHolderProperty
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly BindableProperty IsGroupHolderProperty = BindableProperty.CreateAttached("IsGroupHolder", typeof(bool), typeof(View), false, propertyChanged: OnIsGroupHolderChanged);
+
+        private static readonly BindableProperty RadioButtonGroupProperty = BindableProperty.CreateAttached("RadioButtonGroup", typeof(RadioButtonGroup), typeof(View), null, propertyChanged: OnRadioButtonGroupChanged);
+
+        /// <summary>
         /// Construct RadioButtonGroup
         /// </summary>
         /// <since_tizen> 6 </since_tizen>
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI.Components/Controls/SelectGroup.cs
```diff
@@ -1,5 +1,5 @@
 ﻿/*
- * Copyright(c) 2019 Samsung Electronics Co., Ltd.
+ * Copyright(c) 2019-2021 Samsung Electronics Co., Ltd.
```
bshsqa, line 2: Why don't we not use 2021 only?

src/Tizen.NUI.Components/Controls/RadioButtonGroup.cs
```diff
@@ -85,5 +129,64 @@ public void Remove(RadioButton radio)
...
+                newGroup.OnChildChanged(view, null);
+            }
+        }
+
+        private void OnChildChanged(object sender, EventArgs args)
+        {
+            if (sender is View view)
+            {
+                RemoveAll();
+                foreach (var child in view.Children) if (child is RadioButton radioButton) Add(radioButton);
```
bshsqa, line 179: Is it OK in NUI convention? (I don't know exactly).

src/Tizen.NUI.Components/Controls/SelectGroup.cs
rabbitfor: I was guided to update copyright with a range rather than a single year in another tizen project.
I have searched for it. Some articles indicate that removing the original year can cause problems, so I think we need to discuss it with other members.

> The copyright notice on a work establishes a claim to copyright. The date on the notice establishes how far back the claim is made. This means if you update the date, you are no longer claiming the copyright for the original date and that means if somebody has copied the work in the meantime and they claim its theirs on the ground that their publishing the copy was before your claim, then it will be difficult to establish who is the originator of the work.

src/Tizen.NUI.Components/Controls/RadioButtonGroup.cs
rabbitfor: Fixed!

bshsqa (approved):

src/Tizen.NUI.Components/Controls/SelectGroup.cs
bshsqa: Oh... thank you for your good information.

PR #3241, opened 2021-06-29
[NUI] Fix to support simple MatchParent in RelativeLayout

jaehyun0cho:
This reverts a part of d5ac94d24b3ca635034b57324e527e253e2ed789 to
support simple MatchParent in RelativeLayout.

For easier use, child fills its parent if child has MatchParent without
using RelativeLayout APIs.

### API Changes ###

 - ACR:


huiyueun (approved):
LGTM

PR #3248, opened 2021-07-01
[Camera] Change the way to invoke CDM event

hsgwon:
### Description of Change ###

Change the way to invoke CDM event

### DIFF
--- a/src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
+++ b/src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
@@ -154,8 +154,11 @@ namespace Tizen.Multimedia
         private int connectionCallbackId = 0;
         private void RegisterDeviceConnectionChangedCallback()
         {
+            Log.Debug(CameraLog.Tag, "Enter");
+
             _deviceConnectionChangedCallback = (ref Native.CameraDeviceStruct device, bool status, IntPtr userData) =>
             {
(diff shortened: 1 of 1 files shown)


src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
```diff
@@ -115,12 +114,33 @@ private static string GetString(char[] word)
...
+                _deviceConnectionChanged += value;
+            }
+            remove
+            {
+                _deviceConnectionChanged -= value;
+
+                if (_deviceConnectionChanged == null)
+                {
+                    UnregisterDeviceConnectionChangedCallback();
+                }
```
aferin, line 141: How about unregistering callback first?

                if (_deviceConnectionChanged == value)
                {
                    UnregisterDeviceConnectionChangedCallback();
                }

                _deviceConnectionChanged -= value;

hsgwon: This code intend to remove event first, and then if there's no added event(that is, `_deviceConnectionChanged` is null), unregister native callback also.
If we call `UnregisterDeviceConnectionChangedCallback()` when `_deviceConnectionChanged == value` , the remained event cannot be invoked anymore.

aferin: Okay.

aferin (approved):

PR #3256, opened 2021-07-02
[NUI] Add new features for window

sparrow74:
1. Supports IME window and application.
It is only for internal IME window and application.
These new APIs should be not public.

2. Supports floating window mode.

3. Supports window resized/moved by display server.

### DIFF
--- a/src/Tizen.NUI/src/internal/Application/Application.cs
+++ b/src/Tizen.NUI/src/internal/Application/Application.cs
@@ -1090,6 +1090,19 @@ namespace Tizen.NUI
             return instance;
         }
 
+        public static Application NewApplication(string stylesheet, NUIApplication.WindowMode windowMode, Rectangle positionSize, WindowType type)
+        {
+            if (instance)
+            {
+                return instance;
+            }
+            Application ret = New(1, stylesheet, windowMode, positionSize, type);
+            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+
+            instance = ret;
+            return instance;
+        }
+
         /// <summary>
         /// Ensures that the function passed in is called from the main loop when it is idle.
         /// </summary>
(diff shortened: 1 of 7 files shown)


src/Tizen.NUI/src/internal/Application/Application.cs
```diff
@@ -1090,6 +1090,19 @@ public static Application NewApplication(string[] args, string stylesheet, NUIAp
             return instance;
         }
 
+        public static Application NewApplication(string stylesheet, NUIApplication.WindowMode windowMode, Rectangle positionSize, WindowType type)
+        {
+            if (instance)
+            {
+                return instance;
+            }
+            Application ret = New(1, stylesheet, windowMode, positionSize, type);
```
dongsug-song, line 1099: could you add the first parameter(`1`) meaning with comment? or can be defined as constant like `private const int SomeMeaningfulName = 1;`

src/Tizen.NUI/src/public/Application/NUIApplication.cs
```diff
@@ -157,6 +157,22 @@ public NUIApplication(Size2D windowSize, Position2D windowPosition, ThemeOptions
...
+        /// </summary>
+        /// <param name="styleSheet">The styleSheet URL.</param>
+        /// <param name="windowMode">The windowMode.</param>
+        /// <param name="windowSize">The window size.</param>
+        /// <param name="windowPosition">The window position.</param>
+        /// <param name="type">The default window type.</param>
+        /// <since_tizen> 9 </since_tizen>
+        /// This should be hidden. Because the major purpose is to support IME window as keyboard application.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public NUIApplication(string styleSheet, WindowMode windowMode, Size2D windowSize, Position2D windowPosition, WindowType type) : base(new NUICoreBackend(styleSheet, windowMode, windowSize, windowPosition, type))
```
dongsug-song, line 171: is `stylesheet` the URL of a resource for background image of IME? I couldn't get what `stylesheet` is, so I am just asking.

src/Tizen.NUI/src/internal/Application/Application.cs
sparrow74: It just is dummy value. I referred the other constructor in this file.

src/Tizen.NUI/src/public/Application/NUIApplication.cs
sparrow74: It is not sure how to use by IME. It just supports the service all of them.
After feedback by Input team, I will change this API's interface.

dongsug-song (approved):
looks good to me

PR #3257, opened 2021-07-02
[Camera] Add extra stream format for production

hsgwon:
### Description of Change ###

Add extra stream format for production


src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
```diff
@@ -101,7 +101,7 @@ internal static ReadOnlyCollection<CameraDeviceInformation> GetDeviceInformation
         }
 
         internal static CameraDeviceInformation GetDeviceInformation(Native.CameraDeviceStruct device) =>
-            new CameraDeviceInformation(device.Type, device.device, GetString(device.name), GetString(device.id));
+            new CameraDeviceInformation(device.Type, device.device, device.name, device.id, device.extraStreamNum);
 
         private static string GetString(char[] word)
```
WonyoungChoi, line 106: `GetString()` is deprecated now. It seems possible to delete it.

hsgwon: OK. I'll remove `GetString()` method, which is not used anymore.

aferin (approved):

PR #3258, opened 2021-07-02
[NUI] Fix issue of SetValue to the property which hasn't DefaultValue Create delegate

AdunFang:
[NUI] Fix issue of SetValue to the property which hasn't DefaultValue Create delegate

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

zg2nets (commented):
lgtm

PR #3263, opened 2021-07-05
[NUI] Rename APIs related to context menu.

huayongxu:
### Description of Change ###

This patch is related to the patch of dali-csharp-binder:
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/260245/

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/WebView/WebContextMenu.cs
+++ b/src/Tizen.NUI/src/internal/WebView/WebContextMenu.cs
@@ -64,19 +64,6 @@ namespace Tizen.NUI
         }
 
         /// <summary>
-        /// Gets position of the context menu.
-        /// </summary>
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        public Vector2 Position
-        {
-            get
-            {
-                IntPtr result = Interop.WebContextMenu.GetPosition(SwigCPtr);
-                return new Vector2(result, true);
-            }
-        }
-
-        /// <summary>
         /// Returns the nth item in a context menu.
         /// <param name="index">The position of the item</param>
         /// </summary>
--- a/src/Tizen.NUI/src/internal/Interop/Interop.WebContext.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.WebContext.cs
@@ -30,8 +30,20 @@ namespace Tizen.NUI
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_WebContext_SetProxyUri")]
             public static extern void SetProxyUri(global::System.Runtime.InteropServices.HandleRef jarg1, string jarg2);
 
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_WebContext_GetProxyUri")]
+            public static extern string GetProxyUri(global::System.Runtime.InteropServices.HandleRef jarg1);
+
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_WebContext_SetProxyBypassRule")]
+            public static extern void SetProxyBypassRule(global::System.Runtime.InteropServices.HandleRef jarg1, string jarg2, string jarg3);
+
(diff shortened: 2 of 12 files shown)


zg2nets (approved):
looks good!

src/Tizen.NUI/src/internal/WebView/WebContextMenu.cs
```diff
@@ -63,19 +63,6 @@ public WebContextMenuItemList ItemList
             }
         }
 
-        /// <summary>
-        /// Gets position of the context menu.
-        /// </summary>
```
rabbitfor, line 68: The VD requirements list include "ewk_context_menu_pos_x_get", "ewk_context_menu_pos_y_get".
I thought they are matched with WebContextMenu.Position.
After removing WebContext.Position, which API is patched with the VD requirements?

huayongxu: Mr Youngha Jung said in the mail:
"ewk_context_menu_pos_x_get and ewk_context_menu_pos_y_get return the position of the context menu created using elm popup.
In your case, I think you can return the position of the nui popup."

Both APIs are used for elm popup, not required for NUI.

rabbitfor: Are we using NUI popup to post context menu (or should we implement)? I thought the context menu appears inside of webview so it is rendered by web engine.

huayongxu: Context menu popup could be implemented by NUI apps, just like JavaScript alert/prompt/confirm popups.
When apps users long press or do something else, ContextMenuShown event would be triggered.
When ContextMenuShown event is received, NUI apps would get menu information from WebContextMenu,
for example, how many menu items would be shown, then a popup would be created and shown.
If elm popup is shown, its style is different from that of NUI. And NUI apps might have their own UI styles, 
So I followed the comments of Mr Youngha Jung, and removed WebContext.Position in dali and NUI.

rabbitfor: Could you add example of creating context menu using NUI controls?

huayongxu: I added a MenuView into test/Tizen.NUI.WebViewTest/SimpleWebViewApp.cs .
MenuView is very simple. It would be shown when touched, hidden when touched again.

rabbitfor (approved):
I have questions a lot, but to sync with dali patches I am going to merge this for now.

rabbitfor:
How action do I need to do to show the context menu?

rabbitfor:
When a user select an item in a context menu, how the application pass that information to the web engine?

rabbitfor:
In the example, the application is listening ContextMenuHidden event but doing nothing when it it triggered.
And the application is hide context menu by itself when user touches the menu area.
The example make me think that the closing context menu is doing by application.
If so, why the ContextMenuHidden event exist?

huayongxu:
> In the example, which action do I need to do to show the context menu?

When web page is clicked (touched)  or long-pressed by mouse,
ContextMenuShown (contextmenu,show) event would be triggered.
Currently how to sync long-press of NUI with that of chromium-efl, it might be an issue.
I will discuss it with web engine team by mail.

huayongxu:
> When a user select an item in a context menu, how the application pass that information to the web engine? The example only showing text label and it does not functioning so it is hard to understand how it works. Can you update in the separate PR?

OK.
Regarding the issue how the application pass that information to the web engine,
web engine need provide more APIs to do something.

rabbitfor:
> > When a user select an item in a context menu, how the application pass that information to the web engine? The example only showing text label and it does not functioning so it is hard to understand how it works. Can you update in the separate PR?
> 
> OK.
> Regarding the issue how the application pass that information to the web engine,
> web engine need provide more APIs to do something.

It seems that there already is an API, WebContextMenu.SelectItem(). Doesn't it work?

huayongxu:
> > > When a user select an item in a context menu, how the application pass that information to the web engine? The example only showing text label and it does not functioning so it is hard to understand how it works. Can you update in the separate PR?
> > 
> > 
> > OK.
> > Regarding the issue how the application pass that information to the web engine,
> > web engine need provide more APIs to do something.
> 
> It seems that there already is an API, WebContextMenu.SelectItem(). Doesn't it work?

I forgot it.
This API works. No more API need be added.

PR #3265, opened 2021-07-06
[NUI] Fix Shadow issue when we set BoxShadow = null

hinohie:
When we set view.BoxShadow.Color = new Color(~~);
BoxShadow.Color have same reference with noColor.
So noColor also become changed

We make another refernce when setup Color by noColor.
Now Color doesn't have same reference with noColor.

(Same jobs doing for noOffset and noExtents)

### DIFF
--- a/src/Tizen.NUI/src/public/ViewProperty/Shadow.cs
+++ b/src/Tizen.NUI/src/public/ViewProperty/Shadow.cs
@@ -70,7 +70,7 @@ namespace Tizen.NUI
         /// </summary>
         internal Shadow(PropertyMap propertyMap) : base(propertyMap)
         {
-            Color = noColor;
+            Color = new Color(noColor);
             PropertyValue pValue = propertyMap.Find(ColorVisualProperty.MixColor);
             pValue?.Get(Color);
             pValue?.Dispose();
--- a/src/Tizen.NUI/src/public/ViewProperty/ShadowBase.cs
+++ b/src/Tizen.NUI/src/public/ViewProperty/ShadowBase.cs
@@ -57,8 +57,8 @@ namespace Tizen.NUI
         {
             Debug.Assert(propertyMap != null);
 
-            Offset = noOffset;
-            Extents = noExtents;
+            Offset = new Vector2(noOffset);
+            Extents = new Vector2(noExtents);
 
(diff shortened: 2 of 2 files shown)


src/Tizen.NUI/src/public/ViewProperty/Shadow.cs
```diff
@@ -70,7 +70,8 @@ public Shadow(Shadow other) : this(other == null ? throw new ArgumentNullExcepti
         /// </summary>
         internal Shadow(PropertyMap propertyMap) : base(propertyMap)
         {
-            Color = noColor;
+            Color?.Dispose();
```
rabbitfor, line 73: Since it is constructor and the Color is a member of the Shadow class, so it does not need to dispose.

src/Tizen.NUI/src/public/ViewProperty/ShadowBase.cs
```diff
@@ -57,8 +57,10 @@ protected ShadowBase(PropertyMap propertyMap)
         {
             Debug.Assert(propertyMap != null);
 
-            Offset = noOffset;
-            Extents = noExtents;
+            Offset?.Dispose();
```
rabbitfor, line 60: Ditto.

src/Tizen.NUI/src/public/ViewProperty/ShadowBase.cs
```diff
@@ -57,8 +57,10 @@ protected ShadowBase(PropertyMap propertyMap)
...
+            Extents?.Dispose();
```
rabbitfor, line 61: Ditto.

src/Tizen.NUI/src/public/ViewProperty/Shadow.cs
hinohie: Fix done

src/Tizen.NUI/src/public/ViewProperty/ShadowBase.cs
hinohie: Fix done

rabbitfor (approved):
:)

JoogabYun (approved):

everLEEst (approved):
LGTM.

PR #3268, opened 2021-07-06
[NUI] Fix AlertDialog not to use ViewStyle.

rabbitfor:
ViewStyle is the last applied style which is not the same as the current style.
Hence it is not recommanded to use ViewStyle inside components code.
(Please note that View.ViewStyle is deprecated and better not to use.)

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #3273, opened 2021-07-07
[NUI] Fix to calculate RelativeLayout's grand children position corre…

jaehyun0cho:
…ctly

Previously, RelativeLayout set its children View's size and position but the
children Layout's MeasuredWidth/Height did not match to the children View's size
and position.
This causes that the grand children's position was calculated incorrectly
because the grand children's position was calculated based on the children's
MeasuredWidth/Height.

Now, RelativeLayout sets its children Layout's MeasuredWidth/Height based on the
children View's size and the grand children's position is calculated correctly.

### API Changes ###

 - ACR:


jaehyun0cho:
More investigation is required for the case that grand children have MatchParent Width/HeightSpecification.

PR #3275, opened 2021-07-07
[NUI] Fix to calculate RelativeLayout's children MeasuredWidth/Height correctly

jaehyun0cho:
RelativeLayout sets its children's size in OnLayout().
Therefore, the children's MeasuredWidth/Height are not the same with the
children's size.
This causes that the grand children's MeasuredWidth/Height are calculated
incorrectly.
To resolve the above, RelativeLayout updates its children's MeasuredWidth/Height
based on the children's actual size on RelativeLayout after the RelativeLayout's
MeasuredWidth/Height are calculated.

e.g.
Let parent have RelativeLayout and parent's size be 1920x1080.
Let child have WrapContent with SetFillHorizontal/Vertical true.
Let grand child have MatchParent.
Then, child's size is 1920x1080 but child's MeasuredWidth/Height is 0x0.
Then, grand child's MeasuredWidth/Height is 0x0 and size is 0x0.

### API Changes ###

 - ACR:


huiyueun (approved):
LGTM

jaehyun0cho:
In this patch, RelativeLayout updates its children's MeasuredWidth/Height to be the same with the children's actual size on RelativeLayout.
This patch enables that the grand children's MeasuredWidth/Height are calculated based on the children's actual size on RelativeLayout.

jaehyun0cho:
@huiyueun 
Now, this patch is the final version.
Please review this patch when you are available.
Thank you.

PR #3277, opened 2021-07-08
[Camera] Handle GetDeviceInformation returns null

hsgwon:
### Description of Change ###

Consider there's no connected camera device.

### DIFF
--- a/src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
+++ b/src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
@@ -17,7 +17,6 @@ using System.Linq;
 
 using System;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using System.ComponentModel;
 using Native = Interop.CameraDeviceManager;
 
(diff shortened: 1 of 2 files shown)


src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
```diff
@@ -63,10 +63,16 @@ public CameraDeviceManager()
...
+        public bool IsExternalCameraConnected
+        {
+            get
+            {
+                var deviceList = GetDeviceInformation();
+
+                return deviceList == null ? false : deviceList.Where(d => d.Type == CameraDeviceType.Usb ||
+                                                                          d.Type == CameraDeviceType.Network).Any();
+            }
+        }
```
myroot, line 75: How about use `?.` null conditon operator?
```suggestion
        public bool IsExternalCameraConnected =>
            GetDeviceInformation()?.Where(d => d.Type == CameraDeviceType.Usb ||
                                              d.Type == CameraDeviceType.Network)
                                  .Any() ?? false;
```

src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
```diff
@@ -86,6 +92,11 @@ public ReadOnlyCollection<CameraDeviceInformation> GetDeviceInformation()
 
         internal static ReadOnlyCollection<CameraDeviceInformation> GetDeviceInformation(Native.CameraDeviceListStruct list)
         {
+            if (list.count == 0)
```
myroot, line 95: IMO, I think empty collection is more safe and meaningful than null

hsgwon: Right, I'll return Empty instead of null.

src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
hsgwon: I changed the return type of GetDeviceInformation(), null to Empty.
So, I think we don't need to check anymore whether it's null or not. I'll change this like the following.
```
public bool IsExternalCameraConnected =>
    GetDeviceInformation().Where(d => d.Type == CameraDeviceType.Usb ||
                                      d.Type == CameraDeviceType.Network)
                          .Any();
```

myroot (approved):

PR #3278, opened 2021-07-08
[NUI] Add Api for getting window from view

tscholb:
In some case, The view needs to know which window it is added to.
we can using this api in this case.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window/Window.cs
+++ b/src/Tizen.NUI/src/public/Window/Window.cs
@@ -59,6 +59,25 @@ namespace Tizen.NUI
         }
 
         /// <summary>
+        /// A helper method to get the current window where the view is added
+        /// </summary>
+        /// <param name="view">The View added to the window</param>
+        /// <returns>A Window.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        static public Window Get(View view)
+        {
+            if(view == null)
+            {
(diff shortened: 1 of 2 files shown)


dongsug-song:
please update the relevant csharp-binder patch.

src/Tizen.NUI/src/public/Window/Window.cs
```diff
@@ -58,6 +58,19 @@ internal Window(global::System.IntPtr cPtr, bool cMemoryOwn) : base(cPtr, cMemor
...
 
+        /// <summary>
+        /// A helper method to get the current window where the view is added
+        /// </summary>
+        /// <param name="view">The View added to the window</param>
+        /// <returns>A Window.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        static public Window Get(View view)
+        {
+            Window ret = Registry.GetManagedBaseHandleFromNativePtr(Interop.Window.Get(View.getCPtr(view))) as Window;
```
dongsug-song, line 69: How about adding some code to check the validation of `view`. if view has no body, then `ret` would be `null` ?

tscholb: Thank you. i updated it

dongsug-song (approved):
looks good to me

PR #3281, opened 2021-07-09
[NUI] Make sure that Interop return type is bool

hinohie:

This patch will fix this problem


dongsug-song (approved):
looks good to me, but other member's review also be required~

rabbitfor (approved):
Wow. Thank you for your work.

PR #3287, opened 2021-07-09
[NUI] Fix RelativeLayout Measure size calculation

jaehyun0cho:
Previously, the actual spec size is given to MeasureChild() and MeasureChild()
subtracts Padding size from the actual spec size.

Now, the actual spec size with Padding size is given to MeasureChild() to
display child with the actual spec size.

Moreover, since child's MeasuredWidth/Height are calculated in MeasureChild(),
duplicate MeasureWidth/Height assigning codes are removed.

### API Changes ###

 - ACR:


huiyueun (approved):
LGTM

everLEEst:
LGTM.

everLEEst (approved):
LGTM

PR #3291, opened 2021-07-09
[NUI] Add EncodedImageBuffer class (Load image frome Stream)

hinohie:
This patch will make some simple way to upload image from C# Stream to View.

- EncodedImageBuffer will create by C# Stream.
  EncodedImageBuffer buffer = new EncodedImageBuffer(stream);
- Generate ImageUrl
  ImageUrl url = buffer.GenerateUrl();
- You can use this ImageUrl as image's url
  ImageView view = new ImageView(url.GetUrl());

This patch required below dali patches
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/261019/
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-toolkit/+/259415/

### DIFF
--- a/src/Tizen.NUI/src/public/Images/EncodedImageBuffer.cs
+++ b/src/Tizen.NUI/src/public/Images/EncodedImageBuffer.cs
@@ -1,0 +1,170 @@
+/*
+ * Copyright(c) 2021 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+using System;
+using System.ComponentModel;
+
+namespace Tizen.NUI
+{
+    using global::System;
+    using global::System.ComponentModel;
+    using global::System.Runtime.InteropServices;
+
+    /// <summary>
+    /// Class for Encoded Image Buffer.
+    /// Buffer comes from System.IO.Stream.
+    /// This data will decode internally when you use GeneratedUrl as View's ResourceUrl.
+    /// Note: This class doesn't allow to fix/change anything.
+    /// Only constructor allow to setup data.
+    /// </summary>
+    /// <remarks>Hidden API: Only for inhouse or developing usage. The behavior and interface can be changed anytime.</remarks>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class EncodedImageBuffer : BaseHandle
+    {
+        private ImageUrl mCachedImageUrl = null; // cached Generated Url.
+        private VectorUnsignedChar mCachedBuffer = null; // cached encoded raw buffer
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="stream">The Stream of the image file.</param>
+        /// <exception cref="ArgumentNullException"> Thrown when stream is null. </exception>
+        /// <exception cref="InvalidOperationException"> Thrown when stream don't have any data. </exception>
+        /// <remarks>Hidden API: Only for inhouse or developing usage. The behavior and interface can be changed anytime.</remarks>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public EncodedImageBuffer(System.IO.Stream stream) : this(GetRawBuffrFromStreamHelper(stream))
+        {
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Images/EncodedImageBuffer.cs
```diff
@@ -0,0 +1,151 @@
...
+        /// Get VectorUnsignedChar from System.IO.Stream.
+        /// This funcion is private (exist only for Constructor.)
+        /// </summary>
+        private static VectorUnsignedChar GetRawBuffrFromStreamHelper(System.IO.Stream stream)
+        {
+            if(stream == null)
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+            long streamLength = stream.Length - stream.Position;
```
myroot, line 128: Do not use `Position` property, Only specific streams support it
https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.position?view=net-5.0#remarks

hinohie: Fix done. Thank you for catch my mistake.
I make same mistake at `ImageLoading.cs`. I'll fix it another PR

src/Tizen.NUI/src/public/Images/EncodedImageBuffer.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            if(stream == null)
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+            if(!stream.CanRead)
+            {
+                throw new InvalidOperationException("stream don't support to read");
+            }
+            long streamLength = stream.Length;
```
myroot, line 132: `Length` also not support all types of Stream
https://docs.microsoft.com/ko-kr/dotnet/api/system.io.stream.length?view=net-5.0#--

You do better to copy a stream on MemoryStream

https://github.com/Samsung/TizenFX/blob/90572287f263121d0e1bceaeb1273f99e8d5cd24/src/ElmSharp/ElmSharp/Image.cs#L484-L501

hinohie: If we change like that, Whole stream reading is the only possible works. Is it okay?

src/Tizen.NUI/src/public/Images/EncodedImageBuffer.cs
```diff
@@ -0,0 +1,165 @@
...
+
+            // Allocate buffer that internal DALi engine can read
+            VectorUnsignedChar buffer = new VectorUnsignedChar();
+
+            buffer.Resize((uint)readableStreamLength);
+            var bufferBegin = buffer.Begin();
+            global::System.Runtime.InteropServices.HandleRef bufferRef = SWIGTYPE_p_unsigned_char.getCPtr(bufferBegin);
+
+            // Copy data from memoryStream to buffer
+            System.Runtime.InteropServices.Marshal.Copy(streamData, 0, bufferRef.Handle, readableStreamLength);
```
myroot, line 160: How about use `GetBuffer` instead copy into `streamData`,
https://docs.microsoft.com/ko-kr/dotnet/api/system.io.memorystream.getbuffer?view=net-5.0
```suggestion

            // Allocate buffer that internal DALi engine can read
            VectorUnsignedChar buffer = new VectorUnsignedChar();

            buffer.Resize((uint)streamLength);
            var bufferBegin = buffer.Begin();
            global::System.Runtime.InteropServices.HandleRef bufferRef = SWIGTYPE_p_unsigned_char.getCPtr(bufferBegin);

            // Copy data from memoryStream to buffer
            System.Runtime.InteropServices.Marshal.Copy(memoryStream.GetBuffer(), 0, bufferRef.Handle, streamLength);
```

src/Tizen.NUI/src/public/Images/EncodedImageBuffer.cs
myroot: I don't know exactly means of `Whole stream reading`. 
As I understand you need to know a stream size before create a native buffer.
So reading a whole stream is necessary to know a stream size.

myroot: If we possible to append data into native buffer without a final size, we can use Stream without copy to memoryStream

src/Tizen.NUI/src/public/Images/EncodedImageBuffer.cs
hinohie: Good points. Fix done.

src/Tizen.NUI/src/public/Images/EncodedImageBuffer.cs
hinohie: `VectorUnsignedChar` support `append` operation only for single byte.
(Appending raw-buffers are not support, but appending `VectorUnsignedChar` is support.)
That's why I get final size and copy it ;-;

hinohie: I imagine that some user push `used middle of stream` and try to load images from that.
But... I'll not care about them anymore.

myroot (approved):

myroot (approved):

PR #3294, opened 2021-07-12
[NUI] Fix build erros of Devel.NUI.Tests.

AchoWang:
### Description of Change ###
1. Chage "Url" to "GenerateUrl()"
2. New TCs for ImageUrl.

### API Changes ###

 - ACR:

### DIFF
--- a/test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/public/Images/TSNativeImageSource.cs
+++ b/test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/public/Images/TSNativeImageSource.cs
@@ -11,7 +11,7 @@ namespace Tizen.NUI.Devel.Tests
     using tlog = Tizen.Log;
 
     [TestFixture]
-    [Description("public/Common/NativeImageSource")]
+    [Description("public/Images/NativeImageSource")]
     public class PublicNativeImageSourceTest
     {
         private const string tag = "NUITEST";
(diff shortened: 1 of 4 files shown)


test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/public/Images/TSPixelData.cs
```diff
@@ -76,31 +76,6 @@ public void PixelDataConstructorByIntPtr()
...
-        {
-            tlog.Debug(tag, $"PixelDataUrl START");
-
-            byte[] buffer = new byte[1024];
-
-            var testingTarget = new PixelData(buffer, 1024, 100, 150, PixelFormat.L8, PixelData.ReleaseFunction.Free);
-            Assert.IsNotNull(testingTarget, "Can't create success object PixelData");
-            Assert.IsInstanceOf<PixelData>(testingTarget, "Should be an instance of PixelData type.");
-
-            var result = testingTarget.Url;
```
hinohie, line 96: Could you just change it as `GenerateUrl()`;

test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/public/Images/TSNativeImageSource.cs
```diff
@@ -47,28 +47,6 @@ public void NativeImageSourceConstructor()
...
-        [Property("AUTHOR", "guowei.wang@samsung.com")]
-        public void NativeImageSourceUrl()
-        {
-            tlog.Debug(tag, $"NativeImageSourceUrl START");
-
-            var testingTarget = new NativeImageSource(100, 50, NativeImageSource.ColorDepth.Bits16);
-            Assert.IsNotNull(testingTarget, "Can't create success object NativeImageSource");
-            Assert.IsInstanceOf<NativeImageSource>(testingTarget, "Should be an instance of NativeImageSource type.");
-
-            var result = testingTarget.Url;
```
hinohie, line 65: Could you just change it as `GenerateUrl();`? It will fix everything

AchoWang: Done

test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/public/Images/TSPixelData.cs
AchoWang: Done

hinohie:
Thank you for adding `TSImageUrl.cs` files.
But actually, `ImageUrl` doesn't work like that.

It is return-only value and we cannot make this class in C# area
(You can see ImageUrl.cs does'nt have any public constructor)
And more... `ImageUrl` don't use input as `ImageView`

I think it would be better to leave us to make TC for `ImageUrl`.

AchoWang:
> Thank you for adding `TSImageUrl.cs` files.
> But actually, `ImageUrl` doesn't work like that.
> 
> It is return-only value and we cannot make this class in C# area
> (You can see ImageUrl.cs does'nt have any public constructor)
> And more... `ImageUrl` don't use input as `ImageView`
> 
> I think it would be better to leave us to make TC for `ImageUrl`.

Ok, I remove it, thanks!

dongsug-song (approved):
looks good to me

PR #3313, opened 2021-07-14
[NUI] Fix AppBar Content's Layout properties

jaehyun0cho:
Since LinearLayout's size calculation logic has been fixed, AppBar Content's
Layout properties are fixed to calculate size correctly.

After fixing LinearLayout's size calculation logic, Weight is valid only if
Width/HeightSpecification is MatchParent.

AppBarStyle's ActionView and ActionButton's default size is fixed to show its
own size.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
+++ b/src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
@@ -449,13 +449,13 @@ namespace Tizen.NUI.Components
                 },
                 ActionView = new ViewStyle()
                 {
-                    Size = new Size(-1, 120),
+                    Size = new Size(48, 120),
                     CornerRadius = 0,
                     BackgroundColor = new Color(0, 0, 0, 0),
                 },
                 ActionButton = new ButtonStyle()
                 {
-                    Size = new Size(-1, 120),
+                    Size = new Size(-2, 120),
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
```diff
@@ -449,13 +449,13 @@ public Theme Create()
...
                 {
-                    Size = new Size(-1, 120),
+                    Size = new Size(48, 120),
                     CornerRadius = 0,
                     BackgroundColor = new Color(0, 0, 0, 0),
                 },
                 ActionButton = new ButtonStyle()
                 {
-                    Size = new Size(-1, 120),
+                    Size = new Size(-2, 120),
```
rabbitfor, line 458: Does `-2` mean `Wrap contents`?

jaehyun0cho: yes, right.
In ViewStyle there is no Width/HeightSpecification. So Size -1, -2 are used.

rabbitfor: Thank you to check. I should add them to the ViewStyle.

rabbitfor (approved):
LGTM

PR #3314, opened 2021-07-15
[NUI] Fix Menu's BackgroundColor

jaehyun0cho:
Menu covers the window size to dismiss Menu when outside of Menu is touched.

Therefore, Menu's BackgroundColor should be transparent and Menu Content's
BackgroundColor should be set instead.

### API Changes ###

 - ACR:


rabbitfor (approved):
LGTM

PR #3321, opened 2021-07-18
[NUI] Fix to measure RelativeLayout's children sizes correctly

jaehyun0cho:
Previously, TextLabel and TextField's Ellipsis were not displayed in
RelativeLayout, because TextLabel and TextField's size always covers the full
text string.

Now, if TextLabel and TextField support Ellipsis, then RelativeLayout sets the
exact space size to the TextLabel and TextField. So if the assigned size does
not cover the full text string, then Ellipsis is displayed.

Previously, redundant Meausre() was called again in OnMeasure().

Now, Measure() is called again only for FillHorizontal/Vertical cases to set the
filled size to the MeasuredWidth/Height to calculate grand children's size
correctly.

### API Changes ###

 - ACR:


huiyueun (approved):
LGTM

PR #3322, opened 2021-07-18
[NUI] Change RelativeLayout's default offset values

jaehyun0cho:
To refer the parent layout's area by default, RightRelativeOffset and
BottomRelativeOffset are changed from 0.0f to 1.0f.

### API Changes ###

 - ACR:


huiyueun (commented):
LGTM

PR #3324, opened 2021-07-19
[NUI] Fix the build error when referencing to NUI.XamlBuild of which version is higher than 1.0.11

dongsug-song:
### Description of Change ###
[NUI] Fix the build error when referencing to NUI.XamlBuild of which version is higher than 1.0.11

### API Changes ###
none


everLEEst (approved):
LGTM

PR #3329, opened 2021-07-20
[NUI] Reset transition order to match scene tree order

bshsqa:


dongsug-song (approved):
looks good to me

PR #3330, opened 2021-07-20
[NUI] Fix the Window.VisibilityChanged event's crash issue

dongsug-song:
### Description of Change ###
[NUI] Fix the Window.VisibilityChanged event's crash issue
- Because the signals which binds dali, contain signals as member variables, the cMemoryOwn should be false.
- Without this fix, the WindowEventsTest.cs gets crash when continously pushing left,right direction key.

### API Changes ###
none


huiyueun (approved):
LGTM

PR #3337, opened 2021-07-21
[Tizen.ApplicationManager] Add a new internal method

hjhun:
### Description of Change ###

The AttachWindowBelow() method is added for inhouse applications.
The method is to attach the window of the child application below the
window of the parent application.

Added: 
 - ApplicationManager.AttachWindowBelow()


pjh9216 (approved):
LGTM

PR #3338, opened 2021-07-22
[NUI] Fix to call base.Initialize() of LinearLayouter and GridLayouter

jaehyun0cho:
Previously, base.Initialize() of LinearLayouter and GridLayouter was not
called if source was empty.
This caused Container to be null and caused null reference in
RequestLayout().

Now, it is guaranteed that base.Initialize() of LinearLayouter and
GridLayouter is called.

### API Changes ###

 - ACR:


everLEEst (approved):
Looks good to me.

PR #3339, opened 2021-07-22
[NUI] Added string key to Placeholder PropertyMap

wonrst:
The Placeholder map use string key to set value in NUI.
But, user must use the index key to get value.

This causes confusion and inconvenience to users.
This patch allows users to get value using string key.


dongsug-song (approved):
looks good to me.
but from next time, please add a test code or sample code. 
thank you.

PR #3356, opened 2021-07-29
[NUI] CanvasView: Add constructor that can be created without viewbox

JSUYA:
### Description of Change ###

Add a constructor so that it can be created without viewBox argument.

### API Changes ###

 - ACR:no

Dependency patch
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/261876/


dongsug-song (approved):
looks good to me

PR #3361, opened 2021-07-30
[NUI] CanvasView:: Add ViewBox Property

JSUYA:
### Description of Change ###

Add a property that can change the viewbox value of CanvasView.
If the viewbox is smaller than the view size, the canvas will appear larger.
Conversely, if the viewbox is large, the canvas will look small.

### API Changes ###

 - ACR:No

Dependency patch
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/261995/

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/VectorGraphics/CanvasView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/VectorGraphics/CanvasView.cs
@@ -90,6 +90,29 @@ namespace Tizen.NUI.BaseComponents.VectorGraphics
         }
 
         /// <summary>
+        /// The viewbox of the CanvasView.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Size2D ViewBox
+        {
+            get
+            {
+                Size2D retVal = new Size2D(0, 0);
+                PropertyValue viewBoxPropertyValue = GetProperty(Interop.CanvasView.PropertyViewBoxGet());
+                viewBoxPropertyValue?.Get(retVal);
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/BaseComponents/VectorGraphics/CanvasView.cs
```diff
@@ -89,6 +89,29 @@ protected override void ReleaseSwigCPtr(System.Runtime.InteropServices.HandleRef
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
         }
 
+        /// <summary>
+        /// The viewbox of the CanvasView.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Size2D ViewBox
```
dongsug-song, line 96: Hello, could you please check this first? 
https://code.sec.samsung.net/jira/browse/GRE-2025
we currently recommend NOT to use class type as a property. 
so I think it would be good to change `Size2D ViewBow` property to other type.
This is just an example :
public int ViewBoxSizeWidth, public int ViewBoxSizeHeight, public SetViewBoxSize(Size2D size), and etc.
please check again. thank you.

JSUYA: I updated the patch. Replaced by SetViewBoxSize/GetViewBoxSize methods.

dongsug-song (approved):
looks good to me.
but after dali-csharp-binder's patch is merged, then this PR can be merged.

PR #3365, opened 2021-08-02
[NUI] Apply BorderlineWidth to ImageView and ImageView.Background

hinohie:
Add missing implements about borderline.

```
            roundImageView = new ImageView();
            roundImageView.CornerRadius = 50;
            roundImageView.BackgroundColor = Color.Blue;
            roundImageView.BorderlineWidth = 5;
            roundImageView.ResourceUrl = IMAGE_PATH;
```

Previous version doesn't make borderline like upper samples.
Cause there was some implement missing when we setup `ResoureUrl` first time.
It is fixed now

+

Previous version don't apply borderline to it's background.
But when we use both `Background~~~` and `ResourceUrl`, 
there will occur some logical error
(`CornerRadius` calculated from outer-line of borderline. but when we don't setup `BorderlineWidth` at `Background` layer, gap occured between inner-line of borderline and `Background`.)

So now, we also apply borderline to background of `ImageView` to solve these case
- When we set both `Background~~~` and `ResourceUrl`.
- When we don't set `ResourceUrl` and set only `Background~~~`.

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
@@ -1121,7 +1121,9 @@ namespace Tizen.NUI.BaseComponents
 
         internal override void ApplyBorderline()
         {
-            // Ignore BACKGROUND borderline property. only apply borderline to IMAGE.
+            base.ApplyBorderline();
+
+            // Apply borderline to IMAGE.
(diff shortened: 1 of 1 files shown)


Seoyeon2Kim (approved):
Looks good to me, but I want to get others' reviews.

tscholb (approved):
LGTM :)

PR #3400, opened 2021-08-09
[NUI] Add CursorPositionChanged Event

abdullehghujeh:
Add CursorPositionChanged event, 
which will be called every time the cursor moves.

editor.CursorCursorPositionChanged += (s, e) =>
{
    e.OldCursorPosition; //cursor position before the change
};

related patch : https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/262362

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/TextEditorEvent.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/TextEditorEvent.cs
@@ -315,7 +315,7 @@ namespace Tizen.NUI.BaseComponents
                 CursorPositionChangedEventArgs e = new CursorPositionChangedEventArgs();
 
                 // Populate all members of "e" (CursorPositionChangedEventArgs) with real data
-                e.CursorPosition = oldPosition;
+                e.OldCursorPosition = oldPosition;
 
                 //here we send all data to user event handlers
                 textEditorCursorPositionChangedEventHandler?.Invoke(this, e);
--- a/src/Tizen.NUI/src/public/BaseComponents/TextEditorEvent.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/TextEditorEvent.cs
@@ -318,7 +318,7 @@ namespace Tizen.NUI.BaseComponents
                 e.CursorPosition = oldPosition;
 
                 //here we send all data to user event handlers
-                textEditorCursorPositionChangedEventHandler(this, e);
+                textEditorCursorPositionChangedEventHandler?.Invoke(this, e);
             }
         }
 
--- a/src/Tizen.NUI/src/public/BaseComponents/TextEditorEvent.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/TextEditorEvent.cs
@@ -52,7 +52,7 @@ namespace Tizen.NUI.BaseComponents
         private delegate void ScrollStateChangedCallbackDelegate(IntPtr textEditor, ScrollState state);
 
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
-        private delegate void CursorPositionChangedCallbackDelegate(IntPtr textEditor, uint oldPosition, uint newPosition);
+        private delegate void CursorPositionChangedCallbackDelegate(IntPtr textEditor, uint oldPosition);
 
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
         private delegate void MaxLengthReachedCallbackDelegate(IntPtr textEditor);
@@ -308,15 +308,14 @@ namespace Tizen.NUI.BaseComponents
             }
         }
 
-        private void OnCursorPositionChanged(IntPtr textEditor, uint oldPosition, uint newPosition)
+        private void OnCursorPositionChanged(IntPtr textEditor, uint oldPosition)
         {
             if (textEditorCursorPositionChangedEventHandler != null)
             {
                 CursorPositionChangedEventArgs e = new CursorPositionChangedEventArgs();
 
                 // Populate all members of "e" (CursorPositionChangedEventArgs) with real data
-                e.OldCursorPosition = oldPosition;
-                e.CursorPosition = newPosition;
+                e.CursorPosition = oldPosition;
 
                 //here we send all data to user event handlers
                 textEditorCursorPositionChangedEventHandler(this, e);
--- a/src/Tizen.NUI/src/public/BaseComponents/TextEditorEvent.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/TextEditorEvent.cs
@@ -33,6 +33,9 @@ namespace Tizen.NUI.BaseComponents
         private EventHandler<ScrollStateChangedEventArgs> textEditorScrollStateChangedEventHandler;
         private ScrollStateChangedCallbackDelegate textEditorScrollStateChangedCallbackDelegate;
 
+        private EventHandler<CursorMovedEventArgs> textEditorCursorMovedEventHandler;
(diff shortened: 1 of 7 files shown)


src/Tizen.NUI/src/public/BaseComponents/TextEditorEvent.cs
```diff
@@ -371,6 +426,27 @@ public ScrollState ScrollState
...
 
+        /// <summary>
+        /// The CursorMoved event arguments.
+        /// </summary>
+        public class CursorMovedEventArgs : EventArgs
+        {
+            /// <summary>
+            /// TextEditor.
+            /// </summary>
+            public TextEditor TextEditor { get; set;}
```
myroot, line 437: I think this property is not need, because sender will be TextEditor

src/Tizen.NUI/src/public/BaseComponents/TextEditorEvent.cs
```diff
@@ -371,6 +426,27 @@ public ScrollState ScrollState
...
+
+            /// <summary>
+            /// cursor postion before the move.
+            /// </summary>
+            public uint OldCursorPosition { get; set;}
+
+            /// <summary>
+            /// cursor postion after the move.
+            /// </summary>
+            public uint NewCursorPosition { get; set;}
```
myroot, line 447: How about?
```suggestion
            public uint CursorPosition { get; set; }
```

src/Tizen.NUI/src/public/BaseComponents/TextFieldEvent.cs
```diff
@@ -73,6 +78,32 @@ public partial class TextField
             }
         }
 
+        /// <summary>
+        /// The CursorMoved event.
+        /// </summary>
+        /// This will be public opened after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public event EventHandler<CursorMovedEventArgs> CursorMoved
```
myroot, line 86: How about renamed to `CursorPositionChanged`? because TextEditor provide `CursorPosition` and it inform changes of `CursorPosition`
```suggestion
        public event EventHandler<CursorMovedEventArgs> CursorPositionChanged
```

src/Tizen.NUI/src/public/BaseComponents/TextFieldEvent.cs
```diff
@@ -268,6 +322,27 @@ public TextField TextField
             }
         }
 
+        /// <summary>
+        /// The CursorMoved event arguments.
+        /// </summary>
+        public class CursorMovedEventArgs : EventArgs
```
myroot, line 328: Do not use nested class

❌ DO NOT use public nested types as a logical grouping construct; use namespaces for this.
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/nested-types

src/Tizen.NUI/src/public/BaseComponents/TextEditorEvent.cs
myroot: Remove TextEditor property and extract to public class (not nested) and then share with `TextField` class

abdullehghujeh: Done

src/Tizen.NUI/src/public/BaseComponents/TextEvent.cs
```diff
@@ -47,4 +47,23 @@ public class InputFilteredEventArgs : EventArgs
...
+    /// The CursorPositionChanged event arguments.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class CursorPositionChangedEventArgs : EventArgs
+    {
+        /// <summary>
+        /// cursor postion before the change.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public uint OldCursorPosition { get; set; }
```
myroot, line 61: I can't found any event args that provide old value in NUI. for the consistency how about remove OldCursorPosition?

src/Tizen.NUI/src/public/BaseComponents/TextFieldEvent.cs
```diff
@@ -208,6 +246,21 @@ private void OnTextChanged(IntPtr textField)
...
+            if (textFieldCursorPositionChangedEventHandler != null)
+            {
+                CursorPositionChangedEventArgs e = new CursorPositionChangedEventArgs();
+
+                // Populate all members of "e" (CursorPositionChangedEventArgs) with real data
+                e.OldCursorPosition = oldPosition;
+                e.CursorPosition = newPosition;
+
+                //here we send all data to user event handlers
+                textFieldCursorPositionChangedEventHandler(this, e);
```
myroot, line 260:
```suggestion
                textFieldCursorPositionChangedEventHandler?.Invoke(this, e);
```

src/Tizen.NUI/src/public/BaseComponents/TextFieldEvent.cs
```diff
@@ -208,6 +246,20 @@ private void OnTextChanged(IntPtr textField)
...
         }
 
+        private void OnCursorPositionChanged(IntPtr textField, uint oldPosition)
+        {
+            if (textFieldCursorPositionChangedEventHandler != null)
+            {
+                CursorPositionChangedEventArgs e = new CursorPositionChangedEventArgs();
+
+                // Populate all members of "e" (CursorPositionChangedEventArgs) with real data
+                e.CursorPosition = oldPosition;
```
wonrst, line 256: The name is CursorPosition, but it is actually OldCursorPosition.
This can be confusing to users.
I think this should be "OldCursorPosition".

But you don't have to provide it now, As you know, now we dont't have time.
We have already implemented this in DALi, so you can add it later as needed.

abdullehghujeh: So you mean for now this will be without event args ? remove CursorPositionChangedEventArgs class?

wonrst: I think OldPosition can be implemented easily in the app.
However, we have always been criticized for poor usability.
So I think this may be provided.
What do you think? I think this can be decided by us.
If you remove CursorPositionChangedEventArgs , you should also remove SelectionChangedEventArgs, for consistency.

abdullehghujeh: I think it is better to keep old position as event args, it is more convenient for the user if he is interested in the old value.
for the naming, if I use OldCursorPosition doesn't this imply that there is a NewCursorPosition ?

wonrst:
```
I think it is better to keep old position as event args, it is more convenient for the user if he is interested in the old value.
```
Ok, I agree with you.

```
for the naming, if I use OldCursorPosition doesn't this imply that there is a NewCursorPosition ?
```
However, if there is only CursorPosition in the event args, it seems to cause confusion for users. Isn't it?

abdullehghujeh: yes it could, but reading the docs should clear things.

abdullehghujeh: anyway I have changed it to Old

wonrst (approved):

Seoyeon2Kim (approved):
Looks good to me.

wonrst (approved):

PR #3401, opened 2021-08-09
[NUI] Add CharacterSpacing property 

shrouqsabah:
### Description of Change ###
Add CharacterSpacing property to NUI
The spaces between characters.
A positive value will make the characters far apart (expanded) and a negative value will bring them closer (condensed).

Added:
 - public float CharacterSpacing { get; set; } // Property to  TextLabel, TextEditor & TextField

Related patches on DALi:
 - NUI] Add CharacterSpacing Property : https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/262388
 - Adding Character Spacing : https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/260890

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/TextField.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/TextField.cs
@@ -1929,6 +1929,27 @@ namespace Tizen.NUI.BaseComponents
         }
 
         /// <summary>
+        /// The spaces between characters in Pixels.
+        /// <remarks>
+        /// A positive value will make the characters far apart (expanded) and a negative value will bring them closer (condensed).<br />
+        /// The default value is 0.f which does nothing.
+        ///</remarks>
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public float CharacterSpacing
+        {
+            get
+            {
+                return (float)GetValue(CharacterSpacingProperty);
+            }
+            set
+            {
+                SetValue(CharacterSpacingProperty, value);
+                NotifyPropertyChanged();
+            }
+        }
+
+        /// <summary>
         /// Set InputFilter to TextField.
         /// </summary>
         /// <param name="inputFilter">The InputFilter</param>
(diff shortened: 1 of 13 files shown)


huiyueun (approved):

rabbitfor:
@shrouqsabah If there is a related dali-csharp-binder patch, could you add a link to it?
And Please add "[NUI]" prefix to the commit message.

wonrst:
@shrouqsabah Please resolve the conflict.

dongsug-song:
@shrouqsabah 
Hello, could you check this? if ACR patch, please show the ACR link. 
and please resolve conflict. thank you.

shrouqsabah:
> @shrouqsabah
> Hello, could you check this? if ACR patch, please show the ACR link.
> and please resolve conflict. thank you.

No, this is not an ACR. We  need to wait merge on DALi then merge this. 
@wonrst , we delayed this to be merged after release, right?

Regardless, I will solve conflicts.

wonrst:
For stability, we plan to merge this patch after the M2 release.

shrouqsabah:
Rebased and verified

src/Tizen.NUI/src/public/BaseComponents/TextField.cs
```diff
@@ -1754,6 +1754,24 @@ public EllipsisPosition EllipsisPosition
             }
         }
 
+        /// <summary>
+        /// The spaces between characters.<br />
+        /// A positive value will make the characters far apart (expanded) and a negative value will bring them closer (condensed).<br />
+        /// </summary>
```
myroot, line 1760: 값이 의미하는것이 무엇인지 설명을 추가해주세요
pixel일까요? %일까요?

shrouqsabah: Hi @myroot ,
Yes, it is pixel.
Please, could you check the updated description and let me know what you see.

wonrst (approved):
Looks good to me

everLEEst (approved):
It is okay that we go with pixel by default.
but font size can be different, or scalable,
so it would be more intuitive to use %.
is there any plan to support % unit after this?

shrouqsabah:
> It is okay that we go with pixel by default. but font size can be different, or scalable, so it would be more intuitive to use %. is there any plan to support % unit after this?

We did not think about this option? 
But I think we can add it for future plans.
@wonrst , what do you think ?

wonrst:
> > It is okay that we go with pixel by default. but font size can be different, or scalable, so it would be more intuitive to use %. is there any plan to support % unit after this?
> 
> We did not think about this option? But I think we can add it for future plans. @wonrst , what do you think ?

@shrouqsabah 
Yes, right.
We can add this to future plans.
For scalable UX, we should also consider MinLineSize.

PR #3402, opened 2021-08-10
[NUI] Add a comment about SENSITIVE.

bshsqa:
- If a View is insensitive, the children are also not hittable.
 - This patch adds that at the sensitive property comment.


hinohie (approved):
Looks god to me

hyunjushin (approved):

PR #3408, opened 2021-08-10
[NUI] Fix FlexLayout to measure its child correctly

jaehyun0cho:
FlexLayout calculated child view's size in OnLayout() without
considering child layout's measured size unlike other layouts'
OnLayout().

This caused that the grand child view's size was calculated incorrectly
if the child and grand child had MatchParent Width/HeightSpecification.

To resolve this issue, child layout's measured size is set with the
child view's size calculated by dali-toolkit's YOGA APIs.

### API Changes ###

 - ACR:


huiyueun (approved):

everLEEst (approved):
LGTM.

PR #3416, opened 2021-08-12
[NUI] Support public ime window

sparrow74:
### Description of Change ###
Support IME Keyboard application and window.

1. Unnecessary interfaces are removed.
2. Change method name.
3. Add test case.

### API Changes ###
 - ACR: https://code.sec.samsung.net/jira/browse/TCSACR-436


huiyueun (approved):
LGTM

PR #3417, opened 2021-08-12
[NUI] Fix View to catch Relayout by default

jaehyun0cho:
Previously, View's OnRelayout could be called only if app added Relayout
event callback.

Dali's actor size is calculated after NUI Layout layouts views.
In some special cases, NUI Layout is required to layout views again
based on the Dali's actor size calculation.

To resolve the special cases, layouting is required to be triggered in
View's OnRelayout.
To achieve the above, OnRelayout should be called regardless of app's
Relayout event callback.

### API Changes ###

 - ACR:


dongsug-song:
> LGTM but will wait till other patches

@everLEEst 
안녕하세요. 관련 다른 패치 상황을 확인해 주시고 문제 없으면 머지를 부탁드립니다. 감사합니다.

PR #3418, opened 2021-08-12
[NUI] Fix to support DimensionDependency in Layout

jaehyun0cho:
Previously, View with DimensionDependency could not be calculated
correctly by Layout.
Because the View's ResizePolicy is changed from DimensionDependency to
Fixed when Layout sets the View's Size.

Now, to support View with DimensionDependency, Layout does not change
width or height with DimensionDependency.
Instead, Layout calculates Views when a View is resized to use the size
calculated from DimensionDependency in Dali.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
@@ -830,6 +830,8 @@ namespace Tizen.NUI.BaseComponents
             {
                 onRelayoutEventHandler(this, null);
             }
+
+            Layout?.RequestLayout();
         }
 
         // Callback for View TouchSignal
--- a/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
@@ -584,12 +584,34 @@ namespace Tizen.NUI
                         // the size of the owner view must be reset even the ExcludeLayouting is true.
                         if (Owner.HeightSpecification < 0 || Owner.WidthSpecification < 0)
                         {
-                            Owner.SetSize(right - left, bottom - top);
+                            if (Owner.WidthResizePolicy == ResizePolicyType.DimensionDependency && Owner.HeightResizePolicy != ResizePolicyType.DimensionDependency)
+                            {
+                                Owner.SetSize(Owner.SizeWidth, bottom - top);
+                            }
+                            else if (Owner.WidthResizePolicy != ResizePolicyType.DimensionDependency && Owner.HeightResizePolicy == ResizePolicyType.DimensionDependency)
+                            {
+                                Owner.SetSize(right - left, Owner.SizeHeight);
+                            }
+                            else if (Owner.WidthResizePolicy != ResizePolicyType.DimensionDependency && Owner.HeightResizePolicy != ResizePolicyType.DimensionDependency)
+                            {
+                                Owner.SetSize(right - left, bottom - top);
+                            }
                         }
                     }
                     else
                     {
-                        Owner.SetSize(right - left, bottom - top);
+                        if (Owner.WidthResizePolicy == ResizePolicyType.DimensionDependency && Owner.HeightResizePolicy != ResizePolicyType.DimensionDependency)
+                        {
+                            Owner.SetSize(Owner.SizeWidth, bottom - top);
+                        }
+                        else if (Owner.WidthResizePolicy != ResizePolicyType.DimensionDependency && Owner.HeightResizePolicy == ResizePolicyType.DimensionDependency)
+                        {
+                            Owner.SetSize(right - left, Owner.SizeHeight);
+                        }
+                        else if (Owner.WidthResizePolicy != ResizePolicyType.DimensionDependency && Owner.HeightResizePolicy != ResizePolicyType.DimensionDependency)
+                        {
+                            Owner.SetSize(right - left, bottom - top);
+                        }
                         Owner.SetPosition(left, top);
                     }
                 }


src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
```diff
@@ -584,12 +584,34 @@ private bool SetFrame(float left, float top, float right, float bottom, bool ind
...
                         if (Owner.HeightSpecification < 0 || Owner.WidthSpecification < 0)
                         {
-                            Owner.SetSize(right - left, bottom - top);
+                            if (Owner.WidthResizePolicy == ResizePolicyType.DimensionDependency && Owner.HeightResizePolicy != ResizePolicyType.DimensionDependency)
+                            {
+                                Owner.SetSize(Owner.SizeWidth, bottom - top);
+                            }
+                            else if (Owner.WidthResizePolicy != ResizePolicyType.DimensionDependency && Owner.HeightResizePolicy == ResizePolicyType.DimensionDependency)
+                            {
+                                Owner.SetSize(right - left, Owner.SizeHeight);
```
bshsqa, line 593: This has same issue.
This SetSize() method may change height resize policy too.
It is still problematic.

src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
```diff
@@ -830,6 +830,8 @@ private void OnRelayout(IntPtr data)
             {
                 onRelayoutEventHandler(this, null);
             }
+
+            Layout?.RequestLayout();
```
bshsqa, line 834: NUI layout is works before this OnRelayout callback is called.
This code works for dimension dependency?

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
```diff
@@ -584,12 +584,34 @@ private bool SetFrame(float left, float top, float right, float bottom, bool ind
                         // the size of the owner view must be reset even the ExcludeLayouting is true.
                         if (Owner.HeightSpecification < 0 || Owner.WidthSpecification < 0)
                         {
-                            Owner.SetSize(right - left, bottom - top);
+                            if (Owner.WidthResizePolicy == ResizePolicyType.DimensionDependency && Owner.HeightResizePolicy != ResizePolicyType.DimensionDependency)
+                            {
+                                Owner.SetSize(Owner.SizeWidth, bottom - top);
```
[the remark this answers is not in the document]
bshsqa: And.......
In my knowledge, Size negotiation of DALi for demension dependency will works after this code.
This still look not work for the purpose.

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
```diff
@@ -584,12 +584,34 @@ private bool SetFrame(float left, float top, float right, float bottom, bool ind
                         // the size of the owner view must be reset even the ExcludeLayouting is true.
                         if (Owner.HeightSpecification < 0 || Owner.WidthSpecification < 0)
                         {
-                            Owner.SetSize(right - left, bottom - top);
+                            if (Owner.WidthResizePolicy == ResizePolicyType.DimensionDependency && Owner.HeightResizePolicy != ResizePolicyType.DimensionDependency)
+                            {
```
bshsqa, line 588: I tested now how this works.
If we set width or height to 0, it don't change resize policy of the dimension.
DALi depends to set size zero.

But, what I am wondering is.... when we set Layout View's SizeHeight again, then Owner.SizeWidth is not zero and changes resize policy to Fixed.
And, that time, the size of each entries not changed following newly set SizeHeight.
Is it right to determine entries' size only once?

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
bshsqa, line 593: This has same issue.

I think..., if we want to keep to use ResizePolicy to define View's size without deprecate, we need to consider more case like this.

src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
jaehyun0cho: This code is added to calculate NUI Layout again when ImageView's size is changed from Dali.
However, this code is too simple so it may cause redundant Layout calculation.
I will update the code.

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
jaehyun0cho: Thank you for your comment.
You are right..
If SizeHeight is set only once, it works fine.
However, if SizeHeight is changed, then the SizeWidth becomes Fixed..

I will find another way to fix this.

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
```diff
@@ -584,12 +584,34 @@ private bool SetFrame(float left, float top, float right, float bottom, bool ind
...
+                            else if (Owner.WidthResizePolicy != ResizePolicyType.DimensionDependency && Owner.HeightResizePolicy != ResizePolicyType.DimensionDependency)
+                            {
+                                Owner.SetSize(right - left, bottom - top);
+                            }
                         }
                     }
                     else
                     {
-                        Owner.SetSize(right - left, bottom - top);
+                        if (Owner.WidthResizePolicy == ResizePolicyType.DimensionDependency && Owner.HeightResizePolicy != ResizePolicyType.DimensionDependency)
```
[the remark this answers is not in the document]
jaehyun0cho: I will organize the code when I find a correct way to resolve the issue.

jaehyun0cho:
It is required to find another way instead of this PR.

PR #3437, opened 2021-08-19
[NUI] Fix Dispose(true) issue

myroot:
### Description of Change ###
 It fixes `Dispose(true)` issue

Current Implementation use `DisposeQueue` to release object on main thread

When `Dispose(false)` was called on GC thread, object was added in `DisposeQueue` 
and called on main thread with `Dispose()`

`Dispose()` it call `Dispose(disposing:true)` but current state is disposing false, 
disposing false means "all managed object are invalid so do not access to your field and property that are object types"

So, if API user extend a View class and override `void Dispose(bool disposing)` and access field or property when disposing was true but actually was not, It could be cause undefined behavior

this changes fixes it

Dispose() method was changed but code analyzer make warning to use original code. so this changes increasing compiler warning .

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/Disposable.cs
+++ b/src/Tizen.NUI/src/internal/Common/Disposable.cs
@@ -67,8 +67,15 @@ namespace Tizen.NUI
         /// <since_tizen> 6 </since_tizen>
         public void Dispose()
         {
-            Dispose(true);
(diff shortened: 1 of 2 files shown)


dongsug-song (approved):
looks good to me

PR #3444, opened 2021-08-21
[NUI] Make Unparent() in Dispose() be executed early

dongsug-song:
### Description of Change ###
[NUI] Make Unparent() in Dispose() be executed early
- Unparent() needs to be excuted in explicit dispose() which is also same as disposing = true.
- Also it should be called prior to before the derived class of View deletes the native handle, otherwise when doing Unparent(), the child itself has empty body (native object has been deleted already).

### API Changes ###
none


bshsqa (approved):

rabbitfor (approved):

Seoyeon2Kim (approved):

jaehyun0cho (approved):
LGTM

PR #3458, opened 2021-08-24
[NUI] Fix to measure GridLayout's children sizes correctly

jaehyun0cho:
Previously, if child's measured size is 0, then GridLayout does not
update the child's measured size when the child's size is set by using
GridLayout APIs.
e.g.
SetHorizontalStretch(child, GridLayout.StretchFlags.ExpandAndFill);

Moreover, if child has its own size and ExpandAndFill is set, then
child's size is set to be its own size (its measured size) plus its
expanded size.
e.g.
parent size is 100.
child1's measured size is 50 and ExpandAndFill.
child2's measured size is 0 and ExpandAndFill.
Then, child1's size is 50 + 25 = 75 and child2's size is 25.

To resolve the above, GridLayout updates child's measured size in
OnLayout() and child's size is set to be the bigger one between its
measured size and its expanded size.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Layouting/GridLocations.cs
+++ b/src/Tizen.NUI/src/internal/Layouting/GridLocations.cs
@@ -293,9 +293,20 @@ namespace Tizen.NUI
 
             for (int i = 0; i < edgeList.Length; i++)
             {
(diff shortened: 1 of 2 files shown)


dongsug-song (approved):
looks good to me

PR #3461, opened 2021-08-25
[MediaPlayer] Support Video and UI sync mode

hsgwon:
### Description of Change ###

Support Video and UI sync mode

### API Changes ###

 - TCSACR-447


src/Tizen.Multimedia/Common/Display.cs
```diff
@@ -137,13 +137,32 @@ public Display(Window window)
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Display"/> class with a <see cref="NUI.Window"/> class.
+        /// </summary>
+        /// <param name="window">A <see cref="NUI.Window"/> to display.</param>
+        /// <param name="uiSync">A value indicates whether video and UI is in sync or not.</param>
+        /// <remarks>
+        /// The <see cref="NUI.Window.BackgroundColor"/> must be <see cref="NUI.Color.Transparent"/>
+        /// for the <see cref="Display"/> to be rendered correctly.<br/>
+        /// UI sync is only for <see cref="Tizen.Multimedia.Player"/> and
```
pkosko, line 152: Please check if cref will be properly resolved in doc, regarding below warning.

src/Tizen.Multimedia/Common/Display.cs
```diff
@@ -137,13 +137,32 @@ public Display(Window window)
...
+        /// <see cref="Tizen.Multimedia.Player.DisplaySettings"/> will not work in UI sync mode.
```
pkosko, line 153: Please check if cref will be properly resolved in doc, regarding below warning.

wiertel, line 153: For simple \<see> to work the name should be visible to the compiler, you can add T: to fix it, or use the following snippet to have the full name of DisplaySettings property
```suggestion
        /// UI sync is only for <see cref="T:Tizen.Multimedia.Player"/> and
        /// [](xref:Tizen.Multimedia.Player.DisplaySettings?displayProperty=fullNam
e) will not work in UI sync mode.
```

PR #3466, opened 2021-08-25
[NUI] Implement more request interceptor APIs.

huayongxu:
### Description of Change ###

The related patch in dali-csharp-binder:
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/262954/

### API Changes ###

 - ACR:


src/Tizen.NUI/src/internal/WebView/WebContext.cs
```diff
@@ -33,9 +33,9 @@ public class WebContext : Disposable
         private float timeOffset;
         private ApplicationType applicationType;
         private WebSecurityOriginList securityOriginList;
-        private SecurityOriginListAcquiredCallback securityOriginsAcquiredCallback;
+        private SecurityOriginListAcquiredCallback securityOriginListAcquiredCallback;
         private readonly WebContextSecurityOriginListAcquiredProxyCallback securityOriginListAcquiredProxyCallback;
-        private WebPasswordDataList passwordList;
+        private WebPasswordDataList passwordDataList;
```
rabbitfor, line 38: Is passwordData related with Interceptor API?

rabbitfor: I think the patch is including code that are not related with request interceptor.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
```diff
@@ -46,17 +51,43 @@ public string Url
...
+            {
+                return Interop.WebHttpRequestInterceptor.GetMethod(interceptorHandle);
+            }
+        }
+
+        /// <summary>
+        /// Gets headers of intercepted http request, for example, "schema: https".
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public PropertyMap Headers
```
rabbitfor, line 74: The result data is a set of string key-value pair?

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
```diff
@@ -46,17 +51,43 @@ public string Url
...
+            get
+            {
+                IntPtr mapPtr = Interop.WebHttpRequestInterceptor.GetHeaders(interceptorHandle);
+                return new PropertyMap(mapPtr, true);
             }
         }
 
         /// <summary>
         /// Ignores this request.
+        /// This function can be called only INSIDE WebContext.HttpRequestIntercepted.
```
rabbitfor, line 85: https://docs.tizen.org/application/native/api/mobile/5.0/group__WEBVIEW.html#gae484b204d142e0d117326e5683194f39
> Handling a request is done after using either ewk_intercept_request_ignore, ewk_intercept_request_response_set, ewk_intercept_request_response_body_set, and in some cases after ewk_intercept_request_response_write_chunk.
Further EWK API calls on this Ewk_Intercept_Request instance result in undefined behavior.

The ewk documentation says after calling one of 4 APIs, using the instance would make undefined behavior.
I think we need to specify this restriction.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
```diff
@@ -82,20 +113,64 @@ public bool SetResponseStatus(int statusCode, string customStatusText)
...
+            bool result = Interop.WebHttpRequestInterceptor.AddResponseHeaders(interceptorHandle, PropertyMap.getCPtr(headers));
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
             return result;
         }
 
         /// <summary>
         /// Writes whole response body at once.
         /// <param name="body">Contents of response</param>
-        /// <param name="length">Length of Contents of response</param>
+        /// <param name="length">Length of contents of response</param>
```
rabbitfor, line 136: Length is required? C# `String` provide `Length` property.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
```diff
@@ -82,20 +113,64 @@ public bool SetResponseStatus(int statusCode, string customStatusText)
...
+        }
+
+        /// <summary>
+        /// Writes whole response body at once.
+        /// <param name="headers">Headers of response</param>
+        /// <param name="body">Contents of response</param>
+        /// <param name="length">Length of contents of response</param>
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool AddResponse(string headers, string body, uint length)
```
rabbitfor, line 153: Ditto. I think `Length` is not required.

rabbitfor: As I commented above, the API `ewk_intercept_request_response_set` and `ewk_intercept_request_response_body_set` are the finisher.
So I think using original word `Set` is more suitable than `Add` here.

* AddResponseBody -> SetResponseBody
* AddResponse -> SetResponse

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
```diff
@@ -82,20 +113,64 @@ public bool SetResponseStatus(int statusCode, string customStatusText)
...
+        /// If this function returns false, handling the request is done.
+        /// Any further calls result in undefined behavior.
+        /// User should always check return value, because response to this request might not be needed any more,
+        /// and function can return false even though user still has data to write.
+        /// This function can be called only OUTSIDE WebContext.HttpRequestIntercepted.
+        /// <param name="chunk">Chunk of response</param>
+        /// <param name="length">Length of chunk of response</param>
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool WriteResponseChunk(string chunk, uint length)
```
rabbitfor, line 171: Ditto. `Length`

src/Tizen.NUI/src/internal/WebView/WebContext.cs
huayongxu: Done.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
huayongxu: Yes. I updated comments.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
huayongxu: Done.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
huayongxu: Done.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
huayongxu: Done.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
huayongxu: Done.

zhouleonlei (approved):
LGTM

wb0716 (approved):
Good for me

AchoWang (approved):
Looks good to me.

AdunFang (approved):
It is OK.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
myroot: How about use Dictionary<string,string> instead PropertyMap
ProperyMap is actually not map data structure.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
```diff
@@ -46,56 +51,143 @@ public string Url
...
+            return result;
+        }
+
+        /// <summary>
+        /// Adds HTTP headers to response for intercepted request.
+        /// This function can be used inside or outside WebContext.HttpRequestIntercepted.
+        /// <param name="headers">Map of response headers</param>
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool AddResponseHeaders(PropertyMap headers)
```
myroot, line 134: How about use IDictionary<string,string>?
because, Headers will be comes from WebHttpRequestInterceptor.Headers API and some values need to update to use on `AddResponseHeaders` but PropertyMap does not support update value
```suggestion
        public bool AddResponseHeaders(IDictionary<string,string> header)
```

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
rabbitfor: I agree with him. I think Dictionary is more intuitive in this case.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
myroot: If need to add a special cookie value 

Dictionary version
``` c#
HttpRequestIntercepted += (s, e) => {
    var header = e.HttpRequestInterceptor.Header;
    if (header.ContainKey("Cookie")) {
          header["Cookie"] += ";mycookie=myvalue";
    } else {
          header["Cookie"] += "mycookie=myvalue";
    }
    AddResponseHeaders(header);
};
```

PropertyMap version
``` c#
HttpRequestIntercepted += (s, e) => {
    var header = e.HttpRequestInterceptor.Header;
    PropertyMap newHeader = new PropertyMap();
    bool hasCookie = false;
    for (int i =0; i < header.Count(); i++) {
         var key = header.GetKeyAt(i);
         if (key.StringKey == "Cookie") {
              hasCookie = true;
              header.GetValue(i).Get(out string cookieValues);
              cookieValues += ";mycookie=myvalue";
         } else {
            newHeader.Insert(key.StringKey, header.GetValue(i));
         }
    }
    if (!hasCookie) {
         newHeader.Insert("Cookie", new PropertyValue("mycookie=myvalue"));
    }
    AddResponseHeaders(newHeader);
};
```

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
```diff
@@ -46,56 +51,143 @@ public string Url
...
+        /// To call it, application should have full response headers and body ready for the intercepted request.
+        /// This function can be used inside or outside WebContext.HttpRequestIntercepted.
+        /// After this call, any further call on WebHttpRequestInterceptor results in undefined behavior.
+        /// <param name="headers">Headers of response</param>
         /// <param name="body">Contents of response</param>
-        /// <param name="length">Length of Contents of response</param>
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public bool AddResponseBody(string body, uint length)
+        public bool SetResponse(string headers, string body)
```
myroot, line 167: I wonder different of between headers on this method and `public bool AddResponseHeaders(PropertyMap headers)`

if they was semantically same, why not use same type?

myroot: As I checked, this headers maybe including response status

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
huayongxu: Done.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
huayongxu: Done.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
huayongxu: Yes. headers of http response need include a line like "HTTP/1.1 200 OK".

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
rabbitfor: If you don't mind, could you change `Dictionary` to `IDictionary` as he originally suggested?

src/Tizen.NUI/src/internal/WebView/WebContext.cs
```diff
@@ -115,6 +115,39 @@ protected override void Dispose(DisposeTypes type)
...
+
+        private EventHandler<WebContextHttpRequestInterceptedEventArgs> httpRequestInterceptedEventHandler;
+        private WebContextHttpRequestInterceptedProxyCallback httpRequestInterceptedCallback;
+
+        /// <summary>
+        /// Event for the HttpRequestIntercepted signal which can be used to subscribe or unsubscribe the event handler.<br />
+        /// This signal is emitted when http request would be intercepted.<br />
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public event EventHandler<WebContextHttpRequestInterceptedEventArgs> HttpRequestIntercepted
```
myroot, line 129: I think, Delegate setter is more proper, because http request interceptor is only allow one handler, if multiple event handler was react with this event it could be wrong

How about think it?

```suggestion
        public SetHttpRequestInterceptorDelegate(Action<WebHttpRequestInterceptor> action) {
                    httpRequestInterceptedCallback = (IntPtr interceptor) => {
                        action(WebHttpRequestInterceptor(interceptor, true));
                    }
                    IntPtr ip = Marshal.GetFunctionPointerForDelegate(httpRequestInterceptedCallback);
                    Interop.WebContext.RegisterRequestInterceptedCallback(SwigCPtr, new HandleRef(this, ip));
        }
```

myroot: And I think you need to keep a consistency with Tizen.WebView
https://github.com/Samsung/TizenFX/blob/8bee3b086da6909d53361da7b0e2b1610d0fbdd9/src/Tizen.WebView/Tizen.WebView/Context.cs#L157

myroot (commented):
Could you check Tizen.WebView API?

I think we need to provide a consistency with Tizen.WebView

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
huayongxu: Done.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
```diff
@@ -46,56 +52,172 @@ public string Url
...
         }
 
         /// <summary>
-        /// Ignores this request.
+        /// Gets headers of intercepted http request.
+        /// Headers is a map with string key-value pairs,
+        /// for example, "Accept: text/plain", "Accept-Charset: utf-8", etc.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Dictionary<string, string> Headers
```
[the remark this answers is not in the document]
huayongxu: Done.

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
```diff
@@ -46,56 +52,172 @@ public string Url
...
+                    PropertyKey key = map.GetKeyAt(i);
+                    if (key.Type == PropertyKey.KeyType.String)
+                    {
+                        string outValue;
+                        PropertyValue mapValue = map.GetValue(i);
+                        if (mapValue.Get(out outValue))
+                        {
+                            dictionary.Add(key.StringKey, outValue);
+                        }
+                        mapValue.Dispose();
```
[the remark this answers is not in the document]
huayongxu: Done.

src/Tizen.NUI/src/internal/WebView/WebContext.cs
huayongxu: I replaced event handler with delegate, but I followed naming rule of other APIs of NUI WebView, and renamed it to RegisterHttpRequestedCallback.
What do you think?

myroot: Ok, It also good

src/Tizen.NUI/src/internal/WebView/WebContext.cs
```diff
@@ -520,6 +538,23 @@ public void RegisterMimeOverriddenCallback(MimeOverriddenCallback callback)
...
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public void RegisterHttpRequestInterceptedCallback(HttpRequestInterceptedCallback callback)
+        {
+            httpRequestInterceptedCallback = callback;
+            IntPtr ip = IntPtr.Zero;
+            if (httpRequestInterceptedCallback != null)
+            {
+                ip = Marshal.GetFunctionPointerForDelegate(httpRequestInterceptedProxyCallback);
+            }
+            Interop.WebContext.RegisterRequestInterceptedCallback(SwigCPtr, new HandleRef(this, ip));
```
myroot, line 554: You should store delegate object on somewhere to prevent release by GC
HandleRef could not store object

myroot: >You must manually keep the delegate from being collected by the garbage collector from managed code. The garbage collector does not track references to unmanaged code.

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getfunctionpointerfordelegate?view=net-5.0

>The HandleRef constructor takes two parameters: an Object representing the wrapper, and an IntPtr representing the unmanaged handle. The interop marshaler passes only the handle to unmanaged code, and guarantees that the wrapper (passed as the first parameter to the constructor of the HandleRef) remains alive for the duration of the call.

**remains alive for the duration of the call.**

But you need to hold until WebContext object was alive or callback was set to null

myroot: Please test with GC
``` c#
GC.Collect();
GC.WaitForPendingFinalizers();
```

myroot: Oh, my mistake, I did't check `httpRequestInterceptedCallback`

src/Tizen.NUI/src/internal/WebView/WebContext.cs
```diff
@@ -690,5 +725,15 @@ private void OnPasswordDataListAcquired(IntPtr alist)
...
             passwordDataListAcquiredCallback?.Invoke(passwordDataList);
         }
+
+        private void OnHttpRequestIntercepted(IntPtr interceptor)
+        {
+            if (httpRequestInterceptor != null)
+            {
+                httpRequestInterceptor.Dispose();
+            }
+            httpRequestInterceptor = new WebHttpRequestInterceptor(interceptor, true);
```
myroot, line 735: Why do you store into instance variable? 
Why not local variable?
```suggestion
            var httpRequestInterceptor = new WebHttpRequestInterceptor(interceptor, true);
```

src/Tizen.NUI/src/internal/WebView/WebContext.cs
```diff
@@ -690,5 +725,15 @@ private void OnPasswordDataListAcquired(IntPtr alist)
...
+            httpRequestInterceptedCallback?.Invoke(httpRequestInterceptor);
```
myroot, line 736: `public delegate bool HttpRequestInterceptedCallback(WebHttpRequestInterceptor interceptor);`

callback returning bool value but you do not use it

src/Tizen.NUI/src/internal/WebView/WebContext.cs
huayongxu: If a local variable is used, a warning CA2000 occurs:

Warning	CA2000	Call System.IDisposable.Dispose on object created by 'new WebHttpRequestInterceptor(interceptor, true)' before all references to it are out of scope.

To fix this warning, a instance variable is stored.

src/Tizen.NUI/src/internal/WebView/WebContext.cs
huayongxu: Done. I changed it from bool to void.
Thanks.

src/Tizen.NUI/src/internal/WebView/WebContext.cs
myroot: I think, interceptor object could be disposed when callback was completed
```suggestion
            using (var httpRequestInterceptor = new WebHttpRequestInterceptor(interceptor, true)) {
                      httpRequestInterceptedCallback?.Invoke(httpRequestInterceptor);
            }
```

src/Tizen.NUI/src/internal/WebView/WebContextHttpRequestInterceptedEventArgs.cs
```diff
@@ -21,18 +21,18 @@
 namespace Tizen.NUI
 {
     /// <summary>
-    /// Event arguments that passed via the WebView.HttpRequestIntercepted.
+    /// Event arguments that passed via the WebContext.HttpRequestIntercepted.
     /// </summary>
     [EditorBrowsable(EditorBrowsableState.Never)]
-    public class WebViewHttpRequestInterceptedEventArgs : EventArgs
+    public class WebContextHttpRequestInterceptedEventArgs : EventArgs
```
myroot, line 27: No more need this calss

test/Tizen.NUI.WebViewTest/SimpleWebViewApp.cs
```diff
@@ -725,8 +726,16 @@ private bool OnWebViewKeyEvent(object source, View.KeyEventArgs args)
                 //}
                 else if (args.Key.KeyPressedName == "XF86Red")
                 {
-                    //simpleWebView.HttpRequestIntercepted += OnHttpRequestIntercepted;
-                    //simpleWebView.HttpRequestIntercepted -= OnHttpRequestIntercepted;
+                    if (redKeyPressedCount % 2 == 0)
+                    {
+                        simpleWebView.Context.HttpRequestIntercepted += OnHttpRequestIntercepted;
```
myroot, line 731: need to update sample code

src/Tizen.NUI/src/internal/WebView/WebContextHttpRequestInterceptedEventArgs.cs
huayongxu: Done.

test/Tizen.NUI.WebViewTest/SimpleWebViewApp.cs
huayongxu: Done.

src/Tizen.NUI/src/internal/WebView/WebContext.cs
huayongxu: But WebHttpRequestInterceptor.WriteResponseChunk must be called outside WebContext.HttpRequestInterceptedCallback.

myroot: Where is come from these constraint?

It is chromium-efl API description on ewk_intercept_request_response_write_chunk
I can't find this constraint
```
/**
 * @brief Writes a part of response body.
 *
 * @details This function can be used to write response body in chunks.
 *          Application doesn't have to prepare full response body before
 *          calling this function, it can be used as soon as a part of response
 *          is ready.
 *
 * @remarks If this function returns EINA_FALSE, handling the request is done.
 *          Any further calls on the Ewk_Intercept_Request instance result in
 *          undefined behavior. User should always check return value, because
 *          response to this request might not be needed anymore, and function
 *          can return EINA_FALSE even though user still has data to write.
 *
 *          It is only allowed to use this function *outside* of the
 *          #Ewk_Context_Intercept_Request_Callback.
 *
 *          If a part of response body has been written with this function, you
 *          can't use the following functions anymore:
 *          - ewk_intercept_request_response_set
 *          - ewk_intercept_request_response_body_set
 *
 *          Using this function with too big chunks of data defeats its
 *          purpose. Engine uses chunks of response written with this function
 *          without waiting for end of data, which is its main advantage over
 *          @a ewk_intercept_request_response_body_set.
 *
 * @since_tizen 3.0
 *
 * @param[in] intercept_request Intercept Request instance received from
 *            #Ewk_Context_Intercept_Request_Callback ewk_context callback
 * @param[in] chunk Part of response body for intercepted request
 * @param[in] length Length of response body part
 *
 * @return @c EINA_TRUE on success or EINA_FALSE on failure
 *
 * @pre Before writing response body, you should set response status
 *      and headers using the following functions:
 *      - ewk_intercept_request_response_status_set()
 *      - ewk_intercept_request_response_header_add()
 *      - ewk_intercept_request_response_header_map_add()
 *
 * @post After writing full response body in chunks using this function, call
 *       it again with NULL as @a chunk and 0 as @a length, to signal that
 *       response body is finished.
 *
 *       Use the same arguments if there is a failure with preparing the rest
 *       of response body on application side, and you don't want or can't
 *       continue writing parts.
 */
EXPORT_API Eina_Bool ewk_intercept_request_response_write_chunk(
    Ewk_Intercept_Request* intercept_request, const char* chunk,
    size_t length);

```

huayongxu: In the comments above:
 *          It is only allowed to use this function *outside* of the
 *          #Ewk_Context_Intercept_Request_Callback.

myroot: Oh, Thank you, I find it

src/Tizen.NUI/src/internal/WebView/WebHttpRequestInterceptor.cs
```diff
@@ -46,56 +52,174 @@ public string Url
...
+        /// and the function can return false even though user still has data to write.
+        /// <param name="chunk">Chunk of response</param>
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool WriteResponseChunk(string chunk)
         {
-            bool result = Interop.WebHttpRequestInterceptor.AddResponseBody(SwigCPtr, body, length);
+            if (chunk == null)
+                return false;
+            bool result = Interop.WebHttpRequestInterceptor.WriteResponseChunk(interceptorHandle, chunk, (uint)chunk.Length);
```
myroot, line 222: We could be set null as chunk, it means completes of response

```
 * @post After writing full response body in chunks using this function, call
 *       it again with NULL as @a chunk and 0 as @a length, to signal that
 *       response body is finished.
```

huayongxu: Done.

rabbitfor (approved):
Checked all suggestions are applied

PR #3469, opened 2021-08-26
[NUI] Add RelayoutRequest Binding on View

everLEEst:
To update and process layout on Relayout(),
Add RelayoutRequest Binding on Interop.View and call it View.OnRelayout().

### API Changes ###

 - ACR:


rabbitfor (commented):
There is a `RelayoutRequest` in CustomView. If both are same one, I think the name of APIs should be the same too.

everLEEst:
> There is a `RelayoutRequest` in CustomView. If both are same one, I think the name of APIs should be the same too.

Yes. This patch is orginally raise up the RelayoutRequest from ViewWrapper to View level. but we cannot remove the API which is already added... so add new one here.. currently it is experimental code for testing issue resolving yet, and if DALi side approve all the changes, I'll make update full patch pack.

everLEEst:
NO MERGE,
this patch is related with dali changes, and as dali-developer's opinion, I'll hold this patch till we can find good solution.

PR #3481, opened 2021-08-31
[NUI] Fix to follow coding rule

huiyueun:
- Follow C# Coding rule
 1) reposition the type of class
 2) delete unusued using state
 3) follow event naming rule

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/ProcessorController.cs
+++ b/src/Tizen.NUI/src/internal/Common/ProcessorController.cs
@@ -15,10 +15,7 @@
  *
  */
 
-using Tizen.NUI.BaseComponents;
 using System.Runtime.InteropServices;
-using System.Collections.Generic;
-using System.Diagnostics;
 using System;
 using System.ComponentModel;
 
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/internal/Common/ProcessorController.cs
```diff
@@ -36,41 +33,39 @@ namespace Tizen.NUI
...
-                if(instance == null)
+                if (instance == null)
                 {
                     instance = new ProcessorController();
                 }
                 return instance;
             }
         }
 
+        private ProcessorController() : this(Interop.ProcessorController.New(), true)
```
jaehyun0cho, line 50: It seems that Properties are written after Constructors based on NUI coding rule.

jaehyun0cho: image2020-10-28_14-24-37

jaehyun0cho:
If you don't mind, could you separate this patch into 2 patches~?
e.g. Layout performance fix patch and C# coding rule patch

src/Tizen.NUI/src/internal/Common/ProcessorController.cs
huiyueun: done

huiyueun:
> If you don't mind, could you separate this patch into 2 patches~?
> e.g. Layout performance issue fix patch and C# coding rule patch
> 
> If the above modifications are in one patch, then it might be difficult to catch which modification fixes the Layout performance issue.

OK. I'v modified this patch without fixing bug. after merging this patch, I will upload to fix issue.

jaehyun0cho (approved):
LGTM

PR #3483, opened 2021-08-31
Fix unhandled exception while calling CoRegister

jay-h-park:
Signed-off-by: Jaehong Park jay.h.park@samsung.com

### Description of Change ###
[Nlp] Fix crash while calling CoRegister method. 
Inside of CoRegister method, Exception can be thrown and nowhere can catch & cope with the exception.
https://code.sec.samsung.net/jira/browse/TSDF-3168

### API Changes ###
None


syeonholic (approved):
It looks good to me.

imfine98 (approved):
LGTM

PR #3484, opened 2021-09-01
[NUI] Fix Layout Performance Issue

huiyueun:
Layout Process should be registered only once.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #3494, opened 2021-09-02
[NUI] Fix testhub NUI.Components coredump issue

dongsug-song:
### Description of Change ###
[NUI] Fix testhub NUI.Components coredump issue
- there was no fail but coredumps are remained after finishing NUI.Components test.
- native crash occured when empty native object handle is accessed and the TouchEvent signal is tried to be gotten.
- https://code.sec.samsung.net/jira/browse/TSDF-2266

### API Changes ###
none


Seoyeon2Kim (approved):
Looks good to me! To fix crash issue :)

PR #3525, opened 2021-09-09
[NUI] Fix svace issue of dereferencing possible null handle in TextUtils

jaehyun0cho:
Since PropertyMap.Find can return null, Find().Get() should be replaced with
Find()?.Get().

### API Changes ###

 - ACR:


rabbitfor (approved):
👍

JoogabYun (approved):

PR #3537, opened 2021-09-10
[NUI] Add AdjustViewSize to ImageView

jaehyun0cho:
Up to now, although the position and size of image resource of ImageView
is set by FittingMode, but there is no way to set the size of ImageView
to respect the aspect ratio of its image resource.

If AdjustViewSize is true, then ImageView respects the aspect ratio of
its image resource when it determines its size.

AdjustViewSize works only if the ImageView is added to a View having Layout.
Otherwise, use can set ImageView's size to respect the aspect ratio of its image
resource by using ResizePolicyType.DimensionDependency.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
@@ -336,6 +336,14 @@ namespace Tizen.NUI.BaseComponents
             }
         }
 
+        /// <summary>
+        /// Create internal layout of ImageView
+        /// </summary>
+        internal static LayoutItem CreateImageLayout()
+        {
+            return new ImageLayout();
+        }
+
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
         private delegate void ResourceReadyEventCallbackType(IntPtr data);
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
@@ -1085,6 +1093,14 @@ namespace Tizen.NUI.BaseComponents
             }
         }
 
+        /// <summary>
+        /// Gets or sets the mode to adjust view size to preserve the aspect ratio of the image resource.
+        /// If this is set to be true, then the width or height, which is not set by user explicitly, can be adjusted to preserve the aspect ratio of the image resource.
+        /// AdjustViewSize works only if ImageView is added to a View having Layout.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool AdjustViewSize { get; set; } = false;
+
         internal Selector<string> ResourceUrlSelector
         {
             get => GetSelector<string>(resourceUrlSelector, ImageView.ResourceUrlProperty);
@@ -1465,6 +1481,101 @@ namespace Tizen.NUI.BaseComponents
         private void OnPixelAreaChanged(float x, float y, float z, float w)
         {
             PixelArea = new RelativeVector4(x, y, z, w);
+        }
+
+        private class ImageLayout : LayoutItem
+        {
+            /// <summary>
+            /// Gets or sets the mode to adjust view size to preserve the aspect ratio of the image resource.
+            /// If this is set to be true, then the width or height, which is not set by user explicitly, can be adjusted to preserve the aspect ratio of the image resource.
+            /// </summary>
+            [EditorBrowsable(EditorBrowsableState.Never)]
+            public bool AdjustViewSize
+            {
+                get
+                {
+                    return (Owner as ImageView)?.AdjustViewSize ?? false;
+                }
+                set
+                {
+                    if (Owner is ImageView imageView)
+                    {
+                        imageView.AdjustViewSize = value;
+                    }
+                }
+            }
+
+            /// <inheritdoc/>
+            [EditorBrowsable(EditorBrowsableState.Never)]
+            protected override void OnMeasure(MeasureSpecification widthMeasureSpec, MeasureSpecification heightMeasureSpec)
+            {
+                // To not change the view size by DALi
+                Owner.WidthResizePolicy = ResizePolicyType.Fixed;
+                Owner.HeightResizePolicy = ResizePolicyType.Fixed;
+
+                float specWidth = widthMeasureSpec.Size.AsDecimal();
+                float specHeight = heightMeasureSpec.Size.AsDecimal();
+                float naturalWidth = Owner.NaturalSize.Width;
+                float naturalHeight = Owner.NaturalSize.Height;
+                float minWidth = Owner.MinimumSize.Width;
+                float maxWidth = Owner.MaximumSize.Width;
+                float minHeight = Owner.MinimumSize.Height;
+                float maxHeight = Owner.MaximumSize.Height;
+                float aspectRatio = (naturalWidth > 0) ? (naturalHeight / naturalWidth) : 0;
+
+                // Assume that the new width and height are given from the view's suggested size by default.
+                float newWidth = Math.Min(Math.Max(naturalWidth, minWidth), (maxWidth < 0 ? Int32.MaxValue : maxWidth));
+                float newHeight = Math.Min(Math.Max(naturalHeight, minHeight), (maxHeight < 0 ? Int32.MaxValue : maxHeight));
+
+                // The width and height measure specs are going to be used to set measured size.
+                // Mark that the measure specs are changed by default to update measure specs later.
+                bool widthSpecChanged = true;
+                bool heightSpecChanged = true;
+
+                if (widthMeasureSpec.Mode == MeasureSpecification.ModeType.Exactly)
+                {
+                    newWidth = specWidth;
+                    widthSpecChanged = false;
+
+                    if (heightMeasureSpec.Mode != MeasureSpecification.ModeType.Exactly)
+                    {
+                        if ((AdjustViewSize) && (aspectRatio > 0))
+                        {
+                            newHeight = newWidth * aspectRatio;
+                        }
+                    }
+                }
+
+                if (heightMeasureSpec.Mode == MeasureSpecification.ModeType.Exactly)
+                {
+                    newHeight = specHeight;
+                    heightSpecChanged = false;
+
+                    if (widthMeasureSpec.Mode != MeasureSpecification.ModeType.Exactly)
+                    {
+                        if ((AdjustViewSize) && (aspectRatio > 0))
+                        {
+                            newWidth = newHeight / aspectRatio;
+                        }
+                    }
+                }
+
+                if (widthSpecChanged)
+                {
+                    widthMeasureSpec = new MeasureSpecification(new LayoutLength(newWidth), MeasureSpecification.ModeType.Exactly);
+                }
+
+                if (heightSpecChanged)
+                {
+                    heightMeasureSpec = new MeasureSpecification(new LayoutLength(newHeight), MeasureSpecification.ModeType.Exactly);
+                }
+
+                MeasuredSize.StateType childWidthState = MeasuredSize.StateType.MeasuredSizeOK;
+                MeasuredSize.StateType childHeightState = MeasuredSize.StateType.MeasuredSizeOK;
+
+                SetMeasuredDimensions(ResolveSizeAndState(new LayoutLength(newWidth), widthMeasureSpec, childWidthState),
+                                      ResolveSizeAndState(new LayoutLength(newHeight), heightMeasureSpec, childHeightState));
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1466,5 +1497,108 @@ private void OnPixelAreaChanged(float x, float y, float z, float w)
         {
             PixelArea = new RelativeVector4(x, y, z, w);
         }
+
+        private class ImageLayout : LayoutItem
+        {
+            private bool adjustViewSize = false;
```
everLEEst, line 1503: it is not used.

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1466,5 +1497,108 @@ private void OnPixelAreaChanged(float x, float y, float z, float w)
...
+                bool heightSpecChanged = true;
+
+                if (widthMeasureSpec.Mode == MeasureSpecification.ModeType.Exactly)
+                {
+                    newWidth = specWidth;
+                    widthSpecChanged = false;
+
+                    if (heightMeasureSpec.Mode != MeasureSpecification.ModeType.Exactly)
+                    {
+                        if ((AdjustViewSize) && (aspectRatio > 0))
```
everLEEst, line 1559: I'm usually prefer to use private member value in here, but this value is get it from ImageView, so let it get synchronously every calls. its okay.

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1085,6 +1095,27 @@ public WrapModeType WrapModeV
             }
         }
 
+        /// <summary>
+        /// Gets or sets the mode to adjust view size to preserve the aspect ratio of the image resource.
+        /// If this is set to be true, then the width or height, which is not set by user explicitly, can be adjusted to preserve the aspect ratio of the image resource.
+        /// AdjustViewSize works only if ImageView is added to a View having Layout.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool AdjustViewSize
```
myroot, line 1104: Bindable property로 하는것이 아니면, private변수는 삭제하고 auto generate된 property구현을 사용해도 될것 같습니다.
```suggestion
        public bool AdjustViewSize {get; set;}
```

everLEEst, line 1104: We usually puts IsXXX in boolean data, but I'm not sure IsAdjustedViewSize is correct or not...........

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
myroot, line 1503: 사용되지 않은 변수
```suggestion
```

everLEEst (changes requested):
Please remove unnecessary private member.

src/Tizen.NUI/src/public/Layouting/LayoutGroup.cs
```diff
@@ -188,7 +188,18 @@ private void AddChildToLayoutGroup(View child)
...
+                        child.Layout = (child as TextLabel)?.CreateTextLayout();
+                    }
+                    else if (child is ImageView)
+                    {
+                        child.Layout = (child as ImageView)?.CreateImageLayout();
+                    }
+                    else
+                    {
+                        child.Layout = new AbsoluteLayout();
+                    }
```
myroot, line 202:
```suggestion
                    if (child is TextLabel textLabel)
                    {
                        child.Layout = textLabel.CreateTextLayout();
                    }
                    else if (child is ImageView imageView)
                    {
                        child.Layout = imageView.CreateImageLayout();
                    }
                    else
                    {
                        child.Layout = new AbsoluteLayout();
                    }
```

src/Tizen.NUI/src/public/Layouting/LayoutGroup.cs
```diff
@@ -494,6 +505,10 @@ protected override void OnAttachedToOwner()
                 {
                     view.Layout = (view as TextLabel)?.CreateTextLayout();
                 }
+                else if (view is ImageView)
+                {
+                    view.Layout = (view as ImageView)?.CreateImageLayout();
```
myroot, line 510: 이렇게 되면 AddChildToLayoutGroup에서 다시 한번 ParentLayout에 Add되어서 중복으로 들어가게 됩니다

이 이슈를 참고해서 수정해주세요
https://github.sec.samsung.net/dotnet/NUIBackend/issues/81

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1466,5 +1497,108 @@ private void OnPixelAreaChanged(float x, float y, float z, float w)
...
+            /// <summary>
+            /// Gets or sets the mode to adjust view size to preserve the aspect ratio of the image resource.
+            /// If this is set to be true, then the width or height, which is not set by user explicitly, can be adjusted to preserve the aspect ratio of the image resource.
+            /// </summary>
+            [EditorBrowsable(EditorBrowsableState.Never)]
+            public bool AdjustViewSize
+            {
+                get
+                {
+                    return (Owner is ImageView imageView) ? imageView.AdjustViewSize : false;
```
myroot, line 1514:
```suggestion
                    return (Owner as ImageView)?.AdjustViewSize ?? false;
```

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1466,5 +1497,108 @@ private void OnPixelAreaChanged(float x, float y, float z, float w)
...
+                        {
+                            newWidth = newHeight / aspectRatio;
+                        }
+                    }
+                }
+
+                // Set the new width and height to the view.
+                if ((Owner.SizeWidth != newWidth) || (Owner.SizeHeight != newHeight))
+                {
+                    Owner.SetSize(newWidth, newHeight);
```
myroot, line 1583: NUI의 레이아웃팅을 잘 모르지만, Measure단계에서 size을 실제 View에 반영해도 되나요?
Measure는 언제든지 measure spec이 변경하면서 여러번 호출되는것이 타 플랫폼에서의 동작인데 그럴때마다 실제 size업데이트를 하지는 않아서..

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1466,5 +1497,108 @@ private void OnPixelAreaChanged(float x, float y, float z, float w)
...
+                if (heightMeasureSpec.Mode == MeasureSpecification.ModeType.Exactly)
+                {
+                    newHeight = specHeight;
+                    heightSpecChanged = false;
+
+                    if (widthMeasureSpec.Mode != MeasureSpecification.ModeType.Exactly)
+                    {
+                        if ((AdjustViewSize) && (aspectRatio > 0))
+                        {
+                            newWidth = newHeight / aspectRatio;
```
myroot, line 1575: 지금은 width와 height가 독립적으로 각각 동작하면서 서로 변경한 newHeight와 newWidth를 변경하고 있습니다. 그래서 결과적으로는 height가 기준이 되는 aspect가 반영되는데, 
Measure spec안에서 가능한 크게 표시하는것이 목표라면 aspect에 따라 기준이 되는 길이가 정해져야 할것 같습니다.

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -336,6 +338,14 @@ internal ImageView(global::System.IntPtr cPtr, bool cMemoryOwn, bool shown = tru
             }
         }
 
+        /// <summary>
+        /// Create internal layout of ImageView
+        /// </summary>
+        internal LayoutItem CreateImageLayout()
```
myroot, line 344:
```suggestion
        internal static LayoutItem CreateImageLayout()
```

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
jaehyun0cho: Thank you for your comment.
I have removed it.

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
jaehyun0cho: I understand your point.
I think that "Adjust" is suitable for this case.
Based on the above, we can use IsAdjustingViewSize, but IMHO, AdjustViewSize looks better.

jaehyun0cho: 리뷰 감사합니다.
코멘트주신대로 수정하였습니다.

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
jaehyun0cho: 리뷰 감사합니다.
해당 변수 삭제하였습니다.

src/Tizen.NUI/src/public/Layouting/LayoutGroup.cs
jaehyun0cho: 코멘트주신대로 수정하였습니다.
감사합니다.

src/Tizen.NUI/src/public/Layouting/LayoutGroup.cs
jaehyun0cho: 리뷰 감사드립니다!
중복 설정하는 본 코드는 삭제하였습니다.
TextLabel 관련 중복 코드는 다른 PR로 삭제하도록 하겠습니다.

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
jaehyun0cho: 코멘트주신대로 수정하였습니다.
감사합니다.

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
jaehyun0cho: 지적해주신 내용이 맞습니다!
OnLayout()에서 View의 Size를 설정하도록 되어있습니다.
위 내용 삭제하였습니다.
리뷰 감사합니다!

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
jaehyun0cho: 코멘트주신대로 수정하였습니다.
감사합니다.

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1085,6 +1093,14 @@ public WrapModeType WrapModeV
...
         }
 
+        /// <summary>
+        /// Gets or sets the mode to adjust view size to preserve the aspect ratio of the image resource.
+        /// If this is set to be true, then the width or height, which is not set by user explicitly, can be adjusted to preserve the aspect ratio of the image resource.
+        /// AdjustViewSize works only if ImageView is added to a View having Layout.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool AdjustViewSize { get; set; } = false;
+
```
bshsqa, line 1103: This patch doesn't seem to include a way to handle the non-layout case when AdjustViewSize is true.
I'm currently ignoring the case of removing the layout again, but it's better to keep the behavior if the layout is set during working.

Please set resize policy to dimension dependency in this setter.
(Maybe we need to check details about several cases.)

jaehyun0cho: I totally agree with you and I also want to do that.
However, if AdJustViewSize sets DimensionDependency automatically, then it may misunderstand user's needs.

case 1. User sets AdjustViewSize true, and then user sets View's SizeWidth or SizeHeight later.
case 2. User sets AdjustViewSize true, and then user sets View's WidthResizePolicy or HeightResizePolicy later.

For the both cases, ImageView does not know which should be DimensionDependency when AdjustViewSize is set.

bshsqa: I read the reply of the original issue page after I've written this comment................ Sorry.
Yes, I agree with that, beside the cases you said, we need to consider several complex cases like to reset a layout of a View to null. (You already mentioned before.)
I still think to use specification and resize policy together is not good way. Let's think about the solution.
For the purpose, this patch is acceptable If we adds new patch for them soon.

bshsqa (approved):

everLEEst (approved):
LGTM.

PR #3543, opened 2021-09-12
[NUI] Avoid creating BindablePropertyContext for the special cases

rabbitfor:

### API Changes ###

 - ACR:


dongsug-song:
@everLEEst @AdunFang 
Hello,
could you review this? it's related to BindableProperty so I think your reviews will be helpful.
thank you.

everLEEst (commented):
LGTM but I think it need to check @AdunFang again.
as my understand,
we are planning to do this,
so only make BindablePropertyContexxt when it really binded something.

@AdunFang 
can you plase check it once?

AdunFang (approved):
It is OK.

dongsug-song (approved):
looks good to me

PR #3569, opened 2021-09-14
[NUI] Fix MatchParent child position with LinearLayout Center Alignment

jaehyun0cho:
Previously, LinearLayout calculated MatchParent children's positions
incorrectly when the LinearAlignment was Center.
Because the MatchParent children's sizes were not added to the sum of
all children's sizes.

Now, the MatchParent children's sizes are added to the sum of all
children's sizes and it leads that MatchParent children's positions are
calculated correctly when the LinearAlignment is Center.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM.

PR #3574, opened 2021-09-15
[NUI] ACR patch of typo fix

bshsqa:
- Synchronos -> Synchronous

### API Changes ###

 - ACR:


hyunjushin (approved):

Seoyeon2Kim (approved):
When you're ready to merge the patch, please remove "No Merge" tag.

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -435,8 +436,7 @@ public bool SynchronosLoading
         /// <summary>
         /// Gets or sets whether to synchronous loading the resourceurl of image.<br />
         /// </summary>
```
wiertel, line 438: The change itself is ok. I would like to improve the documentation of the property while we are at it. Please describe what operations, methods, behavior is affected by this property. Is it something like I've put in the suggestion below?
```suggestion
        /// Gets or sets whether loading of the image URL is synchronous.<br />
        /// </summary>
        /// <remarks>
        /// Setting this property to true will cause the following operation to be performed synchronously: <see cref="Reload"/>, <see cref="SetImage"/>.
        /// </remarks>
```

bshsqa: Thank you I fixed the comment based on the comment you wrote.
How about this?

PR #3580, opened 2021-09-16
[NUI] Make ImageView.AddViewSize public

jaehyun0cho:
This is ACR patch to make ImageView.AddViewSize public.

### API Changes ###

 - ACR:


bshsqa (approved):

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1132,7 +1132,7 @@ public WrapModeType WrapModeV
         /// If this is set to be true, then the width or height, which is not set by user explicitly, can be adjusted to preserve the aspect ratio of the image resource.
```
hyunjushin, line 1132: do you mean " can be adjusted while maintaining the aspect ratio"?

jaehyun0cho: It means the ImageView's width or height value can be changed to preserve the aspect ratio of image resource.
I will update the documentation to make users understand better.

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1129,10 +1129,13 @@ public WrapModeType WrapModeV
...
         /// Gets or sets the mode to adjust view size to preserve the aspect ratio of the image resource.
-        /// If this is set to be true, then the width or height, which is not set by user explicitly, can be adjusted to preserve the aspect ratio of the image resource.
+        /// If this is set to be true, then the width or height value, which is not set by user explicitly, can be changed automatically
+        /// to preserve the aspect ratio of the image resource.
         /// AdjustViewSize works only if ImageView is added to a View having Layout.
+        /// e.g. If the image resource size is (100, 100), then the ImageView requests size (100, 100) to its parent layout by default.
+        ///      If the ImageView's HeightSpecification is 50 and AdjustViewSize is true, then the ImageView requests size (50, 50) instead of (100, 50).
         /// </summary>
-        [EditorBrowsable(EditorBrowsableState.Never)]
+        /// <since_tizen> 9 </since_tizen>
```
hyunjushin, line 1138: why don't you set default value as true? keep aspect ratio looks natural. 
And please describe about default value.

jaehyun0cho: I think the default value is false, because AdjustViewSize changes the ImageView size to preserve the image resource's aspect ratio.
Therefore, I think that changing the ImageView size only if user sets AdjustViewSize true.

I have updated the documentation that the default value is false.

PR #3589, opened 2021-09-16
[NUI] Delete dispose() in GetOriginalImageSize

tscholb:
Some applications call this function outside the main thread and this is a problem.
so i delete it temporary

### API Changes ###

 - ACR:


dongsug-song (approved):
LGTM

PR #3590, opened 2021-09-16
[NUI] Apply dp converting in xaml float, int, extents, rectangles

everLEEst:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Xaml/CreateValuesVisitor.cs
+++ b/src/Tizen.NUI/src/internal/Xaml/CreateValuesVisitor.cs
@@ -448,15 +448,15 @@ namespace Tizen.NUI.Xaml
                 }
                 if (nodeType == typeof(Int16))
                 {
-                    return (Int16)GraphicsTypeManager.Instance.ConvertScriptToPixel(valuestring);
+                    return Convert.ToInt16(GraphicsTypeManager.Instance.ConvertScriptToPixel(valuestring));
                 }
                 if (nodeType == typeof(Int32))
                 {
-                    return (Int32)GraphicsTypeManager.Instance.ConvertScriptToPixel(valuestring);
+                    return Convert.ToInt32(GraphicsTypeManager.Instance.ConvertScriptToPixel(valuestring));
                 }
                 if (nodeType == typeof(Int64))
                 {
-                    return (Int64)GraphicsTypeManager.Instance.ConvertScriptToPixel(valuestring);
+                    return Convert.ToInt64(GraphicsTypeManager.Instance.ConvertScriptToPixel(valuestring));
                 }
                 if (nodeType == typeof(Byte))
                 {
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI/src/internal/Xaml/CreateValuesVisitor.cs
```diff
@@ -449,20 +449,50 @@ static object CreateLanguagePrimitive(Type nodeType, IElementNode node)
...
+                        return (short)GraphicsTypeManager.Instance.ConvertToPixel(short.Parse(valuestring.Substring(0, valuestring.LastIndexOf("dp")), CultureInfo.InvariantCulture));
+                    }
+                    else if (valuestring.EndsWith("px"))
+                    {
+                        return (short.Parse(valuestring.Substring(0, valuestring.LastIndexOf("px")), CultureInfo.InvariantCulture));
+                    }
+                    else if (short.TryParse(valuestring, NumberStyles.Number, CultureInfo.InvariantCulture, out retval))
+                    {
                         return retval;
+                    }
```
myroot, line 463: 메소드로 추출해서 리펙토링을 하면 좋을것 같습니다.

그리고 double을 리턴하는 메소드 하나만 뽑아서 나머지는 캐스팅해서 사용하면 될것 같습니다.
```suggestion
                    retval = (short) GraphicsTypeManager.Instance.ParseScableUnit(valuestring);
```

``` c#
        public static double ParseScableUnit(string valuestring)
        {
            if (valuestring == null)
            {
                return 0;
            }
            else if (valuestring.EndsWith("dp"))
            {
                var retval = double.Parse(valuestring.Substring(0, valuestring.LastIndexOf("dp")), CultureInfo.InvariantCulture);
                return retval * GraphicsTypeManager.Instance.Density;
            }
            else if (valuestring.EndsWith("px"))
            {
                return double.Parse(valuestring.Substring(0, valuestring.LastIndexOf("px")), CultureInfo.InvariantCulture);
            }
            else if (double.TryParse(valuestring, NumberStyles.Number, CultureInfo.InvariantCulture, out double retval))
            {
                return retval;
            }
            return 0;
        }
```

src/Tizen.NUI/src/internal/Xaml/CreateValuesVisitor.cs
```diff
@@ -448,21 +470,15 @@ static object CreateLanguagePrimitive(Type nodeType, IElementNode node)
                 }
                 if (nodeType == typeof(Int16))
                 {
-                    short retval;
-                    if (short.TryParse(valuestring, NumberStyles.Number, CultureInfo.InvariantCulture, out retval))
-                        return retval;
+                    return (Int16)ConvertScriptToPixel(valuestring);
```
elishateng, line 473: How about call GraphicsTypeManager.Instance.ConvertScriptToPixel(valuestring) directly ?

everLEEst: I was thought that way firstly.. 
but double and Int64, Uint64 may overflow.(even though that value never overflowed in our gui systems mostly)

myroot: How about change to return double type?

`double GraphicsTypeManager.Instance.ConvertScriptToPixel(string valuestring)`

everLEEst: I change GraphicsTypeManager.Instance.ConvertScriptToPixel(valuestring) all.

Thinking of overflow possibility,
we are getting value from string by user which is very hard to put overflowed value,
also our graphics system are based on float, so it will overflow somewhere else at the end
even we try to prevent well in parse.

so I think it is good enough to ignore overflow possiblity in this case.

AdunFang (approved):
It is OK.

PR #3592, opened 2021-09-17
[NUI] Separate the LayoutTransition from LayoutController

huiyueun:
Separate class to help understanding for layout process.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Layouting/LayoutController.cs
+++ b/src/Tizen.NUI/src/internal/Layouting/LayoutController.cs
@@ -16,9 +16,8 @@
  */
 
 using Tizen.NUI.BaseComponents;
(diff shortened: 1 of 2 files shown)


jaehyun0cho (approved):
This patch not only enhances the maintenance of the code, but also resolves the following issues.
- coreAnimation was not disposed when LayoutController is disposed.
- Whenever coreAnimation was played, Animation callbacks were attached again.

Looks good to me!

huiyueun:
> This patch not only enhances the maintenance of the code, but also resolves the following issues.
> 
> * coreAnimation was not disposed when LayoutController is disposed.
> * Whenever coreAnimation was played, Animation callbacks were attached again.
> 
> Looks good to me!

Thank you.

PR #3599, opened 2021-09-17
[NUI] fix Item measure specification to be largable as much as they need

everLEEst:
### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #3606, opened 2021-09-23
[NUI] fix crash with double enter in TextEditor with TextLabel

abdullehghujeh:
### Description of Change ###
after optimizing layout in this pull request : https://github.com/Samsung/TizenFX/pull/2841
a crash started to appear when entering double enter in TextEditor while its TextChanged event set text to a TextLabel, and these controls are inside parent with layout (example View with LinearLayout).
the crash happens because an extra calls to OnMeasure (compared to before the above optimization pull request), which seems to call GetNaturalSize which is not stable, and in this case it cause a crash.
these extra calls happens because we have moved the condition that check if layout is needed to the LayoutController, and the Measure function in LayoutItem does not check if layout is needed anymore.
so If Measure function of LayoutItem is called from anywhere other than the LayoutController (in this case MeasureChildWithMargins function from inside LayoutGroup is calling it), the code inside it will be executed regardless the layout is needed or not.

so if my above assumption is correct, we can add a condition to check if layout needed in Measure function to make sure we only call its code when needed from all code paths (this pull request do this).
or search for every place that call Measure function and add this condition.

example code : https://github.com/wonrst/NUI-Test/blob/main/text-handle-pos/handle.cs 

please let me know what do you think ?

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
@@ -203,10 +203,16 @@ namespace Tizen.NUI
         /// <since_tizen> 6 </since_tizen>
         public void Measure(MeasureSpecification widthMeasureSpec, MeasureSpecification heightMeasureSpec)
         {
-            OnMeasure(widthMeasureSpec, heightMeasureSpec);
-            OnMeasureIndependentChildren(widthMeasureSpec, heightMeasureSpec);
-            flags = flags | LayoutFlags.LayoutRequired;
-            flags &= ~LayoutFlags.ForceLayout;
+            bool needsLayout = NeedsLayout(widthMeasureSpec.Size.AsDecimal(), heightMeasureSpec.Size.AsDecimal(), widthMeasureSpec.Mode, heightMeasureSpec.Mode);
+
+            needsLayout = needsLayout || ((flags & LayoutFlags.ForceLayout) == LayoutFlags.ForceLayout);
+            if (needsLayout)
+            {
+                OnMeasure(widthMeasureSpec, heightMeasureSpec);
+                OnMeasureIndependentChildren(widthMeasureSpec, heightMeasureSpec);
+                flags = flags | LayoutFlags.LayoutRequired;
+                flags &= ~LayoutFlags.ForceLayout;
+            }
 
             oldWidthMeasureSpec = widthMeasureSpec;
             oldHeightMeasureSpec = heightMeasureSpec;
--- a/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
@@ -205,17 +205,15 @@ namespace Tizen.NUI
         {
             bool needsLayout = NeedsLayout(widthMeasureSpec.Size.AsDecimal(), heightMeasureSpec.Size.AsDecimal(), widthMeasureSpec.Mode, heightMeasureSpec.Mode);
 
-            needsLayout = needsLayout || ((flags & LayoutFlags.ForceLayout) == LayoutFlags.ForceLayout);
             if (needsLayout)
             {
                 OnMeasure(widthMeasureSpec, heightMeasureSpec);
                 OnMeasureIndependentChildren(widthMeasureSpec, heightMeasureSpec);
                 flags = flags | LayoutFlags.LayoutRequired;
                 flags &= ~LayoutFlags.ForceLayout;
+                oldWidthMeasureSpec = widthMeasureSpec;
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
```diff
@@ -203,10 +203,16 @@ internal void AttachToOwner(View owner)
         /// <since_tizen> 6 </since_tizen>
         public void Measure(MeasureSpecification widthMeasureSpec, MeasureSpecification heightMeasureSpec)
         {
-            OnMeasure(widthMeasureSpec, heightMeasureSpec);
-            OnMeasureIndependentChildren(widthMeasureSpec, heightMeasureSpec);
-            flags = flags | LayoutFlags.LayoutRequired;
-            flags &= ~LayoutFlags.ForceLayout;
+            bool needsLayout = NeedsLayout(widthMeasureSpec.Size.AsDecimal(), heightMeasureSpec.Size.AsDecimal(), widthMeasureSpec.Mode, heightMeasureSpec.Mode);
+
+            needsLayout = needsLayout || ((flags & LayoutFlags.ForceLayout) == LayoutFlags.ForceLayout);
```
jaehyun0cho, line 208: This line is not required because NeedsLayout checks this line already.

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
```diff
@@ -203,10 +203,16 @@ internal void AttachToOwner(View owner)
...
+            if (needsLayout)
+            {
+                OnMeasure(widthMeasureSpec, heightMeasureSpec);
+                OnMeasureIndependentChildren(widthMeasureSpec, heightMeasureSpec);
+                flags = flags | LayoutFlags.LayoutRequired;
+                flags &= ~LayoutFlags.ForceLayout;
+            }
 
             oldWidthMeasureSpec = widthMeasureSpec;
```
jaehyun0cho, line 217: Could you move this line into the above conditional braces?
```
    ...
    flags &= ~LayoutFlags.ForceLayout;
    oldWidthMeasureSpec = widthMeasureSpec;
    oldHeightMeasureSpec = heightMeasureSpec;
}
```

jaehyun0cho:
Could you let me know where the crash happens in Tizen.NUI code if you know?
(I mean the exact code line crash happens in Tizen.NUI. I would like to debug more about it.)

Regardless of the crash, I think that this patch also enhances the performance.
Because previously, Measure() was always executed from root to leaf layout even though a layout between root and leaf was not required to be measured.
By this patch, Measure() may be stopped in between root and leaf layout if it is not necessary.

Thank you for this patch!

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
abdullehghujeh: I added it here (took it form the old code before optimization) because it does not exists in the NeedsLayout function !
I'm not familiar with this part if code, so this check is needed in NeedsLayout function ? should I move it there ?

abdullehghujeh:
The crash does not happen in NUI, it is in dali.
after the optimization patch, GetNaturalSize in dali get called for extra unneeded times and this function causes crashes in some cases (we already know about GetNatuarlSize issues and trying to fix it).
so this extra calls cause the crash to show up (before optimization patch there is no crash).

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
jaehyun0cho: In NeedsLayout(), LayoutRequested is checked.
LayoutRequested getter is as follows.
```
get
{
    return (flags & LayoutFlags.ForceLayout) == LayoutFlags.ForceLayout;
}
```

Therefore, you don't need to add the code if you call NeedsLayout().

jaehyun0cho:
> The crash does not happen in NUI, it is in dali.
> after the optimization patch, GetNaturalSize in dali get called for extra unneeded times and this function causes crashes in some cases (we already know about GetNatuarlSize issues and trying to fix it).
> so this extra calls cause the crash to show up (before optimization patch there is no crash).

I see. Thank you for your explanation!

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
abdullehghujeh: thank you

src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
abdullehghujeh: done

jaehyun0cho (approved):
LGTM

PR #3617, opened 2021-09-27
[Build] Print SubmitRequest Tag to Annotations

JSUYA:
### Description of Change ###

Use the notice command of the workflow command.
https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-a-notice-message

This makes the SR Tag more easily exposed.

image
submit/20210927.091225 <- test message, ignore it :)

### API Changes ###

 - ACR:no


dongsug-song (approved):
looks good to me
very good to notify

Seoyeon2Kim (approved):

PR #3623, opened 2021-09-28
[NUI] Fix to update Scrollbar when ScrollableBase size changes

jaehyun0cho:
Previously, if ScrollableBase size changed, Scrollbar was not updated.
So some children was not displayed when ScrollableBase size changed.

Now, when ScrollableBase size changes, Scrollbar is updated in the
ScrollableBase's Relayout callback.
So all children are displayed correctly when ScrollableBase size changes.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM.

PR #3624, opened 2021-09-28
[NUI] Fix tct exception block issues

dongsug-song:
### Description of Change ###
[NUI] Fix tct exception block issues

### API Changes ###
none


everLEEst (approved):
LGTM.

PR #3629, opened 2021-09-29
[NUI] Fix View.DisConnectFromSignals() to return when the native object has no body

dongsug-song:
### Description of Change ###
- Fix View.DisConnectFromSignals() to return when the native object has no body
- Add debugging log which comes only when the NUI_DEBUG_ON is defined

### API Changes ###
none


Seoyeon2Kim (approved):
Looks good to me. This PR is adding a lot of logs to debug more easily than before.

PR #3631, opened 2021-09-30
[NUI] Apply scalable UI in Theme

everLEEst:
### API Changes ###

 - ACR:


everLEEst:
There was holding issue before,
but I tested latest binary, it worked.
I want to make sure it has no issue after applying all SR,
so will test again in new binary and report here again.

PR #3673, opened 2021-10-19
[NUI] Fix GridLayout to fill expanded size correctly

jaehyun0cho:
Previously, child did not fill to the expanded size if the child had
positive Width/HeightSpecification.

e.g. Let child's WidthSpecification = 100
     Let child's HeightSpecification = 100
     Let child's expanded size (row and col's size) be (200, 200)
     Let child's Stretch mode be ExpandAndFill.
     (child should fill to the expanded size)
     Then, child's size became (100, 100) instead of fill to (200, 200)

Changing child's Width/HeightSpecification to assign
MeasuredWidth/Height could cause child's size changed.
Moreover, the changed size was not recalculated to the expanded size.

Now, child's size change is reverted so the child's size is not changed.
But the way of fixing this bug should be modified in a better way later.

e.g. Assigning MeasuredWidth/Height without changing
     Width/HeightSpecification.

### API Changes ###

 - ACR:


jaehyun0cho:
This PR is same with https://github.com/Samsung/TizenFX/pull/3657

PR #3688, opened 2021-10-26
[NUI] Add Transition Group

bshsqa:

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Transition/TransitionItemBase.cs
+++ b/src/Tizen.NUI/src/internal/Transition/TransitionItemBase.cs
@@ -23,14 +23,16 @@ namespace Tizen.NUI
 {
     internal class TransitionItemBase : BaseHandle
     {
+        private bool appearingTransition = true;
+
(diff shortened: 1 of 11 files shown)


src/Tizen.NUI/src/internal/Transition/TransitionItemBase.cs
```diff
@@ -78,7 +80,7 @@ public AlphaFunction AlphaFunction
         /// <summary>
         /// Gets or sets whether the View moves with child or not.
         /// </summary>
-        public bool TransitionWithChild
+        public virtual bool TransitionWithChild
```
dongsug-song, line 83: can I ask why `virtual` is added? not like previous patch?

bshsqa: Because in this patch, TransitionGroupItem override the property.

dongsug-song (approved):
looks good to me

PR #3693, opened 2021-11-01
[Tts] Fix error enum value and doxygen description.

stom-hwang:
### Description of Change ###

Some description and error added by previous ACR-1653 is missed.
This patch adds these missing parts.

### API Changes ###

 - ACR:


stom-hwang:
This needs more validation.

PR #3694, opened 2021-11-01
[NUI] Call RemoveWidget when WidgetView is disposed

tscholb:
when WidgetView is disposed, RemoveWidget must be called.
so i added this patch

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #3718, opened 2021-11-10
[NUI][AT-SPI] Enabled and Disabled events for Accessibility

aswigon:
### Description of Change ###

These events are triggered whenever the value of `Accessibility.IsEnabled` is about to change. In particular, and by design, no event is delivered on application startup or shutdown. Users are encouraged to continue using `IsEnabled` to synchronously query the current accessibility state, and listen to these two events to detect and dynamically react to any later changes of this state if required.

Dependencies:
* https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-adaptor/+/266049/ (merged)
* https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/266314/ (merged)

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.Accessibility.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.Accessibility.cs
@@ -51,6 +51,12 @@ namespace Tizen.NUI
             [DllImport(NDalicPINVOKE.Lib, EntryPoint = "csharp_dali_accessibility_IsEnabled")]
             [return: MarshalAs(UnmanagedType.U1)]
             public static extern bool IsEnabled();
+
+            [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+            public delegate void EnabledDisabledSignalHandler();
+
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "csharp_dali_accessibility_RegisterEnabledDisabledSignalHandler")]
+            public static extern void RegisterEnabledDisabledSignalHandler(EnabledDisabledSignalHandler enabledSignalHandler, EnabledDisabledSignalHandler disabledSignalHandler);
         }
     }
 }
--- a/src/Tizen.NUI/src/internal/Interop/Interop.Accessibility.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.Accessibility.cs
@@ -15,38 +15,41 @@
  *
  */
 
+using System;
+using System.Runtime.InteropServices;
+
 namespace Tizen.NUI
 {
     internal static partial class Interop
     {
         internal static partial class Accessibility
         {
-            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "csharp_dali_accessibility_get_status")]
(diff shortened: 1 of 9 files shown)


Seoyeon2Kim (commented):
This PR would be reviewed and merged after dali-csharp-binder patch gets reviewed at first.

src/Tizen.NUI/src/internal/Interop/Interop.Accessibility.cs
```diff
@@ -15,39 +15,48 @@
...
             public static extern void BridgeDisableAutoInit();
 
-            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "csharp_dali_accessibility_IsEnabled")]
-            [return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.U1)]
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "csharp_dali_accessibility_IsEnabled")]
+            [return: MarshalAs(UnmanagedType.U1)]
             public static extern bool IsEnabled();
+
+            [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+            public delegate void EnabledDisabledSignalHandler();
```
bshsqa, line 56: Does EnabledDisabled mean this single handler handles both of Enabled and Disabled event?
How about AvailabilityStateChanged ~~ ?

aswigon: Actually this `delegate` here is just a type declaration for a pointer to function that takes void and returns void (i.e. `void (*)(void)` in C/C++), so, technically, it doesn't really handle anything.

But two such function pointers are parameters to `RegisterEnabledDisabledSignalHandler` (which get connected to `Dali::Accessibility::Bridge::EnabledSignal()` and `::DisabledSignal()`, respectively). Its exact signature in `dali-csharp-binder` is `void RegisterEnabledDisabledSignalHandler(void (*enabled)(void), void (*disabled)(void))`.

I would love to remove this type alias altogether, but, to my knowledge, C# does not allow to directly specify a function pointer as a parameter like it was done in `dali-csharp-binder` (in other words, omit the `delegate` declaration), that is, without resorting to using `IntPtr` which I also briefly considered.

Should I change `EnabledDisabledSignalHandler` to something like `VoidToVoidFunctionPointer` instead? This would probably better reflect what this `delegate` really represents.

bshsqa: What I really wanted to know is the reason why the delegate has both of Enable and Disable as a name.
But as you said it is used for both of the signal then it is not too weird.

bshsqa (approved):

lol-github (approved):
LGTM

kimcinoo (approved):
No doubts! Good work!

Seoyeon2Kim:
Unfortunately, there are huge changes on DevelNUI branch.
@aswigon , could you resolve the conflicts and have a look at Seungho's comment?

lol-github (approved):
LGTM, Good work!

Seoyeon2Kim (approved):
LGTM!

PR #3719, opened 2021-11-10
[NUI] Fix Weight and size calculation in LinearLayout

jaehyun0cho:
View's Weight is considered in LinearLayout's calculation only if the
View's Specification is MatchParent.
Because Weight is used to divide the parent's space to each child.

Previously, WrapContent View's Weight was also added to the sum of
Weights.
This caused the weighted View's size was calculated incorrectly.

Now, only MatchParent View's Weight is added to the sum of Weights.
As a result, the weighted View's size is calculated correctly.

When a View's Specification was WrapContent, the View's LinearLayout did
not consider its child size correctly.
This caused that the View's Size became 0 incorrectly.

To resolve the above, LinearLayout always considers its child size and
the LinearLayout uses its child size when it calculates its own size.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
I agree about his idea. Weight should be working only for MatchParent case.
But, I'm afraid of the side effect or the existing apps' behavior. (Fhub)
We need to update a guide for it. :)

PR #3723, opened 2021-11-11
[NUI] make a null pointer when signal is disconnected

tscholb:
when signal is disconnected, signal pointer need to be null.
so i added this patch

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #3725, opened 2021-11-11
[NUI] add null check in collectionview

everLEEst:
null value is abnormal in collectionview,
but adding null value check for code management.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #3729, opened 2021-11-12
[NUI] Fix crash issue when signal is disposed

tscholb:
i create this patch to fix the crash issue.
it will be avoid the crash issue.

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #3731, opened 2021-11-12
Revert "[NUI] Ignore ResizePolicy when View size is calculated by Lay…

jaehyun0cho:
…out"

This reverts commit eeaaecd33da254f5447910ecd9b0cab89407d4ba.

By applying the original PR, TextLabel size could not be calculated
correctly because TextLabel size should be calculated in DALi.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM

PR #3732, opened 2021-11-13
[NUI] Check if event handler of the DisposeQueued object is valid

dongsug-song:
### Description of Change ###
[NUI] Check if event handler of the DisposeQueued object is valid

- The objects that are waiting in disposeQueue are not yet signal disconnected, and their states are implict disposed.
- When dali callback comes and the obeject is in the queue, it seems that the problem will occur.
- User event callback checks IsDisposeQueued before Invoked, and if Dispose == False, IsDisposeQueued == true, then event callback is not invoked.
- when Dispose == True, User callback != NULL, throw exception.

### API Changes ###
none


huiyueun (approved):
LGTM

PR #3739, opened 2021-11-15
[NUI][ATSPI] Calculate states based on DALi

kimcinoo:
We do not have to same job calculating states.
This patch sets make the View use DALi states.
If there is NUI specific state, then it will be overwritten.

This depends on following change.
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/266577/
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/AlertDialog.cs
+++ b/src/Tizen.NUI.Components/Controls/AlertDialog.cs
@@ -412,11 +412,11 @@ namespace Tizen.NUI.Components
         /// Informs AT-SPI bridge about the set of AT-SPI states associated with this object.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        protected override AccessibilityStates AccessibilityCalculateStates()
+        protected override AccessibilityStates AccessibilityCalculateStates(ulong states)
         {
-            var states = base.AccessibilityCalculateStates();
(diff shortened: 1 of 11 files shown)


src/Tizen.NUI.Components/Controls/AlertDialog.cs
```diff
@@ -412,11 +412,11 @@ public override void OnInitialize()
         /// Informs AT-SPI bridge about the set of AT-SPI states associated with this object.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        protected override AccessibilityStates AccessibilityCalculateStates()
+        protected override AccessibilityStates AccessibilityCalculateStates(ulong states)
         {
-            var states = base.AccessibilityCalculateStates();
-            FlagSetter(ref states, AccessibilityStates.Modal, true);
-            return states;
+            var _states = base.AccessibilityCalculateStates(states);
```
dongsug-song, line 417: if possible, we are recommended not to use `_`. 
could you remove `_` ?

kimcinoo: Thank you for your review.
Updated.

dongsug-song (approved):
looks good to me

PR #3742, opened 2021-11-17
[NUI] Add some debug log to check reference count of disposed object

dongsug-song:
### Description of Change ###
[NUI] Add some debug log to check reference count of disposed object

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/Disposable.cs
+++ b/src/Tizen.NUI/src/internal/Common/Disposable.cs
@@ -109,11 +109,18 @@ namespace Tizen.NUI
                 //Throw exception if Dispose() is called in separate thread.
                 if (!Window.IsInstalled())
                 {
-                    throw new System.InvalidOperationException("This API called from separate thread. This API must be called from MainThread.");
+                    var process = global::System.Diagnostics.Process.GetCurrentProcess().Id;
+                    var thread = global::System.Threading.Thread.CurrentThread.ManagedThreadId;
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/internal/Common/Disposable.cs
```diff
@@ -109,11 +109,18 @@ protected virtual void Dispose(bool disposing)
...
+                    var me = this.GetType().FullName;
+
+                    throw new global::System.InvalidOperationException("[NUI][Disposable]This API called from separate thread. This API must be called from MainThread. \n" +
+                        $" process:{process} thread:{thread}, disposing:{disposing}, isDisposed:{this.disposed}, isDisposeQueued:{this.isDisposeQueued}, me:{me}\n");
                 }
 
                 if (isDisposeQueued)
                 {
+                    Tizen.Log.Fatal("NUI", $"[Disposable]should not be here! (dead code) this will be removed!");
+                    throw new global::System.Exception($"[NUI] should not be here! (dead code) this will be removed!");
```
jaehyun0cho, line 123: Is it ok if this exception occurs here?

jaehyun0cho (approved):
LGTM

src/Tizen.NUI/src/internal/Common/Disposable.cs
dongsug-song: yes. it must not be came here. if there is no exception, it will be removed in two weeks.

PR #3749, opened 2021-11-18
[NUI] Add some debugging code to check if the object,which has no body, is used

dongsug-song:
### Description of Change ###
[NUI] Add some debugging code to check if the object,which has no body, is used

### API Changes ###
none


Seoyeon2Kim (approved):
Approved. Just for debugging issues.

AchoWang (commented):
Looks good to me.

PR #3751, opened 2021-11-18
[NUI] Make image visual actions protected

bshsqa:

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
Hope the properties not to use anywhere.

bshsqa:
@Seoyeon2Kim It is very a problem....
How do you think which is better to make them "protected internal" again with hidden or to keep current implementation.....

PR #3754, opened 2021-11-18
[Tts][TCSACR-477] Add new API and error enum value

stom-hwang:
### Description of Change ###

Add new API and class for repeating last played text.
- API to repeat the last played text
- Class to store the repeated text and its utterance ID

Add error enum value which presents the next situation.
- Screen reader is off
- Service is reset
- Feature is not supported by current engine.

### API Changes ###

 - ACR: https://code.sec.samsung.net/jira/browse/TCSACR-477


psw0405 (approved):
LGTM

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
```diff
@@ -676,7 +691,7 @@ public State CurrentState
         /// <exception cref="OutOfMemoryException">This exception can be due to out Of memory.</exception>
         /// <exception cref="NotSupportedException">This exception can be due to TTS not supported.</exception>
         /// <pre>
-        /// The State should be created.
+        /// The state must be created.
```
wiertel, line 694: it's great to improve this description, I would suggest rephrasing it to not confuse with creating a state - as if the "created" was a part of verb, but that "The Client must be in the `created` state" or "in the state `created`" or "The Client state must be `created`" or "The state must be \<see cref="State.Created"/\>",
same for other state prerequisites

wiertel (approved):

PR #3756, opened 2021-11-19
[NUI]Add window auxiliary message

sparrow74:
### Description of Change ###

Add window auxiliary message

 When client application added the window's auxiliary hint and if the auxiliary is changed,  display server send the auxiliary message.


sparrow74:
I met some conflict. I will PR that, again.

PR #3758, opened 2021-11-19
[NUI] Add window auixiliary message event

sparrow74:
### Description of Change ###
When client application added the window's auxiliary hint and if the auxiliary is changed, display server send the auxiliary message.

### DIFF
--- a/src/Tizen.NUI/src/public/Window/WindowEvent.cs
+++ b/src/Tizen.NUI/src/public/Window/WindowEvent.cs
@@ -19,6 +19,7 @@ using System;
 using System.ComponentModel;
 using System.Runtime.InteropServices;
 using Tizen.NUI.BaseComponents;
+using System.Collections.Generic;
 
 namespace Tizen.NUI
 {
@@ -58,6 +59,7 @@ namespace Tizen.NUI
         private VoidSignal eventProcessingFinishedSignal;
         private VoidSignal contextLostSignal;
         private VoidSignal contextRegainedSignal;
+        private AuxiliaryMessageEventCallbackType auxiliaryMessageEventCallback;
 
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
         private delegate void WindowFocusChangedEventCallbackType(IntPtr window, bool focusGained);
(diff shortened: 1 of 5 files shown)


src/Tizen.NUI/src/public/Window/WindowEvent.cs
```diff
@@ -1082,5 +1083,108 @@ public void VisibiltyChangedSignalEmit(bool visibility)
...
+
+            List<string> tmpList = new List<string>();
+
+            for (int i = 0; i < size; i++)
+            {
+                string option = "none";
+                tmp.GetElementAt((uint)i).Get(out option);
+                tmpList.Add(option);
+            }
+
```
dongsug-song, line 1134: please call `tmp.Dispose();` here to remove this warning =>

**src/Tizen.NUI/src/public/Window/WindowEvent.cs#L1131
Call System.IDisposable.Dispose on object created by 'tmp.GetElementAt((uint)i)' before all references to it are out of scope**

src/Tizen.NUI/src/public/Window/WindowEvent.cs
```diff
@@ -1082,5 +1083,108 @@ public void VisibiltyChangedSignalEmit(bool visibility)
...
+                        {
+                            AuxiliaryMessageEventCallback = null;
+                        }
+                    }
+                }
+            }
+        }
+
+
+
```
dongsug-song, line 1188: could you add the following in `DisconnectNativeSignals()`

            if (AuxiliaryMessageEventCallback != null)
            {
                using var signal = new WindowAuxiliaryMessageSignal(this);
                signal.Disconnect(AuxiliaryMessageEventCallback);
                AuxiliaryMessageEventHandler= null;
                AuxiliaryMessageEventCallback = null;
            }

thank you.

src/Tizen.NUI/src/public/Window/WindowEvent.cs
sparrow74: Thank you for your detail review.
I updated that.

dongsug-song (approved):
looks good to me

PR #3760, opened 2021-11-22
[NUI] Add window auixiliary message event

sparrow74:
### Description of Change ###

When client application added the window's auxiliary hint and if the auxiliary is changed, display server send the auxiliary message.


dongsug-song (approved):
looks good to me

PR #3767, opened 2021-11-23
[NUI][ATSPI] Navigator - emit accessibility event

kimcinoo:
When new page appears, we need to inform it to the screen-reader to read
its name as a default label.
And we need to inform the currently highlighted component's state on the
disappeared page. If the currently highlighted component is not visible
then it will be cleared by screen-reader.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
+++ b/src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
@@ -197,12 +197,13 @@ namespace Tizen.NUI.Components
             {
                 if (page is DialogPage == false)
                 {
-                   topPage.SetVisible(false);	   
+                   topPage.SetVisible(false);
                 }
 
                 //Invoke Page events
                 page.InvokeAppeared();
                 topPage.InvokeDisappeared();
+                NotifyAccessibilityStatesChangeOfPages(topPage, page);
             };
             transitionFinished = false;
         }
@@ -252,6 +253,7 @@ namespace Tizen.NUI.Components
                 //Invoke Page events
                 newTopPage.InvokeAppeared();
                 topPage.InvokeDisappeared();
+                NotifyAccessibilityStatesChangeOfPages(topPage, newTopPage);
 
                 //Invoke Popped event
                 Popped?.Invoke(this, new PoppedEventArgs() { Page = topPage });
@@ -326,6 +328,7 @@ namespace Tizen.NUI.Components
                 {
                     //Invoke Page events
                     page.InvokeAppeared();
+                    NotifyAccessibilityStatesChangeOfPages(curTop, page);
                 };
                 newAnimation.Play();
             }
@@ -399,6 +402,7 @@ namespace Tizen.NUI.Components
                 {
                     //Invoke Page events
                     newTop.InvokeAppeared();
+                    NotifyAccessibilityStatesChangeOfPages(curTop, newTop);
                 };
                 newAnimation.Play();
             }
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -384,6 +388,7 @@ public Page Pop()
 
                     //Invoke Page events
                     curTop.InvokeDisappeared();
+                    NotifyAccessibilityStatesChangeOfPages(curTop, null);
```
jaehyun0cho, line 391: I think that this line needs to be removed.
Instead, the following code is required at line 407.
```
NotifyAccessibilityStatesChangeOfPages(curTop, newTop);
```

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -399,6 +404,7 @@ public Page Pop()
                 {
                     //Invoke Page events
                     newTop.InvokeAppeared();
+                    NotifyAccessibilityStatesChangeOfPages(null, newTop);
```
jaehyun0cho, line 407: I think this line needs to be changed as follows.
```
NotifyAccessibilityStatesChangeOfPages(curTop, newTop);
```

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -314,6 +316,7 @@ public void Push(Page page)
 
                     //Invoke Page events
                     curTop.InvokeDisappeared();
+                    NotifyAccessibilityStatesChangeOfPages(curTop, null);
```
jaehyun0cho, line 319: I think that this line needs to be removed.
Instead, the following code is required at line 332.
```
NotifyAccessibilityStatesChangeOfPages(curTop, page);
```

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -326,6 +329,7 @@ public void Push(Page page)
                 {
                     //Invoke Page events
                     page.InvokeAppeared();
+                    NotifyAccessibilityStatesChangeOfPages(null, page);
```
jaehyun0cho, line 332: I think this line needs to be changed as follows.
```
NotifyAccessibilityStatesChangeOfPages(curTop, page);
```

kimcinoo: Thank you for your comment.
The property VISIBLE of curTop is changed to FALSE at 315
Is the property VISIBLE of curTop FALSE at this point 332 ?
So it seems that this line needs to be moved to 315.

kimcinoo: I have checked that Disappeared called first and Appeared called next.
So you are right. Updated. Thank you for your review.

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
kimcinoo: done

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
kimcinoo: done

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
kimcinoo: done

jaehyun0cho (approved):
LGTM

PR #3769, opened 2021-11-23
[NUI] Fix crash when View.BackgroundImage is set

jaehyun0cho:
Previously, View.BackgroundImage is set, PropertyValue constructor can
be called with null pointer and this causes crash.

Now, it has been fixed not to pass null pointer to PropertyValue
constructor.

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #3774, opened 2021-11-26
[NUI][ATSPI] set RecyclerViewItem role 

kimcinoo:
From this patch, it is able to set highlight to the DefaultLinearItem.

### API Changes ###

 - ACR:


everLEEst (changes requested):
looks better to move the feature into the super class for applying acc in all items.

kimcinoo:
> looks better to move the feature into the super class for applying acc in all items.

Moved lines to RecycleViewItem.

everLEEst (approved):
Looks good to me.
this OnInitialize method is virtual function,
so sub class might turn highlight off by overriding function! 
:+1:

lol-github (approved):
LGTM

PR #3775, opened 2021-11-26
[NUI][ATSPI] Make DefaultLinearItem use its TextLabel

kimcinoo:
(1) "LabelledBy" relation makes DefaultLinearItem use its TextLabel
    as accessibility name.

(2) when user does 1 finger single tap on the TextLabel, then the
    DefaultLinearItem will have highlighted by "ControlledBy" relation.

This changed could be set on application side.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/RecyclerView/Item/DefaultLinearItem.cs
+++ b/src/Tizen.NUI.Components/Controls/RecyclerView/Item/DefaultLinearItem.cs
@@ -138,6 +138,8 @@ namespace Tizen.NUI.Components
                         layoutChanged = true;
                         Add(itemLabel);
                     }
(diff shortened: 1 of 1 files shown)


everLEEst (commented):
Default Linear Item can have secondary text,
can we read secondary text together when it exist?

kimcinoo:
It is not able to read with this change.
Probably DefaaultLinearItem needs to use a read callback instead of using the relation information.

lol-github:
Precisely, the AccessibilityNameRequested signal can be helpful in case of more complex internal structure of DefaultLinearItem like secondary text:

defaultLinearItem.AccessibilityNameRequested += OnAccessibilityNameRequested;

everLEEst (approved):
I think it is good for now.
secondary text question was just only for checking how application can manage this.
thank you!

PR #3781, opened 2021-12-01
[NUI][AT_SPI] Added Accessibility support to Menu control

m-bialota:
### Description of Change ###

Added AT-SPI2 support to Menu control.

Now in A11y mode the UI context is rebuilt and proper navigation provided.

### API Changes ###

 - ACR: none

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Menu.cs
+++ b/src/Tizen.NUI.Components/Controls/Menu.cs
@@ -394,6 +394,8 @@ namespace Tizen.NUI.Components
             layer.RaiseToTop();
 
             CalculateSizeAndPosition();
+            RegisterDefaultLabel();
+            NotifyAccessibilityStatesChange(AccessibilityStates.Visible | AccessibilityStates.Showing, true);
         }
 
         /// <summary>
@@ -404,6 +406,8 @@ namespace Tizen.NUI.Components
         public void Dismiss()
         {
             Hide();
+            UnregisterDefaultLabel();
+            NotifyAccessibilityStatesChange(AccessibilityStates.Visible | AccessibilityStates.Showing, true);
             Dispose();
         }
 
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI.Components/Controls/Menu.cs
```diff
@@ -640,5 +640,28 @@ private void CalculateScrimPosition()
...
+        /// <summary>
+        /// Initialize AT-SPI object.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override void OnInitialize()
+        {
+            base.OnInitialize();
+            SetAccessibilityConstructor(Role.PopupMenu);
+            AppendAccessibilityAttribute("sub-role", "Alert");
+            Show(); // calls AddPopup()
```
dongsug-song, line 653: Mr. kimcinoo pointed out whether this `Show()` is proper.
@jaehyun0cho 
Hello, Mr. jaehyun0cho
could you check this? I don't think this is required, because this is about ATSPI and ATSPI should not control the Component itself.

jaehyun0cho: @m-bialota How about moving Show() to the end of Post()?
Because Menu is supposed to be displayed when app calls Post().

m-bialota: OK, I moved the call to Post(). it works well too :).

src/Tizen.NUI/src/public/BaseComponents/ViewPublicMethods.cs
```diff
@@ -303,7 +303,7 @@ public void ClearBackground()
         /// <since_tizen> 3 </since_tizen>
         public void Show()
         {
-            if (Accessibility.Accessibility.Enabled && ((GetAccessibilityStates() & AccessibilityStates.Modal) != 0))
+            if ((GetAccessibilityStates() & AccessibilityStates.Modal) != 0)
```
m-bialota, line 306: The Accessibility.Accessibility.Enabled condition has to be removed in this place, as  the existence of any Modal View has to be registered even if Accessibility mode is off, otherwise when Screen Reader is being turned on when the Menu or any other popup-like View is currently shown, the UI context root will not be properly set and the user may navigate out of the context.

m-bialota: This issue needs more investigation, so I removed this change from the current commit.

src/Tizen.NUI.Components/Controls/Menu.cs
```diff
@@ -394,6 +394,7 @@ public void Post(Window window = null)
             layer.RaiseToTop();
 
             CalculateSizeAndPosition();
+            Show(); // calls RegisterPopup()
```
lol-github, line 397: If placing the call to Show() method here is fixing a gap in component's logic then there is no need to explicitly explain the impact on accessibility in the comment (btw, if the following PR https://github.com/Samsung/TizenFX/pull/3801 will be merged first, the comment became obsolete as the accessibility API will change from RegisterPopup to RegisterDefaultLabel).
However, if placing the call to Show() method here is not required by component's logic and it is done just to fulfill accessibility requirement, then IMHO it should be replaced with direct call to RegisterDefaultLabel() API as it is done in Navigator.cs in https://github.com/Samsung/TizenFX/pull/3801 (then the hiding scenario should be handled similarly by direct call to UnregisterDefaultLabel())

aswigon (approved):
LGTM

lol-github (approved):
LGTM

aswigon (approved):

m-bialota:
This patch does not cause regression and works properly with current version of dali from devel/master branch.

Seoyeon2Kim:
@m-bialota , could you resolve the conflicts and update to add [NUI] in the title?
There are huge changes on DevelNUI branch.

lol-github (approved):
LGTM, Good work!

everLEEst (approved):
Looks good to me!

Seoyeon2Kim (approved):
Thanks :)

PR #3793, opened 2021-12-08
[NUI] Set null if callback removed

hinohie:
When signal disconnected and signal become empty, let callback as null.

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #3798, opened 2021-12-08
[NUI] Add to call UnregisterDefaultLabel when Page is popped

Seoyeon2Kim:

### Description of Change ###
 - Currently, Dialog and AlertDialog only call `RegisterDefaultLabel()` on init time and do not call `UnregisterDefaultLabel()`.
 - So, move the code to call Register / Unregister methods when Page is pushed and popped to Navigator.

 - In AT-SPI side, when Navigator pops a page (like `Button` clicked), the page is just removed from Navigator page list.
    That's why the page still has an accessibility highlight and cannot see on the screen.

### API Changes ###
- N/A

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
+++ b/src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
@@ -231,6 +231,9 @@ namespace Tizen.NUI.Components
                    topPage.SetVisible(false);
                 }
 
+                // Need to update Content of the new page
+                ShowContentOfPage(page);
+
                 //Invoke Page events
                 page.InvokeAppeared();
                 topPage.InvokeDisappeared();
@@ -281,6 +284,9 @@ namespace Tizen.NUI.Components
                 Remove(topPage);
                 topPage.SetVisible(true);
 
+                // Need to update Content of the new page
+                ShowContentOfPage(newTopPage);
+
                 //Invoke Page events
                 newTopPage.InvokeAppeared();
                 topPage.InvokeDisappeared();
@@ -329,8 +335,6 @@ namespace Tizen.NUI.Components
             Add(page);
             page.Navigator = this;
 
-            AddPageToList(page);
-
             //Invoke Page events
             page.InvokeAppearing();
             curTop.InvokeDisappearing();
@@ -359,11 +363,18 @@ namespace Tizen.NUI.Components
                 newAnimation.EndAction = Animation.EndActions.StopFinal;
                 newAnimation.Finished += (object sender, EventArgs e) =>
                 {
+                    // Need to update Content of the new page
+                    ShowContentOfPage(page);
+
                     //Invoke Page events
                     page.InvokeAppeared();
                     NotifyAccessibilityStatesChangeOfPages(curTop, page);
                 };
                 newAnimation.Play();
+            }
+            else
+            {
+                ShowContentOfPage(page);
             }
         }
 
@@ -387,7 +398,6 @@ namespace Tizen.NUI.Components
             }
 
             var curTop = Peek();
-            RemovePageToList(curTop);
 
             if (navigationPages.Count == 1)
             {
@@ -434,6 +444,9 @@ namespace Tizen.NUI.Components
                 newAnimation.EndAction = Animation.EndActions.StopFinal;
                 newAnimation.Finished += (object sender, EventArgs e) =>
                 {
+                    // Need to update Content of the new page
+                    ShowContentOfPage(newTop);
+
                     //Invoke Page events
                     newTop.InvokeAppeared();
                     NotifyAccessibilityStatesChangeOfPages(curTop, newTop);
@@ -523,6 +536,8 @@ namespace Tizen.NUI.Components
             //TODO: The following transition codes will be replaced with view transition.
             InitializeAnimation();
 
+            ShowContentOfPage(page);
+
             if (index == PageCount)
             {
                 page.Opacity = 1.0f;
@@ -588,6 +603,8 @@ namespace Tizen.NUI.Components
 
             //TODO: The following transition codes will be replaced with view transition.
             InitializeAnimation();
+
+            HideContentOfPage(page);
 
             if ((page == Peek()) && (PageCount >= 2))
             {
@@ -864,31 +881,23 @@ namespace Tizen.NUI.Components
             }
         }
 
-        // Register dialog(popup) to Accessibility bridge
-        internal void AddPageToList(Page page)
+        // Show and Register Content of Page to Accessibility bridge
+        private void ShowContentOfPage(Page page)
         {
-            ContentPage obj = page as ContentPage;
-            if (obj != null)
+            View content = (page is DialogPage) ? (page as DialogPage)?.Content : (page as ContentPage)?.Content;
+            if (content != null)
             {
-                View view = obj.Content;
-                if (view != null)
-                {
-                    view.Show(); // Calls RegisterDefaultLabel()
-                }
+                content.Show(); // Calls RegisterDefaultLabel()
             }
         }
 
-        // Remove dialog(popup) from Accessibility bridge
-        internal void RemovePageToList(Page page)
+        // Hide and Remove Content of Page from Accessibility bridge
+        private void HideContentOfPage(Page page)
         {
-            ContentPage obj = page as ContentPage;
-            if (obj != null)
+            View content = (page is DialogPage) ? (page as DialogPage)?.Content : (page as ContentPage)?.Content;
+            if (content != null)
             {
-                View view = obj.Content;
-                if (view != null)
-                {
-                    view.Hide(); // Calls UnregisterDefaultLabel()
-                }
+                content.Hide(); // Calls UnregisterDefaultLabel()
             }
         }
     }
--- a/src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
+++ b/src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
@@ -329,6 +329,8 @@ namespace Tizen.NUI.Components
             Add(page);
             page.Navigator = this;
 
+            AddPageToList(page);
+
             //Invoke Page events
             page.InvokeAppearing();
             curTop.InvokeDisappearing();
@@ -385,6 +387,7 @@ namespace Tizen.NUI.Components
             }
 
             var curTop = Peek();
+            RemovePageToList(curTop);
 
             if (navigationPages.Count == 1)
             {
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -440,6 +440,7 @@ public Page Pop()
             else
             {
                 Remove(curTop);
+                curTop.Dispose();
```
hinohie, line 443: When we dispose page here, it will return `Disposed` page.
I think we need to find other good way ;ㅅ;

everLEEst: Agreed.
as far as I know, pop is only intent to pop out the page on navigator, so disposing or keepping it and reusing is application's choice. if we might provide RemovePage() or DeletePage() in Navigator, we could process Dispose() internally as it will not return the page and API intended to remove page object.
please consider remove this line or adding other APIs with discussing @Jaehyun-Cho

Seoyeon2Kim: Thank you for your reviews.
I added `Hide()` as you mentioned here at first, but `RemovePopup()` cannot be called because `Page` is not AccessibilityStates.Modal. And I thought I should match Init and Dispose.
However, as you told, I'll think about something else.

Seoyeon2Kim: Hello all,
I made 'DeletePage()' to call Dispose() internally. Could you check it again?

everLEEst: hmm.. what I meant was Pop() should only processing "pop" action without delete any objects... so if we provide another public API DeletePopup() instead of calling Pop(). or they can dispose popped object which is currently provided.
To solve the problem, how about manaing accessibility list instead of disposing object?
adding accessibility tree when it is added in the navi and remove tree when it popped.

Seoyeon2Kim: Oh I see. I added to call Show/Hide directly instead of adding accessibility tree.
What do you think?

everLEEst: So if object is hided, object will be disqualified accesibility highlight candidate? that seems good solution that I look. 👍 
@jaehyun0cho have you other opinion about this?

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -858,5 +861,33 @@ private void InitializeAnimation()
                 newAnimation = null;
             }
         }
+
+        // Register dialog(popup) to Accessibility bridge
+        internal void AddPageToList(Page page)
```
everLEEst, line 866: I think we could do this for all content page,
and may be internal method is not necessary that only call it one place.
if it necessary,
function name would be better to use show/hide term that other developer can guess what this function actaully does.

thank you for update!

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -385,6 +387,7 @@ public Page Pop()
             }
 
             var curTop = Peek();
+            RemovePageToList(curTop);
```
jaehyun0cho, line 390: It seems that this API hides content of current page.
If there are multiple pages, then popped page is not immediately hidden. Instead, it shows pop animation and then disappears.
Therefore, this API should be called in Finished callback and if (navigationPages.Count == 1).

Isn't it required to call this API in PopWithTransition() and Remove()?

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -329,6 +329,8 @@ public void Push(Page page)
             Add(page);
             page.Navigator = this;
 
+            AddPageToList(page);
```
jaehyun0cho, line 332: It seems that this API shows content of current page.
If there are multiple pages, then there are push animation and then the pushed page is finally displayed.
Therefore, this API should be called in Finished callback.

Isn't it required to call this API in PushWithTransition() and Insert()?

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
jaehyun0cho: I think that it needs to move the APIs to the correct places.
I left my comments on RemovePageToList() and AddPageToList() below.

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
Seoyeon2Kim: Okay, I added Add~ method(Currently, Show~) to PushWithTransition(), Push(), Insert(), PopWithTransition() for the new page, and Pop() for the new page.
Could you check it again?

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
Seoyeon2Kim: I added Remove~ method(Currently, Hide~) to PushWithTransition() for  the disappeared page and Remove().
In Pop() and other methods, Remove() method is already called, so no need to call HideContentOfPage() again.
Please have a look.

Seoyeon2Kim:
Dear @jaehyun0cho, @everLEEst , and @kimcinoo ,
Sorry for the late update.
I rebased this PR on the latest DevelNUI branch  and updated the code according  to reviews.
Could you check them again?

everLEEst (commented):
Looks okay to me but @jaehyun0cho need to check and approve it as maintainer.
thank you for the effort! 👍

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -860,5 +885,25 @@ private void InitializeAnimation()
                 newAnimation = null;
             }
         }
+
+        // Show and Register Content of Page to Accessibility bridge
+        private void ShowContentOfPage(Page page)
```
bshsqa, line 890: I think this design seems a bit of an abuse of authority.
If we want to request to show or to hide Content of each Page, I think it is better to make ShowContent() and HideContent() methods in Page and call them here.
Page.ShowContent() and HideContent() may become virtual.
And real implementation will be written in each inherit class.

everLEEst: agreed.

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -226,10 +226,19 @@ public void PushWithTransition(Page page)
             transitionSet = CreateTransitions(topPage, page, true);
             transitionSet.Finished += (object sender, EventArgs e) =>
             {
+                HideContentOfPage(topPage);
```
jaehyun0cho, line 229: Is this required here?
Because it seems that Push() works fine without this code.

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -226,10 +226,19 @@ public void PushWithTransition(Page page)
...
+
                 if (page is DialogPage == false)
                 {
                    topPage.SetVisible(false);
                 }
+                else
+                {
+                    topPage.SetVisible(true);
```
jaehyun0cho, line 237: Is this code required here?

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
Seoyeon2Kim: Ah
Thanks for checking it. I didn't delete all the old codes I've tested.
Remove it and make the same as Push().

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
Seoyeon2Kim: No, it can be removed now, too.
Thank you for reviewing it.

src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
Seoyeon2Kim: Regarding this, we need a separate patch because Page needs Content then, as you know.
Can we go to another PR about this one?

jaehyun0cho: I agree with @Seoyeon2Kim .
If @bshsqa  and @everLEEst  agrees, we can make a separate PR.

jaehyun0cho (approved):
LGTM
If @bshsqa  and @everLEEst  agrees, then we can merge this PR and create a separate PR.
If this PR is merged, please select "Squash and Merge".

bshsqa (approved):

m-bialota (commented):
This change causes regression in behavior of AlertDialog if its parent is Window.
I am not sure if the following example is a correct usage of AlertDialog, but if the following is called:
`Tizen.NUI.Components.AlertDialog popup = new Tizen.NUI.Components.AlertDialog();
(....)
Window.Instance.GetDefaultLayer().Add(popup);`
then the AlertDialog does not "grab" the focus chain in A11y mode.
Moreover, when the following code is triggered by pressing the button being part of the Navigator Page, the AlertDialog is also not accessible at all.

Seoyeon2Kim:
> This change causes regression in behavior of AlertDialog if its parent is Window. I am not sure if the following example is a correct usage of AlertDialog, but if the following is called: `Tizen.NUI.Components.AlertDialog popup = new Tizen.NUI.Components.AlertDialog(); (....) Window.Instance.GetDefaultLayer().Add(popup);` then the AlertDialog does not "grab" the focus chain in A11y mode. Moreover, when the following code is triggered by pressing the button being part of the Navigator Page, the AlertDialog is also not accessible at all.

I understood what you said.
Currently, AlertDialog and Dialog are guided to be used with Page and Navigator classes. So, I only checked this case.

However, I think adding to Window should be considered as you mentioned. Dialog and AlertDialog inherit from Control class.
I returned the original code of 'Show()' in OnInitialize() of Dialogs. 
Could you check it out again?

But, there's no code to call 'Hide()' in Dialogs, so I think this needs to be fixed in another PR.

m-bialota:
I checked the new version with reverted changes on dialogs. and there is no regression now. Thank you!
Concerning calling Hide() in AlertDialog I think the typical use case is that application developer calls Hide() on "OK" and "Cancel" button callbacks.

Seoyeon2Kim:
> I checked the new version with reverted changes on dialogs. and there is no regression now. Thank you! Concerning calling Hide() in AlertDialog I think the typical use case is that application developer calls Hide() on "OK" and "Cancel" button callbacks.

I see. I need to update the description about it in AlertDialog.
Thank you for checking it ! ;)

PR #3799, opened 2021-12-08
[NUI] Add ContentPageLayout to calculate Content size correctly

jaehyun0cho:
Previously, ContentPage's Layout was AbsoluteLayout so the height of
Content was calculated to the same with the height of ContentPage,
although AppBar existed at the top of ContentPage.

Now, ContentPageLayout has been introduced and ContentPageLayout
calculates the height of Content except the height of AppBar.
Consequently, OnRelayout() is no longer required because
ContentPageLayout calculates the size and position of Content.

### API Changes ###

 - ACR:


everLEEst (approved):
Looks good to me!
Finally this issue is solved! :+1:

PR #3806, opened 2021-12-14
[NUI] refactoring TextMapHelper

wonrst:
- remove duplicate code through function extraction for better readability.
- simplified the code to get the map value from placdeholder property using TextMapHelper
- reduce memory waste by calling Dispose() or using Add() from PropertyMapSetterHelper
- add test cases for newly added internal apis of TextMapHelper


JoogabYun (approved):

dongsug-song (approved):
looks good to me

PR #3811, opened 2021-12-15
[NUI] Fix cannot get Placeholder & InputFilter struct issue.

AchoWang:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/TextMapHelper.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/TextMapHelper.cs
@@ -230,15 +230,14 @@ namespace Tizen.NUI.BaseComponents
         {
             string accepted = "";
             string rejected = "";
-            map.Find(0)?.Get(out accepted);
-            map.Find(1)?.Get(out rejected);
+            map.GetValue(0)?.Get(out accepted);
+            map.GetValue(1)?.Get(out rejected);
 
             var inputFilter = new InputFilter();
             inputFilter.Accepted = accepted;
             inputFilter.Rejected = rejected;
 
             return inputFilter;
-
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/BaseComponents/TextMapHelper.cs
```diff
@@ -230,15 +230,14 @@ public static InputFilter GetInputFilterStruct(PropertyMap map)
         {
             string accepted = "";
             string rejected = "";
-            map.Find(0)?.Get(out accepted);
-            map.Find(1)?.Get(out rejected);
+            map.GetValue(0)?.Get(out accepted);
```
wonrst, line 233: Could you please share an example of the problem?

src/Tizen.NUI/src/public/BaseComponents/TextMapHelper.cs
```diff
@@ -530,14 +529,14 @@ public static Placeholder GetPlaceholderStruct(PropertyMap map)
...
 
-            map.Find(0)?.Get(out text);
-            map.Find(1)?.Get(out textFocused);
-            map.Find(2)?.Get(color);
-            map.Find(3)?.Get(out fontFamily);
-            map.Find(4)?.Get(fontStyle);
-            pointSizeValue = map.Find(5);
-            pixelSizeValue = map.Find(6);
-            map.Find(7)?.Get(out ellipsis);
+            map.Find(0, "text")?.Get(out text);
```
wonrst, line 532: Could you please share an example of the problem?

The placeholder's PropertyMap always has index 0.
So, in the case of this, the property value of index 0 is always returned.

src/Tizen.NUI/src/public/BaseComponents/TextMapHelper.cs
AchoWang: Hello, 
Please refer the case: "TextMapHelperGGetInputFilterStruct" ,
and the path is: TizenFX\test\Tizen.NUI.Tests\Tizen.NUI.Devel.Tests\testcase\public\BaseComponents\TSTextMapHelper.cs

In this case, we will got failed, the out string is empty.

src/Tizen.NUI/src/public/BaseComponents/TextMapHelper.cs
AchoWang: Hello, 
Same as above, please refer the case: "TextMapHelperGetPlaceholderStruct" ,
and the path is: TizenFX\test\Tizen.NUI.Tests\Tizen.NUI.Devel.Tests\testcase\public\BaseComponents\TSTextMapHelper.cs

In this case, we will got failed, the out string is empty too.

wonrst: Hello @AchoWang 
Thanks for reporting the problem.
I fixed the problematic tc.
Please refer PR.
https://github.com/Samsung/TizenFX/pull/3815

AchoWang: Hi @wonrst 
I notice some functions  in TextMapHelper.cs  like GetUnderlineStruct, GetShadowStruct, GetOutlineStruct and so on.
Is it better to consider unification?
Or what's difference between them?

wonrst: This depends on how the DALi toolkit's text component returns the PropertyMap.
Some are index base, some are string key base.
So each property is different.

It would be nice to unify for consistency, 
but this can make to different behavior. And this can causes compatibility issues.
I think it is better to keep it as it is.

AchoWang: OK, I got it!
Thanks very much!

PR #3605, opened 2021-09-23
[NUI] Add GLView

dkdk-ryu:
GLView allows drawing with OpenGL.
It creates a context, a surface and a render thread.
The render thread invokes user's callbacks.

### API Changes ###

 - ACR:

PR #3813, opened 2021-12-16
[NUI] Add GLView

dkdk-ryu:
This PR was supposed to merge into API9. But It is only in master.
So I'd like to apply this commit to API9.
GLView will be public in API10.

The original PR
https://github.com/Samsung/TizenFX/pull/3605


dongsug-song (approved):
GLView has been dropped when the API9 was branched out.
this is integration patch for adding GLView in API9 as hidden API.

PR #3815, opened 2021-12-16
[NUI] fix tc fail for TextMapHelper

wonrst:
Property of InputFilter and Placeholder returns PropertyMap of index base.
So GetHiddenInputStruct() and GetPlaceholderStruct() can parse the index base map.

### DIFF
--- a/test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/public/BaseComponents/TSTextMapHelper.cs
+++ b/test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/public/BaseComponents/TSTextMapHelper.cs
@@ -244,8 +244,8 @@ namespace Tizen.NUI.Devel.Tests
             tlog.Debug(tag, $"TextMapHelperGetInputFilterStruct START");
 
             var map = new PropertyMap();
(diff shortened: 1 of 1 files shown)


AchoWang (approved):
Looks good to me.

AchoWang:
Hi， @wonrst 

I think you should check the TCs continue,  such as  in case TextMapHelperGetTextFitMap, 

"Assert.AreEqual(fontSizeType, "PointSize", "Should be equal!");    "    , will got failed too, 

I check the code, the first character will be lower-cased ( why ?)

so the true result is pointSize , but not PointSize, 

Thanks!

AchoWang:
And case TextMapHelperGetTextFitStruct

Assert.AreEqual(FontSizeType.PixelSize, "PointSize", "Should be equal!");   =>  is not correct, 

ummm...
Anyway, please check them, thanks!

wonrst:
> And case TextMapHelperGetTextFitStruct
> 
> Assert.AreEqual(FontSizeType.PixelSize, "PointSize", "Should be equal!"); => is not correct,
> 
> ummm... Anyway, please check them, thanks!

The issue case has been fixed. Thanks!

dongsug-song (approved):
looks good to me

PR #3817, opened 2021-12-17
[NUI] Improves GetNextFocusableView

JoogabYun:
Currently, GetNextFocusableView of FocusableBase was Index-based of ContentContainer Child.

This is a problem.

Changes the search for the next focus view from index-based to position-based.

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #3819, opened 2021-12-17
[NUI] Fix StopFinal(EndAction) not working defect on lite model

dongsug-song:
### Description of Change ###
[NUI] Fix StopFinal(EndAction) not working defect on lite model

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/AnimatedVectorImageView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/AnimatedVectorImageView.cs
@@ -490,12 +490,9 @@ namespace Tizen.NUI.BaseComponents
                 switch (isMinMaxFrameSet)
                 {
                     case minMaxSetTypes.NotSetByUser:
-                        if (base.CurrentFrame != totalFrameNum - 1)
-                        {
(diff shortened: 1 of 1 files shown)


Seoyeon2Kim (commented):
I approved the same PR of API9, but could you explain this modification in detail?
Is it Okay to remove the conditional? 
if (base.CurrentFrame != totalFrameNum - 1)

JSUYA (approved):
LGTM

dongsug-song:
@Seoyeon2Kim 
@JSUYA 
thank you for reviewing.
the problems are 
1. occurs only on VD TV Lite model, because it is low tier model (low performance, so Lottie animation is blocked).
2. to implement 1., TV.FLUX calls as this sequences, to show "Play()-Stop(StopFinal)" to hide "Stop(StopFinal)".
3. when show() called, Play()-Stop() makes the Lottie not be played but be shown the last frame as still image.
4. sometimes, the frame is stuck at the last frame, so that the condition meets and setting "CurrentFrame = last frame" is skipped.

by removing the redundant condition, making "CurrentFrame = last frame" be called certainly and it solves the problem.

PR #3827, opened 2021-12-29
[NUI] Use ActionUpdateProperty for CornerRadius and Borderline

hinohie:
Due to the DALi's visual policy, we always create new visual
when some properties sensitive with visual's data.

Here, We make visual can change CornerRadius / Borderline
without destroy/create new visual.

### API Changes ###

 - ACR:


dongsug-song (commented):
looks good to me but other's review is also needed.
as I remember, there was some AppBar TCT fails related to similar changes of this.
@jaehyun0cho 
could you review and confirm if this change is not related to the past issue?
thank you.

dongsug-song (approved):
looks good to me

PR #3829, opened 2021-12-31
[NUI] Update ImageView properties lazy

hinohie:
Current Implementation create/destroy visual whenever properties changed.
This patch hold visual-sensitive properties and
create/destroy visual before next EventProcess called.

This patch required this dali code : https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/268723/
Also, this patch required this NUI code : https://github.com/Samsung/TizenFX/pull/3827

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
@@ -36,13 +36,14 @@ namespace Tizen.NUI.BaseComponents
         private EventHandler<ResourceLoadedEventArgs> _resourceLoadedEventHandler;
         private _resourceLoadedCallbackType _resourceLoadedCallback;
 
+        // Collection of image-sensitive properties.
+        private bool _imagePropertyUpdatedFlag = false;
+        private bool _imagePropertyUpdateProcessAttachedFlag = false;
+        private PropertyMap _imagePropertyMap;
         private Rectangle _border;
         private string _resourceUrl = "";
-        private bool _synchronousLoading = false;
-        private string _alphaMaskUrl = null;
         private int _desired_width = -1;
         private int _desired_height = -1;
-        private VisualFittingModeType _fittingMode = VisualFittingModeType.Fill;
         private TriggerableSelector<string> resourceUrlSelector;
         private TriggerableSelector<Rectangle> borderSelector;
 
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1218,81 +1196,116 @@ private void SetBorder(Rectangle value)
...
+                    _imagePropertyMap = new PropertyMap();
+                }
+                _imagePropertyUpdatedFlag = true;
+                _imagePropertyMap[key] = value;
+
+                // Lazy update only if _resourceUrl is not empty and ProcessAttachedFlag is false.
+                if (!string.IsNullOrEmpty(_resourceUrl) && !_imagePropertyUpdateProcessAttachedFlag)
+                {
+                    _imagePropertyUpdateProcessAttachedFlag = true;
+                    ProcessorController.Instance.ProcessorOnceEvent += UpdateImage;
```
dongsug-song, line 1258: Hello, 
I couldn't understand the benefit, so I am just asking. ^^;
for me, it looks like delaying the process of `UpdateImage`.
is this for making secure the order of creating-destroying visuals?
could you explain a little bit more?
thank you.

hinohie: Hello world.
Previous code always try to create new visual for every `UpdateImage` call.
If we set properties 5 times (ex : `ResourceUrl`, `SynchronousLoading`, `DesiredWidth`, `DesiredHeight`, and `CornerRadius`).
It create new visual for every property set call. So there will be called 5 constructor, and 4 destructor of visuals.
And more, every proerty setting time, it also try to get 'ALL' properties of visuals. So it will be called more than 5 CreateVisualMap() function, even we don't need to know other properties.

This patch try to collect properties that  we need to re-create the visuals,
and create visual at end of event thread.
This will be called 1 constructor, and no destructor of visuals.

Also, there is one more benefit. When we want to know just one value of property,
now we don't need to get 'ALL' properties of visuals.

For `CornerRadius` and `Borderline` are special case (they will change shader code.) so I make another PR about it (#3827)

dongsug-song (approved):
Thank you for your explanation.
looks good to me.

PR #3830, opened 2021-12-31
[NUI] text testcase improvement

wonrst:
- remove duplicate code from function extraction
- reduce loc & make better readability
- fix dispose warning CA2000 for prevent waste of memory


JoogabYun (approved):

dongsug-song (approved):
looks good to me

PR #3831, opened 2021-12-31
[NUI] Make SynchronousLoading use index key

hinohie:
Currently, View.cs 's BackgroundImageLoadingSynchronously use string key, instead of index key.

But all other properties about visuals use index key only.
(Instead of animatable properties)

This patch just sync up with standard frame

This patch required to merge this dali code : https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-toolkit/+/268756/


kimcinoo (approved):
No doubts. Nice work!

PR #3837, opened 2022-01-05
[WebRTC] Add new MediaFileSource API

hsgwon:
### Description of Change ###

Add new MediaFileSource API

### API Changes ###

 - TCSACR-479


src/Tizen.Multimedia.Remoting/WebRTC/MediaFileSource.cs
```diff
@@ -42,8 +40,7 @@ public sealed class MediaFileSource : MediaSource
         /// <param name="type">The <see cref="MediaType"/> of file source.</param>
         /// <param name="path">The file path.</param>
         /// <exception cref="ArgumentNullException"><paramref name="path"/> is null.</exception>
```
wiertel, line 42: what about other possible issues: file does not exist, lack of privileges, file contains bad data? are they reported somewhere else, or during playback?

src/Tizen.Multimedia.Remoting/WebRTC/MediaFileSource.cs
```diff
@@ -30,8 +29,7 @@ namespace Tizen.Multimedia.Remoting
     /// </remarks>
```
wiertel, line 29: As it is written, it looks like both privileges are necessary always. It usually depends on the storage used.
Consider changing like this at the beginning: "Depending on where the source file is located either the media storage privilege (http://tizen.org/privilege/mediastorage) is required or the external storage privilege(http://tizen.org/privilege/externalstorage) is required"

hsgwon: OK I'll change it.

src/Tizen.Multimedia.Remoting/WebRTC/MediaFileSource.cs
hsgwon: It will be check when file source is attached to WebRTC instance. - WebRTC.AddSource method

hsgwon:
The branch was diverged so, I'll close this and raise new PR.

PR #3843, opened 2022-01-10
[NUI] Support Layout for Multi-Window

jaehyun0cho:
Previously, Layouts in the default Window were supported.

Now, to support Layout in Multi-Window, Layouts in all Windows are supported.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
Looks good to me!
Thanks for supporting multi-window on Layout :)

PR #3851, opened 2022-01-12
[NUI] Fix bug when ResourceUrl is empty string

hinohie:
Previous code unregist visual when ResourceUrl use empty URL string.
After #3829, this action was changed.
So I fix this bug

This patch required below dali codes:
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-toolkit/+/269231/


dongsug-song (approved):
looks good to me

PR #3869, opened 2022-01-14
[NUI][AT-SPI] Fix accessibility activation for RecyclerViewItem

aswigon:
### Description of Change ###

This change fixes a bug where it was impossible to activate a `RecyclerViewItem` in Accessibility mode.

Because the implementation of `RecyclerViewItem.OnAccessibilityActivated` would look identical to `Button.OnAccessibilityActivated`, to avoid code duplication, accessibility activation is reimplemented in terms of `OnKeyboardEnter` for all controls, just like it is done in DALi.

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Button.Internal.cs
+++ b/src/Tizen.NUI.Components/Controls/Button.Internal.cs
@@ -452,23 +452,5 @@ namespace Tizen.NUI.Components
             ClickEvent?.Invoke(this, nestedEventArgs);
             Clicked?.Invoke(this, eventArgs);
         }
-
-        internal override bool OnAccessibilityActivated()
-        {
-            using (var key = new Key())
-            {
-                key.State = Key.StateType.Down;
-                key.KeyPressedName = "Return";
-
-                // Touch Down
-                OnKey(key);
-
-                // Touch Up
-                key.State = Key.StateType.Up;
-                OnKey(key);
-            }
-
-            return true;
-        }
     }
 }
--- a/src/Tizen.NUI.Components/Controls/Button.cs
+++ b/src/Tizen.NUI.Components/Controls/Button.cs
@@ -855,6 +855,8 @@ namespace Tizen.NUI.Components
         /// <since_tizen> 6 </since_tizen>
         public override bool OnKey(Key key)
         {
+            bool clicked = false;
+
             if (!IsEnabled || null == key)
             {
                 return false;
@@ -872,7 +874,7 @@ namespace Tizen.NUI.Components
             {
                 if (key.KeyPressedName == "Return")
                 {
-                    bool clicked = isPressed && IsEnabled;
+                    clicked = isPressed && IsEnabled;
 
                     isPressed = false;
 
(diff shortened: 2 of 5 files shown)


lol-github (approved):
Very nice! LGTM for me

everLEEst (commented):
Thank you for changes.
Looks good to me but I'm not sure those components are all enough to implements OnKey().
let me check again and reply.

everLEEst (commented):
One more thing to check is backward compatibility issue.
VD has it's own components named FLUX and I'm not sure their components are driven from which parent class.

everLEEst:
> One more thing to check is backward compatibility issue. VD has it's own components named FLUX and I'm not sure their components are driven from which parent class.

they using View so it seems not an issue for VD.

everLEEst:
> Thank you for changes. Looks good to me but I'm not sure those components are all enough to implements OnKey(). let me check again and reply.

there are few more components who implements OnKey() such as Control, RadioButton, Slider, MenuItem...... please check all of them are safe.

Seoyeon2Kim (commented):
I checked that this PR can fix to activate items in the list when Accessibility turns on.
However, as Sanghyun mentioned, I've been worried that virtual function of CustomView would be changed.

m-bialota:
This patch does not cause regression in functional tests.

aswigon:
Here is a summary of components that override `OnKey`:
* `Button` - see below;
* `Control` - does nothing, returns `false`;
* `MenuItem` - derived from `Button`;
* `RadioButton` - derived from `Button`;
* `RecyclerViewItem` - fixed by this patch, because its implementation of `OnKey("Return")` works correctly (and the old override for `OnAccessibilityActivated` did not);
* `SelectButton` - derived from `Button`;
* `Slider` - returns `false` (ignores keys other than `"Left"`, `"Right"`, `"Down"` or `"Up"`);
* `Switch` - derived from `Button`;
* `TabButton` - derived from `Button`;

`Button.OnAccessibilityActivated` already called `Button.OnKey`, this code is now moved up in the hierarchy to `Control`. `OnKey` for `Button`-derived classes always calls `base.OnKey` (to determine the return value), so there is no change of behaviour for `Button` and derived classes.

`Dali::Toolkit::Control` has the same default implementation of `OnAccessibilityActivated` (which calls `OnKeyboardEnter`), and there are no known overriders. With this design, there is a greater chance that Accessibility activation will automatically work for custom controls, because custom controls developers only have to implement keyboard activation `OnKey("Return")` and do not have to think about (or be aware of) accessibility.

everLEEst (approved):
Thank you for the checking any possible issues.
I think it is ready to go then :)

Seoyeon2Kim (approved):
> With this design, there is a greater chance that Accessibility activation will automatically work for custom controls, because custom controls developers only have to implement keyboard activation OnKey("Return") (or OnKeyboardEnter) and do not have to think about (or be aware of) accessibility.

Make sense!
Thanks for checking and fixing the behavior.

dongsug-song (approved):
looks good to me

PR #3877, opened 2022-01-17
[NUI] Support Layout for Multi-Window

jaehyun0cho:
Previously, Layouts in the default Window were supported.

Now, Layouts in all Windows are supported.

To enable Multi-Window, Window.IsSupportedMultiWindow() should be true
and the following code should not be executed.

Information.TryGetValue("http://tizen.org/feature/opengles.surfaceless_context", out isSupported);

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM

Seoyeon2Kim (approved):
LGTM!
Theoretically, these changes look okay to me, but there might be another exception later. Hope that doesn't happen ;)

JoogabYun (approved):

dongsug-song (approved):
looks good to me

PR #3879, opened 2022-01-18
[NUI] Add IsWindowRotating()

sparrow74:
Add IsWindowRotationg function.
It is return with true when window is rotating.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.Window.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.Window.cs
@@ -265,6 +265,10 @@ namespace Tizen.NUI
 
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Window_ExcludeInputRegion")]
             public static extern void ExcludeInputRegion(global::System.Runtime.InteropServices.HandleRef window, global::System.Runtime.InteropServices.HandleRef inputRegion);
+
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Window_IsWindowRotating")]
+            [return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.U1)]
+            public static extern bool IsWindowRotating(global::System.Runtime.InteropServices.HandleRef window);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/internal/Interop/Interop.Window.cs
```diff
@@ -265,6 +265,9 @@ internal static partial class Window
 
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Window_ExcludeInputRegion")]
             public static extern void ExcludeInputRegion(global::System.Runtime.InteropServices.HandleRef window, global::System.Runtime.InteropServices.HandleRef inputRegion);
+
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Window_IsWindowRotating")]
+            public static extern bool IsWindowRotating(global::System.Runtime.InteropServices.HandleRef window);
```
dongsug-song, line 270: could you add this here?
`            [return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.U1)]`

this is for `bool` return type.
- reference : https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/internal/Interop/Interop.Actor.cs#L55

test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/WindowTest.cs
```diff
@@ -41,11 +43,33 @@ void Initialize()
...
+            tm = new Timer(100);
+            tm.Tick += Tm_Tick;
+            tm.Start();
+        }
+
+        private bool Tm_Tick(object source, Timer.TickEventArgs e)
+        {
+            bool rotating = mainWin.IsWindowRotating();
+            log.Fatal(tag, $"window is Rotating: {rotating}");
+            log.Fatal(tag, "window is Rotating: {}"+rotating);
```
hinohie, line 71: I think we dont need to reply this informations twice.

src/Tizen.NUI/src/internal/Interop/Interop.Window.cs
sparrow74: I see, I will update with your comment.

test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/WindowTest.cs
sparrow74: Ah..Thank you for detail review :)

dongsug-song (approved):
looks good to me

PR #3885, opened 2022-01-20
[NUI] Focus movement follows the GetNextFocusableView() method of ScrollableBase

JoogabYun:
### Description of Change ###

Focus movement follows the GetNextFocusableView() method of ScrollableBase

refer : https://github.com/Samsung/TizenFX/pull/3814


everLEEst (approved):
ScrollableBase also an container which moves object to object.. I agreed the point of this patch.
thank you.

dongsug-song (approved):
looks good to me

Seoyeon2Kim (approved):
So, it means CollectionView doesn't need to override GetNextFocusableView method by itself? It can just follow the ScrollableBase behavior.
LGTM

PR #3890, opened 2022-01-20
[Connection] Do not use Lazy for Singleton

chleun-moon:
### Description of Change ###
When Lazy is used, if an exception occurs when a singleton object (ConnectionInternalManager) is created, the exception is cached and thrown continuously.
Same as https://github.com/Samsung/TizenFX/pull/3888


dh79pyun (approved):
Looks good

PR #3893, opened 2022-01-21
[NUI] Get native image PropertyMap instead of cached.

hinohie:
When we try to get cached property map for animated image,
It will not return internal-updated values
something like "CurrentFrame" for animated image.

This patch make we return Current PropertyMap always
instead of cached property map.

Actually, the best way to fix is, create some flags when we
try to get time-sensitive values.
But it need to fixup some codes and test. So just make hot-fix.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/AnimatedImageView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/AnimatedImageView.cs
@@ -295,7 +295,7 @@ namespace Tizen.NUI.BaseComponents
             get
             {
                 int ret = -1;
-                PropertyMap map = Image;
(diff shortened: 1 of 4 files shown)


everLEEst (approved):
Looks good to me.
please check other TCT are not influenced

dongsug-song (approved):
looks good to me

hinohie:
> 

Hi. I checked that TCT relative with this patches are all passed.

Seoyeon2Kim (approved):
LGTM, but could you get a review from Sunghyun who is in charge of ImageView once?

tscholb (approved):
LGTM :)

PR #3894, opened 2022-01-21
[NUI][ATSPI] remove duplicated calculation for states

kimcinoo:
The HIGHLIHGTED state is calculated on parent class ControlAccessible
The get of IsHighlighted is doing exactly same calculation.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
@@ -636,14 +636,9 @@ namespace Tizen.NUI.BaseComponents
         {
             AccessibilityStates accessibilityStates = (AccessibilityStates)states;
 
-            FlagSetter(ref accessibilityStates, AccessibilityStates.Highlightable, this.AccessibilityHighlightable);
-            FlagSetter(ref accessibilityStates, AccessibilityStates.Focusable, this.Focusable);
             FlagSetter(ref accessibilityStates, AccessibilityStates.Focused, this.State == States.Focused);
-            FlagSetter(ref accessibilityStates, AccessibilityStates.Highlighted, this.IsHighlighted);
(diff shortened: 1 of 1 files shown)


everLEEst (approved):
LGTM

src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
```diff
@@ -639,7 +639,6 @@ protected virtual AccessibilityStates AccessibilityCalculateStates(ulong states)
             FlagSetter(ref accessibilityStates, AccessibilityStates.Highlightable, this.AccessibilityHighlightable);
             FlagSetter(ref accessibilityStates, AccessibilityStates.Focusable, this.Focusable);
             FlagSetter(ref accessibilityStates, AccessibilityStates.Focused, this.State == States.Focused);
-            FlagSetter(ref accessibilityStates, AccessibilityStates.Highlighted, this.IsHighlighted);
```
lol-github, line 642: Is "Highlighted" the only state which calculation is duplicated? AFAIK, we should override here only these states which value depends on NUI-specific properties.

kimcinoo: Found HIGHLIGHTABLE and FOCUSABLE are duplicated.

aswigon: HIGHLIGHTED is also duplicated, VISIBLE should also probably be removed from here, because DALi also calculates it (and does it better than always `true`),  DEFUNCT is also duplicated.

kimcinoo: Yes correct. I will updated. Thank you.

m-bialota:
Verified: no regression in functional tests.

Seoyeon2Kim (approved):
LGTM

PR #3896, opened 2022-01-21
[NUI] Fix duplicated unparent issue at Layer

hinohie:
View have some safe-guard when we try to child.Unparent() && Remove(child) dual times.
(child will have null parent after one of them.)

But Layer.Remove didn't have any defence code about that situation.

Tizen.NUI.Components.Menu.Dispose call layer.Remove(this) in Dispose function.
But before Dispose function call, we already called Unparented function somewhere.

This patch will guard that case,
and also guard some null & reference issue.

### DIFF
--- a/src/Tizen.NUI/src/public/Common/Layer.cs
+++ b/src/Tizen.NUI/src/public/Common/Layer.cs
@@ -1,5 +1,5 @@
 /*
- * Copyright(c) 2019 Samsung Electronics Co., Ltd.
+ * Copyright(c) 2022 Samsung Electronics Co., Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -340,6 +340,15 @@ namespace Tizen.NUI
             if (null == child)
             {
                 throw new ArgumentNullException(nameof(child));
+            }
+            if (child.GetParent() == null) // Early out if child parent is null.
+                return;
+
+            if (child.GetParent() != this)
+            {
+                //throw new System.InvalidOperationException("You have deleted a view that is not a child of this layer.");
+                Tizen.Log.Error("NUI", "You have deleted a view that is not a child of this layer.");
+                return;
             }
             Interop.Actor.Remove(SwigCPtr, View.getCPtr(child));
             if (NDalicPINVOKE.SWIGPendingException.Pending)


src/Tizen.NUI/src/public/Common/Layer.cs
```diff
@@ -341,6 +341,14 @@ public override void Remove(View child)
             {
                 throw new ArgumentNullException(nameof(child));
             }
+            if (child.GetParent() == null) // Early out if child parent is null.
+                return;
+
+            if (child.GetParent() != this)
+            {
+                //throw new System.InvalidOperationException("You have deleted a view that is not a child of this view.");
+                Tizen.Log.Error("NUI", "You have deleted a view that is not a child of this view.");
```
dongsug-song, line 350: please change this => `not a child of this view` to `not a child of this layer`

hinohie: Thanks for detail review.

dongsug-song (approved):
looks good to me

src/Tizen.NUI/src/public/Common/Layer.cs
```diff
@@ -341,6 +341,14 @@ public override void Remove(View child)
             {
                 throw new ArgumentNullException(nameof(child));
             }
+            if (child.GetParent() == null) // Early out if child parent is null.
+                return;
+
+            if (child.GetParent() != this)
+            {
+                //throw new System.InvalidOperationException("You have deleted a view that is not a child of this layer.");
+                Tizen.Log.Error("NUI", "You have deleted a view that is not a child of this layer.");
```
jaehyun0cho, line 350: I think this should return at this point after printing Error log.
Because if child's parent is not this, then child cannot exist in Children.
Moreover, I think that the following codes should not be called if child is not a child object.
```
            child.InternalParent = null;
            LayoutCount -= child.LayoutCount;
```

jaehyun0cho (commented):
Please check the comment.
Please fix the typo "gaurd" in description of this PR.

src/Tizen.NUI/src/public/Common/Layer.cs
hinohie: I also catch that points.
But it is not a kind of TCT fix-up patches So I had just plan to do that jobs with another PR.

I saw your PR. it looks god to me, So I also fix it at this PR.

jaehyun0cho (approved):
LGTM

PR #3904, opened 2022-01-24
[NUI][AT-SPI] Clean up APIs related to AccessibilityStates

aswigon:
### Description of Change ###

This PR:
1. Restores the division into `enum AccessibilityState` (enumeration of possible states) and `class AccessibilityStates` (a collection of states), which was wrongly removed. DALi makes the same distinction between `Accessibility::State` and `Accessibility::States`.

2. Restores the AT-SPI standard-compliant values for  `AccessibilityState` enumeration (the same values as in the `at-spi2-core` library and in DALi `accessibility.h`).

3. Adds a convenient indexing syntax (e.g. `states[AccessibilityState.Modal] = true`) for manipulating the states collection without having to perform bitwise operations like `&` and `|` (because using a high level toolkit like `NUI` should not require performing bit operations) or using the cumbersome and roundabout way via `FlagSetter` (e.g. `FlagSetter(ref states, AccessibilityState.Modal, true)`, which is probably inefficient due to the use of generic and dynamic types).

4. Removes `FlagSetter`.

5. Removes `View.DuplicateStates` and returns the bitmask directly from `AccessibilityDelegate.CalculateStates`, to avoid an unnecessary memory allocation.

6. Removes the recently added parameter for `AccessibilityCalculateStates`, because it causes confusion, as this method shouldn't take any arguments (just like `CalculateStates` in DALi). It should only get parent states from `base` and add (or remove) its own states to this set (same as in DALi). The DALi-calculated states are rerouted through a new non-virtual method `AccessibilityCalculateStatesWrapper`.

7. Fixes `EmitAccessibilityStateChangedEvent` to take a single `AccessibilityState` as a parameter (the same as in DALi). `NotifyAccessibilityStatesChange` still takes an `AccessibilityStates` collection as a parameter (the same as in DALi), because it is equivalent to calling `EmitAccessibilityStateChangedEvent` in a loop for all specified states.

8. Updates some `Interop` symbol names. See the commit message for the dali-csharp-binder patch for more information on these changes.

Complementary dali-csharp-binder patch (merge at the same time):
* https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/270040/

The dali-csharp-binder also patch depends on the following dali-adaptor patch:
* https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-adaptor/+/270102/

It is recommended to merge the dali-adaptor patch first, and after that, merge the dali-csharp-binder patch and this PR together.

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
@@ -329,14 +329,20 @@ namespace Tizen.NUI.BaseComponents
         /// Notifies sending notifications about the current states to accessibility clients.
         /// </summary>
         /// <remarks>
-        /// If recursive is true, all children of the Accessibility object will also re-emit the states.
+        /// In essence, this is equivalent to calling EmitAccessibilityStateChangedEvent in a loop for all specified states.
+        /// If recursive mode is specified, all children of the Accessibility object will also re-emit the states.
         /// </remarks>
         /// <param name="states">Accessibility States</param>
         /// <param name="notifyMode">Controls the notification strategy</param>
         [EditorBrowsable(EditorBrowsableState.Never)]
         public void NotifyAccessibilityStatesChange(AccessibilityStates states, AccessibilityStatesNotifyMode notifyMode)
         {
-            Interop.ControlDevel.DaliToolkitDevelControlNotifyAccessibilityStatesChange(SwigCPtr, (ulong)states, (int)notifyMode);
+            if (states is null)
+            {
+                throw new ArgumentNullException(nameof(states));
+            }
+
+            Interop.ControlDevel.DaliToolkitDevelControlNotifyAccessibilityStateChange(SwigCPtr, states.BitMask, (int)notifyMode);
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
         }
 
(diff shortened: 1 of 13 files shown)


lol-github (approved):
Very good job! This clarifies a lot the topic of handling accessibility states. Thanks for doing this before the API has been ultimately frozen.

src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
```diff
@@ -451,14 +457,21 @@ private IntPtr DuplicateString(string value)
...
+            return Interop.ControlDevel.DaliAccessibilityNewRange(range.StartOffset, range.EndOffset, range.Content);
         }
 
-        private IntPtr DuplicateRange(AccessibilityRange range)
+        private static AccessibilityStates AccessibilityInitialStates = new AccessibilityStates();
+
+        private ulong AccessibilityCalculateStatesWrapper(ulong initialStates)
         {
-            return Interop.ControlDevel.DaliAccessibilityNewRange(range.StartOffset, range.EndOffset, range.Content);
+            lock (AccessibilityInitialStates)
```
kimcinoo, line 469: Why do we need 'lock' here?

kimcinoo (commented):
Looks good to me. But I have one question.

src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
aswigon: I made `AccessibilityInitialStates` `static` in order not to use up extra memory per every `View` instance. The AT-SPI bridge in dali-adaptor calls all `Accessible` methods sequentially from a single thread, so the code would work just as well without the `lock`. However, `AccessibilityInitialStates` still technically is a shared resource, and there is a possible race condition with someone else calling `GetAccessibilityStates` from a different thread.

kimcinoo: OK. Cool. There is a try to make separated UI thread as a this year project.

kimcinoo (approved):

dongsug-song (approved):
looks good to me

Seoyeon2Kim (approved):
LGTM! Thanks for your effort!

PR #3906, opened 2022-01-25
[NUI] Rotary Event

JoogabYun:
It can receive Rotary Event on the focused View.
If not consumed, the event is propagated to the parent view.

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/BaseComponents/ViewEvent.cs
```diff
@@ -1368,18 +1368,43 @@ public ControlStateChangedEventArgs(ControlState previousState, ControlState cur
             public ControlState CurrentState { get; }
         }
 
+        /// <summary>
+        /// Handle CustomWheel event type.
+        /// </summary>
+        internal void FeedCustomWheelEvent(object sender, WheelEventArgs arg)
+        {
+            View view = this.GetParent() as View;
```
jaehyun0cho, line 1376: Maybe we can shorten this code as follows.
```
if ((WindowWheelEventHandler == null) || (WindowWheelEventHandler(this, arg) == false))
{
    // Recursively deliver events to its parents, until the event is consumed or the parent is null.
    (GetParent() as View)?.FeedCustomWheelEvent(this, arg);
}
```

PR #3914, opened 2022-01-27
[NUI][ATSPI] Set AccessibilityHidden for DefaultLinearItem Seperator

kimcinoo:
The Seperator of DefaultLinearItem would be not necessary to AT-Client.
This patch is removing Seperator accessible nodes from A11Y tree.

### DIFF
--- a/src/Tizen.NUI.Components/Controls/RecyclerView/Item/DefaultLinearItem.cs
+++ b/src/Tizen.NUI.Components/Controls/RecyclerView/Item/DefaultLinearItem.cs
@@ -150,8 +150,6 @@ namespace Tizen.NUI.Components
                         layoutChanged = true;
                         Add(itemLabel);
                     }
-                    itemLabel.AppendAccessibilityRelation(this, AccessibilityRelationType.ControlledBy);
-                    this.AppendAccessibilityRelation(itemLabel, AccessibilityRelationType.LabelledBy);
                 }
                 return itemLabel;
             }
@@ -305,7 +303,8 @@ namespace Tizen.NUI.Components
                         //need to consider horizontal/vertical!
                         WidthSpecification = LayoutParamPolicies.MatchParent,
                         HeightSpecification = 2,
-                        ExcludeLayouting = true
+                        ExcludeLayouting = true,
+                        AccessibilityHidden = true
                     };
                     layoutChanged = true;
                     Add(itemSeperator);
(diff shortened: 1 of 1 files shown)


lol-github (approved):
LGTM

everLEEst:
could you please rebase the patch?

kimcinoo:
> could you please rebase the patch?
rebased.




> How about setting `AccessibilityHidden = true` also inside `DefaultLinearItem.CreateLabel`? Then, instead of `AppendAccessibilityRelation(.LabelledBy)` pointing to the label, we could have something like:
> 
> ```cs
> protected override string AccessibilityGetName() => Label.Text;
> protected override string AccessibilityGetDescription() => SubLabel.Text;
> ```
> 
> which would give an even more compact AT-SPI tree.

IMHO the "AccessibilityHidden" for the Label could make a problem on the Aurum side if it does not use "LabelledBy"

kimcinoo:
> > IMHO the "AccessibilityHidden" for the Label could make a problem on the Aurum side if it does not use "LabelledBy"
> 
> We do the same thing for `Button` (instead of `AccessibilityHidden` we use `ShouldReportZeroChildren() => true` but that has the same effect of hiding `Button`'s label from the AT-SPI tree). And instead of creating a `LabelledBy` relation, we override `Button.AccessibilityGetName` to provide the label's text directly. I had the same idea in mind for `DefaultLinearItem`.

I got the point. Please review the updates.

aswigon (approved):

Seoyeon2Kim (approved):
LGTM, but could you get a review from @everLEEst ?

lol-github (approved):
LGTM

src/Tizen.NUI.Components/Controls/RecyclerView/Item/DefaultLinearItem.cs
```diff
@@ -360,6 +359,15 @@ public override void OnRelayout(Vector2 size, RelayoutContainer container)
             }
         }
 
+        /// <summary>
+        /// Get accessibility name.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected override string AccessibilityGetName()
+        {
+            return Label.Text + " " + SubLabel.Text;
```
m-bialota, line 368: Should it be space between label and sublabel? I noticed that in EFL list items main and additional texts are separated by  comma.

m-bialota: Apart from this question I have no doubts towards this patch. I see no regression in functional tests.

kimcinoo: > Should it be space between label and sublabel? I noticed that in EFL list items main and additional texts are separated by comma.

Good catch. Thank you. It has been updated.

everLEEst (approved):
Looks good to me!

PR #3916, opened 2022-01-28
[NUI] Catch DllNotFoundException in IsSupportedMultiWindow() for desktop

jaehyun0cho:
To support multi-window, creating a new Window calls
Window.IsSupportedMultiWindow().

Since libcapi-system-info.so.0 is not installed by default on desktop
environment, calling Window.IsSupportedMultiWindow() causes crash by
DllNotFoundException.

Not to cause crash by calling Window.IsSupportedMultiWindow() on
desktop, DllNotFoundException is caught.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window/Window.cs
+++ b/src/Tizen.NUI/src/public/Window/Window.cs
@@ -43,7 +43,14 @@ namespace Tizen.NUI
         static internal bool IsSupportedMultiWindow()
         {
             bool isSupported = false;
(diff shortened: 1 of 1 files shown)


kimcinoo (approved):
Looks valid and good. No doubts!

dongsug-song (approved):
looks good to me

Seoyeon2Kim (approved):
LGTM

tscholb (approved):
LGTM :)

PR #3923, opened 2022-02-03
[NUI] Fix svace issue of RecyclerView LinearLayouter

hinohie:
Check if groupInfo is null. It can be null when count = 0.


everLEEst (approved):
Looks good to me.

Seoyeon2Kim (approved):
LGTM :) To fix svace issue!

PR #3926, opened 2022-02-07
[NUI] Fix FlexLayout size calculation with Padding

jaehyun0cho:
Previously, if FlexLayout had Padding, then child's and grand child's sizes
were calculated incorrectly.
Because in FlexLayout's OnLayout(), child's size was given with applying
Padding. But Padding was applied in FlexLayout's OnLayout() again.
As a result, child's size became smaller by Padding size than it had to be.

Now, if FlexLayout has Padding, then child's and grand child's sizes are
calculated correctly.
The Padding is not applied duplicately while child's size is calculated.

### API Changes ###

 - ACR:


everLEEst (approved):
Loosk goos to me!

dongsug-song (approved):
looks good to me

PR #3945, opened 2022-02-11
[NUI] Fix ArgumentOutOfRangeException when RemoveChild called.

AchoWang:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
@@ -1032,7 +1032,6 @@ namespace Tizen.NUI.BaseComponents
 
             Children.Remove(child);
             child.InternalParent = null;
-            LayoutCount -= child.LayoutCount;
 
             RemoveChildBindableObject(child);
 


src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
```diff
@@ -1032,7 +1032,6 @@ internal void RemoveChild(View child)
 
             Children.Remove(child);
             child.InternalParent = null;
```
dongsug-song, line 1034: is this OK to remove minus count?
what if the parent's `LayoutCount` becomes wrong and `LayoutCount` is used in other place.
is there no problem if the `LayoutCount` is not updated and has wrong value when child is removed?

dongsug-song:
@AchoWang 
could you write the description and could you add sample App or test case when committing a patch?

src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
AchoWang: Hi， Mr. Ssong, 
Thanks for your mention, I checked the sample again, I found the sample which use the method incorrectly, It's my mistake,  so I'll close the pr.

PR #3954, opened 2022-02-16
[NUI] Add an API for evaluating js with callback in WebView.

huayongxu:
### Description of Change ###

Add EvaluateJavaScript method on WebView class

It evaluate javascript code with web engine and asynchronously returning result as string through callback.

#### how to use

``` c#
WebView webview = new WebView();
...
webview.EvaluateJavaScript("2+3", (result) => Console.WriteLine($"Eval : {result}"));
```

### API Changes ###

Add :
``` c#
class WebView {
      public void EvaluateJavaScript(string script, JavaScriptMessageHandler handler)
}
```
 - ACR:


dongsug-song (approved):
looks good to me

PR #3965, opened 2022-02-17
[NUI][AT-SPI] Remove interops for SetAccessibilityConstructor()

aswigon:
### Description of Change ###

The method name `SetAccessibilityConstructor()` in NUI has always been misleading, because it actually never had an `accessibilityConstructor` parameter in the first place. With the recent addition of bindings for `Accessible::GetInterfaces()`, all NUI controls use the same `Accessible` implementation (`NUIViewAccessible`) anyway. However, as this method is currently used in many places, it is kept as a wrapper for setting `AccessibilityRole` (and interface flags) and will be removed if/when we decide to switch to AT-SPI interfaces as native C# interfaces. Then, all usages of `SetAccessibilityConstructor()` would be replaced by simply setting `AccessibilityRole` (and interface flags won't be needed).

Merge together with:
* https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/272010/

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
@@ -465,8 +465,7 @@ namespace Tizen.NUI.BaseComponents
         {
             // We have to store the interface flags until we remove SetAccessibilityConstructor and switch to native C# interfaces
             AtspiInterfaceFlags = (1U << (int)accessibilityInterface);
-            Interop.ControlDevel.DaliToolkitDevelControlSetAccessibilityConstructor(SwigCPtr, (int)role);
-            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+            AccessibilityRole = role;
         }
 
         [EditorBrowsable(EditorBrowsableState.Never)]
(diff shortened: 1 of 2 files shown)


lol-github (approved):
LGTM

src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
```diff
@@ -465,8 +465,7 @@ public void SetAccessibilityConstructor(Role role, AccessibilityInterface access
         {
```
kimcinoo, line 465: there is a commit message "will be removed if/when we decide to switch to AT-SPI interfaces as native C# interfaces. Then, all usages of SetAccessibilityConstructor() would be replaced by simply setting AccessibilityRole (and interface flags won't be needed)."
So this SetAccessibilityConstructor will be chaged to SetAccessibilityInterfaceAndRole or something?

kimcinoo (approved):

src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
aswigon: We intend to remove `SetAccessibilityConstructor` entirely! Please refer to this draft with the intended API that we want to achieve: https://github.com/Samsung/TizenFX/pull/4022.

kimcinoo: I got it. ```AccessibilityRole = XYZ``` will work instead of ```SetAccessibilityConstructor```

Seoyeon2Kim (approved):
Due to this PR, it would be much easier to set its role for one View.

dongsug-song:
already more than 2 reviews have been approved.
so it is merged now.

PR #3974, opened 2022-02-23
[NUI][Test] xaml indexer sample

dongsug-song:
### Description of Change ###

[NUI][Test] xaml indexer sample
- this is just a sample code to test xaml indexer by using markup extension

### API Changes ###
none


everLEEst (approved):
Looks good to me,
but I think it would better to have more samples how to use it.

Seoyeon2Kim (approved):

JoogabYun (approved):

PR #3976, opened 2022-02-23
[NUI] Do not set value of property again when theme is changed.

huayongxu:
Originally although property value of a view is set, if theme is
changed, the value would be restored to the default value.
This is not expected.
This patch is to check if the value has been set already or not,
if set, do not set the value again.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyle.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyle.cs
@@ -565,7 +565,7 @@ namespace Tizen.NUI.BaseComponents
 
                 if (destinationProperty != null)
                 {
-                    SetValue(destinationProperty, sourceValue);
+                    InternalSetValue(destinationProperty, sourceValue);
                 }
             }
 
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -47,6 +47,7 @@ namespace Tizen.NUI.BaseComponents
         private LayoutTransition layoutTransition;
         private TransitionOptions transitionOptions = null;
         private ThemeData themeData;
+        private bool isThemeChanged = false;
 
         // List of constraints
         private Constraint widthConstraint = null;
@@ -3032,8 +3033,10 @@ namespace Tizen.NUI.BaseComponents
         [EditorBrowsable(EditorBrowsableState.Never)]
         protected virtual void OnThemeChanged(object sender, ThemeChangedEventArgs e)
         {
+            isThemeChanged = true;
             if (string.IsNullOrEmpty(styleName)) ApplyStyle(ThemeManager.GetUpdateStyleWithoutClone(GetType()));
             else ApplyStyle(ThemeManager.GetUpdateStyleWithoutClone(styleName));
+            isThemeChanged = false;
         }
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


everLEEst (commented):
could you please add some test for it?
looks good to me the patch though.

myroot:
`SettedPropeties` is also updated when property value was set by Style, so It can't distinguish value was updated by user or style

huayongxu:
> `SettedPropeties` is also updated when property value was set by Style, so It can't distinguish value was updated by user or style

what is it meant by 'property value was set by Style'?
is it 'public virtual void ApplyStyle(ViewStyle viewStyle)' ?
If ApplyStyle is called, it is the same as that when user sets property value.

in this patch, only when theme is changed, the method
protected virtual void OnThemeChanged(object sender, ThemeChangedEventArgs e) 
is called, property value would not be set again.

huayongxu:
> could you please add some test for it? looks good to me the patch though.

Done.

myroot:
>in this patch, only when theme is changed, the method
protected virtual void OnThemeChanged(object sender, ThemeChangedEventArgs e)
is called, property value would not be set again.

Did you know a point of issue that I addressed on github.sec?

1. A view such as Button has a default Style and it's properties was updated by ViewStyle
2. And then, User also set a property, 
3. Theme was changed, a new style is updated to the view
4. in this case, All properties were updated by it even though it was updated by user

Your solution is checking the `SettedPropeties`, it contains properties that value was updated. 

return to my comment, 
When OnThemeChanged was called, 
1. some properties were updated by ApplyStyle
2. and some properties were updated by User (view.XXX = value)
but, both of cases, `SetValue` was used to update a value
and it update `SettedPropeties` so two cases properties are contained to `SettedProperties'

So, as I think, all properties are not updated when OnThemeChanged  because all properties contains on `SettedPropeties`

Could you tell me anythings that i missed?

myroot:
> > could you please add some test for it? looks good to me the patch though.
> 
> Done.

Could you share an animated screen shot like I posted on the issue board? to verify issue was resolved

huayongxu:
> > in this patch, only when theme is changed, the method
> > protected virtual void OnThemeChanged(object sender, ThemeChangedEventArgs e)
> > is called, property value would not be set again.
> 
> Did you know a point of issue that I addressed on github.sec?
> 
> 1. A view such as Button has a default Style and it's properties was updated by ViewStyle
> 2. And then, User also set a property,
> 3. Theme was changed, a new style is updated to the view
> 4. in this case, All properties were updated by it even though it was updated by user
> 
> Your solution is checking the `SettedPropeties`, it contains properties that value was updated.
> 
> return to my comment, When OnThemeChanged was called,
> 
> 1. some properties were updated by ApplyStyle
> 2. and some properties were updated by User (view.XXX = value)
>    but, both of cases, `SetValue` was used to update a value
>    and it update `SettedPropeties` so two cases properties are contained to `SettedProperties'
> 
> So, as I think, all properties are not updated when OnThemeChanged because all properties contains on `SettedPropeties`
> 
> Could you tell me anythings that i missed?

you mean, only when a property is set to a value by user, not ApplyStyle by user, 
the property value would not be changed again. is it right?

In the issue https://github.sec.samsung.net/dotnet/NUIBackend/issues/97,
it is said that only theme is changed, property values would not be changed again.
So ApplyStyle is the same as that case where a property is set to a value.

huayongxu:
> 

In the case,  

1) A value is set to a property by user (view.xxx = xxxx)
2) The property is updated by ApplyStyle by user,  --> Here the property could not be updated. This is an issue.
3) Theme is changed,

myroot:
You seem to have misunderstood the problem. 

1. `ApplyStyle` was called on constructor with default style from defined on theme
2. User update a property, 
3. Theme was change

except result
 Properties are updated with a value on theme except changed by 2

Actual result
 All properties are updated with a value on theme including changed by 2

result of this changes
 All properties (updated by 1 and 2) are not changed

huayongxu:
> > > could you please add some test for it? looks good to me the patch though.
> > 
> > 
> > Done.
> 
> Could you share an animated screen shot like I posted on the issue board? to verify issue was resolved
theme-change

zg2nets (approved):
looks good~

dongsug-song (approved):
looks good to me

hinohie:
Oh.. we finally fix it! Looks God To Me.

hinohie (approved):
Oh.. we finally fix it! Looks God To Me.

JoogabYun (approved):

PR #3983, opened 2022-02-24
[NUI] Save the cropMask value internally in ImageView

tscholb:
Regardless of the cropToMask setting order, the value has been modified to apply properly.
In the existing code, the appclication  set the cropToMask first and setting url later,
there was an issue that the set cropToMask value was not applied properly.
To solve this issue, it has been modified to save this value at the ImageView.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
@@ -268,6 +268,7 @@ namespace Tizen.NUI.BaseComponents
         private string _alphaMaskUrl = null;
         private int _desired_width = -1;
         private int _desired_height = -1;
+         private bool _cropToMask = false;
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1349,6 +1352,13 @@ private void UpdateImage(int key, PropertyValue value)
                 alphaMaskUrl?.Dispose();
             }
 
+            if(key != ImageVisualProperty.CropToMask)
+            {
+                PropertyValue cropToMask = new PropertyValue(_cropToMask);
```
everLEEst, line 1357: what about change this to "using()" ?

everLEEst (approved):
I heard this code only need  for api9,
so no need to maintaining good code.
so then it looks okay to me :)

PR #3984, opened 2022-02-24
[NUI][AT-SPI] make same event usage for Enabled, Disabled

kimcinoo:
An application uses "SayFinised" event as below.

Tizen.NUI.Accessibility.Accessibility.Instance.SayFinished

But the application should use "Enabled", "Disabled" as below.

Tizen.NUI.Accessibility.Accessibility.Enabled
Tizen.NUI.Accessibility.Accessibility.Disabled

Because both are static member used in constructor.

This patch make these events have same usage as "SayFinished".

Tizen.NUI.Accessibility.Accessibility.Instance.Enabled
Tizen.NUI.Accessibility.Accessibility.Instance.Disabled

### DIFF
--- a/src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
+++ b/src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
@@ -43,12 +43,12 @@ namespace Tizen.NUI.Accessibility
         {
             enabledSignalHandler = () =>
             {
-                Enabled?.Invoke(typeof(Accessibility), EventArgs.Empty);
+                isBridgeEnabled?.Invoke(typeof(Accessibility), EventArgs.Empty);
             };
 
             disabledSignalHandler = () =>
             {
-                Disabled?.Invoke(typeof(Accessibility), EventArgs.Empty);
+                isBridgeDisabled?.Invoke(typeof(Accessibility), EventArgs.Empty);
             };
 
             Interop.Accessibility.RegisterEnabledDisabledSignalHandler(enabledSignalHandler, disabledSignalHandler);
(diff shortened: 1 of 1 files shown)


kimcinoo:
> I think we should go the opposite way, i.e. work towards migrating `Accessibility` to a `static class` (meaning a collection of static methods, just like [System.Math](https://docs.microsoft.com/en-us/dotnet/api/system.math?view=net-6.0)) and remove the `Instance` member entirely, especially because all methods / events available in the `Accessibility` class are `static` in DALi anyway.

I got one thing making it hard. It is not able to add 'static' to GetCurrentlyHighlightedView because it is using 'this'
```
        public static View GetCurrentlyHighlightedView()
        {
            var ptr = Interop.ControlDevel.DaliAccessibilityAccessibleGetCurrentlyHighlightedActor();

            return this.GetInstanceSafely<View>(ptr);
        }
```
It looks good if we use 'Accessibility' without 'Instance'
In case of 'AccessibilityManager' or 'FocusManager' we have used 'Instance' 
So it seems that it would be fine with 'Instance'

aswigon:
See:
* https://github.com/Samsung/TizenFX/pull/3987
* https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/271723/

for a more comprehensive solution.

lol-github (commented):
I agree with Artur and I'm opting for #3987

kimcinoo:
Yes agree. The change https://github.com/Samsung/TizenFX/pull/3987 is much better than this.
I did not know hot to remove 'this'. It is very good work.

PR #3987, opened 2022-02-25
[NUI][AT-SPI] Static class Accessibility

aswigon:
### Description of Change ###

This class serves as a collection of unrelated methods, which are all static in DALi anyway, and as such is a good candidate for migration to a C# `static class`.

Complementary dali-csharp-binder patch (merge at the same time):
* https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/271723

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
+++ b/src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
@@ -30,15 +30,10 @@ namespace Tizen.NUI.Accessibility
     [SuppressMessage("Microsoft.Design", "CA1724: Type names should not match namespaces")]
     [SuppressMessage("Microsoft.Design", "CA1001:Types that own disposable fields should be disposable", Justification = "This is a singleton class and is not disposed")]
     [EditorBrowsable(EditorBrowsableState.Never)]
-    public class Accessibility
+    public static class Accessibility
     {
-        #region Constructor, Destructor, Dispose
-        private Accessibility()
-        {
-            dummy = new View();
-            dummy.Name = "dali-atspi-singleton";
-        }
-
+        #region Constructor
+        [SuppressMessage("Microsoft.Performance", "CA1810: Initialize reference type static fields inline", Justification = "Need to call native code")]
         static Accessibility()
         {
             enabledSignalHandler = () =>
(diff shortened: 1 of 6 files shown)


aswigon:
As I mentioned under https://github.com/Samsung/TizenFX/pull/3984, this class is a collection of methods from `Accessibility::Bridge` and `Accessibility::Accessible` which are all static, as well as functions from namespace `Dali::AtspiAccessibility` which do not belong to any class. That's why a `static class` is better than a singleton which does not correspond to an instance of any particular object in DALi.

lol-github (approved):
LGTM,  the code looks cleanlier now due to the reduction of number of parameters and due to the simplification of API usage: Accessibility.Accessibility.Instance.Say() ---> Accessibility.Accessibility.Say()

Optimally would be to have in public API as short form as Accessibility.Say() but as Artur noticed this is blocked by naming conflict with the namespace containing AccessibilityManager class.

src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
```diff
@@ -223,21 +179,21 @@ public void SetHighlightFrameView(View view)
...
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public View GetCurrentlyHighlightedView()
+        public static View GetCurrentlyHighlightedView()
         {
             var ptr = Interop.ControlDevel.DaliAccessibilityAccessibleGetCurrentlyHighlightedActor();
 
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
 
-            return this.GetInstanceSafely<View>(ptr);
+            return dummyHandle.GetInstanceSafely<View>(ptr);
```
kimcinoo, line 188: There is a way without 'this'. Cool.

kimcinoo (approved):
Looks very good to me!

aswigon:
> Looks good to me except one thing. Could you remove this warning?
> 
> src/Tizen.NUI/src/public/Accessibility/Accessibility.cs#L36 Initialize all static fields in 'Accessibility' when those fields are declared and remove the explicit static constructor

Sure.

Seoyeon2Kim (approved):

PR #3992, opened 2022-03-02
[NUI] Fix ImageView.Image property bugs + Make AnimatedImageView use LazyUpdate

hinohie:
1. Apply updated properties before we get `ImageView.Image`.

> What was problem?

Previous LazyUpdate feature have some bugs when we try to get ImageView.Image.
That API get from native engine side properies forcely.
But actually, we didn't send our NUI cached local properties before get from native engine.
So the result value was not matched what we expected want to get.

> How to fix it?

We call `UpdateImage` properties synchronously when we call `ImageView.Image`.
And after get the result of `ImageView.Image`, update our local cached map as it's result.

You can see `Image_GET_VALUE_After_Set_ResourceUrl` TestCase.

2. Apply updated properties before we send animation actions like `ImageView.Play`, `ImageView.Stop`, `ImageView.Pause`.

> What was problem?

These three actions don't update visual. they just trigger the actions, and the result of action calculated internally.
But LazyUpdate feature create the visual lazy. So, when we send these actions, visual can not be created yet.

> How to fix it?

We call `UpdateImage` properties synchronously when we call that action.
Now then, we can assume that there will be exist the visual what we want to control the actions.

3. `AnimatedImageView` also use LazyUpdate feature.

> What was problem?

Similar energe with `ImageView.Play`, `AnimatedImageView` already use it's own LazyUpdate feature internally.
And it's own LazyUpdate feature is not sync with ImageView's LazyUpdate feature.

For example, when we set `ImageView`'s properties (like AlphaMaskURL) and after set `AnimatedImageView` properties and call `SetValues`, it will overwrite 'only' `AnimatedImageView` properties.
and after 1 event frame later, `ImageView`'s LazyUpdate also overwrite 'only' ImageView` properties now.

> How to fix it?

We make this class also use our `ImageView` 's LazyUpdate feature.
First, we make `ImageView.UpdateImage()` as `protected virtual` level.
And also make `_imagePropertyMap` as `internal` level, so we can access this cached properties. (Actually, I change the name as `cachedImagePropertyMap` what future developer can know what it is)
So now, when we need to update only for `AnimatedImageView` property, we can overried the internal function calls.

After this fix, we don't need to `AnimatedImageView`'s `public new Play()` etc API.

Note : We cannot know URLs property changeness. so keep `SetValues` API to properties setup forcely, what previous API already did.

- TODO : 

In `AnimatedImageView.UpdateImage()` set the `Visual.Property.Type` is some kind of trick.
We need to find a good way to seperate normal `ImageView` and `AnimatedImageView`.
But actually, AnimatedImageView is hidden API so rarely app only use this components.
So just keep previous behaviours for legacy, and need to find solution in future.


src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -260,12 +260,33 @@ private PropertyMap InternalImageMap
...
                     PropertyValue image = GetProperty(ImageView.Property.IMAGE);
                     image?.Get(returnValue);
                     image?.Dispose();
+
+                    // Update cached property map
+                    if(returnValue != null)
+                    {
+                        if(_imagePropertyMap != null)
+                        {
+                            _imagePropertyMap.Merge(returnValue);
```
hinohie, line 278: Merge is too expensive functionin dali + I think we don't need to merge here. just dispose and new is okay

src/Tizen.NUI/src/public/BaseComponents/AnimatedImageView.cs
```diff
@@ -372,101 +369,44 @@ private int InternalCurrentFrame
...
+            UpdateImage();
         }
 
-        /// <summary>
-        /// Play animation
-        /// </summary>
-        // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public new void Play()
+        protected override void UpdateImage()
```
hinohie, line 380: Need summary

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1403,6 +1451,7 @@ private void UpdateImageMap()
         {
             PropertyMap imageMap = new PropertyMap();
 
+            // Note : We can't use ImageView.Image property here. Because That propert call UpdateImage internally.
```
hinohie, line 1454: typo

src/Tizen.NUI/src/public/BaseComponents/ImageViewBindableProperty.cs
```diff
@@ -110,10 +110,31 @@ public partial class ImageView
...
+                // Get current properties force.
                 PropertyMap temp = new PropertyMap();
                 Tizen.NUI.Object.GetProperty((HandleRef)imageView.SwigCPtr, ImageView.Property.IMAGE).Get(temp);
+
+                // Update cached property map
+                if(temp != null)
+                {
+                    if(imageView._imagePropertyMap != null)
+                    {
+                        imageView._imagePropertyMap.Merge(temp);
```
hinohie, line 126: Same as here. (Actually it is deprecated API but anyway) Merge is not good here.

test/Tizen.NUI.Devel.Tests.Ubuntu/Tizen.NUI.Devel.Tests/testcase/TSAnimatedImageView.cs
```diff
@@ -0,0 +1,286 @@
...
+
+
+        [Test]
+        [Category("P1")]
+        [Description("internal API test in Ubuntu, AnimatedImageView.LoopCount")]
+        [Property("SPEC", "Tizen.NUI.BaseComponents.AnimatedImageView.LoopCount")]
+        [Property("SPEC_URL", "-")]
+        [Property("CRITERIA", "PRO")]
+        [Property("AUTHOR", "eunkiki.hong@samsung.com")]
+        public void LoopCount_GET_VALUE()
```
hinohie, line 210: SET_GET

src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
hinohie: I think more detail yesterday but... Merge is neccessary job. (`_imagePropertyMap` 's role is actually "cache" . not "sync")
I'll prepare some more reasonable fixed patch next week.

dongsug-song (approved):
looks good to me

everLEEst (approved):
Looks good to me. Well made TC :)

src/Tizen.NUI/src/public/BaseComponents/AnimatedImageView.cs
```diff
@@ -372,101 +372,76 @@ private int InternalCurrentFrame
...
-            SetValues();
-            base.Pause();
+            if (resourceURLs?.Count != 0)
+            {
+                using(PropertyArray indexPropertyArray = new PropertyArray())
+                {
+                    PropertyArray returnedArr = new PropertyArray();
+                    foreach (var iter in resourceURLs)
+                    {
+                        using(PropertyValue index = new PropertyValue(iter))
```
Seoyeon2Kim, line 398: Could you remove this warning message? Or you can suppress this.

src/Tizen.NUI/src/public/BaseComponents/AnimatedImageView.cs#L398
Use recommended dispose pattern to ensure that object created by 'new PropertyValue(iter)' is disposed on all paths.

Seoyeon2Kim:
Oh, the PR is merged :p Wish the message remove later!

src/Tizen.NUI/src/public/BaseComponents/AnimatedImageView.cs
hinohie: Unfortunatly, current code already use 'using' state, so we can assume that it will be Disposed end of this scope.
I don't know why this message exist. So..just ignore.

PR #3998, opened 2022-03-03
[Applications.Common] Add Globalization-invariant mode check

hjhun:
Signed-off-by: Hwankyu Jhun <h.jhun@samsung.com>

### Description of Change ###

In globalization-invariant mode, creating a new CultureInfo() throws a CultureNotFoundException exception. 
If the process runs in invariant mode, the CoreApplication doesn't set CultureInfos. And, the CultureInfoHelper uses ToLowerInvaraint() instead of ToLower() with CultureInfo("en-US");

Issue: https://github.com/Samsung/TizenFX/issues/3994

PR #3999, opened 2022-03-04
[Applications.Common] Add Globalization invariant mode check (#3998)

kkhos:
In globalization-invariant mode, creating a new CultureInfo() throws
a CultureNotFoundException exception. If the process runs in invariant
mode, the CoreApplication doesn't set CultureInfos. And, the
CultureInfoHelper uses ToLowerInvaraint() instead of ToLower() with CultureInfo("en-US");

### API Changes ###

 - ACR:


JoogabYun (approved):

dongsug-song (approved):
integration patch

Seoyeon2Kim (approved):

PR #4003, opened 2022-03-04
[NUI][Samples] Remove error message on Tizen.NUI.Samples

hinohie:
Due to the DaliDemo.cs is just copycat of dali-demo,
it's local environment is not matched with NUI.
So, `{APPLICATION_RESOURCE_PATH}` is not matched as
`Tizen.Applications.Application.Current.DirectoryInfo.Resource`.

Therefore, when we try to SetStyle with image url, we fail to find that image.
That's why Tizen.NUI.Samples show so many error messages.

So I remove that url in demo style file.

Signed-off-by: Eunki Hong <h.pichulia@gmail.com>


dongsug-song (approved):
this is fix of NUI Samples
not tizenfx.

Seoyeon2Kim (approved):
Oh.. Thanks for fixing it!!!

JoogabYun (approved):

PR #4008, opened 2022-03-08
[NUI] Prevent to call button up event when button is not pressed

taehyub:
### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #4022, opened 2022-03-11
[NUI][AT-SPI] Remove SetAccessibilityConstructor()

aswigon:
### Description of Change ###

This PR exposes AT-SPI interfaces as native C# interfaces (e.g. `public class Slider : IAtspiValue`) instead of an enumeration value that has to be passed to `SetAccessibilityConstructor` from `OnInitialize` (e.g. `SetAccessibilityConstructor(AccessibilityInterface.Value)`). Leveraging the native interfaces mechanism allows the C# compiler to check that:

1. The respective interface is declared when trying to implement its method (previously it was possible to override e.g. `AccessibilityGetMinimum` without actually declaring `Value` interface, which is a violation of object-oriented programming principles).

And, conversely, that:

2. All required methods are implemented when declaring a given interface (previously it was possible to declare e.g. `Value` interface and leave `AccessibleGetMinimum` unimplemented, which is also a violation of object-oriented programming principles).

`SetAccessibilityConstructor` is now entirely removed. It has never served the purpose that the name suggested, and the respective DALi function with the same name (which, contrary to NUI, actually had an `accessibilityConstructor` parameter) has recently been removed as well anyway.

Finally, it is now possible to declare more than one AT-SPI interface for any control (e.g. `public class MyControl : IAtspiText, IAtspiValue`).

This change is entirely contained within the TizenFX repository, i.e. it is not dependent on any DALi patches.

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI.Components/Controls/AlertDialog.cs
+++ b/src/Tizen.NUI.Components/Controls/AlertDialog.cs
@@ -463,7 +463,7 @@ namespace Tizen.NUI.Components
         public override void OnInitialize()
         {
             base.OnInitialize();
-            SetAccessibilityConstructor(Role.Dialog);
+            AccessibilityRole = Role.Dialog;
             AppendAccessibilityAttribute("sub-role", "Alert");
             Show(); // calls RegisterDefaultLabel(); Hide() will call UnregisterDefaultLabel()
         }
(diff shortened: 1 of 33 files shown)


lol-github (approved):
I was awaiting this "OO paradigm restoring" patch for a long lime :) Thank you Artur!

m-bialota:
Looks OK!

kimcinoo (approved):
Looks very good to me. Way to go.

dongsug-song (approved):
looks good to me

PR #4036, opened 2022-03-15
[NUI][AT-SPI] Remove interops for SetAccessibilityConstructor()

dongsug-song:
The method name SetAccessibilityConstructor() in NUI has always been
misleading, because it actually never had an 'accessibilityConstructor'
parameter in the first place. With the recent addition of bindings for
Accessible::GetInterfaces(), all NUI controls use the same Accessible
implementation (NUIViewAccessible) anyway. However, as this method is
currently used in many places, it is kept as a wrapper for setting
AccessibilityRole (and interface flags) and will be removed if/when we
decide to switch to AT-SPI interfaces as native C# interfaces. Then, all
usages of SetAccessibilityConstructor() would be replaced by simply
setting AccessibilityRole (and interface flags won't be needed).

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.ControlDevel.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.ControlDevel.cs
@@ -177,8 +177,6 @@ namespace Tizen.NUI
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Accessibility_Accessible_SetHighlightActor")]
             public static extern void DaliAccessibilityAccessibleSetHighlightActor(global::System.Runtime.InteropServices.HandleRef arg1);
 
-            // SetAccessibilityConstructor
-
             // Keep this structure layout binary compatible with the respective C++ structure!
             [EditorBrowsable(EditorBrowsableState.Never)]
             [StructLayout(LayoutKind.Sequential)]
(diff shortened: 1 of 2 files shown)


dongsug-song:
this seems to be the fix patch for yesterday's SR test-hub failure.
and this has been gotten Approval already and been merged in DevelNUI. 
so I merge now

test-hub failure
=>
http://10.113.10.113:8080/job/TIZEN_RPI4_CSHARP/2775/artifact/result/TCT_CSHARP/
=>
System.EntryPointNotFoundException : Unable to find an entry point named '?' in shared library 'libdali2-csharp-binder.so'.
at Tizen.NUI.Interop.ControlDevel.DaliToolkitDevelControlSetAccessibilityConstructor(HandleRef arg1_self, Int32 arg2_role)
at Tizen.NUI.BaseComponents.View.SetAccessibilityConstructor(Role role, AccessibilityInterface accessibilityInterface) in Tizen.NUI.dll:token 0x60022e7 0x11
at Tizen.NUI.BaseComponents.CustomView.OnInitialize() in Tizen.NUI.dll:token 0x6001ba8 0x0
at Tizen.NUI.BaseComponents.CustomView.Initialize() in Tizen.NUI.dll:token 0x6001be7 0x389
at Tizen.NUI.BaseComponents.CustomView..ctor(String typeName, CustomViewBehaviour behaviour) in Tizen.NUI.dll:token 0x6001ba1 0xd
at Tizen.NUI.Tests.CustomViewTests.CustomView_INIT() in /var/lib/jenkins/jobs/CsharpTCT/TCT_7.0/api/tct-suite-vs/Tizen.NUI.Tests/testcase/TSCustomView.cs:line 44

kimcinoo (approved):
way to go

Seoyeon2Kim (approved):

PR #4037, opened 2022-03-16
[NUI] Add GrabTouchAfterLeave to ViewStyle and fix ViewStyle apply bug.

JoogabYun:
### Description of Change ###

1. Add GrabTouchAfterLeave property to ViewStyle.

2. View does not apply ViewStyle. Fix this bug.
   Change the constructor so that defaultViewStyle is applied.
```C#
    internal View(global::System.IntPtr cPtr, bool cMemoryOwn, bool shown = true) : this(cPtr, cMemoryOwn, null, shown)
    {
    }

    internal View(global::System.IntPtr cPtr, bool cMemoryOwn, ViewStyle viewStyle, bool shown = true) : base(cPtr, cMemoryOwn)
    {
        InitializeStyle(viewStyle);
        ...
    }
```

3. Resolves a bug where Views such as ImageView or TextLable did not receive the style of the base View.
   Allows ViewStyles to be merged according to the hierarchy.
```C#
   public ViewStyle GetStyle(Type viewType)
   {
       var currentType = viewType ?? throw new ArgumentNullException(nameof(viewType));
       ViewStyle resultStyle = null;
       do
       {
           var currentStyle = GetStyle(currentType.FullName);
           resultStyle = (resultStyle == null) ? currentStyle : currentStyle?.Merge(resultStyle) ?? resultStyle;
           currentType = currentType.BaseType;
       }
       while (currentType != null);
       return resultStyle;
        }
```

4. Change IThemeCreator and AddPackageTheme to public so that the default theme of the application can be added.
   If you want to change the application default theme, do as follows
```C#
   public class MyDefaultTheme : IThemeCreator
   {
       public Theme Create()
       {
          Theme theme = new Theme();
           theme.AddStyle("Tizen.NUI.BaseComponents.View",  new ViewStyle()
           {
                GrabTouchAfterLeave = true,
           });
           return theme;
       }
   }

   public void Activate()
   {
        ThemeManager.AddPackageTheme(new MyDefaultTheme());

        View viewA = new View(); // A newly created View is created with GrabTouchAfterLeave=true property.

        // Even if a new user theme is applied, it is maintained without overriding the default theme because it is merged with the default theme.
       Theme theme = new Theme();
       theme.AddStyle("Tizen.NUI.BaseComponents.View",  new ViewStyle()
       {
          Size = new Size(600, 600),
       });
       ThemeManager.ApplyTheme(theme);

       View viewB = new View(); // A newly created View is created with 600,600 size and GrabTouchAfterLeave=true properties.
   }
```

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -126,13 +126,14 @@ namespace Tizen.NUI.BaseComponents
             backgroundExtraData = uiControl.backgroundExtraData == null ? null : new BackgroundExtraData(uiControl.backgroundExtraData);
         }
 
-        internal View(global::System.IntPtr cPtr, bool cMemoryOwn, ViewStyle viewStyle, bool shown = true) : this(cPtr, cMemoryOwn, shown)
+        internal View(global::System.IntPtr cPtr, bool cMemoryOwn, bool shown = true) : this(cPtr, cMemoryOwn, null, shown)
         {
-            InitializeStyle(viewStyle);
         }
 
-        internal View(global::System.IntPtr cPtr, bool cMemoryOwn, bool shown = true) : base(cPtr, cMemoryOwn)
+        internal View(global::System.IntPtr cPtr, bool cMemoryOwn, ViewStyle viewStyle, bool shown = true) : base(cPtr, cMemoryOwn)
         {
+            InitializeStyle(viewStyle);
+
             if (HasBody())
             {
                 PositionUsesPivotPoint = false;
@@ -150,7 +151,6 @@ namespace Tizen.NUI.BaseComponents
             {
                 SetVisible(false);
             }
-
         }
 
         internal View(ViewImpl implementation, bool shown = true) : this(Interop.View.NewViewInternal(ViewImpl.getCPtr(implementation)), true)
--- a/src/Tizen.NUI/src/public/Theme/IThemeCreator.cs
+++ b/src/Tizen.NUI/src/public/Theme/IThemeCreator.cs
@@ -14,10 +14,15 @@
  * limitations under the License.
  *
  */
+using System.ComponentModel;
 
 namespace Tizen.NUI
 {
-    internal interface IThemeCreator
+    /// <summary>
+    /// It is an interface that creates a theme.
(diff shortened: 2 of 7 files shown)


src/Tizen.NUI/src/public/Theme/IThemeCreator.cs
```diff
@@ -13,11 +13,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *
- */
```
hinohie, line 16: Super super minor review. Maybe useless fixing line

dongsug-song (approved):
looks good to me!
(b^_^)b

src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -92,7 +92,7 @@ static View()
         /// Creates a new instance of a view.
         /// </summary>
         /// <since_tizen> 3 </since_tizen>
-        public View() : this(Interop.View.New(), true)
+        public View() : this(Interop.View.New(), true, null)
```
hinohie, line 95: This change may occure some strange issues.

(Maybe current codes are not works like this but..) Some subclass may can call `base(cPtr, cMemoryOwn, shown)` constructor directly. in that case, that subclass cannot use default Theme.

Instead of change this line, I hope to change  `internal View(global::System.IntPtr cPtr, bool cMemoryOwn, bool shown = true)`.
Make that function call `this(cPtr, cMemoryOwn, null, shown)`, and move whole jobs into `internal View(global::System.IntPtr cPtr, bool cMemoryOwn, ViewStyle viewStyle, bool shown = true) : base(cPtr, cMemoryOwn)`

src/Tizen.NUI/src/public/Theme/IThemeCreator.cs
JoogabYun: done

src/Tizen.NUI/src/public/BaseComponents/View.cs
JoogabYun: done

JoogabYun:
좀 더 테스트가 필요합니다. 패치는 다시 업데이트 후 리뷰 요청드리겠습니다.

PR #4041, opened 2022-03-16
[NUI] Fix crash issue when visual create failed

hinohie:
Due to the empty array issue, we change that we don't create visual when
UrlList is empty.

But current NUI.VisualView don't consider that visual failure case.

This patch consider case when we fail to create visual, so visual is null in VisualView.

### API Changes ###

 - ACR:


kimcinoo (approved):
No doubts. Good!

Seoyeon2Kim (approved):
LGTM! Thanks :)

dongsug-song (approved):
looks good to me

PR #4045, opened 2022-03-17
[NUI] Fix PointSize convert error in EXaml.

AchoWang:
### Description of Change ###
If use [TypeConverter(typeof(PointSizeTypeConverter))],  PointSize 's convert will through "(float) new PointSizeTypeConverter () ConvertFromInvariantString ();", which is incorrect, and will throw "Value can not be null".

We can verify this with the BackgroundColorTest in NUITizenGallery  , For comparison, Button.PointSize which not use   [TypeConverter(typeof(PointSizeTypeConverter))], will works well.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/TextEditor.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/TextEditor.cs
@@ -295,7 +295,6 @@ namespace Tizen.NUI.BaseComponents
         /// The size of font in points.<br />
         /// </summary>
         /// <since_tizen> 3 </since_tizen>
-        [Binding.TypeConverter(typeof(PointSizeTypeConverter))]
         public float PointSize
         {
             get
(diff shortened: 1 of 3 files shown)


huayongxu (approved):
lgtm

zhouleonlei (approved):
lgtm

AdunFang (approved):
It is OK.

xiaowei-guan (approved):
LGTM

elishateng (approved):
LGTM

wanchao-xu (approved):
Looks good to me

myroot:
Could you analyse this issue more deeply? according to you comment, TypeConverter attribute is not working correctly in EXAML
We need to solve this issue, becuase App developer normally use TypeConverter to convert user defined type. so it should be work correctly

everLEEst (changes requested):
PointSize need to be converted dp and sp properly in application scalable UI. could you please fix the reason of "Value can not be null"?

dongsug-song:
@AchoWang 
Hello, could you check the review comments above?
thank you.

PR #4056, opened 2022-03-22
[NUI][Xaml]Fix issue of Call SetBinding in C# code

AdunFang:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/XamlBinding/BindableObject.cs
+++ b/src/Tizen.NUI/src/public/XamlBinding/BindableObject.cs
@@ -260,25 +260,6 @@ namespace Tizen.NUI.Binding
             }
         }
 
-        internal void SetValueAndForceSendChangeSignal(BindableProperty property, object value)
-        {
-            if (property == null)
-                throw new ArgumentNullException(nameof(property));
-
-            if (true == IsBinded)
-            {
-                if (property.IsReadOnly)
-                    throw new InvalidOperationException(string.Format("The BindableProperty \"{0}\" is readonly.", property.PropertyName));
-
-                SetValueCore(property, value, SetValueFlags.ClearOneWayBindings | SetValueFlags.ClearDynamicResource,
-                    SetValuePrivateFlags.ManuallySet | SetValuePrivateFlags.CheckAccess, true);
-            }
-            else
-            {
-                property.PropertyChanged?.Invoke(this, null, value);
-            }
-        }
-
         /// This will be public opened in next ACR.
         [EditorBrowsable(EditorBrowsableState.Never)]
         public void ForceNotifyBindedInstance(BindableProperty property)
--- a/src/Tizen.NUI/src/public/XamlBinding/BindableObject.cs
+++ b/src/Tizen.NUI/src/public/XamlBinding/BindableObject.cs
@@ -209,19 +209,12 @@ namespace Tizen.NUI.Binding
             SetBinding(targetProperty, binding, false);
         }
 
-        private bool isCreateByXaml = false;
         /// Only used by the IL of xaml, will never changed to not hidden.
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public virtual bool IsCreateByXaml
+        public bool IsCreateByXaml
         {
-            get
-            {
-                return isCreateByXaml;
-            }
-            set
-            {
-                isCreateByXaml = value;
-            }
+            get;
+            set;
         }
 
         /// <summary>
@@ -283,6 +276,48 @@ namespace Tizen.NUI.Binding
             else
             {
                 property.PropertyChanged?.Invoke(this, null, value);
+            }
+        }
+
+        /// This will be public opened in next ACR.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public void ForceNotifyBindedInstance(BindableProperty property)
+        {
+            if (null != property)
+            {
+                BindablePropertyContext context = GetOrCreateContext(property);
+
+                if (null != context?.Binding)
+                {
+                    applying = true;
+                    context.Binding.Apply(true);
+                    applying = false;
+                }
+
+                OnPropertyChanged(property.PropertyName);
+
+                PropertyToGroup.TryGetValue(property, out HashSet<BindableProperty> propertyGroup);
+
+                if (null != propertyGroup)
+                {
+                    foreach (var relativeProperty in propertyGroup)
+                    {
+                        if (relativeProperty != property)
+                        {
+                            var relativeContext = GetOrCreateContext(relativeProperty);
+                            var relativeBinding = relativeContext.Binding;
+
+                            if (null != relativeBinding)
+                            {
+                                applying = true;
+                                relativeBinding.Apply(true);
+                                applying = false;
+                            }
+
+                            OnPropertyChanged(relativeProperty.PropertyName);
+                        }
+                    }
+                }
             }
         }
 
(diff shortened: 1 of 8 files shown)


xuelian-bai:
looks good to me

elishateng (approved):
LGTM

zg2nets (approved):
lgtm

zhouleonlei (approved):
lgtm

src/Tizen.NUI/src/public/XamlBinding/BindableObject.cs
```diff
@@ -272,7 +272,7 @@ internal void SetValueAndForceSendChangeSignal(BindableProperty property, object
             if (property == null)
                 throw new ArgumentNullException(nameof(property));
 
-            if (true == isCreateByXaml)
+            if (true == IsBinded)
```
myroot, line 275: If Some property setter was use `SetValueAndForceSendChangeSignal` and didn't use data binding, it does not raise PropertyChanged event, even though property was changed, is it intended behavior? if you intent this behavior, Could you tell me why?

myroot: Could you tell me why do you add `SetValueAndForceSendChangeSignal` instead `SetValue` I may provide a solution to resolve your origin issue

AdunFang: "If Some property setter was use SetValueAndForceSendChangeSignal and didn't use data binding, it does not raise PropertyChanged event, even though property was changed, is it intended behavior? if you intent this behavior, Could you tell me why?"

AdunFang: The Text of TextEditor/TextField can be changed by key input, so that the object which binded to the Text of TextEditor/TextField will not be notified.

To fix this issue, we listen the TextChanged event in TextEditor/TextField, and notify the binded instance.

myroot: >so that the object which binded to the Text of TextEditor/TextField will not be notified.

Why? I don't understand the reason, What is problem when notified was raised?
 Xamarin.Forms Entry.Text also can be changed by key input but it works same like others property

myroot: Simply, what's issue with using SetValue instead of SetValueAndForceSendChangeSignal on Text of TextField?

src/Tizen.NUI/src/public/XamlBinding/BindableObject.cs
```diff
@@ -267,22 +260,45 @@ public void SetValue(BindableProperty property, object value)
             }
         }
 
-        internal void SetValueAndForceSendChangeSignal(BindableProperty property, object value)
+        /// This will be public opened in next ACR.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public void ForceNotifyBindedInstance(BindableProperty property)
```
myroot, line 265: Could you tell me the purpose of this method?

myroot:
I want to know your original issue that related with Text of TextField, 
Because, Xamarin.forms is no need these code to work correctly

src/Tizen.NUI/src/public/XamlBinding/BindableObject.cs
AdunFang: > > so that the object which binded to the Text of TextEditor/TextField will not be notified.
> 
> Why? I don't understand the reason, What is problem when notified was raised? Xamarin.Forms Entry.Text also can be changed by key input but it works same like others property

Because user can set focus on the TextEditor/TextField, and then change the Text by press remote-control or keyboard, in this situation, the set method of property Text will not be called, so the binded object will not be notified.

To fix this issue, we listen the TextChanged event, in the listener, we set the Text of TextEditor/TextField again.
```
        public override bool IsCreateByXaml
        {
            get
            {
                return base.IsCreateByXaml;
            }
            set
            {
                base.IsCreateByXaml = value;

                if (value == true)
                {
                    this.TextChanged += (obj, e) =>
                    {
                        this.Text = e.TextEditor.Text;  // <==== Set Text again.
                    };
                }
            }
        }
```

When the Text is reset, if call "SetValue", the original value of Text is same to new value, so that it will not notify the binded object
```
                if ((!same || raiseOnEqual))  //<==== orignal value is same to new value, this condition can't be matched.
                {
                    property.PropertyChanged?.Invoke(this, original, value);

                    if (binding != null && !currentlyApplying)
                    {
                        applying = true;
                        binding.Apply(true);
                        applying = false;
                    }

                    OnPropertyChanged(property.PropertyName);  // <==== Binded objects are notified here.
```

so we add the method SetValueAndForceSendChangeSignal, to force notify the binded object.

AdunFang: Now I have removed the method SetValueAndForceSendChangeSignal and added a new method, to make the code more clearly.

myroot:
If the original issue is a binding break issue, it can be resolved through two-way binding. 
All properties that update from native, must use BindingMode.TwoWay

myroot:
I quickly find that wrong declared properties and find it

TextField.TextProperty
TextEditor.TextProperty
Button.IsSelectedProperty
Pagination.SelectedIndexProperty

these properties could be update from UI but does not use two-way binding

AdunFang:
> TextProperty

AdunFang:
> 

Because the input of Key

AdunFang:
> I want to know your original issue that related with Text of TextField, Because, Xamarin.forms is no need these code to work correctly

Hi, Mr. Lee
The code was modified as a result of our yesterday's discussion,
could you please help to reivew it?

dongsug-song:
> > I want to know your original issue that related with Text of TextField, Because, Xamarin.forms is no need these code to work correctly
> 
> Hi, Mr. Lee The code was modified as a result of our yesterday's discussion, could you please help to reivew it?

@AdunFang 
Hello, 
could you share the result of discussion even breifly?
thank you.

AdunFang:
@dongsug-song The result is:

myroot:
>enforce notify the binded objects.

We didn't know way to change notify enforcely.
But I just found to raise PropertyChanged event even though no changes on value.

You can use SetValueFlags.RaiseOnEqual on `SetValueCore` method

AdunFang:
> > enforce notify the binded objects.
> 
> We didn't know way to raise PropertyChanged event enforcely. But I just found to raise PropertyChanged event even though no changes on value.
> 
> You can use SetValueFlags.RaiseOnEqual on `SetValueCore` method

Hi, Mr. Lee

myroot:
> > > enforce notify the binded objects.
> > 
> > 
> > We didn't know way to raise PropertyChanged event enforcely. But I just found to raise PropertyChanged event even though no changes on value.
> > You can use SetValueFlags.RaiseOnEqual on `SetValueCore` method
> 
> Hi, Mr. Lee I think when the Text is changed in DALi, NUI needn't to SetValue again, it only need to notify the binded objects, so I think call SetValue may be affect the performance.

Ok, I see

dongsug-song:
@AdunFang 
could you write the decription?
could you add test case or sample APP to verify this?
thank you.

AdunFang:
> @AdunFang could you write the decription? could you add test case or sample APP to verify this? thank you.

Hi, Mr. Song

dongsug-song (approved):
approved

PR #4060, opened 2022-03-23
[NUI] Fix SVACE issue

dongsug-song:
### Description of Change ###
[NUI] Fix SVACE issue
https://analysishub.sec.samsung.net/service/analyses/1230084

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/AnimatedImageView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/AnimatedImageView.cs
@@ -388,7 +388,7 @@ namespace Tizen.NUI.BaseComponents
         {
             if (!imagePropertyUpdatedFlag) return;
 
-            if (resourceURLs?.Count != 0 && resourceURLs != null)
+            if (resourceURLs != null && resourceURLs.Count != 0)
             {
                 using (PropertyArray indexPropertyArray = new PropertyArray())
                 {


taehyub (approved):
LGTM :)

src/Tizen.NUI/src/public/BaseComponents/AnimatedImageView.cs
```diff
@@ -386,16 +386,16 @@ public void SetValues()
         [EditorBrowsable(EditorBrowsableState.Never)]
         protected override void UpdateImage()
         {
-            if(!imagePropertyUpdatedFlag) return;
+            if (!imagePropertyUpdatedFlag) return;
 
-            if (resourceURLs?.Count != 0)
+            if (resourceURLs?.Count != 0 && resourceURLs != null)
```
everLEEst, line 391: could be better
(resourceURLs != null && resourceURLs.Count != 0)
as we do not check null twice.

src/Tizen.NUI/src/public/Common/BaseHandle.cs
```diff
@@ -680,7 +680,8 @@ private void disposeDebuggingDispose(DisposeTypes type)
 
             if (swigCPtr.Handle != IntPtr.Zero && swigCPtrCopy.Handle != IntPtr.Zero && HasBody())
             {
-                var process = global::System.Diagnostics.Process.GetCurrentProcess().Id;
+                using var currentProcess = global::System.Diagnostics.Process.GetCurrentProcess();
```
everLEEst, line 683: Is this right usage of using..?
I've only seen the below code usages.
using (var currentProcess = global::System.Diagnostics.Process.GetCurrentProcess())
{
             ....
}

everLEEst (commented):
I'm not 100% sure about this patch so leave the comments. if I'm wrong, please ignore it :)

JoogabYun (approved):

src/Tizen.NUI/src/public/Common/BaseHandle.cs
hinohie: and one more thing. If we use `using` tag, `currentProcess` will be Automatically disposed. I don't know exactly but, Is it intention?

jaehyun0cho: I think this svace is about CA1837.
https://docs.microsoft.com/ko-kr/dotnet/fundamentals/code-analysis/quality-rules/ca1837

If so, Environment.ProcessId is better than Process.GetCurrentProcess().Id as follows.
```
var process = global::System.Environment.ProcessId;
```

dongsug-song: > I think this svace is about CA1837. https://docs.microsoft.com/ko-kr/dotnet/fundamentals/code-analysis/quality-rules/ca1837
> 
> If so, Environment.ProcessId is better than Process.GetCurrentProcess().Id as follows.
> 
> ```
> var process = global::System.Environment.ProcessId;
> ```

Thank you for review, but I got build error. 
src/public/Common/BaseHandle.cs(683,58): error CS0117: 'Environment' does not contain a definition for 'ProcessId' [/home/ssong2best/dDrive/github.com/dongsug-song/tizenfx/develNUI/TizenFX/src/Tizen.NUI/Tizen.NUI.csproj]

dongsug-song: > Is this right usage of using..? I've only seen the below code usages. using (var currentProcess = global::System.Diagnostics.Process.GetCurrentProcess()) { .... }

Thank you for review, I can see this =>
in https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement
**"The using declaration, which was introduced in C# 8.0, doesn't require braces:"**
and only process id was needed, so I think it would be OK after process id is gotten, it can be disposed.

dongsug-song: > and one more thing. If we use `using` tag, `currentProcess` will be Automatically disposed. I don't know exactly but, Is it intention?


Thank you for review, I can see this =>
in https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement
"The using declaration, which was introduced in C# 8.0, doesn't require braces:"
**and only process id was needed, so I think it would be OK after process id is gotten, it can be disposed.
only process id was used.**

src/Tizen.NUI/src/public/BaseComponents/AnimatedImageView.cs
dongsug-song: thank you for review.
I will fix it.

src/Tizen.NUI/src/public/Common/BaseHandle.cs
everLEEst: > > and one more thing. If we use `using` tag, `currentProcess` will be Automatically disposed. I don't know exactly but, Is it intention?
> 
> Thank you for review, I can see this => in https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement "The using declaration, which was introduced in C# 8.0, doesn't require braces:" **and only process id was needed, so I think it would be OK after process id is gotten, it can be disposed. only process id was used.**

good to know!! thank you for sharing!

dongsug-song:
I will modify this patch after the sync PR (from master branch) is merged.

src/Tizen.NUI/src/public/BaseComponents/AnimatedImageView.cs
dongsug-song: I have fixed => https://github.com/Samsung/TizenFX/pull/4060/files#diff-7598b59f76bc7253a76f939a74012377377ad684adf2f25600603f3c7c12b3f9R391
Thank you.

PR #4062, opened 2022-03-23
[NUI] Make PropertyMap look like Dictionary.

huayongxu:
### Description of Change ###

1) Some APIs are added like Keys, Values, Contains, Remove, etc.
2) Check whether key is duplicated or not when inserting/adding some key value pair.
3) This patch is related to the patch in csharp-binder:
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/272447/

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Common/PropertyMap.cs
+++ b/src/Tizen.NUI/src/public/Common/PropertyMap.cs
@@ -14,6 +14,7 @@
  * limitations under the License.
  *
  */
+using System.Collections.Generic;
 using System.ComponentModel;
 
 namespace Tizen.NUI
@@ -24,7 +25,6 @@ namespace Tizen.NUI
     /// <since_tizen> 3 </since_tizen>
     public class PropertyMap : Disposable
     {
-
         /// <summary>
         /// The constructor.
         /// </summary>
@@ -49,6 +49,42 @@ namespace Tizen.NUI
         }
 
         /// <summary>
+        /// Retrieves all keys.
+        /// </summary>
+        /// <returns>A list of keys.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public IList<PropertyKey> Keys
+        {
+            get
+            {
+                List<PropertyKey> keys = new List<PropertyKey>();
+                for (uint i = 0; i < Count(); i++)
+                {
+                    keys.Add(GetKeyAt(i));
+                }
+                return keys;
+            }
+        }
+
+        /// <summary>
+        /// Retrieves all values.
+        /// </summary>
+        /// <returns>A list of values.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public IList<PropertyValue> Values
+        {
+            get
+            {
+                List<PropertyValue> values = new List<PropertyValue>();
+                for (uint i = 0; i < Count(); i++)
+                {
+                    values.Add(GetValue(i));
+                }
+                return values;
+            }
+        }
+
+        /// <summary>
         /// The operator to access the element with the specified string key.<br />
         /// If an element with the key does not exist, then it is created.<br />
         /// </summary>
@@ -61,9 +97,12 @@ namespace Tizen.NUI
             {
                 return ValueOfIndex(key);
             }
-            internal set
+            set
             {
-                SetValue(key, value);
+                using (PropertyKey pKey = new PropertyKey(key))
+                {
+                    SetValue(pKey, value);
+                }
             }
         }
 
@@ -80,9 +119,12 @@ namespace Tizen.NUI
             {
                 return ValueOfIndex(key);
             }
-            internal set
+            set
             {
-                SetValue(key, value);
+                using (PropertyKey pKey = new PropertyKey(key))
+                {
+                    SetValue(pKey, value);
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


myroot:
I hope to support this use case
https://github.com/Samsung/TizenFX/blob/1f7646a5a21e7e804449fa6cf62599504f7c141d/src/Tizen.NUI/src/public/Common/PropertyValue.cs#L642

huayongxu:
> I hope to support this use case
> 
> https://github.com/Samsung/TizenFX/blob/1f7646a5a21e7e804449fa6cf62599504f7c141d/src/Tizen.NUI/src/public/Common/PropertyValue.cs#L642

Do you mean that some APIs are added into PropertyValue?
This patch is to add some APIs into PropertyMap.
It is for the issue  https://github.sec.samsung.net/dotnet/NUIBackend/issues/80 .

myroot:
PropertyMap contents could be update from Dali layer such as `PropertyValeu.Get` method/
in this case, your code not work propertly

src/Tizen.NUI/src/public/Common/PropertyMap.cs
```diff
@@ -41,13 +44,43 @@ public PropertyMap() : this(Interop.PropertyMap.NewPropertyMap(), true)
...
         public PropertyMap(PropertyMap other) : this(Interop.PropertyMap.NewPropertyMap(PropertyMap.getCPtr(other)), true)
         {
+            if (other != null)
+            {
+                propertyMap = new Dictionary<PropertyKey, PropertyValue>(other.propertyMap);
+            }
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
         }
 
         internal PropertyMap(global::System.IntPtr cPtr, bool cMemoryOwn) : base(cPtr, cMemoryOwn)
```
myroot, line 54: `propertyMap` not created when use this constructor `internal PropertyMap(global::System.IntPtr cPtr, bool cMemoryOwn) : base(cPtr, cMemoryOwn)`

huayongxu: Done.
propertyMap has been removed.

huayongxu:
> PropertyMap contents could be update from Dali layer such as `PropertyValeu.Get` method/ in this case, your code not work propertly

Done.
propertyMap has been removed, key-value pairs would not be cached.

zhouleonlei (approved):
lgtm

xuelian-bai:
looks good to me

elishateng (approved):
LGTM

src/Tizen.NUI/src/public/Common/PropertyMap.cs
```diff
@@ -48,6 +48,42 @@ internal PropertyMap(global::System.IntPtr cPtr, bool cMemoryOwn) : base(cPtr, c
         {
         }
 
+        /// <summary>
+        /// Retrieves all keys.
+        /// </summary>
+        /// <returns>A collection of keys.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public List<PropertyKey> Keys
```
myroot, line 56:
```suggestion
        public IList<PropertyKey> Keys
```

src/Tizen.NUI/src/public/Common/PropertyMap.cs
```diff
@@ -48,6 +48,42 @@ internal PropertyMap(global::System.IntPtr cPtr, bool cMemoryOwn) : base(cPtr, c
...
+                return keys;
+            }
+        }
+
+        /// <summary>
+        /// Retrieves all values.
+        /// </summary>
+        /// <returns>A collection of values.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public List<PropertyValue> Values
```
myroot, line 74:
```suggestion
        public IList<PropertyValue> Values
```

src/Tizen.NUI/src/public/Common/PropertyMap.cs
```diff
@@ -179,19 +229,72 @@ public PropertyMap Add(KeyValue keyValue)
...
             return this;
         }
 
+        /// <summary>
+        /// Removes the element by the specified key.
+        /// </summary>
+        /// <param name="key">The index key to find.</param>
+        /// <returns>True if the element is removed, false otherwise.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool Remove(PropertyKey key)
```
myroot, line 256: Use exception for the consistency
```suggestion
        public void Remove(PropertyKey key)
```

myroot:
Are you test done?
I wonder `CSharp_Dali_Property_Map_SetValue_IntKey` this method working well when key was not existed on the map

src/Tizen.NUI/src/public/Common/PropertyMap.cs
huayongxu: Done.

src/Tizen.NUI/src/public/Common/PropertyMap.cs
huayongxu: Done

src/Tizen.NUI/src/public/Common/PropertyMap.cs
huayongxu: Done

huayongxu:
> Are you test done? I wonder `CSharp_Dali_Property_Map_SetValue_IntKey` this method working well when key was not existed on the map

Yes. it works.
In NUI side, SetValue is internal, not public:
            internal set
            {
                using (PropertyKey pKey = new PropertyKey(key))
                {
                    SetValue(pKey, value);
                }
            }

There are 2 solutions:
1) Make set public, make insert/add support duplicated keys.
2) Make set internal, make insert/add not check duplicated keys.

This patch is using solution 2).
Maybe solution 1) is better.

What do you think?

myroot:
I don't understand your means, I don't care to use `SetValue`, it should be private. 

And `PropertyMap.Add` and `Insert` are working like a C# Dictonary, 
Dictonary.Add are throw exception when same key was added

My fisrt question is

You are use `SetValue` in Add method, in this case, Key are not existed on map, 
and I also see native dali code, `CSharp_Dali_Property_Map_SetValue_IntKey` it use PropertyMapp's index operator
and it search a key, so I wonder it is working well when key is not exist on the map

huayongxu:
> I don't understand your means, I don't care to use `SetValue`, it should be private.
> 
> And `PropertyMap.Add` and `Insert` are working like a C# Dictonary, Dictonary.Add are throw exception when same key was added
> 
> My fisrt question is
> 
> You are use `SetValue` in Add method, in this case, Key are not existed on map, and I also see native dali code, `CSharp_Dali_Property_Map_SetValue_IntKey` it use PropertyMapp's index operator and it search a key, so I wonder it is working well when key is not exist on the map

Yes, CSharp_Dali_Property_Map_SetValue_IntKey works well if the key does not exist.

src/Tizen.NUI/src/public/Common/PropertyMap.cs
huayongxu: Dictionary.Remove returns false if key does not exist.
So PropertyMap follows this rule.

myroot (approved):

zg2nets (approved):
looks good！

PR #4083, opened 2022-03-30
[NUI][TCSACR-485] Add some methods for Loading

Seoyeon2Kim:

### Description of Change ###
- Make Play / Pause / Stop methods to public API on API level 10.
- These methods are for Loading animation behavior.

### API Changes ###
 - ACR: TCSACR-485
   https://code.sec.samsung.net/jira/browse/TCSACR-483

Added:
 - public void Play()
 - public void Pause()
 - public void Stop()

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Loading.cs
+++ b/src/Tizen.NUI.Components/Controls/Loading.cs
@@ -1,5 +1,5 @@
 ﻿/*
- * Copyright(c) 2021 Samsung Electronics Co., Ltd.
+ * Copyright(c) 2022 Samsung Electronics Co., Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI.Components/Controls/Loading.cs
```diff
@@ -339,6 +336,7 @@ public void Stop()
 
         private void Initialize()
         {
+            Focusable = true;
```
myroot, line 339: 포커스를 갖을 필요가 있나요?

Seoyeon2Kim: 아 넵 해당 라인 지워서 다시 업데이트하겠습니다.
테스트하며 같이 올라갔습니다 :(
리뷰 감사합니다~

myroot (approved):

dongsug-song:
the ACR has been accepted, so this is merged now

src/Tizen.NUI.Components/Controls/Loading.cs
```diff
@@ -1,5 +1,5 @@
 ﻿/*
- * Copyright(c) 2021 Samsung Electronics Co., Ltd.
+ * Copyright(c) 2022 Samsung Electronics Co., Ltd.
```
wiertel, line 2: If it's not a new file, then it would be better to put list or range:  "2021, 2022" OR "2021-2022"

Seoyeon2Kim: Thanks for reviewing this PR.
Yes, it's not a new file. However, as you know, NUI is csharp-binding of native DALi, and DALi has a rule that updates the copyright year when the file is updated. So, as far as I know, NUI also follows this rule.

PR #4094, opened 2022-03-31
[NUI][TEST] sync from nui-dali/NUITizenGallery

dongsug-song:
### Description of Change ###
[NUI][TEST] sync from nui-dali/NUITizenGallery
- this is not NUI code change, but it is NUI sample App change
- code sync from https://github.com/nui-dali/NUITizenGallery
- I tested this PR in local Ubuntu PC and the following command worked fine
- in TizenFX/test/NUITizenGallery folder => in console => $ code NUITizenGallery.code-workspace => "menu - Run - start debugging (F5)" worked fine

### API Changes ###
none


everLEEst (commented):
I think some of test is more latest than dali/NUITIzenGallery.
I'll update both about graphicsTypeTest but I'm not sure about others...

everLEEst (approved):
I think it would better to sync once and update again.
looks good to me.

PR #4096, opened 2022-03-31
[NUI][Tizen.NUI.PerfNew] Add View creation performance benchmark app

hinohie:
Add new sample app that can become performance profiling.

This sample is exactly same as dali-demo perf-new

TODO : Currently, OnAnimationFinished callback not comes well.
We should fix this bug


everLEEst (commented):
Looks good but I hope the test name changed more formal.
Such as Tizen.NUI.PerformanceTest or Tizen.NUI.PerformanceBenchMark

everLEEst (approved):
approved though I commentted about name.

dongsug-song (approved):
looks good to me

PR #4109, opened 2022-04-04
[NUI] Add GetWindow() to View

jaehyun0cho:
In NUI, there are some codes which use default window although the codes
are related to other window.
e.g. layout transition codes are attached to the default window.

To resolve the above issue, View provides GetWindow() internally.
To make View provide GetWindow(), Layer also provides GetWindow().

### API Changes ###

 - ACR:


jaehyun0cho:
https://github.com/Samsung/TizenFX/pull/4110 also covers this PR.

PR #4117, opened 2022-04-05
[NUI] DirectRendering support

jsr184:
### Description of Change ###

Added GLView.BackendMode enum to enable the DIrectRendering via NUI when using GLView.

The change doesn't make DirectRendering backend default. NUI still defaults to the original backend (native image).

### API Changes ###
 - ACR:


adeelkazmi (commented):
Looks like you're basing it off something else - there are too many files to review in this patch.

dongsug-song:
@jsr184 
Hello,
could you resolve the conflicts?
thank you.

PR #4121, opened 2022-04-06
[NUI] Fix to disable ThemeManager in tv profile

dongsug-song:
### Description of Change ###
[NUI] Fix to disable ThemeManager in tv profile
- in some place, `ExternalThemeEnabled` macro has to be inserted to avoid appfw-tizen-theme so loading.

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/public/Application/NUIApplication.cs
+++ b/src/Tizen.NUI/src/public/Application/NUIApplication.cs
@@ -486,7 +486,9 @@ namespace Tizen.NUI
         static public void Preload()
         {
             Interop.Application.PreInitialize();
+#if ExternalThemeEnabled
             ThemeManager.Preload();
+#endif
             IsPreload = true;
         }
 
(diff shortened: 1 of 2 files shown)


everLEEst (approved):
Looks good to me.

Seoyeon2Kim (approved):
LGTM

hinohie:
https://github.com/Samsung/TizenFX/pull/4121/commits/a74d65c6e276450496b05a5104bd5f0062480540#diff-0830238c7f0dfb64c0917e1754f52fcaf49be35400ac82c6ae79d142130c717bR177

Don't we need to block here?

dongsug-song:
> https://github.com/Samsung/TizenFX/pull/4121/files#diff-0830238c7f0dfb64c0917e1754f52fcaf49be35400ac82c6ae79d142130c717bR177
> 
> Don't we need to block here?

yes, you are absolutely right and I even have tested it.
but VD wants secure and absolute fix. and it is related to `Preload` so I just explicitly blocked the code to show the fix to user.

PR #4124, opened 2022-04-07
[NUI] change SetPosition(...) as a hidden API for internal use only

dongsug-song:
### Description of Change ###
[NUI] change SetPosition(...) as a hidden API for internal use only

### API Changes ###
none


hinohie (approved):
I think it will works well. Only move `SetPosition` is reasonable.
(Why `SetSize` didn't make hidden? --> Cause layoutParam issue)

hyunjushin (approved):

everLEEst (commented):
Looks good to me.
this API only used for internal framework app.

PR #4130, opened 2022-04-08
[NUI] Add Null check in Slider Dispose

Seoyeon2Kim:

### Description of Change ###
- 'editModeIndicator' of Slider class is not created all the time.
 So, added Null check for this variable when Dispose() is called.

### API Changes ###
- N/A


dongsug-song (approved):
looks good to me

jaehyun0cho (approved):
LGTM

everLEEst (commented):
I think it would better to check editModeIndicator != null

Seoyeon2Kim:
> I think it would better to check editModeIndicator != null

You're right! Thanks for the review!!!

kimcinoo (approved):

PR #4140, opened 2022-04-11
[NUI][XamlBuild] Fix build error when calling ExitXaml() if XamlOptimization set as 0.

AchoWang:
### Description of Change ###
**Sync from API10.**

This issue was reported by VD, If XamlOptimization set as 0, call ExitXaml would throw the exception : Error CS0103 The name 'ExitXaml' does not exist in the current context.

The reason:
Because the option 0 will not generate the method ExitXaml, so it will cause the build error.

Solution:
Generate an empty 'ExitXaml' method in the *.g.cs file.

After modification:
[global::System.CodeDom.Compiler.GeneratedCodeAttribute("Tizen.NUI.Xaml.Build.Tasks.XamlG", "4.0.0.0")]
internal void ExitXaml() {
}

### API Changes ###

 - ACR:


zhouleonlei (approved):
lgtm

wanchao-xu (approved):
Looks good to me

elishateng (approved):
LGTM

PR #4145, opened 2022-04-13
[NUI] Apply AppBar UX - Size, BackgroundColor, BackButton Icon

jaehyun0cho:
The latest AppBar UX has been applied.
Each component's Size, BackgroundColor and BackButton Icon have been
updated.

### API Changes ###

 - ACR:


src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
```diff
@@ -420,8 +420,8 @@ public Theme Create()
             // AppBar base style
             theme.AddStyleWithoutClone("Tizen.NUI.Components.AppBar", new AppBarStyle()
             {
-                Size = new Size(-1, 120),
-                BackgroundColor = new Color("#EEEFF1"),
+                Size = new Size(-1, 64),
+                BackgroundColor = new Color(0, 0, 0, 0),
```
everLEEst, line 424: How about to use Color.Transparent ?

everLEEst (approved):
we might have to use "sp" for the UX but there is minor issue on using "sp" in theme so I think this is okay. thank you!

src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
jaehyun0cho: Thank you for your comment! I have updated to use Color.Transparent.

PR #4148, opened 2022-04-13
[NUI][TEST] Components function coverage test

dongsug-song:
### Description of Change ###
[NUI] Components function coverage test
- use locally built latest dali and tizenfx packages.
- Picker's Dispose() causes crash => resolved by null check.
- need to fix some BLOCK test cases. (under debugging)
- fix Navigator internal testcase (not using `new Navigator()` but using `GetDefaultNavigator()`)
### API Changes ###
none


taehyub (approved):
LGTM

Seoyeon2Kim (approved):

PR #4151, opened 2022-04-14
[NUI][AT-SPI] Add to set AccessibilityName and Description in some components

Seoyeon2Kim:

### Description of Change ###
- Some NUI Components have text parts and they need to be connected with
 Accessibillity features.
 (`AccessibilityName` and `AccessibiityDescription`)

### API Changes ###
- N/A

### DIFF
--- a/src/Tizen.NUI.Components/Controls/AlertDialog.cs
+++ b/src/Tizen.NUI.Components/Controls/AlertDialog.cs
@@ -1,5 +1,5 @@
 /*
- * Copyright(c) 2021 Samsung Electronics Co., Ltd.
+ * Copyright(c) 2022 Samsung Electronics Co., Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -441,7 +441,7 @@ namespace Tizen.NUI.Components
         }
 
         /// <summary>
-        /// AccessibilityGetName.
+        /// Gets accessibility name.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
         protected override string AccessibilityGetName()
@@ -454,6 +454,15 @@ namespace Tizen.NUI.Components
             {
                 return Message;
             }
+        }
+
+        /// <summary>
+        /// Gets accessibility description.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected override string AccessibilityGetDescription()
+        {
+            return Message;
         }
 
         /// <summary>
--- a/src/Tizen.NUI.Components/Controls/Button.Internal.cs
+++ b/src/Tizen.NUI.Components/Controls/Button.Internal.cs
@@ -37,7 +37,7 @@ namespace Tizen.NUI.Components
         private bool styleApplied = false;
 
         /// <summary>
-        /// Get accessibility name.
+        /// Gets accessibility name.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
         protected override string AccessibilityGetName()
--- a/src/Tizen.NUI.Components/Controls/Button.cs
+++ b/src/Tizen.NUI.Components/Controls/Button.cs
@@ -383,7 +383,6 @@ namespace Tizen.NUI.Components
             internal set
             {
                 buttonText = value;
-                AccessibilityManager.Instance.SetAccessibilityAttribute(this, AccessibilityManager.AccessibilityAttribute.Label, buttonText.Text);
             }
         }
 
--- a/src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
+++ b/src/Tizen.NUI.Components/Controls/Navigation/AppBar.cs
@@ -1,5 +1,5 @@
 ﻿/*
- * Copyright(c) 2021 Samsung Electronics Co., Ltd.
+ * Copyright(c) 2022 Samsung Electronics Co., Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -536,6 +536,15 @@ namespace Tizen.NUI.Components
 
                 return defaultActionContent;
             }
+        }
+
+        /// <summary>
+        /// Gets accessibility name.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected override string AccessibilityGetName()
+        {
+            return Title;
         }
 
         /// <inheritdoc/>
--- a/src/Tizen.NUI.Components/Controls/RecyclerView/Item/DefaultGridItem.cs
+++ b/src/Tizen.NUI.Components/Controls/RecyclerView/Item/DefaultGridItem.cs
@@ -205,7 +205,6 @@ namespace Tizen.NUI.Components
             {
                 itemLabel = value;
                 layoutChanged = true;
-                AccessibilityManager.Instance.SetAccessibilityAttribute(this, AccessibilityManager.AccessibilityAttribute.Label, itemLabel.Text);
             }
         }
 
@@ -474,6 +473,15 @@ namespace Tizen.NUI.Components
             }
 
 
+        }
+
+        /// <summary>
+        /// Gets accessibility name.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected override string AccessibilityGetName()
+        {
+            return itemLabel.Text;
         }
 
         /// <summary>
--- a/src/Tizen.NUI.Components/Controls/RecyclerView/Item/DefaultLinearItem.cs
+++ b/src/Tizen.NUI.Components/Controls/RecyclerView/Item/DefaultLinearItem.cs
@@ -156,7 +156,6 @@ namespace Tizen.NUI.Components
             internal set
             {
                 itemLabel = value;
-                AccessibilityManager.Instance.SetAccessibilityAttribute(this, AccessibilityManager.AccessibilityAttribute.Label, itemLabel.Text);
             }
         }
 
@@ -210,7 +209,6 @@ namespace Tizen.NUI.Components
             internal set
             {
                 itemSubLabel = value;
-                AccessibilityManager.Instance.SetAccessibilityAttribute(this, AccessibilityManager.AccessibilityAttribute.Label, itemSubLabel.Text);
             }
         }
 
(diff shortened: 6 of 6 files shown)


src/Tizen.NUI.Components/Controls/AlertDialog.cs
```diff
@@ -190,6 +190,12 @@ private string InternalTitle
                 {
                     textLabel.Text = title;
                 }
+
+                if (Accessibility.Accessibility.IsEnabled && IsHighlighted)
+                {
+                    AccessibilityName = title;
```
aswigon, line 196: Updating `AccessibilityName` should be moved out of this `if` statement. Otherwise, its value would be outdated if Screen Reader is turned on dynamically after this code is executed (or if the name is changed while the control is not highlighted), resulting in wrong TTS feedback.

src/Tizen.NUI.Components/Controls/AlertDialog.cs
```diff
@@ -190,6 +190,12 @@ private string InternalTitle
...
+                    EmitAccessibilityEvent(AccessibilityPropertyChangeEvent.Name);
```
aswigon, line 197: It is not necessary to check `Accessibility.IsEnabled` before calling `Emit...`, all event emissions are guarded for `IsEnabled` (i.e. `IsUp` in DALi syntax, see `BridgeObject::Emit*()` in `DALI_ADAPTOR/dali/internal/accessibility/bridge/bridge-object.cpp`). This was introduced specifically so that DALi/NUI code would not have to check `IsEnabled` before every `Emit...`.


```c++
void BridgeObject::Emit(Accessible* obj, ObjectPropertyChangeEvent event)
{
  // ...
  
  if(!IsUp() || obj->IsHidden() || obj->GetSuppressedEvents()[AtspiEvent::PROPERTY_CHANGED])
  {
    return;
  }
```

aswigon (commented):
I've only commented on `AlertDialog`, but the same remarks pertain to other controls.

Setting `AccessibilityName` manually and having `protected override string AccessibilityGetName()` are generally mutually exclusive. So one of two approaches has to be taken:

1. Set `AccessibilityName` and remove override for `AccessibilityGetName` (because it becomes dead code).
2. (recommended) Do not set `AccessibilityName` and modify `AccessibilityGetName` to consult `InternalTitle` property instead (I'm assuming that this PR was created because of Screen Reader receiving wrong text in some scenarios).

See `DALI_TOOLKIT/dali-toolkit/devel-api/controls/control-accessible.cpp` (comments mine):

```c++
std::string ControlAccessible::GetName() const
{ // called by Screen Reader to fetch text to be sent to the TTS

  // ...

  if(!controlImpl.mAccessibilityGetNameSignal.Empty())
  { // GetNameSignal -> AccessibilityNameRequested (C#)
    // (highest priority, mostly for application code with dynamic content (e.g. label with current time in minutes and seconds))
    controlImpl.mAccessibilityGetNameSignal.Emit(name);
  }
  else if(!controlImpl.mAccessibilityName.empty())
  { // AccessibilityName (C#)
    // (mostly for application code with static content)
    name = controlImpl.mAccessibilityName;
  }
  else if(auto raw = GetNameRaw(); !raw.empty()) 
  { // GetNameRaw -> View.AccessibilityGetName (C#), not called if AccessibilityName is set!
    // (mostly for control implementations)
    name = raw;
  }
  else
  { // fallback
    name = Self().GetProperty<std::string>(Actor::Property::NAME);
  }
```

Same goes for `AccessibilityDescription` and `AccessibilityGetDescription`.

lol-github (commented):
Agree with Artur's review.

src/Tizen.NUI.Components/Controls/AlertDialog.cs
Seoyeon2Kim: Oh.. I got it.
I just copied other codes in Button.cs, Progress.cs, Slider.cs etc. Are they unnecessary now?
Then, I'll remove them together.
For example,
```C#
// Button.cs
                if (Accessibility.Accessibility.IsEnabled && IsHighlighted && String.IsNullOrEmpty(AccessibilityName) && GetAccessibilityNameSignal().Empty())
                {
                    EmitAccessibilityEvent(AccessibilityPropertyChangeEvent.Name);
                }
```

```C#
// Progress.cs
                if (Accessibility.Accessibility.IsEnabled && IsHighlighted)
                {
                    EmitAccessibilityEvent(AccessibilityPropertyChangeEvent.Value);
                }
```

Seoyeon2Kim: BTW, I'll change other classes like `AccessibilityGetName()`, etc.
Then, I don't need to add `EmitAccessibilityEvent(AccessibilityPropertyChangeEvent.Name)` separately, do I?

Seoyeon2Kim:
@aswigon and @lol-github , Thank you for your detailed explanation :)

I'd like to get your opinions about AccessibilityManager setter on NUI Components.
Some components use `AccessibilityManager.Instance.SetAccessibilityAttribute()` to set Accessibility information. However, as you know, they're internally changed to AT-SPI features.

```c++
void AccessibilityManager::SetAccessibilityAttribute(Actor actor, Toolkit::AccessibilityManager::AccessibilityAttribute type, const std::string& text)
{
  switch(type)
  {
    case Toolkit::AccessibilityManager::ACCESSIBILITY_LABEL:
      actor.SetProperty(Toolkit::DevelControl::Property::ACCESSIBILITY_NAME, text);
      break;

    case Toolkit::AccessibilityManager::ACCESSIBILITY_HINT:
      actor.SetProperty(Toolkit::DevelControl::Property::ACCESSIBILITY_DESCRIPTION, text);
      break;

    case Toolkit::AccessibilityManager::ACCESSIBILITY_TRAIT:
    case Toolkit::AccessibilityManager::ACCESSIBILITY_VALUE:
    default:
      break;
  }
}
```
So... should we change all codes to use AT-SPI methods directly in NUI Components?

lol-github:
> So... should we change all codes to use AT-SPI methods directly in NUI Components?

Yes, I would recommend to do this. And as soon as all references to AccessibilityManager are being eliminated, we should mark this API as deprecated and finally remove it preserving appropriate API withdrawal timing.

Regarding AccessibilityManager::AccessibilityAttribute, as it was shown above, the following mapping should be used to replace it in NUI components:

ACCESSIBILITY_LABEL -> override AccessibilityGetName()
ACCESSIBILITY_HINT -> override AccessibilityGetDescription()
ACCESSIBILITY_TRAIT -> set property AccessibilityRole
ACCESSIBILITY_VALUE -> implement IAtspiValue

src/Tizen.NUI.Components/Controls/AlertDialog.cs
```diff
@@ -441,7 +441,7 @@ private View InternalActionContent
         }
 
         /// <summary>
-        /// AccessibilityGetName.
+        /// Gets AccessibilityName.
```
aswigon, line 444: It does not get `AccessibilityName`, because `AccessibilityName` is empty (and `AccessibilityName` takes priority over `AccessibilityGetName` anyway); it gets the `Title` (or `"accessibility name"` as described in `DefaultLinearItem`).

src/Tizen.NUI.Components/Controls/AlertDialog.cs
```diff
@@ -456,6 +456,15 @@ protected override string AccessibilityGetName()
             }
         }
 
+        /// <summary>
+        /// Gets AccessibilityDescription.
```
aswigon, line 460: ditto (and same in other files)

src/Tizen.NUI.Components/Controls/AlertDialog.cs
aswigon: Checking `Accessibility.IsEnabled` is unnecessary before `EmitAccessibilityEvent` (as already explained, `Emit` checks `IsEnabled` before sending anything over D-Bus). It looks like the existing NUI code predates this DALi patch: https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-adaptor/+/268774/, so it can be simplified now to take advantage of said patch.

However, it really depends on the UX guide and the expected TTS feedback whether to `EmitAccessibilityEvent` and where and when to do so. If (for the currently highlighted control) it is expected to hear the new name / description / value when any of these change, there should be an `EmitAccessibilityEvent` (as can be seen for `Button` and `Progress`). For `Progress` the UX guide specifies that the new value should be read aloud when it changes (and that's why we `Emit...(Value)` there), similarly, `Button` changing its title on click is also expected to cause TTS feedback. But in the general case, not every change of name should be propagated to the Screen Reader in order not to spam the user with TTS feedback (an alert popup title with countdown timer in milliseconds is an extreme example of this).

src/Tizen.NUI.Components/Controls/AlertDialog.cs
Seoyeon2Kim: Got it!

src/Tizen.NUI.Components/Controls/AlertDialog.cs
Seoyeon2Kim: Done.

src/Tizen.NUI.Components/Controls/AlertDialog.cs
Seoyeon2Kim: Now I understood. Thanks.

Seoyeon2Kim:
> > So... should we change all codes to use AT-SPI methods directly in NUI Components?
> 
> Yes, I would recommend to do this. And as soon as all references to AccessibilityManager are being eliminated, we should mark this API as deprecated and finally remove it preserving appropriate API withdrawal timing.
> 
> Regarding AccessibilityManager::AccessibilityAttribute, as it was shown above, the following mapping should be used to replace it in NUI components:
> 
> ACCESSIBILITY_LABEL -> override AccessibilityGetName() ACCESSIBILITY_HINT -> override AccessibilityGetDescription() ACCESSIBILITY_TRAIT -> set property AccessibilityRole ACCESSIBILITY_VALUE -> implement IAtspiValue

Thanks, but as you know, the current Family Hub (by DA division) uses `AccessibilityManager` methods a lot. 
So, we couldn't remove them at once. Instead, we should add deprecated warning messages. :)

kimcinoo (approved):
Looks good to me.

hyunjushin (approved):

dongsug-song (approved):
approved

PR #4154, opened 2022-04-19
[NUI] Minimize size is set too large. And code clean up

JoogabYun:
### Description of Change ###

[NUI] Minimize size is set too large. And code clean up.


dongsug-song (approved):
approved

PR #4160, opened 2022-04-20
[NUI] apply Tizen 7.0 UX on CollectionView and Scrollbar and add samples.

everLEEst:
apply Tizen 7.0 UX on
DefaultLinearItem
DefaultGridItem
DefaultTItleItem
ScrollBar

and add few changes.
bug fix about LInearItem and GridItem when it used StandAlone.
add ScrollbarStyle Thumb internal ImageViewStyle.
EnableControlStatePropagation true on DefaultLienarItem and GridItem.

and add sample in Style Guide.

### API Changes ###

 - ACR:


everLEEst:
Currently it has few bug in the sample.
1. DefaultLinearItem Two Line is not show correctly.
2. DefaultLinearItem Icon and Extra Margin is not properly applied.
3. DefaultGridItem text is not property placed.
4. DefaultGridItem Image cannot be clipped cornerRadius when it's parent is ClipToChildren.

Bug report is issued https://github.com/Samsung/TizenFX/issues/4161.

PR #4165, opened 2022-04-20
[NUI] Fix CustomView.GetNaturalSize() to return Size2D set by user

jaehyun0cho:
Originally, CustomView.GetNaturalSize() returned size set by user
explicitly.

It was changed by https://github.com/Samsung/TizenFX/pull/2515.
By the above PR, CustomView.GetNaturalSize() returned Size2D which can
be set by both user and Layout.

Since the current CustomView.GetNaturalSize() cannot distinguish the
size set by user from the size set by Layout, the size set by Layout
previously can be preserved incorrectly and the size is not updated
incorrectly.

To resolve the above issue, CustomView.GetNaturalSize() returns Size2D
set by user explicitly to make Layout preserve the size only set by user
explicitly.

e.g. problem case

var parent = new Control()
{
    Layout = new AbsoluteLayout(),
    WidthSpecification = LayoutParamPolicies.MatchParent,
    HeightSpecification = LayoutParamPolicies.WrapContent,
    BackgroundColor = Color.Red,
};
window.Add(parent);

var child = new Control()
{
    WidthSpecification = LayoutParamPolicies.MatchParent,
    HeightSpecification = 200,
    BackgroundColor = Color.Blue,
};
parent.Add(child);

var timer = new Timer(1000);
timer.Tick += (object sender, Timer.TickEventArgs args) =>
{
    child.HeightSpecification = 100;
    return false;
};
timer.Start();

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/CustomView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/CustomView.cs
@@ -320,7 +320,7 @@ namespace Tizen.NUI.BaseComponents
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1721: Property names should not match get methods")]
         public new virtual Size2D GetNaturalSize()
         {
-            return (Size2D)GetValue(Size2DProperty);
+            return GetUserSize2D();
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


dongsug-song (approved):
approved

PR #4166, opened 2022-04-20
[NUI] Fix LinearLayout to update max height and width correctly

jaehyun0cho:
Horizontal oriented LinearLayout decides max height among children.
Vertical oriented LinearLayout decides max width among children.

Previously, the max height and width might be calculated before all
children sizes were measured.
So the max height and width might be calculated with children's previous
height and width.

Now, the max height and width should be calculated after all children
sizes are measured.
So the max height and width are calculated with children's current
height and width.

e.g. problem case

var parent = new View()
{
    Layout = new LinearLayout(),
    WidthSpecification = LayoutParamPolicies.MatchParent,
    HeightSpecification = LayoutParamPolicies.WrapContent,
    BackgroundColor = Color.Red,
};
window.Add(parent);

var child = new View()
{
    WidthSpecification = LayoutParamPolicies.MatchParent,
    HeightSpecification = 200,
    BackgroundColor = Color.Blue,
};
parent.Add(child);

var timer = new Timer(1000);
timer.Tick += (object sender, Timer.TickEventArgs args) =>
{
    child.HeightSpecification = 100;
    return false;
};
timer.Start();

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/LinearLayout.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LinearLayout.cs
@@ -342,7 +342,6 @@ namespace Tizen.NUI
                 float childWeight = childLayout.Owner.Weight;
                 Extents childMargin = childLayout.Margin;
                 float childMarginWidth = childMargin.Start + childMargin.End;
-                float childMarginHeight = childMargin.Top + childMargin.Bottom;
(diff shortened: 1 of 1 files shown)


Seoyeon2Kim (approved):
LGTM

dongsug-song (approved):
approved

PR #4167, opened 2022-04-21
[NUI] Fix Svace issue not to do unnecessary null check in DefaultBorder

jaehyun0cho:
Svace issue occurs because null check of BorderWindow is done after
BorderWindow is accessed in DefaultBorder.Dispose().

When close button in DefaultBorder is clicked, BorderWindow is disposed.
When BorderWindow is disposed, DisposeBorder() is called.
In DisposeBorder(), DefaultBorder is disposed.
When DefaultBorder is disposed, BorderWindow is under disposed but
BorderWindow is not set to be null yet.
Therefore, it is not necessary to do null check of BorderWindow in
DefaultBorder.Dispose() to resolve the Svace issue.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window/DefaultBorder.cs
+++ b/src/Tizen.NUI/src/public/Window/DefaultBorder.cs
@@ -673,10 +673,9 @@ namespace Tizen.NUI
             if (disposing)
             {
                 ClearWindowGesture();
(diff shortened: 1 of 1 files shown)


taehyub (approved):
LGTM

PR #4169, opened 2022-04-21
[NUI][AT-SPI] Apply AccessibilityHidden for Scrollbar

Seoyeon2Kim:

### Description of Change ###
- AT-SPI2 tree shows Scrollbar, Control, and ImageView when Scrollbar
class is just created.
- `Control` is trackView in Scrollbar, but is not necessary for at-spi
tree. So, it can be hidden in the tree.
- Applied AccessibilityHidden property to a View.

### API Changes ###
- N/A
-


dongsug-song (approved):
looks good to me

PR #4187, opened 2022-04-25
[NUI] Add IsWindowRotating()

sparrow74:
Add IsWindowRotationg function.
It is return with true when window is rotating.


dongsug-song (approved):
approved

PR #4195, opened 2022-04-26
[NUI][XamlBuild] Change XamlCTask / XamlGTask into NUIXamlCTask /  NUIXamlGTask.

AchoWang:
### Description of Change ###
**Sync code from API10.**

This requirements comes from the build error:

The "ValidateOnly" parameter is not supported by the "XamlCTask" task loaded from assembly: Tizen.NUI.XamlBuild.

Because NUI XamlCTask makes Xamarin, which is already existed in Tizen, so change XamlCTask into NUIXamlCTask, to avoid confusing.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #4196, opened 2022-04-26
[NUI] Apply AccessibilityHidden for Picker

lwc0917:
### Description of Change ###
Screen Reader want to read only currently visible items

### API Changes ###
No API Changes

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Picker.cs
+++ b/src/Tizen.NUI.Components/Controls/Picker.cs
@@ -60,6 +60,7 @@ namespace Tizen.NUI.Components
         //Dummy item count for loop feature. Max value of scrolling distance in 
         //RPI target is bigger than 20 items height. it can adjust depends on the internal logic and device env.
         private const int dummyItemsForLoop = 20;
+        private const int middleItemIdx = 2;
         private int startScrollOffset;
         private int itemHeight;
         private int startScrollY;
@@ -69,11 +70,14 @@ namespace Tizen.NUI.Components
         private int maxValue;
         private int minValue;
         private int lastScrollPosion;
+        private int accessibilityHiddenStartIdx;
         private bool onAnimation; //Scroller on animation check.
         private bool onAlignAnimation;
         private bool displayedValuesUpdate; //User sets displayed value check.
         private bool needItemUpdate; //min, max or display value updated check.
         private bool loopEnabled;
+        private bool isScreenReaderEnabled;
+        private bool isAtspiEnabled;
         private ReadOnlyCollection<string> displayedValues;
         private PickerScroller pickerScroller;
         private View upLine;
@@ -373,6 +377,18 @@ namespace Tizen.NUI.Components
             }
         }
 
+        private void AccessibilityEnabled()
+        {
+            if (loopEnabled) ShowItemsForAccessibility(currentValue - middleItemIdx);
+            else
+            {
+                //Exception case handling condition state.
+                //If user sets 4 items it can scroll but not loop.
+                if (currentValue > (middleItemIdx * 2)) ShowItemsForAccessibility(middleItemIdx + 1);
+                else ShowItemsForAccessibility(middleItemIdx);
+            }
+        }
+
         private void Initialize()
         {
             HeightSpecification = LayoutParamPolicies.MatchParent;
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI.Components/Controls/Picker.cs
```diff
@@ -441,6 +476,7 @@ private void OnScroll(object sender, ScrollEventArgs e)
         private void OnScrollAnimationStarted(object sender, ScrollEventArgs e)
         {
             onAnimation = true;
+            if (isScreenReaderEnabled) Accessibility.Accessibility.ClearCurrentlyHighlightedView();
```
kimcinoo, line 479: If you mve this line into the OnScroll, then it will work as we want. ;-D

src/Tizen.NUI.Components/Controls/Picker.cs
```diff
@@ -413,10 +415,43 @@ private void Initialize()
 
             Focusable = true;
             KeyEvent += OnKeyEvent;
+
+            Accessibility.Accessibility.ScreenReaderEnabled += (s, e) => {
```
kimcinoo, line 419: You do not have to call ```GrabAccessibilityHighlight``` and ```ClearAccessibilityHighlight```,
and you might need to call ```UpdateAccessbilityHiddenItemList``` when the ```ScreenReaderEnabled```.

lwc0917: that's right. i missed it.

src/Tizen.NUI.Components/Controls/Picker.cs
lwc0917: Thank you for this guide!

src/Tizen.NUI.Components/Controls/Picker.cs
```diff
@@ -413,10 +416,76 @@ private void Initialize()
...
+                    //Exception case handling condition state.
+                    //If user sets 4 items it can scroll but not loop.
+                    if (currentValue > (middleItemIdx * 2)) ShowItemsForAccessibility(middleItemIdx + 1);
+                    else ShowItemsForAccessibility(middleItemIdx);
+                }
+            };
+            Accessibility.Accessibility.ScreenReaderDisabled += (s, e) => {
+                isScreenReaderEnabled = false;
+                HideItemsForAccessibility();
+                Accessibility.Accessibility.ClearCurrentlyHighlightedView();
```
kimcinoo, line 434: You do not have call ```ClearCurrentlyHighlightedView```
The screen-reader will clear highlight.

src/Tizen.NUI.Components/Controls/Picker.cs
```diff
@@ -413,10 +416,76 @@ private void Initialize()
...
+            };
+
+            isScreenReaderEnabled = Accessibility.Accessibility.IsScreenReaderEnabled;
+        }
```
kimcinoo, line 438: Do not have to do the same things of line 421 ~ 429, if the screen-reader is enabled after line 437?
```
                if (loopEnabled) ShowItemsForAccessibility(currentValue - middleItemIdx);
                else
                {
                    //Exception case handling condition state.
                    //If user sets 4 items it can scroll but not loop.
                    if (currentValue > (middleItemIdx * 2)) ShowItemsForAccessibility(middleItemIdx + 1);
                    else ShowItemsForAccessibility(middleItemIdx);
                }
```

kimcinoo (commented):
One BIG question.
Do you want to expose items if ```Accessibility.Accessibility.IsEnabled```?
There are three status.
(1) IsEnabled: True, IsScreenReaderEnabled: True - when the screen-reader AT client is working.
(2) IsEnabled: True, IsScreenReaderEnabled: False - when the Aurum AT client is working.
(3) IsEnabled: False, IsScreenReaderEnabled: False - if there is not an AT client.

This path would take care of case (1). You could take care of case (2) as well.
It is up to you. :-]

src/Tizen.NUI.Components/Controls/Picker.cs
lwc0917: I will remove this. but i think the hilight is not gone without this API calling.

src/Tizen.NUI.Components/Controls/Picker.cs
lwc0917: We do for this when value set up.

src/Tizen.NUI.Components/Controls/Picker.cs
```diff
@@ -413,10 +429,78 @@ private void Initialize()
...
+                isAtspiEnabled = true;
+                AccessibilityEnabled();
+            };
+            Accessibility.Accessibility.Disabled += (s, e) => {
+                isAtspiEnabled = false;
+                HideItemsForAccessibility();
+            };
+            Accessibility.Accessibility.ScreenReaderEnabled += (s, e) => {
+                isScreenReaderEnabled = true;
+                AccessibilityEnabled();
```
kimcinoo, line 443: This ```AccessibilityEnabled()``` and ```HideItemsForAccessibility()``` of line 447 are not neccessary,
because there is not the case that the ```IsScreenReaderEnabled``` is ```TRUE``` and ```IsEnabled``` is ```FALSE```.
i.e. If the ```IsScreenReaderEnabled``` is ```TRUE``` then the ```IsEnabled``` should be ```TRUE```.

lwc0917: Thank you for the clear explanation.

Seoyeon2Kim (approved):
Looks good to me, 
@kimcinoo and @everLEEst , could you review this PR please?

kimcinoo (approved):
Looks GR8 to me

m-bialota:
Dear all, 
I tested the Picker behavior with this patch, and I see the following:
- only visible items are being read (there is no auto-scrolling to the items that are not visible on screen).
But also I noticed that:
- changing the picker value is possible only by 2-finger scroll which is impossible in narrow pickers (such as Datepicker subcomponents), moreover you have to hit within the picker area correctly to do it
- reading the current value of the picker is impossible - there is no indication in screen reader output about what is the current value of the picker, so flicking right through the picker gives you only the currently visible elements.

lol-github:
I'd supplement Maria's observations, with few more points:

- it looks that the current UX guideline for Screen Reader and new carousel layout of date picker control (GUI_Tizen 6.0_Screen Reader available here: https://xd.adobe.com/view/1bc8f657-f2ab-4e3f-6bf5-d5b6efdbc8cc-6df9/screen/62766f48-e0e0-48fb-b0ee-1ecb61e66a96/ password: 6TizenGUI) claims only a central (currently selected) value in each category/picker should be highlighted: day, month, year. This seams to vary with this implementation, which allows to highlight all visible values in each particular category/picker. 
- the UX guideline mentioned above is silent about the user interaction (gesture) which should result in changing (increasing/decreasing) of current value of current category/picker in case where we have a group of pickers displayed on the screen at the same time. IMHO such interaction in accessibility mode shouldn't be bound to any particular region of the screen (e.g the region occupied by a particular picker instance) because the visually disabled user simply will not be able to determine such position. Rather, such interaction (gesture) should be allowed to occur anywhere on the screen and be matched to the currently highlighted (active) picker. Maybe this can be achieved (in separate dali-core patch) by overriding GetCurrentGesturedActor() in PanGestureProcessor to return currently highlighted Actor when accessibility mode is on?
- IMHO, determining and reading the current value of the picker in Screen Reader could be possible if Picker would implement (probably in separate patch) IAtspiValue interface.

lol-github:
I'd supplement Maria's observations, with few more points:

- it looks that the current UX guideline for Screen Reader and new carousel layout of date picker control (GUI_Tizen 6.0_Screen Reader) claims only a central (currently selected) value in each category/picker should be highlighted: day, month, year. This seams to vary with this implementation, which allows to highlight all visible values in each particular category/picker.
- the UX guideline mentioned above is silent about the user interaction (gesture) which should result in changing (increasing/decreasing) of current value of current category/picker in case where we have a group of pickers displayed on the screen at the same time. IMHO such interaction in accessibility mode shouldn't be bound to any particular region of the screen (e.g the region occupied by a particular picker instance) because the visually disabled user simply will not be able to determine such position. Rather, such interaction (gesture) should be allowed to occur anywhere on the screen and be matched to the currently highlighted (active) picker. Maybe this can be achieved (in separate dali-core patch) by overriding GetCurrentGesturedActor() in PanGestureProcessor to return currently highlighted Actor when accessibility mode is on?
- IMHO, determining and reading the current value of the picker in Screen Reader could be possible if Picker would implement (probably in separate patch) IAtspiValue interface.

src/Tizen.NUI.Components/Controls/Picker.cs
```diff
@@ -413,10 +429,76 @@ private void Initialize()
...
+                if (currentValue > (middleItemIdx * 2))
+                    for (int i = accessibilityHiddenStartIdx; i < (accessibilityHiddenStartIdx + (maxValue - minValue)); i++)
+                        itemList[i].AccessibilityHidden = true;
+                else
+                    for (int i = accessibilityHiddenStartIdx; i < (accessibilityHiddenStartIdx + (maxValue - minValue + 1)); i++)
+                        itemList[i].AccessibilityHidden = true;
+            }
+        }
+
+        private void ShowItemsForAccessibility(int startIdx)
```
lol-github, line 469: What about introducing public property that would allow to control how many items on both sides of the middle item should be shown for Accessibility? Value 0 would mean that only a middle item is shown to Accessibility, 1 - will extend the range by 1 item on each side of middle item, etc. The default value of this property could be adjusted to show to Accessibility all visible items (what would match the implementation from this patch), but when needed (for example in date picker) the range of items available for accessibility could be narrowed.

PR #4208, opened 2022-04-28
[NUI] Add SetFocusFinderRootView() and ResetFocusFinderRootView() api

JoogabYun:
### Description of Change ###

If you enable DefaultAlgorithm of FocusManager, you can automatically move the focus to the nearest View.

However, this is always root based on Window.

We want to change this root.

So I add two apis

1. SetFocusFinderRootView(View root)

  - This will only look for focusable Views within that View tree when looking for the next focus.

2. ResetFocusFinderRootView()

  - the root becomes the window again.

Dependency
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/274424/
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-toolkit/+/274422/

### API Changes ###

 - ACR:


dongsug-song (approved):
looks good to me

PR #4210, opened 2022-04-29
[NUI] Fix Window.Add and Remove to use Layer.Add and Remove directly

jaehyun0cho:
Previously, Window.Add and Remove did not use Layer.Add and Remove.
This caused LayoutCount was updated duplicately.
e.g.
- Calling window.Remove(view) twice decreases LayoutCount duplicately.
Moreover, users could not get ChildAdded and ChildRemoved events from
Window's RootLayer.

Now, Window.Add and Remove use Layer.Add and Remove directly.

### API Changes ###

 - ACR:


AchoWang (commented):
Looks good to me.

PR #4214, opened 2022-04-29
[NUI] Fix MarginProperty and PaddingProperty with Layout

jaehyun0cho:
View.Margin and Padding refers View.Layout.Margin and Padding.

So if View.Margin and Padding are set with Layout, then
View.Layout.Margin and Padding are set and View.Margin and Padding
become Extents(0, 0, 0, 0).

Previously, the problem was that View.MarginProperty and PaddingProperty
did not support the above logic properly.

Now, to resolve the above issue, View.MarginProperty and PaddingProperty
support the above logic properly.
i.e. Setting View.MarginProperty and PaddingProperty with Layout set
     value to View.Layout.Margin and Padding and also set
     Extents(0, 0, 0, 0) to View.Margin and Padding.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -2100,35 +2100,11 @@ namespace Tizen.NUI.BaseComponents
         {
             get
             {
-                // If View has a Layout then padding in stored in the base Layout class
-                if (Layout != null)
-                {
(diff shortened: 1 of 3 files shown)


jaehyun0cho:
This PR should be applied after https://github.com/Samsung/TizenFX/pull/4160

everLEEst (approved):
Looks good to me.

PR #4215, opened 2022-05-02
[NUI] Fix crash issue in widget

tscholb:
As the widget passes a string to use the creation function, string types makes crash.
To avoid this problem, we use Intptr instead of string when widget sender its Id.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #4216, opened 2022-05-02
[NUI] Apply Slider UX - Thumb and Tracks

Seoyeon2Kim:

### Description of Change ###
- The latest Slider UX have been applied. (Based on API10)
 The colors and sizes of Thumb(Handle), Progress, and Track are changed.

### API Changes ###
- N/A


jaehyun0cho (approved):
LGTM

everLEEst (approved):
Looks good to me.
but it seems size is increased quite a lot than before version so it may break other layout.

Seoyeon2Kim:
> Looks good to me. but it seems size is increased quite a lot than before version so it may break other layout.

Indeed. I worried about the comment you left.
That's why I modified the sample of Styleguide, but later I need to check other devices(profiles) on 7.0 version.

PR #4219, opened 2022-05-02
[NUI] Fix RelativeLayout to calculate ellipsis text from line target

jaehyun0cho:
It is commonly used that child view is positioned with its size from
line target in RelativeLayout.

If ellipsis text was positioned from line target in RelativeLayout, then
it was not positioned to the place where users intended.

e.g. Position textLabel at the bottom of parent.
     The target is the bottom line of the parent.
RelativeLayout.SetTopTarget(textLabel, parent);
RelativeLayout.SetTopRelativeOffset(textLabel, 1.0f);
RelativeLayout.SetBottomTarget(textLabel, parent);
RelativeLayout.SetBottomRelativeOffset(textLabel, 1.0f);
RelativeLayout.SetVerticalAlignment(textLabel, RelativeLayout.Alignment.End);

Previously, in the above situation, RelativeLayout could not position
ellipsis text at the bottom of the parent.
Instead, the ellipsis text was positioned outside of the parent.

Now, in the above situation, RelativeLayout position ellipsis text at
the bottom of the parent.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/RelativeLayout.cs
+++ b/src/Tizen.NUI/src/public/Layouting/RelativeLayout.cs
@@ -393,67 +393,80 @@ namespace Tizen.NUI
                 LayoutItem childLayout = LayoutChildren[i];
                 if (childLayout != null)
                 {
-                    bool ellipsisText = false;
+                    bool ellipsisTextWidth = false;
+                    bool ellipsisTextHeight = false;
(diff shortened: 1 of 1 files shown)


jaehyun0cho:
This PR resolves the issue https://github.com/Samsung/TizenFX/issues/4161

everLEEst (approved):
Looks good to me.

PR #4221, opened 2022-05-03
[NUI] Update Tizen 7.0 UX for Button and Check- sizes and text color

jaehyun0cho:
- change button and check sizes
- change icon size
- set space size between icon and text
- change text color

### API Changes ###

 - ACR:


taehyub (approved):
LGTM :)

PR #4222, opened 2022-05-03
[NUI] Fix CheckBox and Button.Icon sizes in theme

jaehyun0cho:
Since CheckBox.Text size can be increased, CheckBox size should not be
set in theme.
CheckBox.ItemSpacing is set to be 8px.
Since Button.Icon is able not to be set, Button.Icon size should not be
set in theme.

### API Changes ###

 - ACR:


taehyub (approved):
LGTM :)

Seoyeon2Kim (approved):

PR #4229, opened 2022-05-03
[NUI] Update control theme whose UX is not defined by Tizen 7.0

jaehyun0cho:
Tizen 7.0 UX does not include all controls.
However, to give the same look and feel to all controls, control theme
whose UX is not defined by Tizen 7.0 is also updated like others.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #4231, opened 2022-05-04
[NUI] Fix TabButton's size calculation logic with the latest UX

jaehyun0cho:
Based on the latest UX, TabButton's sub components' sizes are different
when Text/IconURL is set and when Text/IconURL is not set.

Therefore, TextButton's sub components' sizes should be updated when
Text/IconURL is set.

If sub components' sizes are updated in OnRelayout() like previous
calculation logic, then the sizes cannot be applied by NUI Layout.

The reason is as follows.
- Changing subcomponents' sizes calls NUI Layout's RequestLayout().
- NUI Layout's RequestLayout() calls Awake() to awake idling DALi main
  loop.
- DALi main loop calls NUI Layout's Process() and then calls DALi
  Relayout().
- NUI Layout's Process() applies changed sizes by NUI Layout.

Consequently, if NUI Layout's RequestLayout() is called in OnRelayout(),
then DALi main loop is finished without calling NUI Layout's Process()
again.

As a result, to resolve the above problem, TabButton's sub components'
sizes are updated when Text/IconURL is set.

Related samples are also updated.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

Seoyeon2Kim (approved):
LGTM!!! Great

PR #4240, opened 2022-05-10
[NUI] Fix Navigator not to make ContentPage.Content hide

jaehyun0cho:
ContentPage.Content becomes hidden when push/pop were finished to
unregister accessibility.
i.e. Show() registers accessibility.
     Hide() unregisters accessibility.

Previously, ContentPage.Content was invisible when push/pop were
started because the invisible state remained since the previous
push/pop were finished.
e.g. PageTransitionSample in Tizen.NUI.Samples

Now, ContentPage.Content is visible when push/pop are started.
i.e. SetVisible(true) is called for ContentPage.Content to make it
     visible but not to register accessibility at that time.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

everLEEst (approved):
LGTM.

PR #4244, opened 2022-05-10
[NUI] Fix ThumbColor property to change its own color

Seoyeon2Kim:

### Description of Change ###
- Currently, when the user sets ThumbColor, the background color of
 thumb is changed.
- Fixed to change the thumb color itself, not the background color.

### API Changes ###
- N/A

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Slider.cs
+++ b/src/Tizen.NUI.Components/Controls/Slider.cs
@@ -613,14 +613,40 @@ namespace Tizen.NUI.Components
         {
             get
             {
-                return thumbImage?.Color;
+                return thumbColor;
             }
             set
             {
                 if (null != thumbImage)
                 {
-                    thumbImage.BackgroundColor = value;
                     thumbColor = value;
+
+                    if (thumbImage.ResourceUrl != null)
+                    {
+                        thumbImage.ResourceUrl = null;
+                    }
+
+                    using (PropertyMap map = new PropertyMap())
+                    {
+                        // To remove CA2000 warning messages, use `using` statement.
+                        using (PropertyValue type = new PropertyValue((int)Visual.Type.Color))
+                        {
+                            map.Insert((int)Visual.Property.Type, type);
+                        }
+                        using (PropertyValue color = new PropertyValue(thumbColor))
+                        {
+                            map.Insert((int)ColorVisualProperty.MixColor, color);
+                        }
+                        using (PropertyValue radius = new PropertyValue(0.5f))
+                        {
+                            map.Insert((int)Visual.Property.CornerRadius, radius);
+                        }
+                        using (PropertyValue policyType = new PropertyValue((int)VisualTransformPolicyType.Relative))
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI.Components/Controls/Slider.cs
```diff
@@ -613,14 +613,20 @@ private Color InternalThumbColor
...
                 {
-                    thumbImage.BackgroundColor = value;
                     thumbColor = value;
+
+                    PropertyMap map = new PropertyMap();
+                    map.Insert((int)Visual.Property.Type, new PropertyValue((int)Visual.Type.Color));
+                    map.Insert((int)ColorVisualProperty.MixColor, new PropertyValue(thumbColor));
+                    map.Insert((int)Visual.Property.CornerRadius, new PropertyValue(0.5f));
+                    map.Insert((int)Visual.Property.CornerRadiusPolicy, new PropertyValue((int)VisualTransformPolicyType.Relative));
+                    thumbImage.Image = map;
```
Seoyeon2Kim, line 629: Hello, @Samsung/nui guys!
I'd like to get your opinions.
This change is currently to set `Image` property of `ImageView`. However, it can be `Color` property of `View` instead.
```
                if (null != thumbImage)
                {
                    thumbColor = value;
                    thumbImage.Color = value;
                }
```

But, when I tested with `Color` property, the image quality is a little low at first before moving the thumb.
So, I chose `Image` property like this.
What do you think?
This code is to create PropertyMap instance locally and change color visual.

hinohie: `thumbImage.Color` is the color of Actor.. so If we setup like this, the result color becom "mixed" as "original image" x "Color".

I think we need to think about the fixed-way.

1. To make similar grammer as `View.BackgroundImage` vs `View.BackgroundColor`
2. Make default thumbImage color as white, and make default thumbColor as blue, so just keep to use this property as "mixed" role.

Seoyeon2Kim: As we discussed, I use `Image` property as it is now. (it means `ImageVisual` for thumb color)
Just updated to remove warning messages!

hinohie (approved):
Looks God To Me. It will works in normal usecase.

jaehyun0cho (approved):
LGTM

jaehyun0cho (changes requested):
Oh, by applying this patch, Thumb color is changed correctly initially, but the Thumb color seems blended with the image color.

Seoyeon2Kim:
> Oh, by applying this patch, Thumb color is changed correctly initially, but the Thumb color seems blended with the image color.

Oh :( Thanks for checking the bug!!! I fixed it.

everLEEst (approved):
Looks good to me.
but please keep monitoring this patch as it might be changing previous ThumbColor behavior.
thank you.

jaehyun0cho (approved):
LGTM

PR #4256, opened 2022-05-11
[NUI] Remove useless API in FlexLayout

hinohie:
FlexLayout is not BaseHandle.
We don't need to follow up API pattern of BaseHandle.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #4263, opened 2022-05-13
[WebRTC] Add new DataChannel buffered amount APIs

hsgwon:
### Description of Change ###

Add new DataChannel buffered amount APIs

User can set threshold value using `BufferedAmountLowThreshold` property and the amount of buffered data is lower than threshold value, `BufferedAmountLow` event will be raised.

### API Changes ###

 - TCSACR-480


aferin (approved):
LGTM

PR #4265, opened 2022-05-13
[NUI][AT-SPI] Make GetAccessibilityRelations() usable

aswigon:
### Description of Change ###

`GetAccessibilityRelations()` used to return an `AddressCollection`, however the `Address` type is useless in NUI, because (1) it is not possible to convert an `Address` to a `View`, and (2) it is not possible to obtain an `Address` of a `View`. This PR makes `GetAccessibilityRelations()` return a dictionary mapping `AccessibilityRelationType` to a collection of `View`. `Address` and `AddressCollection` are removed altogether with this PR.

Dependencies (merge together with this PR):
* https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-adaptor/+/275033/
* https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-toolkit/+/275035/
* https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/275034/

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
@@ -16,115 +16,13 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Runtime.InteropServices;
 using Tizen.NUI;
 
 namespace Tizen.NUI.BaseComponents
 {
-    /// <summary>
-    /// Address represents an unique object address on accessibility bus.
-    /// </summary>
-    [EditorBrowsable(EditorBrowsableState.Never)]
-    public class Address
-    {
-        /// <summary>
-        /// Creates an initialized Address.
-        /// </summary>
-        /// <param name="bus">Accessibility bus</param>
-        /// <param name="path">Accessibility path</param>
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        public Address(string bus, string path)
-        {
-            Bus = bus;
-            Path = path;
-        }
-
-        /// <summary>
-        /// Gets or sets accessibility bus.
-        /// </summary>
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        public string Bus { get; set; }
-
-        /// <summary>
-        /// Gets or sets accessibility path.
-        /// </summary>
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        public string Path { get; set; }
-    }
-
-    /// <summary>
-    /// AddressCollection.
-    /// </summary>
-    [EditorBrowsable(EditorBrowsableState.Never)]
-    public class AddressCollection : SafeHandle
-    {
-        /// <summary>
-        /// Creates an initialized AddressCollection.
-        /// </summary>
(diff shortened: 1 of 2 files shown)


lol-github (approved):
LGTM, far less code and usable output from GetAccessibilityRelations()

src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
```diff
@@ -285,15 +183,44 @@ public void ClearAccessibilityRelations()
...
         }
 
+        private static void GetAccessibilityRelationsCallback(int relationType, IntPtr relationTarget, IntPtr userData)
+        {
+            var handle = GCHandle.FromIntPtr(userData);
+            var list = (List<KeyValuePair<int, IntPtr>>)handle.Target;
+
+            list.Add(new KeyValuePair<int, IntPtr>(relationType, relationTarget));
+        }
+
```
kimcinoo, line 223: How did you get this idea? Very GR8!! Thumbs UP!!
We have a ```TizenFX/test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/AtspiSample.cs```
It would be better we have an example code for this. It could be separated a PR.

kimcinoo (commented):
GR8!

src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
```diff
@@ -285,15 +183,44 @@ public void ClearAccessibilityRelations()
...
+            Interop.ControlDevel.DaliToolkitDevelControlGetAccessibilityRelations(SwigCPtr, callback, GCHandle.ToIntPtr(listHandle));
+            listHandle.Free();
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+
+            var result = new Dictionary<AccessibilityRelationType, List<View>>();
+
+            foreach (var pair in list)
+            {
+                var key = (AccessibilityRelationType)pair.Key;
+                var value = this.GetInstanceSafely<View>(pair.Value);
```
hinohie, line 203: Wow! You don't miss this point!.
`pair.value` mallocated inside of csharp-binder side. But due to this instance getter, It will be safe in memory leak.
LGTM

hinohie (approved):

src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
aswigon: > How did you get this idea?

The idea of using a single interop with a callback function called for each item in the collection is based on this code example for `GCHandle` and `EnumWindows`: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.gchandle?view=net-6.0. The advantage is that there is only one interop needed (the previous version with `AddressCollection` required many).

> It would be better we have an example code for this. It could be separated a PR.

OK, I'll add that to my TODO list. BTW, we are also developing more Aurum tests for relations.

dongsug-song (approved):
approved

PR #4271, opened 2022-05-17
[NUI] Replace the part that throws an exception with an error message to keep the behavior of previous version

dongsug-song:
### Description of Change ###
[NUI] Replace the part that throws an exception with an error message to keep the behavior of previous version
- related to http://jira.vd.sec.samsung.net/browse/UNI-3344
- throwing an exception has been added in API10 first, this would cause backward compatibility violation, so the exception is replaced by error message

### API Changes ###
none


Seoyeon2Kim (approved):
LGTM

jaehyun0cho (approved):
LGTM

PR #4272, opened 2022-05-17
[NUI] Fix MenuItem to add TextLabel and Icon always

jaehyun0cho:
Previously, MenuItem added TextLabel and Icon only if their properties
were set properly.
e.g. Icon is added only if IconURL is set properly.

Now, MenuItem adds TextLabel and Icon regardless of their properties
like Button does.
Because TextLabel and Icon are MenuItem's default feature so they
should always be added to MenuItem.

Icon's default size in theme is removed until Button has a proper way
to support default Icon size.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/MenuItem.cs
+++ b/src/Tizen.NUI.Components/Controls/MenuItem.cs
@@ -29,8 +29,6 @@ namespace Tizen.NUI.Components
     {
         private bool selectedAgain = false;
 
-        private bool styleApplied = false;
-
         /// <summary>
         /// Creates a new instance of MenuItem.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI.Components/Controls/MenuItem.cs
```diff
@@ -66,31 +64,6 @@ protected override void Dispose(DisposeTypes type)
             base.Dispose(type);
         }
 
-        /// <summary>
-        /// Applies style to MenuItem.
-        /// </summary>
-        /// <param name="viewStyle">The style to apply.</param>
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        public override void ApplyStyle(ViewStyle viewStyle)
```
Seoyeon2Kim, line 74: Just question! 
Why do you remove ApplyStyle() override method here?
No need to exist anymore?

jaehyun0cho: It was added to check "styleApplied" in methods.
But all the methods are removed in this PR.
So yes, it is not required to exist any more.

everLEEst (commented):
I do not agree this is good idea to change default behavior of control's children's existence by UX changes...
if we really should change their default behavior by UX changes,
I think it need to be controlled by theme not by the code.

jaehyun0cho:
> I do not agree this is good idea to change default behavior of control's children's existence by UX changes... if we really should change their default behavior by UX changes, I think it need to be controlled by theme not by the code.

Yes, I totally agree with you.
In the discussion about adding children policy, people had the same opinion so they preferred keep adding children regardless of the states of control's properties.
So this PR keeps adding children regardless of the state of control's properties.

Previously, MenuItem added TextLabel and Icon conditionally but it should be changed and all other controls which does the same thing should be changed as well.

dongsug-song (approved):
approved

dongsug-song:
this will be revised later when the other component(Button)'s change is done.

PR #4295, opened 2022-05-27
[NUI] Fix bug that PreMultipliedAlpha not applied to Visual

hinohie:
Due to the PreMultipliedAlpha property used
both ImageView and ImageVisual internally,
(+ This property don't required re-create new Visual)

We need to make some new way to send this changness
news into ImageVisual.


dongsug-song (approved):
approved

PR #4296, opened 2022-05-27
[NUI]First use own bindingcontext

elishateng:
### Description of Change ###

1. First use own bindingcontext,if not, then use inherit bindingcontext

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

myroot:
It change from original code you need to check side effect
https://github.com/xamarin/Xamarin.Forms/blob/70e57274a2a1a010795eb1f9d2e1d857d7bfb4eb/Xamarin.Forms.Core/BindableObject.cs#L30

myroot:
If you have test set of binding features, please do test it and show the result

elishateng:
> If you have test set of binding features, please do test it and show the result

Hi, This patch will be reverted in https://github.com/Samsung/TizenFX/pull/4298.

PR #4297, opened 2022-05-30
[NUI] Add border resizing policy and add api to show/hide border when OverlayMode is true.

JoogabYun:
### Description of Change ###

[NUI] Add border resizing policy and add api to show/hide border when OverlayMode is true.

1. It can be proportionally resized or fixed according to the border resizing policy.
2. If OverlayMode is true, you can show/hide the border through api.
   Through this, you can show/hide the border by touch or specific key event.

### DIFF
--- a/src/Tizen.NUI/src/public/Window/BorderWindow.cs
+++ b/src/Tizen.NUI/src/public/Window/BorderWindow.cs
@@ -36,9 +36,7 @@ namespace Tizen.NUI
         private Layer borderWindowRootLayer = null;
         private Layer borderWindowBottomLayer = null;
         private bool isBorderWindow = false;
-        private bool isInterceptTouch = false;
 
-        private Timer overlayTimer;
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI/src/public/Window/BorderWindow.cs
```diff
@@ -84,6 +82,26 @@ public enum BorderDirection
...
+        public enum BorderResizePolicyType
+        {
+          /// <summary>
+          /// The window can be resized freely.
+          /// </summary>
+          Free = 0,
+          /// <summary>
+          /// The window is resized according to the ratio.
+          /// </summary>
+          Ratio = 1,
```
hyunjushin, line 99: How about KeepRatio?

hyunjushin (approved):

Seoyeon2Kim (approved):
LGTM

PR #4313, opened 2022-06-03
[NUI] Fix save/restore key focus defects when push/pop Diaglog

dongsug-song:
### Description of Change ###
[NUI] Fix save/restore key focus defects when push/pop Diaglog
- Fix this bug => https://github.com/Samsung/TizenFX/pull/4308#issuecomment-1144818753
- Dialog test => 
- Pop/PushWithTransition test => 

### API Changes ###
none


JoogabYun (approved):

everLEEst (approved):
Looks good to me.
but I wonder this is really navigator should have to control manually.
we might need some algorithm such as,
all object need to be store previous focused children and when they focused back,
they can restore prev focused children or not.

need more discussion regarding this feature though.

PR #4315, opened 2022-06-03
[NUI] Add CloseDelegate for closing the application by the border.

JoogabYun:
### Description of Change ###

When closing, it is necessary to distinguish between the application case and the window case.
So, in the case of Application, we add a delegate so that it can be properly closed through the delegate.


dongsug-song (approved):
approved

Seoyeon2Kim (approved):
LGTM

PR #4319, opened 2022-06-07
[NUI] Remove GetDescription() in TextLabelBindableProperty

wonrst:
remove GetDescription() for better performance.
and there is no need to convert string to enum.
because the HORIZONTAL_ALIGNMENT and VERTICAL_ALIGNMENT properties of DALi
can be set to an integer or string, and there is no burden if it is an integer(enum).

https://code.sec.samsung.net/confluence/pages/viewpage.action?pageId=269926545&focusedCommentId=269928026


dongsug-song (approved):
approve

jaehyun0cho (approved):
LGTM

PR #4321, opened 2022-06-07
[MediaTool] Add new Create API to support increasing ref count

hsgwon:
### Description of Change ###

Add a new Create() method to creates a new instance with increasing internal ref count.
This can be useful when user share current MediaPacket instance to other module and don't know the exact time to dispose it.
In this case, user creates a new instance with Create(MediaPacket) method and shares it to other module.
And then, each MediaPacket instances can be disposed independently without concerning other instance's life cycle.

### API Changes ###

 - TCSACR-489


hsgwon:
Please refer the following PR for the previous discussion.
https://github.com/Samsung/TizenFX/pull/4276

myroot (approved):

src/Tizen.Multimedia/MediaTool/MediaPacket.cs
```diff
@@ -51,6 +51,16 @@ internal MediaPacket(MediaFormat format)
             _buffer = new Lazy<IMediaBuffer>(GetBuffer);
         }
 
+        /// <summary>
+        /// Initializes a new instance of the MediaPacket class from a original MediaPacket.
```
wiertel, line 55: it's only internal, but consider "from a original" -> "from an original" or just "from original"

src/Tizen.Multimedia/MediaTool/MediaPacket.cs
```diff
@@ -656,6 +657,22 @@ public static MediaPacket Create(MediaFormat format)
             return new SimpleMediaPacket(format);
         }
 
+        /// <summary>
+        /// Creates an object of the MediaPacket based on the original MediaPacket and increase ref count of both MediaPacket instances.
```
wiertel, line 661: 1. "and increase" -> "and increases"
2. I'm a little confused about the word "both", I was expecting that the refcount is attached to underlying native media packet and both C# MediaPacket instances share that native one. If there is only one refcount and it is not possible that "both MediaPacket instances" have a different value of ref count, then please consider changing the description to sth like "and increases the ref count of media packet internal data" / "media packet buffer" / "native media packet" / "native media packet handle". 
3. I'm wondering where we should write "reference counter" instead of short "ref count", at least once, probably here. But the term "ref count" is well known, so it would be ok to keep only that short term.

src/Tizen.Multimedia/MediaTool/MediaPacket.cs
```diff
@@ -656,6 +657,22 @@ public static MediaPacket Create(MediaFormat format)
...
+        /// </summary>
+        /// <remarks>
+        /// After this instance creation, the internal data of both MediaPacket instances are the same and those ref count is greater than the original MediaPacket by 1.\n
```
wiertel, line 664: "those" need to be changed, if it references "MediaPacket instances" then "theirs" would be better, if it references "internal data" then some restructuring would be in place.

src/Tizen.Multimedia/MediaTool/MediaPacket.cs
hsgwon: I'll fix it to "from an original".

src/Tizen.Multimedia/MediaTool/MediaPacket.cs
hsgwon: How about this?
```
This method can be useful when user share MediaPacket instance to other module and don't know the exact time to dispose it.\n
In this case, user create a new instance with <see cref="Create(MediaPacket)"/> and share it to other module.\n
And then, each MediaPacket instances can be disposed independently without concerning other instance's life cycle.
```

src/Tizen.Multimedia/MediaTool/MediaPacket.cs
wiertel: yes, that would be good

src/Tizen.Multimedia/MediaTool/MediaPacket.cs
wiertel: great, just change "user create ... and share" -> "user creates ... and shares"

PR #4344, opened 2022-06-13
[NUI] Fix Key constructor's cMemoryOwn as false

dongsug-song:
### Description of Change ###
[NUI] Fix Key constructor's `cMemoryOwn` as false
- dali returns `Key` or `Touch`'s `IntPtr`and dali keeps and maintains the `Key` or `Touch`'s instance so the nui's wrapping `Key` class should not `delete` or not `free` the `IntPtr`
- `cMemoryOwn = true` causes native crash, so it needs be fixed as `false`

### API Changes ###
none


jaehyun0cho (approved):
LGTM

taehyub (approved):
LGTM

PR #4347, opened 2022-06-14
[NUI] Open text selection APIs

wonrst:
https://code.sec.samsung.net/jira/browse/TCSACR-461

This patch should have been applied in API9, but it was missing.
I made this PR to complete the pending ACR.


dongsug-song (approved):
this is a missing patch when API9's branching out.
the ACR has already been completed.
approve.

PR #4349, opened 2022-06-14
[NUI] Slider track make correct value.

hinohie:
Due to the track touch didn't apply offset of track,
the track touch didn't make correct value.

This patch apply the offset of bg track to the touch position.
Now the positions become correct value.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Slider.Internal.cs
+++ b/src/Tizen.NUI.Components/Controls/Slider.Internal.cs
@@ -568,59 +568,93 @@ namespace Tizen.NUI.Components
             }
         }
 
-        private int BgTrackLength()
+        private int GetBgTrackLength()
         {
             int bgTrackLength = 0;
+
+            int lowIndicatorOffset = GetBgTrackLowIndicatorOffset();
+            int highIndicatorOffet = GetBgTrackHighIndicatorOffset();
+
+            if (direction == DirectionType.Horizontal)
+            {
+                bgTrackLength = this.Size2D.Width - lowIndicatorOffset - highIndicatorOffet;
+            }
+            else if (direction == DirectionType.Vertical)
+            {
+                bgTrackLength = this.Size2D.Height - lowIndicatorOffset - highIndicatorOffet;
+            }
+
+            return bgTrackLength;
+        }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


jaehyun0cho (approved):
LGTM

src/Tizen.NUI.Components/Controls/Slider.Internal.cs
```diff
@@ -623,6 +623,51 @@ private int BgTrackLength()
...
+                if (direction == DirectionType.Horizontal)
+                {
+                    bgTrackOffset = (int)(thumbImage.Size.Width * 0.5f);
+                }
+                else if (direction == DirectionType.Vertical)
+                {
+                    bgTrackOffset = (int)(thumbImage.Size.Height * 0.5f);
+                }
+            }
+            else if (type == IndicatorType.Image)
```
everLEEst, line 642: Code on Image and Text seems very similar so could be merge them one single bracket by changing Size of type.

src/Tizen.NUI.Components/Controls/Slider.Internal.cs
```diff
@@ -623,6 +623,51 @@ private int BgTrackLength()
             return bgTrackLength;
         }
 
+        private int BgTrackOffset()
```
everLEEst, line 626: method name would be better to have verb such as GetBgTrackOffset()

everLEEst (commented):
some suggestions.
although code looks good to me.

src/Tizen.NUI.Components/Controls/Slider.Internal.cs
hinohie: If than, I'd better change the name of `BgTrackLength` also. I'll try to clean up it.

src/Tizen.NUI.Components/Controls/Slider.Internal.cs
hinohie: done.

src/Tizen.NUI.Components/Controls/Slider.Internal.cs
hinohie: done.

everLEEst (approved):
Looks good to me :)

PR #4351, opened 2022-06-15
[NUI] Add dispose.

JoogabYun:
### Description of Change ###

Dispose seems to be missing after creating and using value.


dongsug-song (approved):
you can use `using` like this. 
`using var val = new Uint16Pair((uint)positionSize.Width, (uint)positionSize.Height);`
but this is ok, approve.

PR #4353, opened 2022-06-15
[NUI] Fix not to do NUI.Components preloading on TV target

dongsug-song:
### Description of Change ###
[NUI] Fix not to do NUI.Components preloading on TV target
- NUI.Components preloading is not required in TV target because TV doesn't use NUI.Components.
- change Preload() method from Control class' to DefaultThemeCreator class'.
- tested in tv target and checked it works normally.

### API Changes ###
none


jaehyun0cho (approved):
LGTM

PR #4367, opened 2022-06-22
[NUI] Make the velocity of slider as 1% at key event

hinohie:
Previous code can move only 1 units when we try to change value
by Key event.

This patch make the value moved for 1%. It will change as same rates
even min/max value difference is big, or small.

TODO : Currently, we make it as const value 1%.
Should we need to make this value as property?


dongsug-song (approved):
approve

Issue #4369, opened 2022-06-23
[NUI] AccessibilityCalculateStatesWrapper Missing null check


myroot:
https://github.com/Samsung/TizenFX/blob/356b2ee819f33e19f15e668868339feebd086d98/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityWrappers.cs#L95-L109

view could be null, but didn't check null

PR #4370, opened 2022-06-23
[NUI] Fix AccessibilityCalculateStatesWrapper

myroot:
### Description of Change ###
 Fixes : #4369

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityWrappers.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityWrappers.cs
@@ -99,7 +99,7 @@ namespace Tizen.NUI.BaseComponents
                 return 0UL;
             
             ulong bitMask = 0UL;
-            
+
             lock (AccessibilityInitialStates)
             {
                 AccessibilityInitialStates.BitMask = initialStates;
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityWrappers.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityWrappers.cs
@@ -95,8 +95,11 @@ namespace Tizen.NUI.BaseComponents
         private static ulong AccessibilityCalculateStatesWrapper(IntPtr self, ulong initialStates)
         {
             View view = GetViewFromRefObject(self);
+            if (view == null)
+                return 0UL;
+            
             ulong bitMask = 0UL;
-
+            
             lock (AccessibilityInitialStates)
             {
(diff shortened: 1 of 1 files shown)


hinohie (approved):

aswigon:
That's very interesting. This static method is called from `nui-view-accessible.cpp`:

```c++
Accessibility::States NUIViewAccessible::CalculateStates()
{
  std::uint64_t states = ControlAccessible::CalculateStates().GetRawData64();

  states = CallMethod<Interface::ACCESSIBLE>(mTable->calculateStates, states);

  return Accessibility::States{states};
}
```

and I thought that `CallMethod` already contained all necessary assertions:

```c++
template<Interface I, typename R, typename... Args>
R NUIViewAccessible::CallMethod(R (*method)(RefObject*, Args...), Args... args) const
{
  DALI_ASSERT_DEBUG(method);
  DALI_ASSERT_ALWAYS(GetInterfaces()[I]);

  return method(Self().GetObjectPtr(), args...);
}
```

Note also that `Self` (defined in `actor-accessible.h`) used in the above code asserts for non-null Actor handle:

```c++
  Dali::Actor Self() const
  {
    auto handle = mSelf.GetHandle();

    // It is a bug if the Accessible outlives its Actor
    DALI_ASSERT_ALWAYS(handle);

    return handle;
  }
```

With every `NUIViewAccessible` instance being tied to a specific `View` (or a more derived class) instance, I cannot imagine why `view` could ever be null on entry to `AccessibilityCalculateStatesWrapper`. There are 36 `Accessiblity*Wrapper` methods in total all of them have always safely assumed a non-null `view` that comes from the conversion from `IntPtr`.

myroot:
> and I thought that CallMethod already contained all necessary assertions:

these assertion is only check native layer condition


and As I know C# handle is just wrapper handle of native actor. 
So, C# handle could be released even though native actor still alived.
I have got this situation several times.

I don't know when `AccessibilityCalculateStatesWrapper` was called, if you let me know when `AccessibilityCalculateStatesWrapper` was called, I can reproduce it with TestCase

myroot:
@dongsug-song Please merge it

myroot:
@aswigon Do you have any opinion on my comment?

aswigon:
It may be better to create a `private static View dummyView` and return it from `GetViewFromRefObject` to ensure that the return value is non-null (along with an error log, of course).

dongsug-song (approved):
approve

myroot:
> OK, but why check null in only one of 36 methods that use `GetViewFromRefObject`?

Because It is not my job. I just want to resolve my app issue

Code owner (@aswigon? or @dongsug-song?) need to finish this works

PR #4374, opened 2022-06-28
[NUI] Add function to set maximum/minimum size

sparrow74:
It is to set the maximized size when window is maximized or the size is increased by RequestResizeToServer.
Although the limitation is set by this function, window's size can be over the limitation by SetPositionSize() or SetSize().
The Set minimum size is similar to set maximum size.
Also it is only useful for tizen world.


JoogabYun (approved):

dongsug-song (approved):
approve

PR #4392, opened 2022-07-05
[NUI] When hiding the overlaid border, the timer is also initialized.

JoogabYun:
### Description of Change ###

When hiding the overlaid border, the timer is also initialized.

If the timer is not initialized, the border will not be visible when OverlayBorderShoew is performed again while the timer is alive.


Seoyeon2Kim (approved):
LGTM

jaehyun0cho (approved):
LGTM

PR #4394, opened 2022-07-07
[NUI] set default value for AutoScroll in TextLabel

wonrst:
NUI TextLabel no longer use dali-theme.
So set default values in NUI theme.
This values are the same as the values in prev dali-theme.


jaehyun0cho (approved):
LGTM

taehyub (approved):
LGTM

PR #4398, opened 2022-07-11
[NUI] Fixed border behavior of windows launched at full size

JoogabYun:
### Description of Change ###

For windows launched in full size, we need to calibrate the window as much as the border size for Maximized to work well.

Otherwise, the border area will be located outsize of screen.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

taehyub (approved):
LGTM :)

PR #4405, opened 2022-07-12
[Build] Add dali2-csharp-binder to BuildRequires

jaehyun0cho:
To depend on the latest dali2-csharp-binder, dali2-csharp-binder is
added to BuildRequires in spec.

### API Changes ###

 - ACR:

### DIFF
--- a/packaging/csapi-tizenfx.spec
+++ b/packaging/csapi-tizenfx.spec
@@ -30,6 +30,7 @@ BuildArch:   noarch
 AutoReqProv: no
 
 BuildRequires: dotnet-build-tools
+BuildRequires: pkgconfig(dali2-csharp-binder)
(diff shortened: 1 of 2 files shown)


JoogabYun (approved):

wonrst (approved):

hinohie:
Don't we need to add the depedency into `csapi-tizenfx.spec.in` instead of `csapi-tizenfx.spec`? This patch result will be overwrited

hinohie:
I think we should change both immediately.

jaehyun0cho:
@hinohie Thank you for your comment! I have updated spec.in file as well!

taehyub (approved):
LGTM :)

everLEEst (approved):
LGTM! great job!

PR #4407, opened 2022-07-12
[NUI] Make webcontext/cookiemanager static.

huayongxu:
### Description of Change ###

The related patch at csharp-binder side:
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/277709/

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.WebView.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.WebView.cs
@@ -30,6 +30,12 @@ namespace Tizen.NUI
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_WebView_New_3")]
             public static extern global::System.IntPtr New3(int jarg1, string[] jarg2);
 
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_WebView_GetContext")]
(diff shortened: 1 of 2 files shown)


myroot:
The orignal ewk API can create a new web context and it could be used for a new web view.

a default web context could be provide as static but some webview can use another web context instead of default web context.
so you should not assumed that all webview has same web context

huayongxu:
This patch is to fix the issue 
https://github.sec.samsung.net/dotnet/NUIBackend/issues/169
web context is shared among web views, is not created by web view itself.
when web engine(chromium-efl) is integrated into dali-extension,  default context is a singleton actually,

myroot:
> web context is shared among web views, is not created by web view itself.

It is a default behavior. as default WebView use a default web context. but original EWK API provide to create web view with a new web context.
http://ewebkit.github.io/ewebkit-doc/ewebkit2/html/d6/daf/ewk__view_8h.html#aa06060a9c00fa46acd92095de688ed61

And I asked the web team to add these API to C#.

myroot:
Anyway, currently NUI WebView API is not support to create WebContext, so this changes acceptable. 
But in the future, this structure should be refactored to allow multiple web context

zg2nets (approved):
looks good to me

elishateng (approved):
LGTM

AchoWang (approved):
Looks good to me.

dongsug-song (approved):
approved

PR #4413, opened 2022-07-14
[NUI] Add InsetsChanged event to Window

jaehyun0cho:
Window insets are the inner size between window content and window edge.

Window insets are required when status bar, virtual keyboard, or
clipboard appears on window to avoid window content obscured by them.

To notify the window insets changes, InsetsChanged event is added to
Window.

### API Changes ###

 - ACR:


jaehyun0cho:
The following patches should be verified and merged before verifying and merging this PR.
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-adaptor/+/277884/
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/277885/

everLEEst (commented):
We discussed & reviewed this issue offline.
Occlusion seems very general terms and our usage is very limited..
we need to find better candidate..

everLEEst (approved):
Looks good to me.
after few researching, @Jaehyun-Cho find good candidate term that used in other UI systems.

PR #4389, opened 2022-07-04
[NUI][AT-SPI] Store AccessibilityAttributes in NUI

aswigon:
### Description of Change ###

Having a modifiable attribute dictionary in NUI removes the performance penalty of having to go through interops on every modification. A new override for `GetAttributes` in `NUIViewAccessible` will now collect the attributes stored in NUI (through `AccessibilityDelegate.GetAttributes`). Moreover, the `class` attribute is now calculated lazily in order to reduce the creation cost of `View`.

Merge together with (ABI change):
* https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/277288/

### API Changes ###

 - ACR: None

PR #4414, opened 2022-07-15
[NUI][AT-SPI] Fix AutomationId for BaseComponents

aswigon:
### Description of Change ###

Due to an oversight in https://github.com/Samsung/TizenFX/pull/4389, the `automationId` attribute disappeared from the AT-SPI tree for some `View`s (notably, `BaseComponents`). This PR fixes this bug.

Three bindings for `AppendAccessibilityAttribute`, `RemoveAccessibilityAttribute` and `ClearAccessibilityAttributes` were recently removed. They were generally no longer necessary with the addition of `AccessibilityAttributes` to `View` (collected on demand by `NUIViewAccessible::GetAttributes`). However, some `View`s (those not backed by `ViewWrapperImpl`\*) use `Accessible` implementations other than `NUIViewAccessible`, e.g. `TextEditor` uses `TextEditorAccessible` or `WidgetView` uses `WidgetViewAccessible`, which are internal to other repositories and cannot call into C# to collect attributes, nor bindings for them can be created. For the one specific use case of `AutomationId`, a binding function needs to be restored.

Please note that `AccessibilityAttributes` is not the only C# API that won't work for `View`s that don't create a `ViewWrapperImpl`\* (and the `NUIViewAccessible` created by it, e.g. `BaseComponents` or Flux controls). Other APIs that will not work for such `View`s include:
* `AccessibilityGetName`, `AccessibilityGetDescription` etc.,
* `AccessibilityCalculateStates` etc.,
* `IAtspiSelection`, `IAtspiText`, `IAtspiValue` etc.

because the C++ implementations (from the respective `Accessible`-derived class) will be used instead (e.g. `ControlAccessible::GetAttributes`, `ControlAccessible::CalculateStates`, `TextEditorAccessible::InsertText` etc.).

\* Or potentially also `ViewImpl` if an [extra patch is applied](https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/277964/) (not necessary at the moment).

Merge together with:
* https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/277986/

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.ControlDevel.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.ControlDevel.cs
@@ -390,6 +390,9 @@ namespace Tizen.NUI
 
             [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Accessibility_SetAccessibilityDelegate")]
             public static extern IntPtr DaliAccessibilitySetAccessibilityDelegate(IntPtr arg1_accessibilityDelegate, int arg2_accessibilityDelegateSize);
+
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Accessibility_SetAccessibilityAttribute")]
+            public static extern void DaliAccessibilitySetAccessibilityAttribute(HandleRef arg1_control, string arg2_key, string arg3_value);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


m-bialota (approved):
Looks OK and fixes the issue!

aswigon:
Superseded by https://github.com/Samsung/TizenFX/pull/4426.

PR #4428, opened 2022-07-20
[NUI] Fix to change IsSelected when ControlState is changed

jaehyun0cho:
Previously, IsSelected property was not changed when its ControlState
was changed between Normal and Selected.

Now, IsSelected property is also changed when ControlState is changed.
So, SelectedChanged event's IsSelected is synchronized with its look.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Switch.cs
+++ b/src/Tizen.NUI.Components/Controls/Switch.cs
@@ -349,18 +349,15 @@ namespace Tizen.NUI.Components
         [EditorBrowsable(EditorBrowsableState.Never)]
         protected override void OnControlStateChanged(ControlStateChangedEventArgs controlStateChangedInfo)
         {
+            if (controlStateChangedInfo == null) throw new ArgumentNullException(nameof(controlStateChangedInfo));
             base.OnControlStateChanged(controlStateChangedInfo);
 
-            if (!IsSelectable)
+            if (IsSelectable)
             {
-                return;
-            }
-
-            bool previousSelected = controlStateChangedInfo.PreviousState.Contains(ControlState.Selected);
-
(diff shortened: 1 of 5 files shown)


everLEEst (approved):
Looks good to me.
we might need to reconsider whole design of ControlState is next generation of NUI,
but I agree this is the best what we can do for current NUI without breaking compatibility.
thanks for the work!

src/Tizen.NUI.Components/Controls/Switch.cs
```diff
@@ -351,16 +351,12 @@ protected override void OnControlStateChanged(ControlStateChangedEventArgs contr
...
             {
-                return;
-            }
-
-            bool previousSelected = controlStateChangedInfo.PreviousState.Contains(ControlState.Selected);
-
-            if (previousSelected != IsSelected)
-            {
-                OnSelect();
+                if (controlStateChangedInfo.PreviousState.Contains(ControlState.Selected) != controlStateChangedInfo.CurrentState.Contains(ControlState.Selected))
```
Seoyeon2Kim, line 356: As you can see, there's a build warning message here.
`In externally visible method 'void Switch.OnControlStateChanged(ControlStateChangedEventArgs controlStateChangedInfo)', validate parameter 'controlStateChangedInfo' is non-null before using it. If appropriate, throw an ArgumentNullException when the argument is null or add a Code Contract precondition asserting non-null argument.`

Probably, controlStateChangedInfo needs to check NULL before use?

jaehyun0cho: Thank you for your comment!
I have added the check code.

Seoyeon2Kim (approved):
LGTM!
Thanks for fixing it :)

PR #4430, opened 2022-07-20
[NUI] Add PropagatableControlStates Property for controlling ControlState propagation from parent.

everLEEst:
Parent View can propagate ControlState by setting EnableControlStatePropagation(default is true),
but child view cannot control propgations.
we know that some child view want to block the propagations,
or want to propagate specific states only.

this new property PropagatableControlStates will give child view to decide which states will be propagated and which isn't from the parent view.

For examples,

```childView.PropagatableControlStates = ControlState.All;```
will allow every states to be propagated(which is by default.)

```childView.PropagatableControlStates = ControlState.Disabled;```
will allow only DIsabled states to be propagated, others will be standalone.

```item.Extra.PropagatableControlStates = ControlState.Create();```
will disallow every states to be propagated.

you can use + operator for more than two states to be propagatable.

```cihldView.PropagatableControlStates = ControlState.Disabled + ControlState.Pressed;```

you can use - for remove state on current PropatableControlStates, but default value All has high priority of others,
so you need to remove All firstly.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -69,6 +69,7 @@ namespace Tizen.NUI.BaseComponents
         private Size internalSize = null;
         private Size2D internalSize2D = null;
         private int layoutCount = 0;
+        private ControlState propagatableControlStates = ControlState.All;
 
         static View()
         {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by jaehyun0cho — APPROVED** [review]
Thank you for this PR!
By this PR, each child view can decide whether it synchronizes its ControlState with its parent.

File: src/Tizen.NUI/src/public/BaseComponents/View.cs [hunk 1]
```diff
@@ -262,7 +263,91 @@ protected set
                 {
                     foreach (View child in Children)
                     {
-                        child.ControlState = value;
+                        ControlState Allowed = child.PropagatableControlStates;
```
Seoyeon2Kim, line 266: Just my curiosity :
Is `Allowed` a grammar of C# or just a variable here?
If it's a variable of ControlState, how about changing its name to `allowed`?

src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -262,7 +263,91 @@ protected set
...
+                        if (Allowed.Contains(ControlState.All))
+                        {
+                            child.ControlState = value;
+                        }
+                        else
+                        {
+                            ControlState prevControlState = child.ControlState;
+                            ControlState newControlState = prevControlState;
```
Seoyeon2Kim, line 274: I'm wondering why `child.ControlState` does not set directly to `newControlState`.
Will `prevControlState` be used again next time?

src/Tizen.NUI/src/public/BaseComponents/View.cs
everLEEst: sorry for the confusion. yes it should be renamed allowed. let me update quickly.

Seoyeon2Kim (approved):
LGTM!
Thanks for updating it.

PR #4432, opened 2022-07-21
[Tizen.Log] Replace getting filename with a primitive (fast) way

ppdl:
This enhances speed up to 5x faster than using library functions when
it comes to a burst call.

Signed-off-by: Youngjae Cho <y0.cho@samsung.com>

### API Changes ###

 - ACR:


hyotaekshim (approved):
LGTM

PR #4435, opened 2022-07-21
[NUI] fix ScrollableBase focus moving bug on non-scrollable case

everLEEst:
currently we do not check item can be moved focus to next candidate,
so when scrollableBase is non-scrollable,
it directly move next object not the item.

here I add to next focusable item existence.

note : this code makes unfocusable in none-focusable - non-scrollable case,
but not sure we can call normal behavior.
need to more discuss about this.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

jaehyun0cho (approved):
LGTM

JoogabYun (approved):

PR #4436, opened 2022-07-21
[NUI][DO NOT MERGE] Add View constructor for ViewImpl backend

aswigon:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -166,6 +166,14 @@ namespace Tizen.NUI.BaseComponents
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -166,6 +166,14 @@ internal View(ViewImpl implementation, bool shown = true) : this(Interop.View.Ne
             }
         }
 
+        public struct WantToUseViewImpl
```
everLEEst, line 169: As this struct is open public, I think we need to name it more carefully...
WantToUse is too informal and ViewImpl is not open public so I'm not sure using this name in public struct is good enough.

PR #4439, opened 2022-07-22
[NUI] Add UpdateProperty() api and revert "Fixed border behavior of windows launched at full size" patch

JoogabYun:
### Description of Change ###

1. Make the Border property updatable.

For now, it only applies when set in the constructor.

Modify so that the property can be applied when the property is set again.

2. revert "https://github.com/Samsung/TizenFX/pull/4398"

### DIFF
--- a/src/Tizen.NUI/src/public/Window/BorderWindow.cs
+++ b/src/Tizen.NUI/src/public/Window/BorderWindow.cs
@@ -48,6 +48,11 @@ namespace Tizen.NUI
         private float borderHeight = 0;
         private int screenWidth = 0;
         private int screenHeight = 0;
+
+        // for config
+        private Size2D minSize = null;
+        private Size2D maxSize = null;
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Window/BorderWindow.cs
```diff
@@ -120,6 +125,37 @@ public enum BorderResizePolicyType
...
+            {
+                if (minSize != borderInterface.MinSize)
+                {
+                    using Size2D mimimumSize = new Size2D(borderInterface.MinSize.Width, borderInterface.MinSize.Height + (int)borderHeight);
+                    SetMimimumSize(mimimumSize);
+                    minSize = borderInterface.MinSize;
+                }
+                if (maxSize != borderInterface.MaxSize)
+                {
+                    using Size2D maximuSize = new Size2D(borderInterface.MaxSize.Width, borderInterface.MaxSize.Height + (int)borderHeight);
```
everLEEst, line 144: typo of maximumSize?

PR #4440, opened 2022-07-22
[NUI] fix AnimatedVectorImageView's current frame set defect

dongsug-song:
### Description of Change ###
[NUI] fix AnimatedVectorImageView's current frame set defect
- AnimatedVectorImageView is hidden API and used only for VD
- CurrentFrame should be possible to set at out of the min/max range
- The CSFS's setting icon should be white color outline when losing the focus => this is fixed

### API Changes ###
none


taehyub (approved):
LGTM

everLEEst (commented):
LGTM.

everLEEst (approved):

PR #4449, opened 2022-07-27
[NUI] Fix API/ABI break on WebView

myroot:
### Description of Change ###
 `Context` / `CookieManager` change to instance property

 Why I change to instance property, It originally could be have a different value on each webview.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #4452, opened 2022-07-29
[NUI] Remove Previous ModelView, and modify new ModelView

bshsqa:
This patch removes previous ModelView in NUI that will not be used anymore.
And, newly added Scene3D.ModelView is modifed in this patch too.
Some methods are changed to properties.

### API Changes ###

 - ACR:


src/Tizen.NUI.Scene3D/src/ModelView.cs
```diff
@@ -76,58 +70,74 @@ public ModelView(ModelView modelView) : this(Interop.ModelView.NewModelView(Mode
...
 
         /// <summary>
-        /// Retrieves model root Actor.
+        /// Fits the model to the Control size.
         /// </summary>
-        /// <returns>Root View of the model.</returns>
         // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public View GetModelRoot()
+        public bool FitSize
```
everLEEst, line 92: have more information on  property name would be better.
how about FitSizeOnControl ?
or what about making SizeMode enum and give an options like FitOnControl?

I understand the name came from DALi, so if we must maintain original name, and also it is just simple change method to property...  it is okay.

src/Tizen.NUI.Scene3D/src/ModelView.cs
```diff
@@ -76,58 +70,74 @@ public ModelView(ModelView modelView) : this(Interop.ModelView.NewModelView(Mode
...
 
         /// <summary>
-        /// Fits the model to the Control size.
+        /// Moves the model to the center of control.
         /// </summary>
-        /// <param name="fit">True to fit model size to control.</param>
         // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public void FitSize(bool fit)
+        public bool FitCenter
```
everLEEst, line 111: is it possible to use FitSize and FitCenter both? is there any differences by using FitSize only?

src/Tizen.NUI.Scene3D/src/ModelView.cs
```diff
@@ -76,58 +70,74 @@ public ModelView(ModelView modelView) : this(Interop.ModelView.NewModelView(Mode
...
+        public View ModelRoot
         {
-            ModelView ret = new ModelView(Interop.ModelView.ModelViewAssign(SwigCPtr, ModelView.getCPtr(modelView)), false);
-            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
-            return ret;
+            get
+            {
+                View ret = new View(Interop.ModelView.GetModelRoot(SwigCPtr), false);
+                if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+                return ret;
```
everLEEst, line 83: could you please check this message?

get_ModelRoot creates an exception of type Exception, an exception type that should not be raised in a property. If this exception instance might be raised, use a different exception type, convert this property into a method, or change this property's logic so that it no longer raises an exception.

bshsqa: Thank you I added a method.

src/Tizen.NUI.Scene3D/src/ModelView.cs
bshsqa: Thank you I changed the name and added more comments.

src/Tizen.NUI.Scene3D/src/ModelView.cs
bshsqa: Thank you I added more comments.

everLEEst (approved):
Looks good to me.

Seoyeon2Kim (commented):
A minor comment :
@bshsqa , could you update 'Description of Change' of this PR?
such as the reason for changing the ModelView briefly.

Seoyeon2Kim (approved):
LGTM

PR #4461, opened 2022-08-03
[NUI] Modify GetWidgetViewFromPtr to use ManagedBaseHandle

tscholb:
When WidgetView uses its callback, WidgetView gets a crash problem because of referencing object.
To avoid this problem, WidgetView uses ManageBadeHandle.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
LGTM

everLEEst (approved):
Looks good to me.

jaehyun0cho (approved):
LGTM

PR #4466, opened 2022-08-05
[NUI] Temporarily use InterceptTouchEvent instead of FocusChanged.

JoogabYun:
### Description of Change ###

Registering the FocusChanged event at the time of OnCreate only in the emulator causes a crash.

Temporarily use InterceptTouchEvents until emulator problem is fixed.

This is a patch that will be reverted back once the emulator fix is ​​complete.

### API Changes ###

 - ACR:


everLEEst (changes requested):
Looks good to me.

Seoyeon2Kim (approved):
LGTM

everLEEst (approved):

PR #4469, opened 2022-08-08
[NUI] Fix Key constructor's cMemoryOwn as false

dongsug-song:
### Description of Change ###
[NUI] Fix Key constructor's cMemoryOwn as false
- same as https://github.com/Samsung/TizenFX/pull/4344
- dali returns Key or Touch's IntPtrand dali keeps and maintains the Key or Touch's instance so the nui's wrapping Key class should not delete or not free the IntPtr
- cMemoryOwn = true causes native crash, so it needs be fixed as false

### API Changes ###
none


taehyub (approved):
LGTM

PR #4471, opened 2022-08-08
Spannbale: Create core classes and ColorSpan

shrouqsabah:
### Description of Change ###
Spannable: it is a text module. It allows application developers to modify text style dynamically at runtime on character level.
The application can attach and unattach spans on range of text for text-controller.

BaseSpan: Base structure for different text-spans.
ColorSpan: Span to change fore color of font

### API Changes ###

Added:
 - Tizen.NUI.Text.Spans.BaseSpan // Class
 - Tizen.NUI.Text.Spans.ColorSpan // Class
 - Tizen.NUI.Text.Spans.ColorSpanBuilder // Class
 - Tizen.NUI.Text.Spans.ISpan // interface
 -  Tizen.NUI.Text.Spannable// Class
- public static void AttachSpan(TextLabel textLabel, BaseSpan span, int startIndex, int endIndex) //Method
- public static void UnAttachSpan(TextLabel textLabel, BaseSpan span)
- public static void BuildSpannedText(TextLabel textLabel)
- public static void AttachSpan(TextField textField, BaseSpan span, int startIndex, int endIndex)
- public static void UnAttachSpan(TextField textField, BaseSpan span)
- public static void BuildSpannedText(TextField textField)
- public static void AttachSpan(TextEditor textEditor, BaseSpan span, int startIndex, int endIndex)
- public static void UnAttachSpan(TextEditor textEditor, BaseSpan span)
- public static void BuildSpannedText(TextEditor textEditor)

### Testcases ###
Added the below to (Tizen.NUI.Devel.Tests.Ubuntu) and (Tizen.NUI.Tests):
- Tizen.NUI.Devel.Tests.PublicBaseSpanTest
- Tizen.NUI.Devel.Tests.PublicColorSpanTest
- Tizen.NUI.Devel.Tests.PublicColorSpanBuilderTest
- Tizen.NUI.Devel.Tests.PublicSpannableTest
- Tizen.NUI.Devel.Tests.TestUtils

**This PR depends on the below patches:**

- https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/279035
- https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/276651

**Here's an example how to use it**
- https://github.com/shrouqsabah/NUI-Test/tree/spannable_core_colorspan/text-spannable


src/Tizen.NUI/src/public/Text/Spans/ColorSpan.cs
```diff
@@ -0,0 +1,71 @@
...
+        internal ColorSpan(global::System.IntPtr cPtr, bool cMemoryOwn) : base(cPtr, cMemoryOwn)
+        {
+        }
+
+        /// <summary>
+        /// The fore-color in color-span.
+        /// </summary>
+        /// This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Vector4 ForeColor
```
myroot, line 38:
```suggestion
        public Color ForegroundColor
```

src/Tizen.NUI/src/public/Text/Spannable.cs
```diff
@@ -0,0 +1,255 @@
...
+
+
+        /// <summary>
+        /// Un-apply the given SpanStyle on all ranges of text.<br />
+        /// </summary>
+        /// <param name="textLabel">The TextLabel control containing the text.</param>
+        /// <param name="span">TThe span of style to apply it on text from startIndex to endIndex</param>
+        // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static void UnAttachSpan(TextLabel textLabel, BaseSpan span)
```
myroot, line 83:
```suggestion
        public static void DetachSpan(TextLabel textLabel, BaseSpan span)
```

src/Tizen.NUI/src/public/Text/Spans/ColorSpanBuilder.cs
```diff
@@ -0,0 +1,75 @@
...
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class ColorSpanBuilder : Disposable
+    {
+
+        /// <summary>
+        /// Creates a new ColorSpanBuilder object<br />
+        /// </summary>
+        // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static ColorSpanBuilder Initialize( )
```
myroot, line 34: The builder pattern is a good pattern when they have many configurable fields. 
But, ColorSpan has only one filed

So, it not good for builder pattern
I think it just enough factory method
```suggestion
        public static ColorSpan WithForegroundColor(Color color)
```

myroot: on android also builder pattern was not used to create span object
The builder pattern was only used to create spanableStrings, not spans.

src/Tizen.NUI/src/public/Text/Spans/ColorSpan.cs
shrouqsabah: Done.

src/Tizen.NUI/src/public/Text/Spannable.cs
shrouqsabah: Done

src/Tizen.NUI/src/public/Text/Spans/ColorSpanBuilder.cs
shrouqsabah: The factory method is added according to your suggestion. I agree with you this is easier for developer.

But, we need to keep Builder because we are adding more spans and most of them have many attributes.
We need to provide the same experience on all spans.
Also, we will add more APIs in builder to allow developer construct ColorSpan with other options. 

Thanks for your time and feedback.

myroot: Could you give me one example that has many attribute?

myroot: I think We can split a complex span into individual spans, for example FontSpan could be split into FontSizeSpan, FontFaceSpan, FontStyleSpan, FontColorSpan.

anyway, I think a complex span is not normal in span so Builder pattern should not apply to a common pattern on span

shrouqsabah: Here's a PR for FontSpan. But it is under testing.
https://github.com/shrouqsabah/TizenFX/pull/3

We can not split complex span into simple spans:
1) We have to integrate spans with tags in markup. Each tag should be mapped to span. ( we have features depend on this integration)
2) If we split complex spans then the developer will create maybe 3 to 5 spans for same range(s), imagine the number of spans and number of calling for attach. (we want to create a general span like the one in markup)
3) We can help developer through builders to create spans in many different options (especially we are using fluent builder)
 
Thanks for your time and feedback

myroot: OK, I respect you

dongsug-song:
@shrouqsabah 
Hello, 
Some patches has been merged only into the master branch, so the master branch and the DevelNUI branch were conflicted.
So I did hard reset the DevelNUI branch.
Could you rebase the patch and resolve the conflict?
thank you.

the following tip could be helpful :
(execute from the work branch)
1) git fetch upstream
2) git rebase --onto=upstream/DevelNUI HEAD~1
3) git push -f
(number of commits on the work branch has to be given after '~', for this case it is 1)

myroot:
I don't know why need `BuildSpannedText` method could you explain the reason?

shrouqsabah:
> I don't know why need `BuildSpannedText` method could you explain the reason?

Because we do not want to call Relayout and Rendering functions when the developer Attach span or/and Detach span.
This is better for the performance of application.

Usually the developer will attach various spans on many and different ranges. Do Relayout and rendering after each attach/detach could cause performance issue on application level.

So the developer can attach/detach many spans then call 'BuildSpannedText' to call Relayout and Rendering functions.

PR #4481, opened 2022-08-12
[NUI] Append window move evnet

sparrow74:
Append window move event
This event occurs when window's position is changed.


dongsug-song (approved):
approve

PR #4486, opened 2022-08-16
[NUI] Fix typo & bind error at FrameUpdateCallbackInterface

hinohie:
There was some mis-bind API and typo error in FrameUpdateCallbackInterface.cs.

This patch fix this error, and also some API (like BakeColor, GetColor)
bind correctly.

### DIFF
--- a/src/Tizen.NUI/src/public/Common/FrameUpdateCallbackInterface.cs
+++ b/src/Tizen.NUI/src/public/Common/FrameUpdateCallbackInterface.cs
@@ -72,11 +72,10 @@ namespace Tizen.NUI
             {
                 return false;
             }
-            bool ret = Interop.FrameUpdateCallbackInterface.FraemCallbackInterfaceGetPosition(proxyIntPtr, id, Vector3.getCPtr(position));
+            bool ret = Interop.FrameUpdateCallbackInterface.FrameCallbackInterfaceGetPosition(proxyIntPtr, id, Vector3.getCPtr(position));
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Common/FrameUpdateCallbackInterface.cs
```diff
@@ -141,40 +137,37 @@ protected bool BakeSize(uint id, Vector3 size)
...
+            bool ret = Interop.FrameUpdateCallbackInterface.FrameCallbackInterfaceBakeSize(proxyIntPtr, id, Vector3.getCPtr(size));
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
             return ret;
         }
 
-
         /// This will be public opened in next tizen after ACR done. Before ACR, need to be hidden as inhouse API.
         [EditorBrowsable(EditorBrowsableState.Never)]
-        protected bool SetScale(uint id, Vector3 scale)
+        protected bool GetScale(uint id, Vector3 scale)
```
jaehyun0cho, line 147: is out keyword required for scale?

jaehyun0cho: As @hinohie replied, out keyword is not required here because Vector3.getCPtr is used.

jaehyun0cho (approved):
LGTM

src/Tizen.NUI/src/public/Common/FrameUpdateCallbackInterface.cs
hinohie: We use `Vector3.getCPtr` Internally. We don't need to use out keyword here.

PR #4487, opened 2022-08-16
[NUI][API9] Fix typo & bind error at FrameUpdateCallbackInterface

hinohie:
There was some mis-bind API and typo error in FrameUpdateCallbackInterface.cs.

This patch fix this error, and also some API(like BakeColor, GetColor)
bind correctly.


jaehyun0cho (approved):
LGTM

PR #4488, opened 2022-08-16
[Tts][TCSACR-497] Add new API for service state information

stom-hwang:
Signed-off-by: Suyeon Hwang <stom.hwang@samsung.com>

### Description of Change ###

Add new API for service state information

### API Changes ###

 - ACR: https://code.sec.samsung.net/jira/browse/TCSACR-497

### DIFF
--- a/src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
+++ b/src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
@@ -212,25 +212,21 @@ namespace Tizen.Uix.Tts
         /// <summary>
         /// Ready state.
         /// </summary>
-        /// <since_tizen> 10 </since_tizen>
         Ready = 0,
 
         /// <summary>
         /// Synthesizing state.
         /// </summary>
-        /// <since_tizen> 10 </since_tizen>
         Synthesizing = 1,
 
         /// <summary>
         /// Playing state.
         /// </summary>
-        /// <since_tizen> 10 </since_tizen>
         Playing = 2,
 
         /// <summary>
         /// Unavailable state.
         /// </summary>
-        /// <since_tizen> 10 </since_tizen>
         Unavailable
     };
 
@@ -624,7 +620,7 @@ namespace Tizen.Uix.Tts
         }
 
         /// <summary>
-        /// Event to be invoked when state of TTS service changes.
+        /// Event to be invoked when the state of TTS service changes.
         /// </summary>
         /// <since_tizen> 10 </since_tizen>
         public event EventHandler<ServiceStateChangedEventArgs> ServiceStateChanged
@@ -755,13 +751,10 @@ namespace Tizen.Uix.Tts
         /// <summary>
         /// Gets the current state of TTS service.
         /// </summary>
-        /// <since_tizen> 10 </since_tizen>
         /// <value>
         /// The current state of TTS service.
         /// </value>
-        /// <returns>
-        /// Current TTS service state value.
-        /// </returns>
+        /// <since_tizen> 10 </since_tizen>
         public ServiceState CurrentServiceState
         {
             get
--- a/src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
+++ b/src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
@@ -204,6 +204,37 @@ namespace Tizen.Uix.Tts
     };
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
```diff
@@ -203,6 +203,37 @@ public enum Voice
...
+    /// <summary>
+    /// Enumeration for the states of TTS service.
+    /// </summary>
+    /// <since_tizen> 10 </since_tizen>
+    public enum ServiceState
+    {
+        /// <summary>
+        /// Ready state.
+        /// </summary>
+        /// <since_tizen> 10 </since_tizen>
```
wiertel, line 215: you don't need the \<since_tizen\> in enum field if the field was introduced in the initial version of the enum - i.e. if the version here is the same as the \<since_tizen\> of the containing enum

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
```diff
@@ -589,6 +623,51 @@ public TtsClient()
             }
         }
 
+        /// <summary>
+        /// Event to be invoked when state of TTS service changes.
```
wiertel, line 627: just a suggestion:
```suggestion
        /// Event to be invoked when the state of TTS service changes.
```

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
```diff
@@ -673,6 +752,32 @@ public State CurrentState
...
+        /// <summary>
+        /// Gets the current state of TTS service.
+        /// </summary>
+        /// <since_tizen> 10 </since_tizen>
+        /// <value>
+        /// The current state of TTS service.
+        /// </value>
+        /// <returns>
+        /// Current TTS service state value.
+        /// </returns>
```
wiertel, line 764: \<returns\> for a property is ignored as not related to this element, you can remove those 3 lines

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
```diff
@@ -203,6 +203,37 @@ public enum Voice
...
+        /// Synthesizing state.
+        /// </summary>
+        /// <since_tizen> 10 </since_tizen>
+        Synthesizing = 1,
+
+        /// <summary>
+        /// Playing state.
+        /// </summary>
+        /// <since_tizen> 10 </since_tizen>
+        Playing = 2,
```
wiertel, line 228: just wondering - can those state overlap? i.e. parallel synth and stream play in real time?

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
stom-hwang: Thank you for checking. I will remove it soon.

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
stom-hwang: In the aspect of real behavior of TTS service engine, yes it is possible to parallely synth and stream play in some times.
We just want to seperate the timing between the begining of the synthesis and the begining of the playing. That is, `Synthesizing` state means that the TTS service engine only synthesizes the sound data without playing.
TTS service engine will play sound data when the synthesized data is enough to play and then the service state will be changed to `Playing`.

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
stom-hwang: Thank you for checking. I will remove it soon.

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
stom-hwang: Thank you for suggestion. I will fix it soon.

stom-hwang (commented):
Thank you for checking.
I will fix it.

syeonholic (approved):
It looks good to me.

syeonholic (approved):
It looks good to me.

wonnamy (approved):
It looks good to me.

urmain (approved):
LGTM

PR #4493, opened 2022-08-19
[NUI] Make sure that Interop return type is bool

JoogabYun:
### Description of Change ###

This patch will fix this problem

### API Changes ###

 - ACR:


sparrow74:
LGTM :)

dongsug-song (approved):
approve

PR #4495, opened 2022-08-19
[NUI] Add UpdateProperty api for BorderWindow

JoogabYun:
### Description of Change ###

Changing the BorderView's SizeHeight should also change the Window size.

For this, we add an UpdateProperty() api.


everLEEst (approved):
Looks good to me.

PR #4496, opened 2022-08-19
[NUI] Make WidgetView not propagate touch event to the below

jaehyun0cho:
HandleControlStateOnTouch is overridden by WidgetView with return value
true not to propagate touch event to the below.

### API Changes ###

 - ACR:


everLEEst (approved):
Looks good to me.

Seoyeon2Kim (approved):
LGTM

PR #4518, opened 2022-08-29
[NUI] Border lines make the window invisible when maximized.

JoogabYun:
### Description of Change ###

When the window is full size, the content area should be full size. So it makes the border line disappear.
When it becomes partial size again, the border line is visible.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):

dongsug-song (approved):
approve

PR #4520, opened 2022-08-29
[NUI] Fix ContentPageLayout to measure AppBar prior to Content

jaehyun0cho:
Content's height is calculated correctly after AppBar's height is
calculated.

Therefore, AppBar is measured and then Content is measured.

### API Changes ###

 - ACR:


JoogabYun (approved):
approved

dongsug-song (approved):
approve

PR #4530, opened 2022-08-30
[NUI] Add Outlined and TextOnly Button styles

jaehyun0cho:
### Description of Change ###

The latest Tizen 7.0 UX introduces Outlined and TextOnly style Buttons.
Outlined style Button has transparent background and colored borderline.
TextOnly style Button has a text and transparent background.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
+++ b/src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
@@ -84,6 +84,29 @@ namespace Tizen.NUI.Components
                 }
             });
 
+            // TextOnly Button style
+            theme.AddStyleWithoutClone("Tizen.NUI.Components.Button.TextOnly", new ButtonStyle()
+            {
+                SizeHeight = 48,
+                MinimumSize = new Size(120, 48),
+                Padding = new Extents(32, 32, 8, 8),
+                ItemHorizontalAlignment = HorizontalAlignment.Center,
+                ItemVerticalAlignment = VerticalAlignment.Center,
+                BackgroundColor = Color.Transparent,
+                Text = new TextLabelStyle()
+                {
+                    TextColor = new Selector<Color>()
+                    {
+                        Normal = new Color("#FF6200"),
+                        Pressed = new Color("#FFA166"),
+                        Focused = new Color("#FF7119"),
+                        Selected = new Color("#FF8133"),
+                        Disabled = new Color("#CACACA"),
+                    },
+                    PixelSize = 24,
+                }
+            });
+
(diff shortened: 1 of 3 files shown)


jaehyun0cho:
This PR should be merged with Rebase and merge.

dongsug-song:
@jaehyun0cho 
thank you for the patch. could you explain why these components are needed?
could you update it in description?
thank you.

jaehyun0cho:
> @jaehyun0cho thank you for the patch. could you explain why these components are needed? could you update it in description? thank you.

Thank you for your comment.
Tizen 7.0 UX introduces OutlinedButton and TextButton.

everLEEst (changes requested):
We don't need OutlineButton, TextButton class for this.
adding style in the default theme is enough.

jaehyun0cho:
> We don't need OutlineButton, TextButton class for this. adding styles in the default theme are enough.

Thank you for your comment.
I have removed OutlinedButton and TextButton classes.
Instead, I have modified the samples to use "Tizen.NUI.Components.OutlinedButton" and "Tizen.NUI.Components.TextButton" styles.

everLEEst (approved):
Overall looks good to me.
but before pushing this patch, I think we need to decide naming policy of styles.
I think we have two good solutions.
```c#
Tizen.NUI.Components.Button
Tizen.NUI.Components.OutlinedButton
Tizen.NUI.Components.TextButton
```
or
```
Tizen.NUI.Components.Button  (.Default)
Tizen.NUI.Components.Button.Outlined
Tizen.NUI.Components.Button.TextOnly //Text might quite confusing in this case.
```

jaehyun0cho:
> Overall looks good to me. but before pushing this patch, I think we need to decide naming policy of styles. I think we have two good solutions.
> 
> ```cs
> Tizen.NUI.Components.Button
> Tizen.NUI.Components.OutlinedButton
> Tizen.NUI.Components.TextButton
> ```
> 
> or
> 
> ```
> Tizen.NUI.Components.Button  (.Default)
> Tizen.NUI.Components.Button.Outlined
> Tizen.NUI.Components.Button.TextOnly //Text might quite confusing in this case.
> ```

Good point!
I prefer the second way.
Reviewers, please leave your opinions~

jaehyun0cho:
I have updated this PR to use the style naming rule as follows.
"Tizen.NUI.Components.Button.Outlined"
"Tizen.NUI.Components.Button.TextOnly"

dongsug-song (approved):
very good idea~
we can avoid to create new classes which seem to be heavy and rich.

PR #4534, opened 2022-09-01
[MediaContent] Add CreateThumbnail and deprecate CreateThumbnailAsync

hsgwon:
### Description of Change ###

Add CreateThumbnail and deprecate CreateThumbnailAsync.
The performance of creating thumbnail is improved too much in Native FW. So we recommend to use Sync API.

### API Changes ###

 - TCSACR-507


src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoCommand.cs
```diff
@@ -962,6 +964,47 @@ private async Task<string> CreateThumbnailAsyncCore(string mediaId, Cancellation
...
         #endregion
 
+        /// <summary>
+        /// Creates the thumbnail image for the given media.
+        /// If the thumbnail already exists for the given media, the existing path will be returned.
+        /// </summary>
+        /// <privilege>http://tizen.org/privilege/content.write</privilege>
+        /// <privilege>http://tizen.org/privilege/mediastorage</privilege>
+        /// <privilege>http://tizen.org/privilege/externalstorage</privilege>
+        /// <param name="mediaId">The media ID to create the thumbnail.</param>
```
wiertel, line 974: consider "The media ID to create the thumbnail from." or "The ID of the media for which the thumbnail will be created."
missing `<returns>`

wiertel (changes requested):

src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoCommand.cs
hsgwon: Changed description and added missing tag.

aferin (approved):

PR #4535, opened 2022-09-01
[NUI] Add BorderlineColorSelector to ViewStyle

jaehyun0cho:
To support Selector<Color> for BorderlineColor, BorderlineColorSelector
has been added.
BorderlineColorSelector is applied to Outlined Button style.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
@@ -2000,17 +2000,29 @@ namespace Tizen.NUI.BaseComponents
         /// BorderlineColor Property
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public static readonly BindableProperty BorderlineColorProperty = BindableProperty.Create(nameof(BorderlineColor), typeof(Color), typeof(View), null, propertyChanged: (bindable, oldValue, newValue) =>
-        {
-            var view = (View)bindable;
-            (view.backgroundExtraData ?? (view.backgroundExtraData = new BackgroundExtraData())).BorderlineColor = (Color)newValue;
-            view.ApplyBorderline();
-        },
-        defaultValueCreator: (bindable) =>
-        {
-            var view = (View)bindable;
-            return view.backgroundExtraData == null ? Color.Black : view.backgroundExtraData.BorderlineColor;
-        });
+        public static readonly BindableProperty BorderlineColorProperty = BindableProperty.Create(nameof(BorderlineColor), typeof(Color), typeof(View), null,
+            propertyChanged: (bindable, oldValue, newValue) =>
+            {
+                var view = (View)bindable;
+
+                view.themeData?.selectorData?.BorderlineColor?.Reset(view);
+
+                if (newValue is Selector<Color> selector)
+                {
+                    if (selector.HasAll()) view.SetBorderlineColor(selector.All);
+                    else view.EnsureSelectorData().BorderlineColor = new TriggerableSelector<Color>(view, selector, view.SetBorderlineColor, true);
+                }
+                else
+                {
+                    view.SetBorderlineColor((Color)newValue);
+                }
+            },
+            defaultValueCreator: (bindable) =>
+            {
+                var view = (View)bindable;
+                return view.backgroundExtraData == null ? Color.Black : view.backgroundExtraData.BorderlineColor;
+            }
+        );
 
         /// <summary>
         /// BorderlineOffset Property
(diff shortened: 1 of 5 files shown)


src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
```diff
@@ -2675,6 +2688,27 @@ private void SetBackgroundImageBorder(Rectangle value)
...
+            PropertyMap map = Background;
+
+            if (map.Empty())
+            {
+                return;
+            }
+
+            map[Visual.Property.BorderlineColor] = new PropertyValue(value);
+
+            Tizen.NUI.Object.SetProperty((System.Runtime.InteropServices.HandleRef)SwigCPtr, View.Property.BACKGROUND, new PropertyValue(map));
```
hinohie, line 2709: We'd better use virtual API `ApplyBorderline` instead of this thing.

`ApplyBorderline` automatically create `Transparent` background if nobody settup Background things.
and, `ImageView` has it's own borderline logic (due to the `Image` drawed on `ImageView.Property.IMAGE` level.)
And also, It did'nt newly create new Visual.

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
```diff
@@ -2000,17 +2000,30 @@ public partial class View
...
+                else
+                {
+                    view.SetBorderlineColor((Color)newValue);
+                }
+            },
+            defaultValueCreator: (bindable) =>
+            {
+                var view = (View)bindable;
+
+                return view.backgroundExtraData?.BorderlineColor;
```
hinohie, line 2024: Previous code return `Color.Black` when nobody setup `backgroundExtraData`. 
But now, we return `null`.
Maybe not an big problem, but anyway we need to know the action is changed

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
jaehyun0cho: Thank you for your comment!
I have replaced the code with ApplyBorderline();

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
jaehyun0cho: Thank you for your comment!
I have updated the code like previous code.

dongsug-song (approved):
approve

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
```diff
@@ -2675,6 +2687,27 @@ private void SetBackgroundImageBorder(Rectangle value)
...
 
+        private void SetBorderlineColor(Color value)
+        {
+            if (value == null)
+            {
+                return;
+            }
+
+            (backgroundExtraData ?? (backgroundExtraData = new BackgroundExtraData())).BorderlineColor = value;
+
```
hinohie, line 2698: 'View.ApplyBorderline()' will use the value of 'backgroundExtraData'. Please remove 'PropertyMap map' usimg codes. we dont need them now.

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
```diff
@@ -2675,6 +2687,27 @@ private void SetBackgroundImageBorder(Rectangle value)
...
+            PropertyMap map = Background;
+
+            if (map.Empty())
+            {
+                return;
```
hinohie, line 2703: ...or, Should we ignore Selector result when backround didnt setup?
anyway, for this code now... we cannot setup borderline-only case. (Which is requested feature.)
I think we should think about this case

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
jaehyun0cho: Thank you for your comment!
I have removed the PropertyMap code.

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
jaehyun0cho: I think that checking Background map can be removed because ApplyBorderline() checks Background map as follows. If it is empty, then Transparent background color is automatically set.

```
        internal virtual void ApplyBorderline()
        {
            if (backgroundExtraData == null) return;

            // ActionUpdateProperty works well only if BACKGROUND visual setup before.
            // If view don't have BACKGROUND visual, we set transparent background color in default.
            using (PropertyMap backgroundPropertyMap = new PropertyMap())
            {
                using (PropertyValue propertyValue = Object.GetProperty(SwigCPtr, Property.BACKGROUND))
                {
                    propertyValue?.Get(backgroundPropertyMap);
                    if (backgroundPropertyMap.Empty())
                    {
                        // BACKGROUND visual doesn't exist.
                        SetBackgroundColor(Color.Transparent);
                        // SetBackgroundColor function apply borderline internally.
                        // So we can just return now.
                        return;
                    }
                }
            }
```

hinohie (approved):
Looks god to tme

PR #4539, opened 2022-09-06
[NUI] Update summary of RegisterHttpRequestInterceptedCallback

huayongxu:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/WebView/WebContext.cs
+++ b/src/Tizen.NUI/src/internal/WebView/WebContext.cs
@@ -537,6 +537,9 @@ namespace Tizen.NUI
 
         /// <summary>
         /// Registers callback for http request interceptor.
+        /// This callback is not called on UI(Main) thread, so users should be cautious
+        /// when accessing their data also used on UI thread.
+        /// No other than WebEngineRequestInterceptor API should be used in the callback.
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/internal/WebView/WebContext.cs
```diff
@@ -537,6 +537,9 @@ public void RegisterMimeOverriddenCallback(MimeOverriddenCallback callback)
 
         /// <summary>
         /// Registers callback for http request interceptor.
+        /// This callback is not called on UI(Main) thread, so users should be cautious
+        /// when accessing their data also used on UI thread.
+        /// No other than WebEngineRequestInterceptor API should be used in the callback.
```
myroot, line 542: When will this constraint apply? Because in the 9/15 binary, it is called on UI Thread.

huayongxu: In order to fix the issue https://github.sec.samsung.net/dotnet/NUIBackend/issues/173 ,
the constraint above is applied.
But it seems that there is a problem in dali release process,
the patch is not released now.

dongsug-song (approved):
approved

PR #4546, opened 2022-09-07
[NUI] prevent the case when touchData is null

JoogabYun:
### Description of Change ###

If touchData is null, GetTouchFromPtr throws an exception.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

taehyub (approved):
LGTM

PR #4561, opened 2022-09-13
[NUI.Scene3D] Create and Register Scene3D.Camera

hinohie:
Due to the Scene3D.SceneView's default camera didn't created in NUI side, we need to Regist this camera information into Registry.
If we already register it into Registry, use that.
Else, register new Camera into Registry.


dongsug-song (approved):
approved

PR #4568, opened 2022-09-14
[NUI] Add Set/Get IBL Scale Factor for SceneView and Model

bshsqa:

### Description of Change ###
Add Set/Get IBL Scale Factor for SceneView and Model.

### API Changes ###

 - ACR:


everLEEst (approved):
Looks good tome

dongsug-song (approved):
approved

PR #4584, opened 2022-09-21
[NUI] Modify the return value of OnAccessibilityPan()

JoogabYun:
### Description of Change ###

When return true, panGestureEvent is consumed.
When return false, panGestureEvent is propagated to the parent.


dongsug-song (approved):
approved

Seoyeon2Kim (approved):
lgtm!
Thanks.

PR #4589, opened 2022-09-22
[NUI] disable acessibilityPanGesture when scroll disabled.

everLEEst:
### Description of Change ###

Accessibiliy call OnPanGesture directly without check ScrollEnabled variable.
this patch fix to check ScrollEnabled value before sending OnPanGesture.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #4592, opened 2022-09-24
[NUI] Change property get/set not to use PropertyValue

dongsug-song:
### Description of Change ###
[NUI] Change property get/set not to use PropertyValue
- this should be merged with https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/281975/
- this is the first patch and it will be continuously changed and modified.
- `NUI_PROPERTY_CHANGE_1` this macro will be deleted after all changes are completed.
- start with View and expand it to ImageView, Text, etc.
- local testing has been done on NikeM2 target.

### API Changes ###
none


Seoyeon2Kim (approved):
LGTM!

PR #4594, opened 2022-09-26
[NUI] BorderLineThickness should be taken into account when setting MimimumSize and MaximumSize.

JoogabYun:
### Description of Change ###

 BorderLineThickness should be taken into account when setting MimimumSize and MaximumSize.

If BorderLineThickness is not calculated, there will be a problem with window resize when it is reduced by MimimumSize.


taehyub (approved):
LGTM

PR #4600, opened 2022-09-26
[NUI] Modify internal window position type

sparrow74:
In DALi, window position data struct only has unisigned int.
    It means this can not support the negative coornidates.
    This PR support that.


dongsug-song:
@sparrow74 
Hello,
The TizenFX master branch has been changed for API11(tizen_7.5).
So the DevelNUI branch also did a hard reset with the master branch to sync this.
I think you need to rebase this PR or re-upload the new PR.
I am sorry for the inconvenience.

sparrow74:
This patch's contents is overwritten by the other patch.
So, it is not needed and will be closed.

PR #4612, opened 2022-09-29
[Stc] Fix datatype conversion issue

a-nijhara:
### Description of Change
* time_t type has different size on 32bit/64bit. so need to change type of interop method call parameter


SeonahMoon (approved):
Looks good to me.

dh79pyun (approved):
Looks good

myroot (approved):

PR #4614, opened 2022-09-30
[NUI.Scene3D] Add ProjectionDirection in Camera

hinohie:
Add new property to control major direction of camera FieldOfView.
It will effort to FieldOfView and OrthographicSize.

reference link : https://developer.apple.com/documentation/scenekit/scncamera/2878134-projectiondirection?language=objc

### DIFF
--- a/src/Tizen.NUI.Scene3D/src/public/Controls/Camera.cs
+++ b/src/Tizen.NUI.Scene3D/src/public/Controls/Camera.cs
@@ -32,23 +32,6 @@ namespace Tizen.NUI.Scene3D
     /// <since_tizen> 10 </since_tizen>
     public partial class Camera : View
     {
-        /// <summary>
-        /// Enumeration for the projectionMode.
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI.Scene3D/src/public/Controls/Camera.cs
```diff
@@ -581,6 +662,38 @@ internal bool GetInvertYAxis()
             return ret;
         }
 
+        /// <summary>
```
bshsqa, line 665: What is the use case of these method?
The input parameter verticalFov and horizontalFov are not the member variable..

hinohie: Thank you for remind me. It was some helper method to compute FoV.
It will be helpful if user want to use both Vertical FoV and Horizontal FoV.
Now I make it as public static method now.

bshsqa (approved):

PR #4615, opened 2022-10-04
[NUI] Support native new window position

sparrow74:
[NUI] Supports new native window position data type.
    Previous native window data type can not support negative window coordinates.
    New native window position data type support that.


dongsug-song (approved):
approved

PR #4620, opened 2022-10-04
[NUI] Fix Rotation Property set/get usage

hinohie:
Rotation don't use Vector3. It use Quaternion internally.


dongsug-song (approved):
approve

PR #4624, opened 2022-10-05
[NUI] Apply Tizen 7.0 UX on Slider ValueIndicator

Seoyeon2Kim:

### Description of Change ###
- The latest Slider UX for ValueIndicator has been applied. (Based on API10) 
   1) `ValueIndicatorImage` has become a child of Thumb.
   2) The style of `ValueIndicatorText` is added to the default theme.
   3) Updated the position and size for value indicator.

### API Changes ###
- N/A

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Slider.Internal.cs
+++ b/src/Tizen.NUI.Components/Controls/Slider.Internal.cs
@@ -300,13 +300,14 @@ namespace Tizen.NUI.Components
                 {
                     WidthResizePolicy = ResizePolicyType.Fixed,
                     HeightResizePolicy = ResizePolicyType.Fixed,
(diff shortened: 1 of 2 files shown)


jaehyun0cho (approved):
LGTM

tscholb (approved):

everLEEst (approved):
Looks good to me, but I have one question.
this thumbImage have same lifecycle of silder and won't be changed?
cause I can only see Add and Disposed is called when slider disposed.

Seoyeon2Kim:
> 

Thanks for reviewing it.
But, could you explain the question in detail please?
`thumbImage` is a child of `bgTrackImage`, which is a child of Slider control itself.
And they're created when View init time ( `ApplyStyle()` ).

PR #4627, opened 2022-10-05
[NUI] Fix Apps crash issue when terminated

dongsug-song:
### Description of Change ###
[NUI] Fix Apps crash issue when terminated
- change the Window's events codes as following View's events pattern.
- locally tested with RPI4 target.

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/public/Window/WindowEvent.cs
+++ b/src/Tizen.NUI/src/public/Window/WindowEvent.cs
@@ -100,20 +100,19 @@ namespace Tizen.NUI
                 if (windowFocusChangedEventHandler == null)
                 {
                     windowFocusChangedEventCallback = OnWindowFocusedChanged;
-                    windowFocusChangedSignal = WindowFocusChangedSignal();
-                    windowFocusChangedSignal?.Connect(windowFocusChangedEventCallback);
+                    using WindowFocusSignalType signal = new WindowFocusSignalType(Interop.Window.FocusChangedSignal(SwigCPtr), false);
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Window/WindowEvent.cs
```diff
@@ -100,22 +100,24 @@ public partial class Window
...
             remove
             {
                 windowFocusChangedEventHandler -= value;
-
-                if (windowFocusChangedEventHandler == null && windowFocusChangedSignal?.Empty() == false && windowFocusChangedEventCallback != null)
+                if (windowFocusChangedEventHandler == null)
                 {
-                    windowFocusChangedSignal?.Disconnect(windowFocusChangedEventCallback);
-                    if (windowFocusChangedSignal?.Empty() == true)
+                    using WindowFocusSignalType signal = new WindowFocusSignalType(Interop.Window.FocusChangedSignal(SwigCPtr), false);
```
hinohie, line 113: It will be dead code if `windowFocusChangedEventCallback` is `null`. If you don't mine, Cause you use `if (windowFocusChagnedEventHandler == null && windowFocusChangedEventCallback != null)` at line 111?

hinohie (approved):

src/Tizen.NUI/src/public/Window/WindowEvent.cs
dongsug-song: Thank you very much for the review.
I have changed all by following your comment.
thank you again.

PR #4628, opened 2022-10-05
[NUI] Fix Apps crash issue when terminated

dongsug-song:
### Description of Change ###
[NUI] Fix Apps crash issue when terminated
- change the Window's events codes as following View's events pattern.
- locally tested with RPI4 target.

### API Changes ###
none


hinohie (approved):
Maybe minor. Don't we need to remove unused private member?

dongsug-song:
> Maybe minor. Don't we need to remove unused private member?

yes. your are absolutely right, I will remove them on next patch. 
thank you for review.

PR #4632, opened 2022-10-06
[NUI] Fix to calculate Picker's current value correctly

jaehyun0cho:
Since scrolling position is floating value, current index should be calculated with rounding integer.

### API Changes ###

 - ACR:


jaehyun0cho:
This PR should be investigated more.

src/Tizen.NUI.Components/Controls/Picker.cs
```diff
@@ -557,9 +557,9 @@ private void OnScrollAnimationEnded(object sender, ScrollEventArgs e)
                 {
                     PageAdjust(e.Position.Y);
                 }
-                if (currentIdx != ((int)(-e.Position.Y / itemHeight) + middleItemIdx))
+                if (currentIdx != ((int)(-e.Position.Y / itemHeight + 0.5) + middleItemIdx))
```
myroot, line 560: 명시적으로 Round 메소드를 사용하는게 코드 가독성 측면에서 더 좋지 않을까요?
```suggestion
                if (currentIdx != ((int)(Math.Round(-e.Position.Y / itemHeight)) + middleItemIdx))
```

PR #4638, opened 2022-10-09
[NUI] Change property get/set not to use PropertyValue (patch set3)

dongsug-song:
### Description of Change ###
[NUI] Change property get/set not to use PropertyValue (patch set3)

### API Changes ###
none


Seoyeon2Kim (approved):
lgtm

PR #4642, opened 2022-10-11
[WebRTC] Disable MediaPacket native handle destroy

hsgwon:
### Description of Change ###

Disable MediaPacket native handle destroy when Native FW destroys native media packet handle.


hancepark (approved):
LGTM

PR #4647, opened 2022-10-12
[NUI] Apply dotnet-nui type

dkdk-ryu:
NUI apps have to set dotnet-nui app type in tizen-manifest.xml. Then the apps are preloaded from dotnet-loader.
It improves the launching performance.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #4649, opened 2022-10-12
[NUI] Fix picker, scroller animation gives abnormal position value

lwc0917:
### Description of Change ###
This is hot-fix for if the velocity has very small value, result is not updated even progress done.

### API Changes ###
No API Changes


jaehyun0cho (approved):
LGTM

hinohie (approved):
Hot fix checked. We should change custom alpha function somedays.

PR #4650, opened 2022-10-12
[NUI] Include json files

taehyub:
Include json files
There are no "loading" lottie animation json file, so broken image is shown


jaehyun0cho (approved):
LGTM

PR #4653, opened 2022-10-13
[NUI] Remove AnyContainerBase type

hinohie:
AnyContainerBase is private data. We don't need to access there and also, we don't need to change it now.
Actually, nobody use it. So remove this API

### API Changes ###

 - ACR:


everLEEst (approved):
Looks good to me.

jaehyun0cho (approved):
LGTM

PR #4654, opened 2022-10-13
[NUI] If the view had focus when it was unparented, the focus is cleared.

JoogabYun:
### Description of Change ###

When the view is unparented, the focus is not cleared. So, when app re-add the view and SetCurrentFocusView, app do not receiving the FocusGained event.


jaehyun0cho (approved):
LGTM

hinohie (approved):

PR #4655, opened 2022-10-13
[NUI] Change property get/set not to use PropertyValue (patch set4)

dongsug-song:
### Description of Change ###
[NUI] Change property get/set not to use PropertyValue (patch set4)
- this should be merged with https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/282909/
- NUI_PROPERTY_CHANGE_X this macro will be deleted after all changes are completed.
- local testing has been done on NikeM2 target.

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
@@ -18,6 +18,7 @@
 using System;
 using System.ComponentModel;
 using global::System.Diagnostics;
+using Tizen.NUI;
 
 namespace Tizen.NUI.BaseComponents
 {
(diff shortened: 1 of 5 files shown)


Seoyeon2Kim (approved):
LGTM

src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
```diff
@@ -1221,6 +1644,33 @@ protected override void Dispose(DisposeTypes type)
             rotationGestureDetector?.Dispose();
             rotationGestureDetector = null;
 
+#if NUI_PROPERTY_CHANGE_1
```
jaehyun0cho, line 1647: Is this NUI_PROPERTY_CHANGE_1 or NUI_PROPERTY_CHANGE_3?
Because internalCurrentParentOrigin is defined with NUI_PROPERTY_CHANGE_3.

dongsug-song: thank you for review.
you are right, (3 is right)
it is fixed.

PR #4656, opened 2022-10-13
[NUI] Picker: Fix scroll area height calc logic and Expand default scroll area

lwc0917:
### Description of Change ###
Expand default scroll area to support app requirment.
but we guide do not use more than 100 selections for picker.
Picker not designed for many items. it's not ListView.

Fix scroll area height calc logic
picker size -> itemHeight * 5
Picker size can change by user.

### API Changes ###
No API Changes


dongsug-song (approved):
approved

PR #4659, opened 2022-10-13
[NUI] Picker: Protect delegate from GC

lwc0917:
### Description of Change ###
Fix below Err.
Process terminated. A callback was made on a garbage collected delegate of type

### API Changes ###
No API Changes


jaehyun0cho (approved):
LGTM

taehyub (approved):
LGTM

myroot:
It can't solve GC issue
because GC.KeepAlive is just keep obejct just before call `GC.KeepAlive`

>References the specified object, which makes it ineligible for garbage collection from the start of the current routine to the point where this method is called.

https://learn.microsoft.com/en-us/dotnet/api/system.gc.keepalive?view=net-6.0


`customScrollAlphaFunction` object will be released when `Picker` object released but if Animation was not finished, crash will be occurred.

so, you need to stop animation before destroy picker object

It is related issue with https://github.sec.samsung.net/dotnet/NUIBackend/issues/162

PR #4662, opened 2022-10-14
[NUI] Add null checking in InternalSet/GetPropertyXXX(), Change to use internalName

dongsug-song:
### Description of Change ###
[NUI] Add null checking in InternalSet/GetPropertyXXX(), Change to use internalName
 - make the InternalSet/GetPropertyXXX() same as the Object.PropertySet/Get() pattern.
 - change for `View.Name` property to use `View.internalName`.
 - tested locally with NikeM2 target

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/Object.cs
+++ b/src/Tizen.NUI/src/internal/Common/Object.cs
@@ -15,6 +15,8 @@
  *
  */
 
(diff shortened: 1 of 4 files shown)


jaehyun0cho (approved):
LGTM

PR #4664, opened 2022-10-14
[NUI] Fix Linear Layouter crash and update issue

everLEEst:
i### Description of Change ###

1. Linear Layouter get crashed when group has no children.
2. Linear, Grid Layouter update is not working well because delayed
update timer is not started.
3. remove spaces.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #4674, opened 2022-10-20
[NUI] Reduce PropertyValue creation during Animation

hinohie:
There was some PropertyValue creation during data conversion.
Let we reduce new PropertyValue class creation so GC feel happy


dongsug-song (approved):
looks good to me

PR #4675, opened 2022-10-20
[NUI][API10] Reduce PropertyValue creation during Animation

hinohie:
There was some PropertyValue creation during data conversion. Let we reduce new PropertyValue class creation so GC feel happy

This patch is copy of https://github.com/Samsung/TizenFX/pull/4674


dongsug-song (approved):
approved

PR #4676, opened 2022-10-25
[NUI] Fix issue item is not unrealized after insertion.

everLEEst:
After Item Insertion,
Previous index of visible item is not updated properly,
so unrealization goes wrong.

it fix visible indexes most cases except rangemove.
range move is to complex for re-calculate indexes,
so it unrealize every items and reset the visible item indexes.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

Seoyeon2Kim (approved):
LGTM~

PR #4678, opened 2022-10-26
[NUI] 1. Move the Transparency setting to OnMaximize().

JoogabYun:
### Description of Change ###

1. Move the Transparency setting to OnMaximize().
2. Fixed the wrong condition when resizing.
3. backgroundImage and backgroundColor are exclusive.
   So if the user has set the backgroundImage then the backgroundColor should not be reset.
4. BorderLineThickness can be changed dynamically.

### DIFF
--- a/src/Tizen.NUI/src/public/Window/BorderWindow.cs
+++ b/src/Tizen.NUI/src/public/Window/BorderWindow.cs
@@ -55,6 +55,7 @@ namespace Tizen.NUI
         // for config
         private Size2D minSize = null;
         private Size2D maxSize = null;
+        private uint borderLineThickness;
         private BorderResizePolicyType borderResizePolicy = BorderResizePolicyType.Free;
         #endregion //Fields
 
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Window/BorderWindow.cs
```diff
@@ -168,6 +169,29 @@ internal void UpdateProperty()
...
+                        borderView.Padding = new Extents(start, end, top, bottom);
+                        if (IsMaximized() == true)
+                        {
+                            borderView.OnMaximize(true);
+                        }
+                    }
+
+                    ResizedEventArgs e = new ResizedEventArgs();
+                    e.WindowSize = WindowSize;
+                    OnBorderWindowResized(this, e);
```
everLEEst, line 193: Is this code intended to call internal function only?
seems below code is more general for resize event..
```suggestion
                    Resized?.Invoke(this, e);
```

src/Tizen.NUI/src/public/Window/DefaultBorder.cs
```diff
@@ -101,7 +102,18 @@ private enum CurrentGesture
         /// The thickness of the border.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public uint BorderLineThickness {get; set;}
+        public uint BorderLineThickness
```
everLEEst, line 105: I think we might need BindableProperty for this.

src/Tizen.NUI/src/public/Window/DefaultBorder.cs
```diff
@@ -345,7 +356,11 @@ public virtual void CreateBorderView(View borderView)
             {
                 return;
             }
-            borderView.BackgroundColor = DefaultBackgroundColor;
+
+            if (borderView.BackgroundImage == string.Empty)
```
everLEEst, line 360: please consider the github actions build warnings.

src/Tizen.NUI/src/public/Window/DefaultBorder.cs
```diff
@@ -360,16 +375,19 @@ public virtual void CreateBorderView(View borderView)
 
         private bool OnBorderViewTouched(object sender, View.TouchEventArgs e)
         {
-            if (e.Touch.GetState(0) == PointStateType.Started)
+            if (borderView.BackgroundImage == string.Empty)
```
everLEEst, line 378: here also... and so on other 3 also

everLEEst (approved):
LGTM. thanks for fast update.

jaehyun0cho (approved):
LGTM

PR #4685, opened 2022-10-27
[NUI] Fix crash on ScrollableBase.Dispose()

jaehyun0cho:
In ScrollableBase.Dispose(), propertyNotification is disposed as well. To dispose propertyNotification, RemovePropertyNotifications was called inappropriately and it caused crash.

To resolve the above, RemovePropertyNotification is called by ContentContainer with propertyNotification like AddPropertyNotification is called by ContentContainer with propertyNotification.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM.

JoogabYun (approved):

PR #4687, opened 2022-10-27
[NUI] Modify Menu to help customizing class

jaehyun0cho:
To help application to customize Menu, Menu's size calculation should be modified not to set its size to its Content's size.

### API Changes ###

 - ACR:


everLEEst (approved):
Looks good to me.

PR #4689, opened 2022-10-28
[NUI] Add constructor with style instance

jaehyun0cho:
### Description of Change ###

To customize and apply app's own style instance, constructor with style
instance should be added.

For now, ApplyStyle() modifies the given properties only based on the
default style unlike applying style in constructor.
e.g. Let default style contain properties A, B, C.
     Let custom style contain properties B, C.
     Then ApplyStyle(custom style) contain default style's A and custom
     style's B and C.

Consequently, to apply custom style without applying default style,
constructor with style instance is required.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Dialog.cs
+++ b/src/Tizen.NUI.Components/Controls/Dialog.cs
@@ -48,15 +48,30 @@ namespace Tizen.NUI.Components
 
         private View content = null;
 
+        private void Initialize()
+        {
+            Layout = new AbsoluteLayout();
+
+            this.Relayout += OnRelayout;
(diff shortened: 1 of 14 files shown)


JoogabYun (approved):

everLEEst (commented):
I'm not sure necessity of constructor with ControlStyle.
Components like menu who have their own style(MenuStyle) need styling constructor,
but the Components who doesn't have their own Style might useless to set ControlStyle on Components...

everLEEst (approved):
After discussing about the usage on ControlStyle, I agreed on this patch.

PR #4696, opened 2022-11-01
[NUI] Add native handle null check in Uint16Pair

dongsug-song:
### Description of Change ###
[NUI] Add native handle null check in Uint16Pair

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/Uint16Pair.cs
+++ b/src/Tizen.NUI/src/internal/Common/Uint16Pair.cs
@@ -15,8 +15,8 @@
  *
  */
 
-using System;
-using System.ComponentModel;
+using global::System;
+using global::System.Runtime.InteropServices;
 
 namespace Tizen.NUI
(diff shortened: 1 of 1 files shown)


hinohie (approved):
Approve it. BTW, This patch doesn't resolve crash anyway. Is this patch'es goal is just catch the crash callstack/?

dongsug-song:
> Approve it. BTW, This patch doesn't resolve crash anyway. Is this patch'es goal is just catch the crash callstack/?

Hello, thank you for review.
yes, you are absolutely right, as you know, the crash may be resolved by App side which uses this object. 
because in the past, in this kind of case, it didn't throw exception and showing error msg with just returning, but this way is not resolving root cause so for now it throws exception to tell application.
this is same informing way as this =>
https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/internal/Common/Object.cs#L31

PR #4709, opened 2022-11-03
[NUI] Lets you know if the window is maximized.

JoogabYun:
### Description of Change ###

[NUI] Lets you know if the window is maximized.


dongsug-song (approved):
approved

PR #4715, opened 2022-11-03
[NUI] Modify TabContent to help customizing class

jaehyun0cho:
To help application to customize TabContent, some methods and properties have been modified.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/TabContent.cs
+++ b/src/Tizen.NUI.Components/Controls/TabContent.cs
@@ -63,10 +63,22 @@ namespace Tizen.NUI.Components
         protected int SelectedIndex { get; set; }
 
         /// <summary>
+        /// A list of content views.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected IList<View> Views
+        {
+            get
+            {
+                return views;
+            }
+        }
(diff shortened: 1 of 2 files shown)


everLEEst (approved):
LGTM.

Seoyeon2Kim (changes requested):
Could you remove the warning messages of this PR?
There are two messages :
1. At line 67, `Change 'Views' to be read-only by removing the property setter.`
2. At line 170, `Rename virtual/interface member TabContent.Select(int) so that it no longer conflicts with the reserved language keyword 'Select'.`

If you cannot change these features, then please suppress the messages using the tags.

jaehyun0cho:
> Could you remove the warning messages of this PR? There are two messages :
> 
> 1. At line 67, `Change 'Views' to be read-only by removing the property setter.`
> 2. At line 170, `Rename virtual/interface member TabContent.Select(int) so that it no longer conflicts with the reserved language keyword 'Select'.`
> 
> If you cannot change these features, then please suppress the messages using the tags.

Thank you for your comment!
I have changed as you suggested.

Seoyeon2Kim (approved):
LGTM!
Thank you for your support even though you're out of office now! lol

PR #4717, opened 2022-11-04
Add wheelEvent and WheelScrollDistance property on ScrollableBase

everLEEst:
### Description of Change ###

Add wheelEvent on ScrollableBase for wheel scroll move.
WheelScrollDistance Property is also added for modify distance by each wheel event.

this patch use WheelEvent on View not OnWheel()
because currently OnWheel is deactivated.
need more discussion about using OnWheel().

### API Changes ###

 - ACR:


JoogabYun (approved):

everLEEst:
Please merge it with squash.

PR #4737, opened 2022-11-09
[NUI] Move Matrix API as public (hidden)

hinohie:
Add and bind some matrix operator + remove useless items

And also, move this matrix class under public
so normal NUI Apps also can use this feature

Removed :
 * AsFloat() : Due to ignore unsafe float point accesser.

Added :
 * operator * override.
 * MultiplyAssign (static Matrix.Multiply API cannot do something like this.)
 * Value setter/getter by index (mat[9] == mat.ValueOfIndex(9) == mat.ValueOfIndex(1, 2). Set by mat.SetValueAtIndex(1, 2, 3.0f)
   * Since NUI.Matrix system is column-major, we need to some help for App developer to know the row/column index.

More detail : https://github.com/dalihub/dali-core/blob/master/dali/public-api/math/matrix.h#L38-L58
TODO : Apply comments on this classes. It might be doing on another patches.

Required dali patch : https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/283173


dongsug-song (approved):
approved

PR #4747, opened 2022-11-14
[Applications.PackageManager] Add ClearUserData(package, path)

upple:
Adds:
 - ClearUserData(packageId, filePath)

Signed-off-by: ChangGyu Choi <uppletaste@gmail.com>

### Description of Change ###
PackageManager has an API that removes all user data from a specific package.
There is a requirement to remove only certain file or directory
Data management becomes more detailed because only specific data can be removed by adding this API.

### API Changes ###

 - ACR: TCSACR-522

### DIFF
--- a/src/Tizen.Applications.PackageManager/Tizen.Applications/PackageManager.cs
+++ b/src/Tizen.Applications.PackageManager/Tizen.Applications/PackageManager.cs
@@ -353,6 +353,31 @@ namespace Tizen.Applications
         }
 
         /// <summary>
+        /// Clears the application's user file or directory from the file path.
+        /// </summary>
+        /// <remarks>
+        /// A file or directory under user data folder in the internal storage is removed.
+        /// </remarks>
+        /// <param name="packageId">ID of the package.</param>
+        /// <param name="filePath">The path of the file or directory in the package user data folder.</param>
+        /// <exception cref="OutOfMemoryException">Thrown when there is not enough memory to continue the execution of the method.</exception>
+        /// <exception cref="System.IO.IOException">Thrown when the method failed due to an internal IO error.</exception>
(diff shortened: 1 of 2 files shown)


src/Tizen.Applications.PackageManager/Tizen.Applications/PackageManager.cs
```diff
@@ -352,6 +352,31 @@ public static void ClearDataDirectory(string packageId)
             }
         }
 
+        /// <summary>
+        /// Clears the application's user file or directory from the file path.
```
wiertel, line 356: This description is ok, but please consider replacing it with
"Removes a file or directory specified with \<paramref name="filePath"/\> from user data internal storage for the application."
and maybe rename parameter from "filePath" to "path" - as it would fit better when it can be a directory path.
I'm not sure about the term "internal storage" - unless it really is limited to internal and not external storage.
If you write it as suggested, then the \<remarks\> can be removed.

wiertel (changes requested):
Please change the PR title to be more specific, e.g. "Add ClearUserData(package, path)"

src/Tizen.Applications.PackageManager/Tizen.Applications/PackageManager.cs
upple: Thank you for the review!
A applied your suggestion. 👍

upple:
@wiertel 
Hello.
Could you review this?

hjhun:
> Please change the PR title to be more specific, e.g. "Add ClearUserData(package, path)"

Done. Please check this change.

upple:
@wiertel 
Please approve this PR.

wiertel (approved):

PR #4755, opened 2022-11-17
[NUI] Support sub node's Animation

bshsqa:
This patch allows to access to the child Actor of View to animate Properties.

e.g., animation.AnimateTo(bicycle.FindChildAnimatableByName("soundbar_off"), "positionY", -50);

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/PropertyHelper.cs
+++ b/src/Tizen.NUI/src/internal/Common/PropertyHelper.cs
@@ -71,16 +71,21 @@ namespace Tizen.NUI
         ///<summary>
         /// Returns a Property if stringProperty is a valid index
         ///</summary>
-        internal static SearchResult Search(View view, string stringProperty)
+        internal static SearchResult Search(Animatable animatable, string stringProperty)
         {
             var propertyName = LowerFirstLetter(stringProperty);
 
-            return SearchProperty(view, propertyName) ?? SearchVisualProperty(view, propertyName);
+            if(animatable is View)
+            {
+                View view = animatable as View;
+                return SearchProperty(view, propertyName) ?? SearchVisualProperty(view, propertyName);
+            }
+            return SearchProperty(animatable, propertyName);
         }
 
-        private static SearchResult SearchProperty(View view, string lowercasePropertyString)
+        private static SearchResult SearchProperty(Animatable animatable, string lowercasePropertyString)
         {
-            Property property = new Property(view, lowercasePropertyString);
+            Property property = new Property(animatable, lowercasePropertyString);
 
             if (property.propertyIndex == Property.InvalidIndex)
             {
@@ -89,7 +94,7 @@ namespace Tizen.NUI
             }
 
             OOConverter converter = null;
-            if (view.GetPropertyType(property.propertyIndex).Equals(PropertyType.Float))
+            if (animatable.GetPropertyType(property.propertyIndex).Equals(PropertyType.Float))
             {
                 converter = ObjectIntToFloat;
             }
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewPublicMethods.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewPublicMethods.cs
@@ -217,6 +217,25 @@ namespace Tizen.NUI.BaseComponents
         }
 
         /// <summary>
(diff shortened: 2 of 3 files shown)


src/Tizen.NUI/src/internal/Common/PropertyHelper.cs
```diff
@@ -71,14 +71,19 @@ internal static Property GetPropertyFromString(Animatable handle, string stringP
...
+            if(animatable is View)
+            {
+                View view = animatable as View;
+                return SearchProperty(view, propertyName) ?? SearchVisualProperty(view, propertyName);
+            }
+            return SearchProperty(animatable, propertyName);
         }
 
-        private static SearchResult SearchProperty(View view, string lowercasePropertyString)
+        private static SearchResult SearchProperty(Animatable view, string lowercasePropertyString)
```
jaehyun0cho, line 86: Could you change view to animatable?

src/Tizen.NUI/src/public/BaseComponents/ViewPublicMethods.cs
```diff
@@ -446,6 +465,36 @@ public View FindChildByName(string viewName)
...
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Animatable FindChildAnimatableByName(string childName)
+        {
+            //to fix memory leak issue, match the handle count with native side.
+            IntPtr cPtr = Interop.Actor.FindChildByName(SwigCPtr, childName);
+            Animatable ret = Registry.GetManagedBaseHandleFromNativePtr(cPtr) as Animatable;
+            if(ret == null)
+            {
+                // Register new camera into Registry.
+                ret = new Animatable(cPtr, true);
```
Seoyeon2Kim, line 485: Could you remove this warning message?
`Call System.IDisposable.Dispose on object created by 'new Animatable(cPtr, true)' before all references to it are out of scope.`

You can fix it to use `using` statement.
```suggestion
                using (ret = new Animatable(cPtr, true))
```
or to use `try` and `finally` statement.

Seoyeon2Kim (changes requested):
Please add the description in detail of this PR.

src/Tizen.NUI/src/public/BaseComponents/ViewPublicMethods.cs
bshsqa: We need to return the ret.
If we use using we can expect the ret cannot be used outside of this method.

Seoyeon2Kim: I apologize that I was in such a hurry that I wrote the wrong changes before I entered the meeting!
I mean you can change this code like this :
```suggestion
                using (var animatable = new Animatable(cPtr, true))
                {
                    ret = animatable;
                }
```

bshsqa: It didn't works......

Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'target')
   at Tizen.NUI.Animation.AnimateTo(Animatable target, String property, Object destinationValue, AlphaFunction alphaFunction) in /home/seungho/Shared/NUI/TizenFX/src/Tizen.NUI/src/public/Animation/Animation.cs:line 784

Target handle is missing.

jaehyun0cho: As @bshsqa said, ret should be returned, it seems that "using" cannot be used in this case.
Because outside of "using" block, animatable is disposed and ret is assigned to the disposed handle..

I think that try catch can resolve this case but I am not sure if try catch is the right way because we need to apply try catch every case similar to this case..

jaehyun0cho (approved):
LGTM

src/Tizen.NUI/src/public/BaseComponents/ViewPublicMethods.cs
Seoyeon2Kim: So, here try catch statement is the best way, but it may affect the future cases. :(
We need to think about this warning message after this commit is merged.

Seoyeon2Kim (approved):
Approved PR this time.
NUI team should discuss how to handle `using` statement.

PR #4756, opened 2022-11-17
[NUI] Implement missing properties for LottieAnimationView.

hinohie:
1. Send 'RedrawInScaleDown' property when we change URL.
2. Send these initialized informations when we create Image


dongsug-song (approved):
approved

PR #4772, opened 2022-11-24
[NUI] Fix TextEditor default theme

hinohie:
Since we don't use dali's default style if NUI theme defined,
We may need to re-define something dali's default style in DefaultThemeCommon.

`TextEditor`'s `EnableScrollBar` property was some kind of this case.
Now we make `TextEditor`'s scrollbar relative vaules matched with dali's default style.


wonrst (approved):
This is definitely an amazing patch

dongsug-song (approved):
approved

PR #4775, opened 2022-11-25
[NUI] Fix Layouter to handle empty group case.

everLEEst:
If we create empty group it gets crash and moreover,
if we add more items in the group item didn't show correctly.
this patch will fix by re-initialize when new item added.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/RecyclerView/Layouter/GridLayouter.cs
+++ b/src/Tizen.NUI.Components/Controls/RecyclerView/Layouter/GridLayouter.cs
@@ -210,7 +210,8 @@ namespace Tizen.NUI.Components
 
             bool failed = false;
             //Final Check of FirstIndex
-            if (colView.InternalItemSource.Count - 1 < firstIndex)
+            if ((colView.InternalItemSource.Count - 1 < firstIndex) ||
+                (colView.InternalItemSource.IsFooter(firstIndex) && (colView.InternalItemSource.Count - 1) == firstIndex))
             {
                 StepCandidate = 0F;
                 failed = true;
@@ -517,7 +518,7 @@ namespace Tizen.NUI.Components
             // Insert Single item.
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (colView == null) return;
-            if (isSourceEmpty || StepCandidate == 0)
+            if (isSourceEmpty || StepCandidate <= 1)
             {
                 Initialize(colView);
             }
@@ -669,7 +670,7 @@ namespace Tizen.NUI.Components
              // Insert Group
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (colView == null) return;
-            if (isSourceEmpty || StepCandidate == 0)
+            if (isSourceEmpty || StepCandidate <= 1)
             {
                 Initialize(colView);
             }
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI.Components/Controls/RecyclerView/Layouter/GridLayouter.cs
```diff
@@ -210,11 +210,19 @@ public override void Initialize(RecyclerView view)
 
             bool failed = false;
             //Final Check of FirstIndex
-            while (colView.InternalItemSource.IsHeader(firstIndex) ||
+            if (colView.InternalItemSource.Count - 1 < firstIndex)
```
jaehyun0cho, line 213: how about changing the condition to be the same with the condition in the while loop?
```
if (colView.InternalItemSource.Count - 1 <= firstIndex)
```

src/Tizen.NUI.Components/Controls/RecyclerView/Layouter/LinearLayouter.cs
```diff
@@ -204,10 +204,19 @@ public override void Initialize(RecyclerView view)
             bool failed = false;
 
             //Final Check of FirstIndex
-            while (colView.InternalItemSource.IsHeader(firstIndex) ||
+
+            if (colView.InternalItemSource.Count - 1 < firstIndex)
```
jaehyun0cho, line 208: how about changing the condition to be the same with the condition in the while loop?
```
if (colView.InternalItemSource.Count - 1 <= firstIndex)
```

src/Tizen.NUI.Components/Controls/RecyclerView/Layouter/GridLayouter.cs
everLEEst: This is intentional difference.

if firstIndex is exceed the count is certainly failed,
but if firstIndex is last item, and if that item is not Header,Footer,GroupHeader,GropuFooter, then it can be really our first founded item.
that is why this condition is different.

jaehyun0cho (approved):
LGTM

jaehyun0cho (approved):
LGTM

src/Tizen.NUI.Components/Controls/RecyclerView/Layouter/GridLayouter.cs
```diff
@@ -509,7 +518,7 @@ public override void NotifyItemInserted(IItemSource source, int startIndex)
             // Insert Single item.
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (colView == null) return;
-            if (isSourceEmpty)
+            if (isSourceEmpty || StepCandidate == 0)
```
hinohie, line 521: I cannot imagine s.t. `isSourceEmpty` is false, but `StepCandidate` is 0. (Since we make StepCanditate as 1 at line 280) Is this branch is right way to determine empty group case?

everLEEst: StepCandidate is for normal item candidate.
if source has a group but chlidren is empty, isSourceEmpty is false but children is empty so StepCandidate should be 0.
and NotifyItemInserted is called, we don't have stepCandidate, so we need to re-intialize to find out this source has Children or not and re-calculate StepCandidate.

PR #4777, opened 2022-11-25
[NUI] Support back navigation

jaehyun0cho:
### Description of Change ###

To support back navigation, EnableBackNavigation and OnBackNavigation of Navigator and Page are added.
Moreover, for better usage, back navigation has been modified as follows.
If navigator has more than one page, then the peek page is popped when back navigation proceeds.
Otherwise, the current program is exited when back navigation proceeds.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
+++ b/src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
@@ -841,6 +841,43 @@ namespace Tizen.NUI.Components
         }
 
         /// <summary>
+        /// Sets the default navigator of the given window.
+        /// SetDefaultNavigator does not remove the previous default navigator from the window.
+        /// Therefore, if a user does not want to show the previous default navigator, then it should be removed from the window manually.
+        /// </summary>
+        /// <exception cref="ArgumentNullException">Thrown when the argument window is null.</exception>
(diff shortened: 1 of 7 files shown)


src/Tizen.NUI.Components/Controls/Navigation/Navigator.cs
```diff
@@ -696,6 +727,14 @@ public Page Peek()
             return navigationPages[navigationPages.Count - 1];
         }
 
+        /// <summary>
+        /// Gets or sets if Navigator proceeds back navigation when back button or back key is pressed and released.
+        /// Back navigation pops the peek page if Navigator has more than one page.
+        /// If Navigator has only one page, then the current program is exited.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool EnableBackNavigation { get; set; } = true;
```
everLEEst, line 736: shouldn't it be bindableProperty?

jaehyun0cho: Thank you for your comment!
You are right. I have updated it to be a bindable property.

dongsug-song (approved):
looks good to me

taehyub (approved):
LGTM :)

PR #4782, opened 2022-11-29
[NUI] fix to be able to set null to AnimatedVectorImageView.ResoureUrl property

dongsug-song:
### Description of Change ###
[NUI] fix to be able to set null to AnimatedVectorImageView.ResoureUrl property

### API Changes ###
none


hinohie (approved):
Oh.. LGTM

PR #4790, opened 2022-12-01
[NUI] Fix Navigator Insert to keep pages order correctly

jaehyun0cho:
Previously, pages order was not kept correctly when a page was inserted. Because of that, inserted page overlaps the next page's page transition. e.g.
- Let page1 be pushed.
- Let page2 be inserted before page1.
- Pop page1. Then page1's page transition is hidden by page2 because page2 is added later so page2 overlaps page1.

Now, to resolve the above issue, the order of the inserted page is updated based on its page index value.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM.

hinohie (approved):

PR #4793, opened 2022-12-02
[NUI] fix group remove error.

everLEEst:
when Nth empty item group is removed,
we didn't restore parentIndex while searching group item,
so N+1 group is not updated, as it changed Nth.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

hinohie (approved):

PR #4796, opened 2022-12-02
[NUI] Add implicit dispose check

dongsug-song:
### Description of Change ###
[NUI] Add implicit dispose check
- Add implicit disposed status check code for TouchEvent and PropertySet event handler which is frequently used

### API Changes ###
none


hinohie (approved):
LGTM. To get some debug log

PR #4804, opened 2022-12-05
[NUI] Fix Navigator to enable multiple Pop()

jaehyun0cho:
A popped page is removed in the Pop() animation's Finished callback.

Previously, Navigator could not support multiple Pop().
Because Pop() simply stopped the previous Pop() animation without
calling the animation's Finished callback.

Now, Pop() stops the previous Pop() animation with calling the
animation's Finished callback by emitting FinishedSignal().
Consequently, Navigator can support multiple Pop() now.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #4817, opened 2022-12-08
[NUI][API10] Animated view's size cache into NUI

hinohie:
When view's size animated, the DALi known size and NUI stored size (mean, userSizeWidth and widthPolicy like) will not be matched.

If NUI use fixed size policy (mean, It "May" not use NUI.Layout system so can use animation feature.) Let we make NUI keep dali-side size's information and keep it.

### API Changes ###

 - ACR:


dongsug-song (approved):
great!
approved

PR #4819, opened 2022-12-09
[NUI] Fix WidthSpecification and HeightSpecification not to set Size2D

hinohie:
We don't need to change the width value when we try to set the height. Also, widthPolicy might not matched with SizeWidth when we use NUI.Animation

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
@@ -733,12 +733,6 @@ namespace Tizen.NUI.BaseComponents
                 }
                 Object.InternalRetrievingPropertyVector2ActualVector3(view.SwigCPtr, View.Property.SIZE, view.internalSize2D.SwigCPtr);
 
-                // Update NUI stored internal size informations only both widthPolicy and heightPolicy are fixed.
-                if(view.widthPolicy >= 0 && view.heightPolicy >= 0)
-                {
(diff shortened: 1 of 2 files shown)


dongsug-song (approved):
approved

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
```diff
@@ -733,12 +733,6 @@ public partial class View
                 }
                 Object.InternalRetrievingPropertyVector2ActualVector3(view.SwigCPtr, View.Property.SIZE, view.internalSize2D.SwigCPtr);
 
-                // Update NUI stored internal size informations only both widthPolicy and heightPolicy are fixed.
```
jaehyun0cho, line 736: could you restore this?

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
```diff
@@ -1580,12 +1574,6 @@ public partial class View
                 }
                 Object.InternalRetrievingPropertyVector3(view.SwigCPtr, View.Property.SIZE, view.internalSize.SwigCPtr);
 
-                // Update NUI stored internal size informations only both widthPolicy and heightPolicy are fixed.
```
jaehyun0cho, line 1583: could you restore this?

jaehyun0cho (approved):
LGTM

PR #4837, opened 2022-12-15
Revert "[NUI] fix the defect that the focus of VitualRemote App is stuck very rarely"

dongsug-song:
### Description of Change ###
This reverts commit 0017dd0e49ffbc78d8c51c6f068afd6e8fdbc8d0.
- the root cause of this defect was VirtualRemote App had misused BaseHandle.Reset()
- BaseHandle.Reset() will be deprecated because it is not proper for nui. (it resets native dali BaseHandle's ref count)

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/LottieAnimationView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/LottieAnimationView.cs
@@ -176,7 +176,7 @@ namespace Tizen.NUI.BaseComponents
                         }
                     }
                 }
-                //Tizen.Log.Info(tag, $"  [ERROR][{GetId()}](LottieAnimationView) Fail to get URL from dali >");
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/BaseComponents/LottieAnimationView.cs
```diff
@@ -176,7 +176,7 @@ private string InternalURL
                         }
                     }
                 }
-                //Tizen.Log.Info(tag, $"  [ERROR][{GetId()}](LottieAnimationView) Fail to get URL from dali >");
+                Tizen.Log.Error(tag, $"  [ERROR][{GetId()}](LottieAnimationView) Fail to get URL from dali >");
```
hinohie, line 179: Isn't it a good code?

hinohie (approved):

PR #4839, opened 2022-12-15
[NUI] add ImageUrl property on DefaultGridItem.

everLEEst:
### Description of Change ###

Add ImageUrl Property on DefaultGridItem.
DefaultGridItem Image is readonly property,
so in XAML ResourceUrl is unaccessible.
this patch will fix those issue by directly binding resourceUrl on ImageUrl of DefaultGridItem.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

hinohie (approved):
Maybe minor. Can't we use property's name as `ResourceUrl` to sync with `ImageView` usage?

PR #4845, opened 2022-12-16
[NUI] make BaseHandle.Reset() not work

dongsug-song:
### Description of Change ###
[NUI] make BaseHandle.Reset() not work
- this came from native dali and it is not suitable for nui.
- it makes BaseHandle's reference count as 0, but this action could cause memory leak in nui side.
- in nui side, calling Dispose() explicitly or doing nothing to delegate disposing action to the DisposeQueue is recommended.

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/public/Common/BaseHandle.cs
+++ b/src/Tizen.NUI/src/public/Common/BaseHandle.cs
@@ -432,10 +432,12 @@ namespace Tizen.NUI
         /// Resets the handle.
         /// </summary>
         /// <since_tizen> 3 </since_tizen>
+        /// <remark>
+        /// This will be deprecated in next version, please use Dipose() instead.
+        /// </remark>
         public void Reset()
         {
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Common/BaseHandle.cs
```diff
@@ -434,8 +434,7 @@ public bool GetTypeInfo(Tizen.NUI.TypeInfo info)
         /// <since_tizen> 3 </since_tizen>
```
Seoyeon2Kim, line 434: Understood what you mean, but it's a public method. So, I think it would be better to add <remarks> or any description here, too.
What do you think?

dongsug-song: yes, you are absolutely right. I have added remark comment. thank you.

Seoyeon2Kim (approved):
LGTM!
Thank you for your commit!

PR #4847, opened 2022-12-16
[NUI] Fix FlexLayout OnMeasure to calculate size correctly

jaehyun0cho:
Previously, the measured size of FlexLayout with WrapContent could be calculated like MatchParent.
The above issue happened in the following case.
Tizen.NUI.Components.DialogPage.ShowAlertDialog("Title", "Message", null);

To resolve the above issue without breaking backward compatibility, FlexLayout OnMeasure has been fixed to use ResolveSizeAndState() instead of GetDefaultSize() when calculating the measured size.

To test the issue, showing a dialog without action button case has been added to Tizen.NUI.StyleGuide.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
Thank you for fixing the issue and the detailed description.
I'll test the example you've updated. LGTM!

everLEEst (approved):
LGTM.

PR #4850, opened 2022-12-19
[NUI] Pass SelectionChangedEventArgs by default as SelectionChangedCommandParameter.

everLEEst:
### Description of Change ###

In SelectionChangedCommand,
SelectionChangedCommandParamater is the only accessible value for changing selection.
if user sets SelectionChangedCommandParamater to SelectedItem of CollectionView,
Parameter will be updated after SelectedItemProperty is updated,
So SelectionChangedCommand which calls in the SelectedItemProperty,
cannot get updated SelectedItem value.

Here is a simple solution for that.
If nothing exist in SelectionChangedCommandParameter,
CollectionView will pass SelectionChangedEventArgs by default.
So user can get PreviousSelection and CurrentSelection in the model.

This change makes the model dependent on SelectionChangedEventArgs,
which I wish to avoid but we don't have a proper solution for now.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/RecyclerView/CollectionView.cs
+++ b/src/Tizen.NUI.Components/Controls/RecyclerView/CollectionView.cs
@@ -47,8 +47,8 @@ namespace Tizen.NUI.Components
 
                     oldValue = colView.selectedItem;
                     colView.selectedItem = newValue;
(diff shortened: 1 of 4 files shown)


dongsug-song (approved):
approved

PR #4860, opened 2022-12-23
[NUI] fix testhub crash issue

dongsug-song:
### Description of Change ###
[NUI] fix testhub crash issue

- Add PanGesture and PropertyNotification class's Dispose()
- tested OK by sample of DisposeWithoutEventUnsubscribedTest.cs

### API Changes ###
none

### DIFF
--- a/src/Tizen.NUI/src/public/Common/PropertyNotification.cs
+++ b/src/Tizen.NUI/src/public/Common/PropertyNotification.cs
@@ -30,7 +30,7 @@ namespace Tizen.NUI
     {
 
         private DaliEventHandler<object, NotifyEventArgs> propertyNotificationNotifyEventHandler;
-        private NotifyEventCallbackDelegate propertyNotificationNotifyEventCallbackDelegate;
+        private NotifyEventCallbackDelegate propertyNotificationNotifyEventCallback;
 
         /// <summary>
         /// Create a instance of PropertyNotification.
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI/src/public/Common/PropertyNotification.cs
```diff
@@ -66,24 +66,29 @@ internal PropertyNotification(global::System.IntPtr cPtr, bool cMemoryOwn) : bas
...
             {
-                // Restricted to only one listener
                 if (propertyNotificationNotifyEventHandler == null)
                 {
-                    propertyNotificationNotifyEventHandler += value;
-
-                    propertyNotificationNotifyEventCallbackDelegate = new NotifyEventCallbackDelegate(OnPropertyNotificationNotify);
-                    this.NotifySignal().Connect(propertyNotificationNotifyEventCallbackDelegate);
+                    propertyNotificationNotifyEventCallback = OnPropertyNotificationNotify;
+                    using PropertyNotifySignal signal = new PropertyNotifySignal(Interop.PropertyNotification.NotifySignal(SwigCPtr), false);
```
Seoyeon2Kim, line 72: Great, It's very reasonable to use `using` statement here not to make warning message :)

src/Tizen.NUI/src/public/Events/PanGestureDetector.cs
```diff
@@ -71,19 +71,25 @@ internal PanGestureDetector(global::System.IntPtr cPtr, bool cMemoryOwn) : base(
...
             remove
             {
                 detectedEventHandler -= value;
-
-                if (detectedEventHandler == null && DetectedSignal().Empty() == false)
+                if (detectedEventHandler == null)
                 {
-                    DetectedSignal().Disconnect(detectedCallback);
+                    using PanGestureDetectedSignal signal = new PanGestureDetectedSignal(Interop.PanGestureDetector.DetectedSignal(SwigCPtr), false);
+                    if (signal?.Empty() == false)
```
Seoyeon2Kim, line 85: Just my curiosity :
From now on, should we check `Empty()` before calling `Disconnect` and null check?
AFAIK, we normally call Disconnect and set null without checking `Empty()` of signal.

test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/DisposeWithoutEventUnsubscribedTest.cs
```diff
@@ -27,7 +28,7 @@ public class DisposeWithoutEventUnsubscribedTestAndAnimationDisposeTest : IExamp
         public void Activate()
         {
             DisposeWithoutUnsubscribedEventTest();
-            AnimationDisposeTest();
+            //AnimationDisposeTest();
```
Seoyeon2Kim, line 31: Thanks for adding an updated example.
BTW, why do you comment at this line and 296 line? Probably, for history?

Seoyeon2Kim (commented):
I added some minor comments.

src/Tizen.NUI/src/public/Events/PanGestureDetector.cs
dongsug-song: yes, you are right, it is not very necessary. but it is secure to check again the native dali signal is disconnected surely.
and also in Dispose(), the callback null check is performed so before Dispose(), setting null to callback should be done carefully. (because if the signal is not disconnected here and callback is set to null, then signal disconnection in Dispose() will be missed. this is why the double checking has been added.

test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/DisposeWithoutEventUnsubscribedTest.cs
dongsug-song: good point~ 
it is difference purpose so I added more comments to remind easily.
thank you.

src/Tizen.NUI/src/public/Events/PanGestureDetector.cs
Seoyeon2Kim: I see. Now this commit can be a leading code for double checking. Also, I can refer to it later~
Thank you for checking them!

test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/DisposeWithoutEventUnsubscribedTest.cs
Seoyeon2Kim: You updated the comment! Thanks for the feedback. :)
It's really good at scenario.

Seoyeon2Kim (approved):
LGTM!
This PR is not only fixing the crash issue, but also providing an example of Dispose()

PR #4873, opened 2022-12-29
[NUI][DO NOT MERGE] Add Cancel Camera Transition in SceneView

bshsqa:
Signed-off-by: seungho baek <sbsh.baek@samsung.com>

### API Changes ###

 - ACR:


hinohie:
I think `SkipTransition` is more useful API then Cancel.

PR #4880, opened 2023-01-05
[NUI] Support Device orientation and window orientation event.

sparrow74:
Supporting  Device orientation and Window Orienation event
when device orientation is changed or window orientation is changed.
Window orientation event is emitted by display server.
To emit the Window Orientation event, AddAvailableOrientation()
or SetPreferredOrientation() should be called before the device is rotated.
Otherwise, Device Orientation event is emitted by Application framework
and any condition is not needed.


dongsug-song (approved):
approved

PR #4882, opened 2023-01-06
[NUI] Add method to DisposeRecursively. + Fix DisposeTest demo works

hinohie:
Previous Container.cs can dispose recursively only if it is created by xaml. (DisposeIncludeChildren)
And also, that API only for internal.

Now we add some useful API that user can dispose all connected children recursively by one button.

```
View a = new View();
View b = new View();

a.Add(b);
a.DisposeRecursively();

// Now both a and b are disposed.
```

Signed-off-by: Eunki Hong <h.pichulia@gmail.com>


dongsug-song (approved):
approved

PR #4897, opened 2023-01-17
[NUI] Make PixelBuffer & PixelData as Disposable

hinohie:
Since they were BaseHandle, there was some Fatal log printed when we try to use them at worker thread.

For example, PixelBuffer buffer = ImageLoader.LoadImageFromFile() was invalid.

This patch make them as Disposable.
So they will not be registed into Registry.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #4899, opened 2023-01-18
[NUI] Fix bug where window size didn't change when borderline was dynamically changed.

JoogabYun:
### Description of Change ###

When I dynamically changed the borderline, only the padding changed and the window size did not change.
Fix this so that the window size also changes.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window/BorderWindow.cs
+++ b/src/Tizen.NUI/src/public/Window/BorderWindow.cs
@@ -137,41 +137,9 @@ namespace Tizen.NUI
         {
             if (borderInterface != null)
             {
-                float height = 0;
-                if (hasTopView) height += topView.SizeHeight;
-                if (hasBottomView) height += bottomView.SizeHeight;
-
-                if (height != borderHeight)
-                {
-                    float diff = height - borderHeight;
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Window/BorderWindow.cs
```diff
@@ -137,10 +137,34 @@ internal void UpdateProperty()
         {
             if (borderInterface != null)
             {
+                if (borderLineThickness != borderInterface.BorderLineThickness)
+                {
+                    int diffBorderLine = (int)(borderInterface.BorderLineThickness - borderLineThickness);
```
hinohie, line 142: currently, `borderLineThickness` is `uint` so it's not a big deal..but anyway,

Could we cast unsigned-to-signed value by value, for safety? if `borderLineThickness` type changed, it will make int overflow.

```suggestion
                    int diffBorderLine = (int)borderInterface.BorderLineThickness - (in t)borderLineThickness;
```

JoogabYun: 수정했습니다.~

hinohie (approved):

PR #4902, opened 2023-01-18
[NUI] Reorder Disposable's ReleaseSwigCPtr + Allow to dispose seperat…

hinohie:
…e thread

Reorder the `ReleaseSwigCPtr` call time and `SwigCPtr` value setup. It will be matched as `BaseHandle.cs` doing now.

Plus

Make we don't print any fatal log when we try to dispose `Disposable` class. For example, `Vector2` or `Uint16Pair` don't need to be used only for main thread. Only some UI releative things has some dependency issue of Main / UI thread.


dongsug-song (approved):
approved

PR #4903, opened 2023-01-18
[NUI] fix tct block issue in text field and editor

wonrst:
GetInputMethodContext() in ReleaseSwigCPtr crashes due to current changes in BaseHandle.
Move DestroyContext() to Dispose to avoid this.


dongsug-song (approved):
approved

PR #4908, opened 2023-01-18
[NUI] Window resizing is performed in one step if possible.

JoogabYun:
### Description of Change ###

If you resize the window several times, there may be a problem with the configure event with the window server.
So avoid resizing the window as many times as possible.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #4915, opened 2023-01-19
[Privilege] Deprecate privacy related API

yl33:
### Description of Change ###
- Remove privacy related API

### API Changes ###

 - ACR: https://code.sec.samsung.net/jira/browse/TCSACR-524


src/Tizen.Security/Tizen.Security/Privilege.cs
```diff
@@ -153,6 +153,7 @@ public static string GetDescription(string apiVersion, string privilege, Package
         /// <exception cref="System.OutOfMemoryException">Thrown when out of memory occurs.</exception>
         /// <exception cref="System.InvalidOperationException">Thrown when an internal error occurs.</exception>
         /// <exception cref="System.NotSupportedException">The required feature is not supported.</exception>
+        [Obsolete("Deprecated since API level 11.")]
```
wiertel, line 156: it's better to include information about replacement, is there any a way a user can achieve similar result with other API?

earth-table-coffee: There's no replacement of it. Would it be better to add 'Please don't use it' message?

yl33: There' no replacement of it. Would it be better to add 'Please don't use it' message instead?

kd0228-kim (approved):

src/Tizen.Security/Tizen.Security/Privilege.cs
yl33: @wiertel Please review this PR.

yl33:
something wrong with auto rebase, will open new PR

PR #4920, opened 2023-01-25
[Applications.Common] Remove log prints in the thread

hjhun:
### Description of Change ###

Currently, the CoreApplication prints the log for improving the performance of ToLower() and ToUpper().
This causes the resource contention. Because creating the thread needs the resource as like the fork() in the linux system.
This request removes the thread creation for removing the resource contention.


jusung-son (approved):

sung-su (approved):
Nice catch.

PR #4935, opened 2023-01-27
[NUI.Gadget] Add NUIGadget class for inhouse developers

hjhun:
### Description of Change ###

The NUIGadget is the UI Gadget model for the NUIApplication. The NUIApplication is able to handle gadgets using the NUIGadgetManager.
The dll file of the gadget that is derived by the NUIGadget should be installed to 'res/allowed' or 'res/global' directory.
If the developer wants to allow it only to a specific application, the developer should install it in the 'res/allowed' directory.
And, the developer adds the allowed application information to the tizen-manifest.xml file.
This feature is for inhouse developers.  The NUIGadget is available for platform level signed resource packages.


src/Tizen.NUI.Gadget/Tizen.NUI/NUIGadgetManager.cs
```diff
@@ -0,0 +1,294 @@
...
+                Log.Warn("NUIGadgetAssembly.Load(): " + info.ResourcePath + info.ExecutableFile + " ++");
+                assembly = Assembly.Load(File.ReadAllBytes(info.ResourcePath + info.ExecutableFile));
+                Log.Warn("NUIGadgetAssembly.Load(): " + info.ResourcePath + info.ExecutableFile + " --");
+            }
+            catch (FileLoadException e)
+            {
+                throw new InvalidOperationException(e.Message);
+            }
+
+            NUIGadget gadget = (NUIGadget)assembly.CreateInstance(className);
```
pjh9216, line 146: If the developer accidentally does not inherit NUIGadget type, it may not be possible to change to the type.
How about using 'as' and check the result?
```cs
NUIGadget gadget = assembly.CreateInstance(className) as NUIGadget;
if (gadget == null)
    throw new InvalidOperationException(...); 
```

src/Tizen.NUI.Gadget/Tizen.NUI/NUIGadget.cs
```diff
@@ -0,0 +1,265 @@
...
+using Tizen.NUI.BaseComponents;
+
+namespace Tizen.NUI
+{
+    /// <summary>
+    /// This class represents a NUIGadget controlled lifecycles.
+    /// </summary>
+    /// <since_tizen> 11 </since_tizen>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class NUIGadget
```
pjh9216, line 29: Why don't you add 'abstract' to prevent direct creation of instances?

src/Tizen.NUI.Gadget/Tizen.NUI/NUIGadgetManager.cs
hjhun: Done. Thanks.

src/Tizen.NUI.Gadget/Tizen.NUI/NUIGadget.cs
hjhun: I thought if the developer does not override the OnCreate() method, it's meaningless. So I didn't consider adding the 'abstract' keyword to the class.
Done. Thanks.
I think adding 'abstract' keyword is better than the previous implementation.

src/Tizen.NUI.Gadget/Tizen.NUI/NUIGadget.cs
```diff
@@ -0,0 +1,265 @@
...
+        /// <since_tizen> 11 </since_tizen>
+        public NUIGadgetLifecycleState State
+        {
+            internal set;
+            get;
+        }
+
+        internal void Create()
+        {
+            MainView = OnCreate();
```
pjh9216, line 96: I think making view is not an option.
How about changing return type and handling exceptions at caller?
```cs
internal bool Create()
{
    MainView = OnCreate();
    if (MainView == null)
       return false;
    return true;
}

```

hjhun: Done. Thanks.

I've updated the patchset.
If the Create() method returns 'false', the Add() method throws the InvalidOperationException.

pjh9216 (approved):
LGTM

hjhun:
> LGTM

Thank you for the review.
I've modified the NUIGadgetManager implementation to use the CoreApplication.Post() instead of the TizenUISynchronizationContext.Post().

pjh9216 (approved):

PR #4940, opened 2023-01-28
[NUI][Scene3D] Fix memory leak issue for Model's Animation

hinohie:
Since we create new handle at csharp binder, we should call DeleteHandle function.

Note : Extensions.GetInstanceSafely delete the getted cPtr's reference inside of that function. So we cannot use it.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #4941, opened 2023-01-28
[NUI][Scene3D][API10] Fix memory leak issue for Model's Animation

hinohie:
Since we create new handle at csharp binder, we should call DeleteHandle function.


dongsug-song (approved):
approved

PR #4945, opened 2023-01-29
[NUI][API10] Fix ImageUrl.Dispose issue

hinohie:
Since we update ImageView lazy, ImageUrl's lifecycle might not matched what user think.

```
ImageUrl imageUrl = ~~~.GenerateUrl();
ImageView imageView = new ImageView(imageUrl.ToString());
imageUrl.Dispose(); ///< We should allow this situation.
```

To make scene, We allow to call UpdateImage(); synchoronously if we use GeneratedUrl().

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #4946, opened 2023-01-30
[NUI] Fix NUI button not working when style disabled

rabbitfor:
There was a flag that prevent state updating while applying style. But the flag seemed to have a bug : when style is disabled, the state updating is never going to be done.

This patch fixes the bug and changes the flag's name to be more contextual: styleApplied -> styleApplying.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

hinohie (approved):
Test done. Looks God To Me.

PR #4955, opened 2023-02-01
[NUI] Fix to use SizeHeight instead of Size with width 0

jaehyun0cho:
Previously, although users sets size with width 0, the width is updated by dali's size calculation (e.g. text calculation) instead of width 0 set by user.

The following patch fixed the above bug.
8033b0ed4a31ebcd9449c2e0c98074247502c9be

By applying the above patch, some size assignment in theme should be fixed to set size height instead of size with width 0.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM

Seoyeon2Kim (approved):
LGTM!
화이팅!

PR #4958, opened 2023-02-02
[NUI] Revive View constructor with style working again in TV

rabbitfor:
And hide PROFILE_TV flags to make clear code

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/Theme/ThemeManager.cs
```diff
@@ -324,7 +332,7 @@ internal static void ApplyExternalPlatformTheme(string id, string version)
             }
 
             // If no platform theme has been applied and the base theme can cover the given one, do nothing.
-            if (platformTheme == null && baseTheme.HasSameIdAndVersion(id, version))
+            if (platformTheme == null && baseTheme != null && baseTheme.HasSameIdAndVersion(id, version))
```
jaehyun0cho, line 335: baseTheme can be nullable?

dongsug-song (approved):
approved

src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
```diff
@@ -1560,10 +1560,12 @@ private void DisConnectFromSignals()
         [EditorBrowsable(EditorBrowsableState.Never)]
         protected virtual void InitializeStyle(ViewStyle style = null)
         {
-#if PROFILE_TV
-            // tv profile doesn't use NUI Style, so do nothing and just return here!
-            return;
-#endif
+            if (style == null && ThemeManager.InitialThemeDisabled)
```
jaehyun0cho, line 1563: Is ok if style is not null and ThemeManager.InitialThemeDisabled is false?

src/Tizen.NUI/src/public/Theme/ThemeManager.cs
rabbitfor: It was not in the previous code. But to reduce memory usage (for VD), I decided to make it nullable. I also edited all lines referencing `baseTheme` to avoid null exception.

src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
rabbitfor: For example, 
```
var Button = new Button(new ButtonStyle() {
/* style description */
});
```
In this case, the `style` is not null.
If it is a TV application, `InitialThemeDisabled` is false.

The code should not pass the style processing in this condition for cases like this.

jaehyun0cho: oh right! This patch is to apply style for this case!

jaehyun0cho (approved):
LGTM

PR #4962, opened 2023-02-03
[NUI] Disable compiler warnings complaining about theme off in TV profile

rabbitfor:
There are some code that are unreachable according to device profile.
It's a configuration on compilation which needs a change in the constant to support different devices.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

hinohie (approved):

PR #4964, opened 2023-02-03
[Applications.RPCPort] Add a new Parcel constructor

hjhun:
### Description of Change ###

A new constructor of the Parcel is added.
It has a parameter that is the 'withHeader'. If it's false, the parcel object does not have the header.

### API Changes ###

 - ACR: TCSACR-526

Added:
 - Parcel(bool withHeader);


src/Tizen.Applications.Common/Tizen.Applications.RPCPort/Parcel.cs
```diff
@@ -158,13 +158,24 @@ public class Parcel : IDisposable
         /// <summary>
         /// Constructor for this class.
         /// </summary>
+        /// <param name="withoutHeader">If it's true, the parcel object does not have the header. (Since Tizen 11)</param>
         /// <exception cref="InvalidIOException">Thrown when an internal IO error occurs.</exception>
         /// <since_tizen> 5 </since_tizen>
-        public Parcel()
+        public Parcel(bool withoutHeader = false)
```
wiertel, line 164: This would break ABI - meaning that applications built for older TizenFX dll will crash on executing this constructor with newer dll. Even though the app could be compiled against changed new TizenFX without any changes.

BTW, why not   withHeader=true   to avoid negations, logic inversion?

The solution is to add second constructor (the parameter could also be optional) and maintain them both from now on.
```
public Parcel(bool withoutHeader) {
...
}
public Parcel() : this(false) {}
```
It would be more tricky if you don't want to maintain both of them and would rather end up with only one with optional parameter. In that case please deprecate the non-parameter constructor, add new one with mandatory parameter. Later, after 2 API levels, you could remove the deprecated one and make the parameter of the other constructor optional if you want. This way app developers will be warned to switch to the constructor with parameter, and if they switch their apps will be compatible with dll prior removing `Parcel()` and when it has been removed finally.

wiertel (changes requested):
ABI break

src/Tizen.Applications.Common/Tizen.Applications.RPCPort/Parcel.cs
hjhun: Thank you first. I did not consider ABI break. That makes a big problem for exsiting applications.
I will define a new constructor as per your suggestion and use this keyword to fix the problem in the current implementation.

hjhun:
> ABI break

Fixed.

hjhun:
Dear @wiertel ,
Please review this request.

wiertel (approved):

PR #4972, opened 2023-02-07
[NUI.Scene3D] Do not dispose camera internally

hinohie:
SceneView's camera is not an instance object anymore. It can be holded by app side.
So, if we dispose it inside of CameraTransition Function, It become sudden dead.

Actually, Dali side camera object still alive, but the NUI.Camera handle is dereferenced. So, If app try to access that camera, crash occured.


Seoyeon2Kim (approved):
LGTM!

jaehyun0cho (approved):
LGTM

rabbitfor:
If warnings are ignorable, you'd better use 'SuppressMessageAttribute'

hinohie:
> If warnings are ignorable, you'd better use 'SuppressMessageAttribute'

Thank you for your guide. I made one more PR #4982 now. It will remove all warning messages in NUI.Scene3D.

PR #4978, opened 2023-02-09
[NUI] Fix Picker's HeightSpecification

jaehyun0cho:
Previously, HeightSpecifications of Picker, DatePicker, TimePicker are MatchParent.
So the Pickers' SizeHeights are calculated based on their parent's SizeHeight.
This causes the recursive size calculation problem if Picker's parent has WrapContent HeightSpecification. (e.g. Dialog)

Now, the Pickers' SizeHeights are set in DefaultThemeCommon as follows. Picker's SizeHeight is set in DefaultThemeCommon.
DatePicker's Pickers SizeHeights are set in DefaultThemeCommon. TimePicker's Pickers SizeHeights are set in DefaultThemeCommon.

Therefore, Picker does not need to have MatchParent HeightSpecification.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/DatePicker.cs
+++ b/src/Tizen.NUI.Components/Controls/DatePicker.cs
@@ -287,8 +287,6 @@ namespace Tizen.NUI.Components
 
         private void Initialize()
         {
-            HeightSpecification = LayoutParamPolicies.MatchParent;
-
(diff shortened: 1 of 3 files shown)


hinohie (approved):

Seoyeon2Kim (approved):
LGTM~

PR #4990, opened 2023-02-14
 [NUI] DragAndDrop : don't reuse drag window

taehyub:


src/Tizen.NUI/src/public/DragAndDrop/DragAndDrop.cs
```diff
@@ -76,79 +101,74 @@ public void StartDragAndDrop(View sourceView, View shadowView, DragData dragData
...
-            //Update Window Directly
-            mDragWindow.VisibiltyChangedSignalEmit(true);
-            mDragWindow.RenderOnce();
+                    callback((DragSourceEventType)sourceEventType);
+                };
 
-            sourceEventCb = (sourceEventType) =>
-            {
-                if ((DragSourceEventType)sourceEventType == DragSourceEventType.Finish)
+                if (!Interop.DragAndDrop.StartDragAndDrop(SwigCPtr, View.getCPtr(sourceView), Window.getCPtr(mDragWindow), dragData.MimeType, dragData.Data,
```
juns-kim, line 156: The "StartDragandDrop" API is a request for the Display server to allow "mDragWindow" to belong to the cursor.
After this API call, any change to "mDragWindow" can change the position of "mDragWindow".
For prevent mis-positioning of the mDragWindow, It's better that called after the mDragWindow.Show().

juns-kim (commented):
It could cause drag window blinking issue. please check the comment :)

src/Tizen.NUI/src/public/DragAndDrop/DragAndDrop.cs
taehyub: modified the codes following this guide

PR #4993, opened 2023-02-15
[NUI] Fix Switch thumb positioning

rabbitfor:
When the size of thumb or track changes, the it should update the position of the thumb properly.

And to prevent unnecessary switch state updating while applying style, upgrade styleApplying flag to int type.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
LGTM~

PR #4994, opened 2023-02-15
[NUI] Improve behaviors of ThemeManager and add an event

rabbitfor:
* ThemeChanging event: An event called right after the theme changed and before applying it to the view.
* ClearAllAppliedThemes: Since every themes passed by ApplyThemes() are cumulative, user may want to clear all.
* Open Theme.indexer public: This is for C_XAML.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #4996, opened 2023-02-15
[NUI] Fix Button Theme not to set fixed size

jaehyun0cho:
To increase Button width based on the Button's text, Button Theme sets MinimumSize instead of fixed Size.

### API Changes ###

 - ACR:


rabbitfor (approved):
LGTM

PR #5001, opened 2023-02-16
[NUI] Improve behaviors of ThemeManager and add an event

rabbitfor:
* ThemeManager.AppendTheme: While ApplyTheme() clears previous applied user theme, AppendTheme() appends it to the previous one.
* ThemeChanging event: An event called right after the theme changed and before applying it to the view.
* Open Theme.indexer public: This is for C_XAML.
* Theme implements IResourcesProvider: to support XAMLResources in xaml code

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #5002, opened 2023-02-17
[NUI] Fix Slider ValueIndicatorImage and Thumb size

Seoyeon2Kim:
### Description of Change ###

- Update Slider ValueIndicatorImage / Text to change its size depending on the content.
- Thumb size uses natural size of its image size.
- Update Slider style guide sample for ValueIndicator

### API Changes ###
- N/A

### DIFF
--- a/src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
+++ b/src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
@@ -273,7 +273,8 @@ namespace Tizen.NUI.Components
                 },
                 Thumb = new ImageViewStyle()
                 {
-                    Size = new Size(40, 40),
+                    WidthResizePolicy = ResizePolicyType.UseNaturalSize,
+                    HeightResizePolicy = ResizePolicyType.UseNaturalSize,
                     ResourceUrl = new Selector<string>()
                     {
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
```diff
@@ -284,7 +285,9 @@ public Theme Create()
                 },
                 ValueIndicatorImage = new ImageViewStyle()
                 {
-                    Size = new Size(43, 40),
+                    SizeHeight = 40,
+                    WidthResizePolicy = ResizePolicyType.FitToChildren,
```
rabbitfor, line 289: When an indicator image has a fixed height but variable width, can it be ellipse in some cases?

rabbitfor (approved):
LGTM

src/Tizen.NUI.Components/Theme/DefaultThemeCommon.cs
Seoyeon2Kim: Yes, the default value is just drawing the border line and `ValueIndicatorImage` property has `ValueIndicatorText` as a child.
So, the width could be resized depending on the text length. :)
Thank you for your review!

PR #5009, opened 2023-02-17
[NUI.Scene3D] Use Model defined camera

hinohie:
Make API that we can use gltf model file defined camera.

Since Tizen.NUI.Scene3D.Camera is one of View & This camera should be children of SceneView to render the scene,
Camera cannot be owned by Model itself.

So, we support only for 'Generate' or 'Apply' camera parameters loaded from gltf model.

This patch required below dali patches merged.

https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/288530
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/288531


jaehyun0cho (approved):
LGTM

PR #5010, opened 2023-02-19
[NUI.Scene3D] Change native default camera create API

hinohie:
Change default camera generate API as 3D engine style.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #5012, opened 2023-02-20
[NUI] Remove unused code in NUI

AdunFang:
### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -711,24 +690,6 @@ public Color BorderlineColor
             }
         }
 
-        /// <summary>
-        /// The color selector for the borderline of the View.
-        /// Like BackgroundColor, color selector typed BorderlineColor should be used in ViewStyle only.
-        /// So this API is internally used only.
-        /// </summary>
-        internal Selector<Color> BorderlineColorSelector
```
hinohie, line 719: As I know, BorderlineColorSelector added due to the APP side request.
@Jaehyun-Cho  Could you check this is removable?

src/Tizen.NUI/src/public/BaseComponents/ViewEnum.cs
```diff
@@ -199,11 +199,9 @@ internal class Property
             internal static readonly int ParentOrigin = Interop.ActorProperty.ParentOriginGet();
             internal static readonly int ParentOriginX = Interop.ActorProperty.ParentOriginXGet();
```
hinohie, line 200: Is there any usage with `ParentOriginX` and `ParentOriginY`, instead of `ParentOriginZ`?

src/Tizen.NUI/src/public/BaseComponents/View.cs
rabbitfor: I'm afraid it is used in `ViewStyle` class. Please check `ViewStyle.BorderlineColorSelector`.
When `ApplyStyle()`, all properties in ViewStyle class are copied to the corresponding View.
You need to remove both `View.BorderlineColorSelector` and `ViewStyle.BorderlineColorSelector`, or none of them.
But please note that `ViewStyle.BorderlineColorSelector` is currently used in `Tizen.NUI.Components.DefaultThemeCreator.Create()` to decorate outlined button.

AdunFang: Hi, Ms. Yang
Because the View doesn't has the property BorderlineColorSelectorProperty, so it will not copy from ViewStyle to View.

AdunFang: Hi, Ms. Yang
Because the View doesn't has the property BorderlineColorSelectorProperty, so it will not copy from ViewStyle to View.

rabbitfor: `ViewStyle.BorderlineColorSelector` is currently used in `Tizen.NUI.Components.DefaultThemeCreator.Create()` to decorate outlined button.
As fas as I know, if you remove `View.BorderlineColorSelector`, the look of outlined button will change because the value of `ViewStyle.BorderlineColorSelector will not copy to the `View.BorderlineColorSelector.

PR #5014, opened 2023-02-20
[NUI] Fix button bugs and improve switch.

rabbitfor:
* Button's
IconOrientation bug In the previous code, when the IconOrientation property changed, the LayoutItems() was called and it created a new layout. But in case the button was inside another layout, the behaviour described above caused unexpect look. Therefore I fixed it by changing orientation related properties to the existing layout instead of creating a new layout.

* Improve switch and extension relationship
The switch extension does not register any event handler to the switch components anymore.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #5023, opened 2023-02-20
[NUI] Fix Slider thumb position by adding SliderStyle.ThumbMaximumSize

jaehyun0cho:
Slider thumb position is calculated by using the track length. The track length was calculated by subtracting the thumb image size from the slider size.
Thumb image size changes by its control state.
e.g. normal size is (26, 26) and pressed size is (40, 40). So the right end position of the thumb changes by its control state.

To resolve the above, the thumb position is calculated by using the maximum size of the thumb image all the time.
e.g. pressed size (40, 40) is used for the thumb position calculation.

To use the maximum size of the thumb image, ThumbMaximumSize is added to SliderStyle.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Style/SliderStyle.cs
+++ b/src/Tizen.NUI.Components/Style/SliderStyle.cs
@@ -227,6 +227,15 @@ namespace Tizen.NUI.Components
             set => SetValue(TrackPaddingProperty, value);
         }
 
+        /// <summary>
+        /// Gets or sets the maximum size of the thumb.
+        /// This is used to calculate the track length and thumb position.
+        /// The track length is calculated by subtracting the maximum size of the thumb from the slider size.
+        /// The thumb position is calculated by using the track length.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Size ThumbMaximumSize { get; set; } = new Size();
+
         /// <inheritdoc/>
         /// <since_tizen> 8 </since_tizen>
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI.Components/Style/SliderStyle.cs
```diff
@@ -227,6 +227,15 @@ public Extents TrackPadding
...
         }
 
+        /// <summary>
+        /// Gets or sets the maximum size of the thumb.
+        /// This is used to calculate the track length and thumb position.
+        /// The track length is calculated by subtracting the maximum size of the thumb from the slider size.
+        /// The thumb position is calculated by using the track length.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Size ThumbMaximumSize { get; set; } = new Size();
```
Seoyeon2Kim, line 237: Just my curiosity :
Without bindableproperty, is it possible to declare `new` as Size constructor?

Seoyeon2Kim (commented):
Thank you for update Slider implementation.
I have a question, 
is this PR to prevent any corner cases that can occur when user arbitrarily puts in a huge thumb image, as Eunki mentioned last time?
Or, as you know, there are low/high indicator issues now and is it solving it?

jaehyun0cho:
As @Seoyeon2Kim mentioned, this PR has problems.
Instead of this PR, the following PR is appropriate.

PR #5024, opened 2023-02-20
[NUI] Fix Slider thumb position by adding SliderStyle.ThumbMaximumSize

jaehyun0cho:
Slider thumb position is calculated by using the track length. The track length was calculated by subtracting the thumb image size from the slider size.
Thumb image size changes by its control state.
e.g. normal size is (26, 26) and pressed size is (40, 40). So the right end position of the thumb changes by its control state.

To resolve the above, the thumb position is calculated by using the maximum size of the thumb image all the time.
e.g. pressed size (40, 40) is used for the thumb position calculation.

To use the maximum size of the thumb image, ThumbMaximumSize is added to SliderStyle.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Slider.Internal.cs
+++ b/src/Tizen.NUI.Components/Controls/Slider.Internal.cs
@@ -71,6 +71,11 @@ namespace Tizen.NUI.Components
 
         // To store the thumb size of normal state
         private Size thumbSize = null;
+        // To store the maximum size of the thumb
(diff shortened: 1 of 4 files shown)


jaehyun0cho:
As @Seoyeon2Kim mentioned, this PR has problems.
Instead of this PR, the following PR is appropriate.
https://github.com/Samsung/TizenFX/pull/5028

PR #5027, opened 2023-02-21
[NUI] Resize slider normal thumb to calculate thumb position correctly

jaehyun0cho:
Previously, slider thumb size for normal and pressed were different. This caused that the thumb position was calculated incorrectly.

Now, slider thumb for normal is resized to be the same with the thumb for pressed.
By the above, the thumb position is calculated correctly.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM

Seoyeon2Kim (approved):
LGTM!
Thanks for the detailed explanation :)

PR #5038, opened 2023-02-23
[NUI] Fix Menu not to show Content if there is no menu item

jaehyun0cho:
Menu.Content is a container of menu items.

Previously, Menu.Content was displayed even if there is no menu item. Now, Menu.Content is displayed only if menu item exists.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Menu.cs
+++ b/src/Tizen.NUI.Components/Controls/Menu.cs
@@ -165,13 +165,22 @@ namespace Tizen.NUI.Components
 
             set
             {
+                if (Content == null)
+                {
+                    Content = CreateDefaultContent();
+                    if (styleApplied && (menuStyle != null))
+                    {
+                        Content.ApplyStyle(menuStyle.Content);
+                    }
+                }
+
                 if (menuItems != null)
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI.Components/Controls/Menu.cs
```diff
@@ -559,12 +568,7 @@ private View GetRootView()
         // If there is not enought space, then menu's size can be also resized.
         private void CalculateMenuPosition()
         {
-            if ((Anchor == null) || (Content == null))
-            {
-                return;
-            }
-
-            if (Items == null)
+            if (Anchor == null)
```
Seoyeon2Kim, line 571: To remove null check of `Content` and `Items` is because they are not used in this method, right?

Seoyeon2Kim (approved):
LGTM

src/Tizen.NUI.Components/Controls/Menu.cs
jaehyun0cho: Yes, you are right. Moreover, this PR enables that Content and Items are not set.

PR #5039, opened 2023-02-23
[NUI] Fix Menu not to show Content if there is no menu item

jaehyun0cho:
Menu.Content is a container of menu items.

Previously, Menu.Content was displayed even if there is no menu item. Now, Menu.Content is displayed only if menu item exists.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
LGTM~

PR #5040, opened 2023-02-26
[NUI.Scene3D][API10] Use Model defined camera

hinohie:
Make API that we can use gltf model file defined camera.

Since Tizen.NUI.Scene3D.Camera is one of View & This camera should be children of SceneView to render the scene, Camera cannot be owned by Model itself.

So, we support only for 'Generate' or 'Apply' camera parameters loaded from gltf model.


dongsug-song (approved):
approved

PR #5048, opened 2023-03-06
[NUI] Add Dispose to LongPressGestureDetector and modify ReleaseSwigCPtr

JoogabYun:
### Description of Change ###

[NUI] Add Dispose to LongPressGestureDetector and modify ReleaseSwigCPtr.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Events/LongPressGestureDetector.cs
+++ b/src/Tizen.NUI/src/public/Events/LongPressGestureDetector.cs
@@ -195,14 +195,45 @@ namespace Tizen.NUI
             return ret;
         }
 
+        /// <summary>
+        /// override it to clean-up your own resources.
+        /// </summary>
+        /// <param name="type"></param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected override void Dispose(DisposeTypes type)
+        {
+            if (disposed)
+            {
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Events/LongPressGestureDetector.cs
```diff
@@ -195,14 +195,45 @@ internal LongPressGestureDetectedSignal DetectedSignal()
...
+            //because the execution order of Finalizes is non-deterministic.
+
+            if (HasBody())
             {
-                DetectedSignal().Disconnect(detectedCallback);
+                if (detectedCallback != null)
+                {
+                    using LongPressGestureDetectedSignal signal = new LongPressGestureDetectedSignal(Interop.LongPressGestureDetector.DetectedSignal(GetBaseHandleCPtrHandleRef), false);
+                    signal?.Disconnect(detectedCallback);
+                    detectedCallback = null;
```
hinohie, line 227: Maybe not only for this patch but... Shouldn't we need to make `detectedEventHanlder` as `null` after call this?

dongsug-song (approved):
approved

hinohie (approved):
Approve but.. left one minor comments for future.

PR #5050, opened 2023-03-06
[NUI] Remove duplicate size setter

hinohie:
When if we set View.Size, It will call WidthPolicy / HeightPolicy internal. And then it will call SizeWidth and SizeHeight.

This patch ignore some duplicated size setting each other.

Test for 10,000 color view creation on Ubuntu

before : 1294.67 ms
after : 1159.00 ms


dongsug-song (approved):
approved

PR #5051, opened 2023-03-06
[NUI] Add getter property about WindowPositionSizeWithBorder and WindowSizeWithBorder

JoogabYun:
### Description of Change ###

[NUI] Add getter property about WindowPositionSizeWithBorder and WindowSizeWithBorder

In the case of a border window, the size including the border area is required.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window/BorderWindow.cs
+++ b/src/Tizen.NUI/src/public/Window/BorderWindow.cs
@@ -460,6 +460,44 @@ namespace Tizen.NUI
             return direction;
         }
 
+        /// <summary>
+        /// Gets position and size of the border window
+        /// </summary>
+        /// <returns>The total window size including the border area in the default window.</returns>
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Window/BorderWindow.cs
```diff
@@ -460,6 +460,43 @@ public BorderDirection GetDirection(float xPosition, float yPosition)
...
+            get
+            {
+                return GetRealWindowSize();
+            }
+        }
+
+        private Size2D GetRealWindowSize()
+        {
+            using var val = new Uint16Pair(Interop.Window.GetSize(SwigCPtr), true);
+            using Size2D size = new Size2D(val.GetWidth(), val.GetHeight());
```
hinohie, line 495: I think we should not make this `size` as 'using' Because we will return this value, and user will use this thing outsize of this functin.

dongsug-song (approved):
approved

PR #5054, opened 2023-03-06
[NUI] Do not connect default signals when View created

hinohie:
Since Window.ViewAdded signal required to all common view's creation time,
It will reduce each view's creation time.
And more, ViewAdded signal only works for default windows previously.

Now we make the signal ownership from window.

And additionaly, register custom HitTest signal only if subclass required.

TODO : The role of signal conflict with View.AddedToWindow.
Can we just remove this ViewAdded signal?

TODO : Can we just skip default view constructor's signal connection?


dongsug-song (approved):
approved

JoogabYun (approved):

PR #5055, opened 2023-03-06
[NUI] Remove duplicate URL settup

hinohie:
When we create ImageView with url, It will create image visual twice. To avoid this situation, make we replace internal url value for creation time.

And also, setup some missing values when we inputed desired size.

Test for load 10,000 ImageView with 50 kind of difference images in Ubuntu

Before : 2,764 ms
After : 1,659 ms


bshsqa (approved):
It looks work well.

PR #5057, opened 2023-03-07
[NUI] Add getter property about WindowPositionSizeWithBorder and Wind…

JoogabYun:
…owSizeWithBorder

In the case of a border window, the size including the border area is required.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
LGTM~

PR #5072, opened 2023-03-10
[NUI] Fix collectionView Selection on Item.

everLEEst:
### Description of Change ###

RecyclerViewItem need to update collectionview SelectedItem and SelectedItems list when it selected standalone.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/RecyclerView/Item/RecyclerViewItem.cs
+++ b/src/Tizen.NUI.Components/Controls/RecyclerView/Item/RecyclerViewItem.cs
@@ -50,6 +50,35 @@ namespace Tizen.NUI.Components
                     {
                         instance.UpdateState();
                     }
+                    if (instance.ParentItemsView is CollectionView colView)
+                    {
+                        var context = instance.BindingContext;
+                        if (colView.SelectionMode is ItemSelectionMode.Single ||
+                            colView.SelectionMode is ItemSelectionMode.SingleAlways)
+                        {
+                            if (newSelected && colView.SelectedItem != context)
+                            {
+                                colView.SelectedItem = context;
+                            }
+                            else if (!newSelected && colView.SelectedItem == context)
+                            {
+                                colView.SelectedItem = null;
+                            }
+                        }
+                        else if (colView.SelectionMode is ItemSelectionMode.Multiple)
+                        {
+                            var selectedList = colView.SelectedItems;
+                            bool contains = selectedList.Contains(context);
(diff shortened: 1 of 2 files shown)


dongsug-song (approved):
approved

src/Tizen.NUI.Components/Controls/RecyclerView/Item/RecyclerViewItem.cs
```diff
@@ -123,7 +152,11 @@ public RecyclerViewItem(RecyclerViewItemStyle itemStyle) : base(itemStyle)
         public bool IsSelectable
         {
             get => (bool)GetValue(IsSelectableProperty);
-            set => SetValue(IsSelectableProperty, value);
+            set
+            {
+                SetValue(IsSelectableProperty, value);
+                OnPropertyChanged("IsSelectable");
```
rabbitfor, line 158: How about this?
```
OnPropertyChanged(nameof(IsSelectable));
```
I think it'd be better to use `nameof` here for 2 reasons.

1. The intention of your code is to get property changed notice for the property named "IsSelectable". With the string value, it is not clear if there is a property named `IsSelectable` or not.
2. `nameof` is evaluated at compile time and has no effect at run time. Please check [here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/nameof).

src/Tizen.NUI.Components/Controls/RecyclerView/Item/RecyclerViewItem.cs
```diff
@@ -50,6 +50,35 @@ public partial class RecyclerViewItem : Control
                     {
                         instance.UpdateState();
                     }
+                    if (instance.ParentItemsView is CollectionView colView)
```
rabbitfor, line 53: How about replacing the abbreviation variable name with a more appropriate one ? :)
`colView` ->`collectionView`

src/Tizen.NUI.Components/Controls/RecyclerView/Item/RecyclerViewItem.cs
everLEEst: thank you. I will update this with another patch.

src/Tizen.NUI.Components/Controls/RecyclerView/Item/RecyclerViewItem.cs
everLEEst: thank you. I will update this with another patch.

PR #5074, opened 2023-03-10
[NUI] fix previous patch by coding rules

everLEEst:
### Description of Change ###

use nameof instead of string.
use full name of view as variable.

### API Changes ###

 - ACR:


rabbitfor (approved):
Thank you for the quick response for the review of #5072.
Looks good to me!

PR #5079, opened 2023-03-14
[NUI] Add internal property to avoid binding canceling on property set

everLEEst:
### Description of Change ###

assign a value directly into the property internally may have the potential issue of binding canceling.
 to prevent this, add internal property and use them insteadly.

see more detail about issue on below,
https://code.sec.samsung.net/confluence/display/GFX/NUI+Binding+Issue

This patch include below class property changes :

View.IsEnabled
Button.IsSelected
RecyclerViewItem.IsSelected

and all the usage of this properties.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

hinohie (approved):
Looks god to me

rabbitfor (approved):
Looks good to me!
It was nice to get a link to the document including detail descriptions about the issues related with the patch. :)

PR #5088, opened 2023-03-16
Add new cache APIs

akash1-kumar:
### Description of Change

Add new APIs for caching feature.

### API Changes
ACR: https://code.sec.samsung.net/jira/browse/TCSACR-531


src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
```diff
@@ -843,6 +844,213 @@ private void UnregisterProgressChangedEvent()
...
         }
+
+        /// <summary>
+        /// Sets the 'enabled' state of the cache feature.
+        /// </summary>
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <param name="enable">The enable value</param>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <remarks>
+        /// Enable/disable for _downloadId
```
HONGCHAEHEE, line 855: '_downloadId' is a private value. It's not appropriate to show in the API level.

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
```diff
@@ -843,6 +844,213 @@ private void UnregisterProgressChangedEvent()
...
+            }
+        }
+
+        /// <summary>
+        /// Returns the 'enabled' state of the cache feature.
+        /// </summary>
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <remarks>
+        /// Get Enable/disable for _downloadId
```
HONGCHAEHEE, line 876: '_downloadId' is a private value. It's not appropriate to show in the API level.

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
```diff
@@ -843,6 +844,213 @@ private void UnregisterProgressChangedEvent()
...
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <param name="size">The max size of the cache (MB)</param>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <remarks>The default size is 1000 MB.</remarks>
+        /// <exception cref="ArgumentException">Thrown when it is failed due to an invalid parameter.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when it is failed due to an invalid operation.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when a permission is denied.</exception>
+        /// <exception cref="NotSupportedException">Thrown when not supported.</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        internal void SetCacheMaxSize(uint size)
```
HONGCHAEHEE, line 923: Make type 'public' to be called from an user.
(This API is for internal use, only authorized users can use it.)

internal void SetCacheMaxSize(uint size) > public void SetCacheMaxSize(uint size)

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
```diff
@@ -843,6 +844,213 @@ private void UnregisterProgressChangedEvent()
...
+        /// Gets the max size(MB) of the cache and return it.
+        /// </summary>
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <exception cref="ArgumentException">Thrown when it is failed due to an invalid parameter.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when it is failed due to an invalid operation.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when a permission is denied.</exception>
+        /// <exception cref="NotSupportedException">Thrown when not supported.</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        internal uint GetCacheMaxSize()
```
HONGCHAEHEE, line 942: Make type 'public' to be called from an user.

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
```diff
@@ -843,6 +844,213 @@ private void UnregisterProgressChangedEvent()
...
+        /// Clears all cache.
+        /// </summary>
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <exception cref="ArgumentException">Thrown when it is failed due to an invalid parameter.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when it is failed due to an invalid operation.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when a permission is denied.</exception>
+        /// <exception cref="NotSupportedException">Thrown when not supported.</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        internal void ResetAllCache()
```
HONGCHAEHEE, line 963: Make type 'public' to be called from an user.

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
```diff
@@ -843,6 +844,213 @@ private void UnregisterProgressChangedEvent()
...
+        /// </summary>
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <param name="path">The path of the cache stored.</param>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <exception cref="ArgumentException">Thrown when it is failed due to an invalid parameter.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when it is failed due to an invalid operation.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when a permission is denied.</exception>
+        /// <exception cref="NotSupportedException">Thrown when not supported.</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        internal void SetCachePath(string path)
```
HONGCHAEHEE, line 983: Make type 'public' to be called from an user.

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
```diff
@@ -843,6 +844,213 @@ private void UnregisterProgressChangedEvent()
...
+        /// Returns the path of the cache stored.
+        /// </summary>
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <exception cref="ArgumentException">Thrown when it is failed due to an invalid parameter.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when it is failed due to an invalid operation.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when a permission is denied.</exception>
+        /// <exception cref="NotSupportedException">Thrown when not supported.</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        internal string GetCachePath()
```
HONGCHAEHEE, line 1002: Make type 'public' to be called from an user.

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
```diff
@@ -843,6 +844,213 @@ private void UnregisterProgressChangedEvent()
...
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <param name="time">The life cycle of the cache (seconds).</param>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <remarks>The default time is 48 * 3600 seconds.</remarks>
+        /// <exception cref="ArgumentException">Thrown when it is failed due to an invalid parameter.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when it is failed due to an invalid operation.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when a permission is denied.</exception>
+        /// <exception cref="NotSupportedException">Thrown when not supported.</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        internal void SetCacheLifeCycle(uint time)
```
HONGCHAEHEE, line 1025: Make type 'public' to be called from an user.

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
```diff
@@ -843,6 +844,213 @@ private void UnregisterProgressChangedEvent()
...
+        /// Returns the life cycle of the cache(in seconds).
+        /// </summary>
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <exception cref="ArgumentException">Thrown when it is failed due to an invalid parameter.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when it is failed due to an invalid operation.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when a permission is denied.</exception>
+        /// <exception cref="NotSupportedException">Thrown when not supported.</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        internal uint GetCacheLifeCycle()
```
HONGCHAEHEE, line 1044: Make type 'public' to be called from an user.

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
HONGCHAEHEE: For set/get APIs, how about using C# Property? Please refer to 'public string DestinationPath {get; set;}'

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
HONGCHAEHEE: For set/get APIs, how about using C# Property? Please refer to 'public string DestinationPath {get; set;}'

HONGCHAEHEE (changes requested):

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
akash1-kumar: > '_downloadId' is a private value. It's not appropriate to show in the API level.

Updated. Thanks.

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
akash1-kumar: Changed methods to C# Properties.

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
akash1-kumar: Changes methods to C# properties.

HONGCHAEHEE:
It would be better to add a class that includes methods that does not work for a request id.
For below APIs, could you implement them in a new class?

public void ResetCache()
public uint MaxCacheSize
public void ResetAllCache()
public string CachePath
public uint CacheLifeCycle

src/Tizen.Content.Download/Tizen.Content.Download/CacheManager.cs
```diff
@@ -0,0 +1,163 @@
...
+                DownloadErrorFactory.ThrowException(ret, "Failed to reset cache");
+            }
+        }
+
+        /// <summary>
+        /// Max size(MB) of the cache.
+        /// </summary>
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <remarks>The default size is 1000 MB.</remarks>
```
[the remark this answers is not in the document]
akash1-kumar: Okay. I will remove the comment.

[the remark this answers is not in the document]
akash1-kumar: Done.

src/Tizen.Content.Download/Tizen.Content.Download/CacheManager.cs
```diff
@@ -0,0 +1,162 @@
...
+                DownloadErrorFactory.ThrowException(ret, "Failed to reset cache");
+            }
+        }
+
+        /// <summary>
+        /// Max size(MB) of the cache.
+        /// </summary>
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <exception cref="ArgumentException">Thrown when it is failed due to an invalid parameter.</exception>
```
wiertel, line 52: standard meaning of ArgumentException is that there is something wrong with the value of method argument, this is a property. If that exception can actually be thrown, then let it stay, but the description could be changed to sth like say e.g. that setting to invalid value was rejected. Is there something like invalid value? I guess any value would work.

oh, it's not a public API, never mind then

src/Tizen.Content.Download/Tizen.Content.Download/CacheManager.cs
```diff
@@ -0,0 +1,162 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Clears all cache.
+        /// </summary>
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <exception cref="ArgumentException">Thrown when it is failed due to an invalid parameter.</exception>
```
wiertel, line 84: this method does have parameters

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
```diff
@@ -843,6 +844,37 @@ private void UnregisterProgressChangedEvent()
                 DownloadErrorFactory.ThrowException(ret, "Unsetting ProgressChanged callback failed");
             }
         }
+
+        /// <summary>
+        /// Enabled state of the cache feature.
+        /// </summary>
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <exception cref="ArgumentException">Thrown when it is failed due to an invalid parameter.</exception>
```
wiertel, line 853: see my comment for CacheManager.MaxCacheSize, it applies here too. But this one is public.

src/Tizen.Content.Download/Tizen.Content.Download/CacheManager.cs
```diff
@@ -0,0 +1,162 @@
...
+        /// Life cycle of the cache.
+        /// </summary>
+        /// <since_tizen> 7.5 </since_tizen>
+        /// <privilege>http://tizen.org/privilege/download</privilege>
+        /// <exception cref="ArgumentException">Thrown when it is failed due to an invalid parameter.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when it is failed due to an invalid operation.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when a permission is denied.</exception>
+        /// <exception cref="NotSupportedException">Thrown when not supported.</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        static public uint CacheLifeCycle
```
wiertel, line 140: please provide the meaning of those values in documentation, or even better introduce an enum instead of using int/uint

src/Tizen.Content.Download/Tizen.Content.Download/CacheManager.cs
```diff
@@ -0,0 +1,162 @@
...
+                int ret = Interop.Download.SetDownloadCacheMaxSize(value);
+                if (ret != (int)DownloadError.None)
+                {
+                    DownloadErrorFactory.ThrowException(ret, "Failed to set cache max size");
+                }
+            }
+        }
+
+        /// <summary>
+        /// Clears all cache.
```
wiertel, line 80: Please say more about ResetCache() and ResetAllCache() - do they apply to user cache, all system download cache, per app download cache?

wiertel (changes requested):

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
```diff
@@ -843,6 +844,37 @@ private void UnregisterProgressChangedEvent()
...
+        /// <exception cref="InvalidOperationException">Thrown when it is failed due to an invalid operation.</exception>
+        /// <exception cref="UnauthorizedAccessException">Thrown when a permission is denied.</exception>
+        /// <exception cref="NotSupportedException">Thrown when not supported.</exception>
+        public bool CacheEnable
```
wiertel, line 857: Please consider changing the name to "UsesCache", "IsCacheEnabled", or similar that is better for boolean property. EnableCache() would be good name for method, UsesCache is better for property, so the the code be like `request.UsesCache = true`. See note about booleans at  [Member naming convention](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-properties)

src/Tizen.Content.Download/Tizen.Content.Download/CacheManager.cs
akash1-kumar: Lifecycle value is in 'seconds' so it can have a range of values possible. So, declaring enum doesn't seems feasible to me.

Will remarks
 <summary>
'Lifecycle of the cache in seconds.
 </summary>

<remarks>
The time after which the cache will be invalid. New requests corresponding to the cache entry need to fetch the data from server and update the cache for future requests to be served from cache.
</remarks>

src/Tizen.Content.Download/Tizen.Content.Download/CacheManager.cs
```diff
@@ -0,0 +1,162 @@
...
+
+using System;
+using System.ComponentModel;
+
+namespace Tizen.Content.Download
+{
+    /// <summary>
+    /// The CacheManager class provides the functions to manage cache properties.
+    /// </summary>
+    /// <since_tizen> 7.5 </since_tizen>
```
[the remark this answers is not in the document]
akash1-kumar: Updated.

src/Tizen.Content.Download/Tizen.Content.Download/CacheManager.cs
akash1-kumar: Updated to "Thrown when it is failed due to setting an invalid value."

src/Tizen.Content.Download/Tizen.Content.Download/CacheManager.cs
akash1-kumar: Removed the line.

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
akash1-kumar: yes. updated.

src/Tizen.Content.Download/Tizen.Content.Download/CacheManager.cs
akash1-kumar: Updated to " Life cycle of the cache in seconds (Unix time). The time after which the stored cache will be invalid/stale"

src/Tizen.Content.Download/Tizen.Content.Download/CacheManager.cs
akash1-kumar: Updated.

src/Tizen.Content.Download/Tizen.Content.Download/Request.cs
akash1-kumar: Changed to UsesCache.

wiertel (approved):

HONGCHAEHEE (approved):

PR #5093, opened 2023-03-17
[NUI] Refactor ButtonExtension

rabbitfor:
### 1. Rename and change signatures of some methods in ButtonExtension. ###

* OnCreateIcon -> ProcessIcon
* OnCreateText -> ProcessText
* OnCreateOverlayImage -> ProcessOverlayImage
* OnCreateThumb -> ProcessThumb

These methods are designed for futher processing of the button's subcomponents.
To better match the design intent, the methods were better renamed: `OnCreateXXX` to `ProcessXXX`.

Also, the parameters and return type were not optimal form and needed refactoring.

```C#
// Before
// Previously, the method returned processed image which can be the given image or a new one.
// The caller doesn't know if the image has replaced to a new one or not.
ImageView OnCreateIcon(Button, ImageView);

// After
// The method returns the boolean result indicating the given reference of image is replaced or not.
bool ProcessIcon(Button, ref ImageView);
```

### 2. Extract Helper class to provide common methods used by multiple classes. ###

Most part of `LottieButtonExtension` and `LottieSwitchExtension` are duplicated, but it is impossible to take advantage of inheritance because C# does not allow them to have multiple base classes.
Previously, `LottieSwitchExtension` used `LottieButtonExtension`'s static methods to solve duplicate part.
For better structure, this patch extract the common part and make a new class: `LottieExtensionHelper`.

* Before

* After

image

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Button.Internal.cs
+++ b/src/Tizen.NUI.Components/Controls/Button.Internal.cs
@@ -34,7 +34,7 @@ namespace Tizen.NUI.Components
         private EventHandler<StateChangedEventArgs> stateChangeHandler;
 
         private bool isPressed = false;
-        internal bool styleApplying = false;
+        internal int styleApplying = 0;
 
         /// <summary>
         /// Gets accessibility name.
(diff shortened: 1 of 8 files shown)


taehyub (approved):
LGTM :)

jaehyun0cho (approved):
LGTM

PR #5101, opened 2023-03-20
[Tizen.Log] Replace dlog_print with dlog_print_dotnet

chanwoochoi:
### Description of Change ###
Replace dlog_print with dlog_print_dotnet which is used for only csharp api.

### API Changes ###
 - ACR: No API changed

### DIFF
--- a/src/Tizen.Log/Interop/Interop.Dlog.cs
+++ b/src/Tizen.Log/Interop/Interop.Dlog.cs
@@ -50,10 +50,10 @@ internal static partial class Interop
             DLOG_PRIO_MAX,
         }
 
-        [DllImport(Libraries.Dlog, EntryPoint = "dlog_print", CallingConvention = CallingConvention.Cdecl)]
+        [DllImport(Libraries.Dlog, EntryPoint = "dlog_print_dotnet", CallingConvention = CallingConvention.Cdecl)]
         internal static extern int Print(LogPriority prio, string tag, string fmt, string msg);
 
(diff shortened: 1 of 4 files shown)


myroot:
Could you share details of dlog_print_dotnet? or commits of dlog_print_dotnet

myroot (approved):

chanwoochoi:
> Could you share details of dlog_print_dotnet? or commits of dlog_print_dotnet

There are no operation difference of between dlog_print and dlog_print_dotnet. 
But, On the dlog side, there are sepaprate property to enable/disable the log as following:
- dlog_print()             : enable_apps=0 or 1 in /etc/dlog.conf (if enable_apps = 0, the log via dlog_print() is not stored.)
- dlog_print_dotent()  : enable_dotent_api=0 or 1 in /etc/dlog.conf (if enable_dotnet_api = 0, the log via dlog_print_dotnet() is not stored.)


Actually, dlog_print() has been used for application level log on multiple point such as csharp api, native c api and so on.
When enabling the 'enable_apps' property in /etc/dlog.conf, it should open the all application level log including csharp/c native and so on.

But, dlog_print_dotnet() will be only used in csharp api. In the dlog, there are no expressed symbol in the header file.
It try to hide the dlog_print_dotnet symbol from the other user of dlog because dlog_print_dotnet should be used in TizenFX.git.

Finally, VD will open the only csharp api log by enabling the 'enable_dotnet_api=1' and display other log level except for dlog_print_dotent log.

sung-su (approved):
LGTM

PR #5109, opened 2023-03-24
[NUI][AT-SPI] Add ViewAccessibilityMode

aswigon:
### Description of Change ###

The accessibility mode helps select the right Accessible implementation for a given view. It does not affect controls derived from CustomView (or any of its subclasses), e.g. Tizen.NUI.Components, which already behave in the custom mode. The default mode is kept for Tizen.NUI.BaseComponents. However, this patch add a possibility of opting into the custom mode for controls derived directly from View (e.g. FluxView), and by doing so, to be able to implement accessibility support for them in the C# language.

Dependencies:
- https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/290142/

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -116,10 +116,54 @@ namespace Tizen.NUI.BaseComponents
         }
 
         /// <summary>
+        /// Accessibility mode for controlling View's Accessible implementation.
+        /// It is only relevant when deriving custom controls from View directly,
+        /// as classes derived from CustomView (or any of its subclasses) get the
+        /// Custom mode by default.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public enum ViewAccessibilityMode
+        {
+            /// <summary>
+            /// Default accessibility implementation. Overriding View.Accessibility...()
+            /// virtual methods will have no effect.
+            /// </summary>
+            [EditorBrowsable(EditorBrowsableState.Never)]
+            Default,
+            /// <summary>
+            /// Custom accessibility implementation. Overriding View.Accessibility...()
+            /// will be necessary to provide accessibility support for the View.
+            /// </summary>
+            [EditorBrowsable(EditorBrowsableState.Never)]
+            Custom,
+        }
+
+        private static IntPtr NewWithAccessibilityMode(ViewAccessibilityMode accessibilityMode)
+        {
+            switch (accessibilityMode)
+            {
+                case ViewAccessibilityMode.Custom:
+                {
+                    return Interop.View.NewCustom();
+                }
+                case ViewAccessibilityMode.Default:
+                default:
+                {
+                    return Interop.View.New();
+                }
+            }
+        }
+
+        /// <summary>
         /// Creates a new instance of a view.
         /// </summary>
         /// <since_tizen> 3 </since_tizen>
-        public View() : this(Interop.View.New(), true)
+        public View() : this(ViewAccessibilityMode.Default)
+        {
+        }
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public View(ViewAccessibilityMode accessibilityMode) : this(NewWithAccessibilityMode(accessibilityMode), true)
         {
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
         }
--- a/src/Tizen.NUI/src/internal/Interop/Interop.View.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.View.cs
@@ -26,6 +26,9 @@ namespace Tizen.NUI
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_View_New")]
             public static extern global::System.IntPtr New();
 
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_View_NewCustom")]
+            public static extern global::System.IntPtr NewCustom();
+
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_new_View__SWIG_0")]
             public static extern global::System.IntPtr NewView();
 


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -115,11 +115,52 @@ static View()
...
+        private static IntPtr NewWithAccessibilityMode(ViewAccessibilityMode accessibilityMode)
+        {
+            switch (accessibilityMode)
+            {
+            default:
+            case ViewAccessibilityMode.Default:
+                return Interop.View.New();
+
+            case ViewAccessibilityMode.Custom:
+                return Interop.View.NewCustom();
```
hinohie, line 150:
```suggestion
            case ViewAccessibilityMode.Custom:
            {
                return Interop.View.NewCustom();
            }
            case ViewAccessibilityMode.Default:
            default:
            {
                return Interop.View.New();
            }
```

Wrap by brace
+
Places 'default' as end of cases (It is my personal taste)

src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -115,11 +115,52 @@ static View()
...
+            }
+        }
+
         /// <summary>
         /// Creates a new instance of a view.
         /// </summary>
         /// <since_tizen> 3 </since_tizen>
-        public View() : this(Interop.View.New(), true)
```
dongsug-song, line 122: Hello, please change as like below.

        /// <summary>
        /// Creates a new instance of a view.
        /// </summary>
#if PROFILE_TV
        public View() : this(Interop.View.New(), true)
#else
        public View() : this(ViewAccessibilityMode.Custom)
#endif     
        {
        }
=> VD will not use ATSPI and the performance and memory consumption are more important, so that this should **NOT** be applied to VD.

and these patches (tizenfx and csharp-binder) shoud be submitted to tizen_7.0 (API10) also.

src/Tizen.NUI/src/public/BaseComponents/View.cs
aswigon: Reordering done & braces added. Also indentation added so that the braces are all at different level.s

src/Tizen.NUI/src/public/BaseComponents/View.cs
aswigon: In the context of non-TV profiles, I disagree with changing the default mode to `Custom` (as you suggested: `public View() : this(ViewAccessibilityMode.Custom)`). This is thought as an opt-in functionality. Note that other `View` constructors as well as all `BaseComponents`'s constructors behave as if the `ViewAccessibilityMode.Default` mode was specified.

`BaseComponents`'s accessibility is implemented in C++ (so it is the `Default` mode) and it is intentional that there is no way to change that. It would be confusing, and against object-oriented programming principles, if the `View` constructor added some functionality (`Custom` mode) that would simply be lost in its subclasses.

And since the default mode has to be `Default`, as explained above, there is absolutely no point in splitting the code into basically two identical paths:
```c#
#if PROFILE_TV
// create with Default mode (Interop.View.New())
#else
// create with Default mode (Interop.View.New())
#endif
```

dongsug-song: OK, I see, then could you check the below?

- except VD, ATSPI need to be enabled as a default. because regardless of screen-reader on/off, "custom" is required by the constructor to use atspi, so it cannot be changed after instance is created.
- users can only use public constructors to create derived classes. So adding arguments to the public constructors.


#if PROFILE_TV
        /// <summary>
        /// Creates a new instance of a view.
        /// </summary>
        public View() : this(Interop.View.New(), true)
        {
        }

        /// This will be public opened in next release of tizen after ACR done. Before ACR, it is used as HiddenAPI (InhouseAPI).
        public View(ViewStyle viewStyle) : this(Interop.View.New(), true, viewStyle)
        {
        }

        /// <summary>
        /// Create a new instance of a View with setting the status of shown or hidden.
        /// </summary>
        /// <param name="shown">false : Not displayed (hidden), true : displayed (shown)</param>
        /// This will be public opened in next release of tizen after ACR done. Before ACR, it is used as HiddenAPI (InhouseAPI).
        public View(bool shown) : this(Interop.View.New(), true)
        {
            SetVisible(shown);
        }
#else
        /// <summary>
        /// ViewAccessibilityMode
        /// </summary>
        public enum ViewAccessibilityMode
        {
            /// <summary>
            /// Use Toolkit::Internal::Control as the View's backend (comes with DevelControl::ControlAccessible)
            /// </summary>
            Default,
            /// <summary>
            /// Use SlimCustomViewImpl as the View's backend (comes with NUIViewAccessible)
            /// </summary>
            Custom,
        }

        private static IntPtr NewWithAccessibilityMode(ViewAccessibilityMode accessibilityMode)
        {
            switch (accessibilityMode)
            {
                case ViewAccessibilityMode.Custom:
                    return Interop.View.NewCustom();

                default:
                case ViewAccessibilityMode.Default:
                    return Interop.View.New();
            }
        }

        /// <summary>
        /// Creates a new instance of a view.
        /// </summary>
        /// <param name="accessibilityMode">Default is Custom and the Custom enables ATSPI for derived class of View</param>
        public View(ViewAccessibilityMode accessibilityMode = ViewAccessibilityMode.Custom) : this(NewWithAccessibilityMode(accessibilityMode), true)
        {
        }

        /// This will be public opened in next release of tizen after ACR done. Before ACR, it is used as HiddenAPI (InhouseAPI).
        public View(ViewStyle viewStyle, ViewAccessibilityMode accessibilityMode = ViewAccessibilityMode.Custom) : this(NewWithAccessibilityMode(accessibilityMode), true, viewStyle)
        {
        }

        /// <summary>
        ///  Create a new instance of a View with setting the status of shown or hidden.
        /// </summary>
        /// <param name="shown">false : Not displayed (hidden), true : displayed (shown) </param>
        /// <param name="accessibilityMode">Default is Custom and the Custom enables ATSPI for derived class of View</param>
        /// This will be public opened in next release of tizen after ACR done. Before ACR, it is used as HiddenAPI (InhouseAPI).
        public View(bool shown, ViewAccessibilityMode accessibilityMode = ViewAccessibilityMode.Custom) : this(NewWithAccessibilityMode(accessibilityMode), true)
        {
            SetVisible(shown);
        }
#endif

aswigon: > except VD, ATSPI need to be enabled as a default. because regardless of screen-reader on/off, "custom" is required by the constructor to use atspi, so it cannot be changed after instance is created.

⇒ **Incorrect**. Both `Default` and `Custom` modes **provide** AT-SPI support, the difference is what programming language is used to **customize** AT-SPI behaviour. For `Tizen.NUI.BaseComponents` the mode is `Default`, hence the language is C++ (creating a subclass of `Dali::Toolkit::DevelControl::ControlAccessible` is needed). For `Components` the mode is equivalent to `Custom` (internally another mechanism is used, called `ViewWrapperImpl`), hence the language is C# (overriding `View`'s virtual methods is needed).

The purpose of this PR is to enable some other controls (notably, `FluxView`) to use C# rather than C++ to **customize** AT-SPI. VD is unaffected by this PR, they are already using AT-SPI-enabled DALi and NUI code anyway, they simply do not use the functionality.

aswigon: So, in other words, `ViewAccessibilityMode.Default` really needs to remain the default mode for `View` (its behaviour without this PR is identical to its behaviour with this PR and mode `Default`). `BaseComponents` also **need** the `Default` mode for proper AT-SPI support (it is already implemented in C++ for `BaseComponents`).

aswigon:
> ok. I see. but please apply this (I think this is acceptable for you) => #if PROFILE_TV public View() : this(Interop.View.New(), true) #else public View() : this(ViewAccessibilityMode.Default) #endif

But what's the point? `this(ViewAccessibilityMode.Default)` and `this(Interop.View.New(), true)` are 100% equivalent, and using the preprocessor only makes the code uglier for no reason at all...

Are you worried that the C# compiler will not reduce the call to `NewWithAccessibilityMode(ViewAccessibilityMode.Default)` to simply `Interop.View.New()`? I can check the disassembled CLI code... but remember that there are also JIT / AOT optimizations applied and the mode is a constant expression in this context!

dongsug-song (approved):
approved

dongsug-song:
please submit the patch in tizen_7.0 (API10) also

aswigon:
> please submit the patch in tizen_7.0 (API10) also

You're welcome: https://github.com/Samsung/TizenFX/pull/5118

PR #5111, opened 2023-03-27
[NUI] Add interceptTouchEvent at WindowEvent

JoogabYun:
### Description of Change ###

[NUI] Add interceptTouchEvent at WindowEvent

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window/WindowEvent.cs
+++ b/src/Tizen.NUI/src/public/Window/WindowEvent.cs
@@ -32,6 +32,7 @@ namespace Tizen.NUI
     {
         private WindowFocusChangedEventCallbackType windowFocusChangedEventCallback;
         private RootLayerTouchDataCallbackType rootLayerTouchDataCallback;
+        private RootLayerTouchDataCallbackType rootLayerInterceptTouchDataCallback;
         private WheelEventCallbackType wheelEventCallback;
         private EventCallbackDelegateType1 stageKeyCallbackDelegate;
         private InterceptKeyEventDelegateType stageInterceptKeyCallbackDelegate;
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/mono_crash.2207df118e.0.json
rabbitfor, line 1: Hello! It seems that this file was uploaded by mistake.

JoogabYun: thank you :) I deleted the wrong file

src/Tizen.NUI/src/public/Window/WindowEvent.cs
```diff
@@ -135,6 +136,40 @@ public partial class Window
...
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public event ReturnTypeEventHandler<object, TouchEventArgs, bool> InterceptTouchEvent
+        {
+            add
+            {
+                if (rootLayerInterceptTouchDataEventHandler == null)
+                {
+                    rootLayerInterceptTouchDataCallback = OnWindowInterceptTouch;
+                    using TouchDataSignal signal = new TouchDataSignal(Interop.ActorSignal.ActorInterceptTouchSignal(Layer.getCPtr(GetRootLayer())), false);
+                    signal?.Connect(rootLayerInterceptTouchDataCallback);
```
rabbitfor, line 154: I'm not familiar with this part, so I'm just curious about this line. Is there any possibilty that the signal is null?
In my opinion, the signal is just created so that it seems not possible to be null.

JoogabYun: This is the same as any other event handling logic. 
The signal null check is known as Mr dongsug-song's intention.
Isn't this code in case the native object is destroyed?

rabbitfor: Thank you for the explanation. :+1:

rabbitfor (approved):

PR #5113, opened 2023-03-27
[Connection] Fix race condition

chleun-moon:
### Description of Change ###
A race condition occurred when two threads subscribed ConnectionManager.ConnectionTypeChanged.
Only the first event subscription register the C# side callback to the native capi.
So the subscription should be protected by lock.


HONGCHAEHEE (approved):
It looks good.

SeonahMoon (approved):

PR #5116, opened 2023-03-28
[NUI] Register HitTest callback only if GrabTouchAfterLeave = true

hinohie:
Since useless HitTest checkup function make performance down for standard cases, Let we make we use custom callback only if GrabTouchAfterLeave case.


taehyub (approved):
LGTM 👍

PR #5118, opened 2023-03-28
[NUI][AT-SPI] Add ViewAccessibilityMode

aswigon:
### Description of Change ###

The accessibility mode helps select the right Accessible implementation for a given view. It does not affect controls derived from `CustomView` (or any of its subclasses), e.g. `Tizen.NUI.Components`, which already behave in the custom mode. The default mode is kept for `Tizen.NUI.BaseComponents`. However, this patch add a possibility of opting into the custom mode for controls derived directly from `View` (e.g. `FluxView`), and by doing so, to be able to implement accessibility support for them in the C# language.

(Note: This is a backport of https://github.com/Samsung/TizenFX/pull/5109)

Dependencies:
- https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/290143/ (merged)

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -116,10 +116,54 @@ namespace Tizen.NUI.BaseComponents
         }
 
         /// <summary>
+        /// Accessibility mode for controlling View's Accessible implementation.
+        /// It is only relevant when deriving custom controls from View directly,
+        /// as classes derived from CustomView (or any of its subclasses) get the
+        /// Custom mode by default.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public enum ViewAccessibilityMode
+        {
+            /// <summary>
+            /// Default accessibility implementation. Overriding View.Accessibility...()
+            /// virtual methods will have no effect.
+            /// </summary>
+            [EditorBrowsable(EditorBrowsableState.Never)]
+            Default,
+            /// <summary>
+            /// Custom accessibility implementation. Overriding View.Accessibility...()
+            /// will be necessary to provide accessibility support for the View.
+            /// </summary>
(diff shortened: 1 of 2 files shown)


Seoyeon2Kim (approved):
LGTM! :)

src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -115,11 +115,55 @@ static View()
...
         }
 
+        /// <summary>
+        /// Accessibility mode for controlling View's Accessible implementation.
+        /// It is only relevant when deriving custom controls from View directly,
+        /// as classes derived from CustomView (or any of its subclasses) get the
+        /// Custom mode by default.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public enum ViewAccessibilityMode
```
rabbitfor, line 125: How about moving this enum declaration out of class?
In my opinion, `View.ViewAccessibilityMode` feels like a duplicate a lot, for example,
```
/// Current
var view = new View(View.ViewAccessibilityMode.Custom)
```
If you move it out of class, it can be simple.
```
/// Suggestion1
var view = new View(ViewAccessibilityMode.Custom)
```

Furthermore, if you rename it to `AccessibilityMode` the code will be much simpler.
```
/// Suggestion2
var view = new View(AccessibilityMode.Custom)
```

Could you consider these?
I understand this patch is just for back-porting, so I'd like you to create a new patch if you want to modify it.

rabbitfor: Thank you for the explaination.
As far as I understand, it is for `FluxView`. If that's the only case, I think it's okay to leave it as it is now.

Seoyeon2Kim:
The backport of this PR was already merged in DevelNUI!
So, I'll merge it now. Thanks for your work.

PR #5123, opened 2023-03-28
[NUI] refactoring collectionView and Layouters for extended class

everLEEst:
to create customized collectionView and Layouters, some properties must be extendable.
this change includes properties access modifier
which is internal or private to protected or public.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/RecyclerView/Layouter/GridLayouter.cs
+++ b/src/Tizen.NUI.Components/Controls/RecyclerView/Layouter/GridLayouter.cs
@@ -46,6 +46,35 @@ namespace Tizen.NUI.Components
         private Timer requestLayoutTimer = null;
         private bool isSourceEmpty;
 
+        /// <inheritdoc/>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected new IGroupableItemSource Source
+        {
+            get
+            {
+                if (colView) return colView.InternalSource;
+                else return null;
+            }
+        }
+
+        /// <summary>
+        /// Span Size
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected int SpanSize
+        {
+            get => spanSize;
+        }
+
+        /// <summary>
+        /// Size Candidate
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected (float Width, float Height) SizeCandidate
+        {
+            get => sizeCandidate;
+        }
+
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Tizen.NUI.Components/Controls/RecyclerView/Layouter/GridLayouter.cs
```diff
@@ -46,6 +46,35 @@ public class GridLayouter : ItemsLayouter
         private Timer requestLayoutTimer = null;
         private bool isSourceEmpty;
 
+        /// <inheritdoc/>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected new IGroupableItemSource Source
+        {
+            get
+            {
+                if (colView) return colView.InternalSource;
```
rabbitfor, line 55: Unless the intention of `if (colView)` is to check if the BaseHandle's cptr exists, I think it would be better to fix it like this.
```suggestion
                if (colView != null) return colView.InternalSource;
```
Or just replace the whole getter code to
```
                return colView?.InternalSource;
```

FYI, the code `if (colView)` will call `BaseHandle::operator true`. Please refer here:https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/public/Common/BaseHandle.cs#L141

everLEEst: Thank you. will fix this.

src/Tizen.NUI.Components/Controls/RecyclerView/Layouter/GridLayouter.cs
```diff
@@ -46,6 +46,29 @@ public class GridLayouter : ItemsLayouter
...
+        /// Size Candidate
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected (float Width, float Height) SizeCandidate => sizeCandidate;
+
+        /// <summary>
+        /// Visible ViewItem.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected new List<GroupInfo> GroupItems => groups;
```
rabbitfor, line 69: How about declaring `ItemLayouter.GroupItems` as a virtual property? And then `GridLayouter.GroupItems` can `override` it.

everLEEst: sure. it makes sense. thank you.

rabbitfor (approved):

PR #5124, opened 2023-03-28
[NUI] Add view helper API to rotate : LookAt

hinohie:
Add helper API to rotate View to lookAt specific position.
It will be helpful if user have some problem to calculate
Orientation property

Note : This API only for change View.Orientation. Not for constraint.

This PR depend on below dali patches :
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-core/+/290532
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/290533


Seoyeon2Kim (approved):
LGTM.
If there's a dependent patch of dali-csharp-binder, please write down the patch url together.

PR #5132, opened 2023-03-29
[NUI] Fix signal bugs in View and Window.

rabbitfor:
* Do not try to disconnect callback unless the callback is not null.
* Ensure a signal has valid cptr before calling Connect/Disconnect.

### API Changes ###

 - ACR:


dongsug-song (approved):
very good!
approved

hinohie (approved):
Wow! codes looks much better

PR #5133, opened 2023-03-29
[NUI] Fix size calculation for TextLabel with Margin in RelativeLayout

jaehyun0cho:
Previously, RelativeLayout calculated the size for TextLabel with Margin incorrectly by calling MeasureChildWithMargins().
Because TextLabel size was decreased by its Margin size incorrectly.

Now, RelativeLayout calculates the size for TextLabel with Margin correctly by calling MeasureChild().
Because TextLabel size is not decreased by its Margin size.

### API Changes ###

 - ACR:


hinohie (approved):
Seems good points

PR #5137, opened 2023-03-31
[NUI] Remove regacy Interop files

rabbitfor:
* Button
* CheckBoxButton
* Popup
* ProgressBar
* PushButton
* RadioButton
* Slider
* ToggleButton
* Signals that are not removed in #5136

Note that this patch is a set with: https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/290691/

### API Changes ###

 - ACR:


dongsug-song (approved):
very good!!!
approved

PR #5147, opened 2023-04-03
[NUI.Samples] Add sample for spring like animation

hinohie:
Simple sample with custom alpha function

Animation3


dongsug-song (approved):
great!!!

PR #5151, opened 2023-04-04
[NUI][AT-SPI] Promote DoAction implementation from CustomView to View

aswigon:
### Description of Change ###

This is a follow-up to #5109 that enables controls derived from View, but not CustomView, that use the ViewAccessibilityMode.Custom (e.g. FluxView), to customize their behaviour related to the AT-SPI Action interface.

The implementation of AccessibilityDoAction() and its helper methods is moved up from CustomView to View. Note that the events that AccessibilityDoAction invokes have already been declared at the View level, in ViewAccessibilityEvent.cs (e.g. AccessibilityActivated), so they do not have to be moved anywhere in this PR.

### API Changes ###

 - ACR: None


loleksak (approved):
LGTM

PR #5155, opened 2023-04-05
[NUI] Add Default Image & transition effect

tscholb:
### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/BaseComponents/ImageView.cs
```diff
@@ -1154,6 +1154,44 @@ public bool AdjustViewSize
...
         private bool adjustViewSize = false;
 
+        /// <summary>
+        /// ImageView PlaceHolderURL, type string.
+        /// This is one of mandatory property. Even if not set or null set, it sets empty string ("") internally.
+        /// When it is set as null, it gives empty string ("") to be read.
+        /// </summary>
+        /// <since_tizen> 11 </since_tizen>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public string PlaceHolderURL
```
hinohie, line 1164:
```suggestion
        public string PlaceHolderUrl
```

How about to use `Url` instead of `URL`? Since, we are already using `ResourceUrl` for ImageView.

hinohie (approved):
Wait until dali-side codes merged but... anyway looks good to me for now

PR #5162, opened 2023-04-06
[Tizen.Content.Download] Update protection level for CacheManager class

akash1-kumar:
Updates protection level to 'public' for CacheManager class. So that it can be used by applications.


HONGCHAEHEE (approved):
It looks good.

SeonahMoon (approved):

PR #5165, opened 2023-04-10
[NUI] Collect Renderer Property enum as public hidden

hinohie:
Since the enum type was internally hidden, User don't know what value should we use for some Renderer's property.

Until do Modify ACR, make them as hidden


src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
```diff
@@ -0,0 +1,465 @@
+﻿// Copyright (c) 2019 Samsung Electronics Co., Ltd.
```
Seoyeon2Kim, line 1: 2023

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
```diff
@@ -0,0 +1,465 @@
...
+using System.ComponentModel;
+
+namespace Tizen.NUI
+{
+    /// <summary>
+    /// Enumeration for face culling mode.
+    /// </summary>
+    /// This will be public opened in next tizen after ACR done. Before ACR, need to be hidden as inhouse API.
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public enum FaceCullingModeType
```
Seoyeon2Kim, line 24: Just my curiosity :
Is it okay to open these all enumerations as public? Because they are originally internal values in NUI.
If they're public APIs in DALi, I think it's available. :)

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
```diff
@@ -0,0 +1,465 @@
...
+    /// <summary>
+    /// Enumeration for blend equation.
+    /// </summary>
+    /// This will be public opened in next tizen after ACR done. Before ACR, need to be hidden as inhouse API.
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public enum BlendEquationType
+    {
+        /// <summary>
+        /// The source and destination colors are added to each other.
+        /// </summary>
```
Seoyeon2Kim, line 89: Please add 'EditorBrowsable' tag here, too.

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
```diff
@@ -0,0 +1,465 @@
...
+        Add = 0x8006,
+
+        /// <summary>
+        /// Subtracts the destination from the source.
+        /// </summary>
```
Seoyeon2Kim, line 94: ditto

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
```diff
@@ -0,0 +1,465 @@
...
+        Subtract = 0x800A,
+
+        /// <summary>
+        /// Subtracts the source from the destination.
+        /// </summary>
```
Seoyeon2Kim, line 99: ditto

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
```diff
@@ -0,0 +1,465 @@
...
+        ReverseSubtract = 0x800B,
+
+        //OpenGL es 3.0 enumeration
+        /// This will be public opened in next tizen after ACR done. Before ACR, need to be hidden as inhouse API.
```
Seoyeon2Kim, line 103: Probably you don't add descriptions for `<summary>` yet below?

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
```diff
@@ -0,0 +1,465 @@
...
+    /// <summary>
+    /// Enumeration for blend factor.
+    /// </summary>
+    /// This will be public opened in next tizen after ACR done. Before ACR, need to be hidden as inhouse API.
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public enum BlendFactorType
+    {
+        /// <summary>
+        /// Match as GL_ZERO
+        /// </summary>
```
Seoyeon2Kim, line 167: All enum values have to get `[EditorBrowsable(EditorBrowsableState.Never)]` tag or ACR `version`.
It's WIP PR now, so it might not be added  yet.?
Same as below.

Seoyeon2Kim (commented):
When you're ready to get reviews, please remove '[WIP] tag in the tile

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
hinohie: Oh...! what a nice point! Fix done.

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
hinohie: All of enum values (instead of some `BlendEquationType` cases) are public. We can make them as public.
Actually, they don't have any reason to be internal :(....

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
hinohie: This can be public open. In other words, The enums what I write `EditorBrowsable` in this scopes, are not ready to be public.

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
hinohie: ditto

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
hinohie: ditto

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
hinohie: So sad story, Below values dont have any descriptions even DALi side. they are also kind of hidden properties at DALi. So just skip it.

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
hinohie: I don't know detail but... Should we add that tags for each enum? For example, `NUIConstants.cs`'s `ResizePolicyType` mark since_tizen only at `enum` define, and not add tags for each values.

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
Seoyeon2Kim: I see. All the enum values are in internal folders, but they can be public and users are able to use the enum directly.

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
Seoyeon2Kim: AFAIK, all values should have each `since_tizen` version tag, but I found that some values have, but some values don't, as you said... :(

For example, the document for `LayoutParamPolicies` is like this :
https://docs.tizen.org/application/dotnet/api/TizenFX/API10/api/Tizen.NUI.BaseComponents.LayoutParamPolicies.html

On the other hands, the document for `States`, which does not write down its version, is like this :
https://docs.tizen.org/application/dotnet/api/TizenFX/API10/api/Tizen.NUI.BaseComponents.View.States.html

`LayoutParamPolicies` and `States` enums are all for `View` and they're included in `ViewEnum.cs`.
So, I think we need to get other NUI members' opinions for our unity.

Seoyeon2Kim (commented):
Some comments are added again. :)

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
```diff
@@ -0,0 +1,465 @@
...
+    /// Enumeration for face culling mode.
+    /// </summary>
+    /// This will be public opened in next tizen after ACR done. Before ACR, need to be hidden as inhouse API.
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public enum FaceCullingModeType
+    {
+        /// <summary>
+        /// None of the faces should be culled
+        /// </summary>
+        NONE = 0,
```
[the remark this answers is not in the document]
hinohie: Good point! It is the best timming to change naming format of each properties. I'll push relative commit on this PR.

src/Tizen.NUI/src/public/Rendering/RendererConstants.cs
hinohie: https://github.com/Samsung/TizenFX/pull/5165/commits/1a5ba365f3ce94601dd44c99868e94cbb6965104

Thank you for your comment. After some offline discussion, I decide to write `[EditorBrowsable(EditorBrowsableState.Never)]` tag for all properties.

Seoyeon2Kim: Thank you for adding it all. :)
I'll check it again.

Seoyeon2Kim (approved):
LGTM !:)

PR #5166, opened 2023-04-10
[NUI] Rotation class with EulerAngle (pitch/yaw/roll)

hinohie:
Add API that we can set Rotation by pitch/yaw/roll + Get Rotation informations as pitch/yaw/roll

Rotation apply order is x, y, and z. (pitch->yaw->roll)

Relative dali patch :
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/291144

### DIFF
--- a/test/Tizen.NUI.Devel.Tests.Ubuntu/Tizen.NUI.Devel.Tests/testcase/public/Common/TSRotation.cs
+++ b/test/Tizen.NUI.Devel.Tests.Ubuntu/Tizen.NUI.Devel.Tests/testcase/public/Common/TSRotation.cs
@@ -1,0 +1,115 @@
+﻿using global::System;
+using NUnit.Framework;
+using NUnit.Framework.TUnit;
+using Tizen.NUI;
+
+namespace Tizen.NUI.Devel.Tests
+{
+    using tlog = Tizen.Log;
+
+    [TestFixture]
+    [Description("public/Common/Rotation")]
+    public class RotationTest
+    {
+        private const string tag = "NUITEST";
+
+        [SetUp]
+        public void Init()
+        {
+            tlog.Info(tag, "Init() is called!");
+        }
+
+        [TearDown]
+        public void Destroy()
+        {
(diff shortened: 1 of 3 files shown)


dongsug-song (approved):
approved

test/Tizen.NUI.Devel.Tests.Ubuntu/Tizen.NUI.Devel.Tests/testcase/public/Common/TSRotation.cs
```diff
@@ -0,0 +1,115 @@
...
+        public void RotationGetEulerAngle()
+        {
+            tlog.Debug(tag, $"RotationGetEulerAngle START");
+
+            var expectPitchRadian = new Radian(new Degree(30.0f));
+            var expectYawRadian = new Radian(new Degree(45.0f));
+            var expectRollRadian = new Radian(new Degree(60.0f));
+            var testingTarget = new Rotation(expectPitchRadian, expectYawRadian, expectRollRadian);
+            Assert.IsNotNull(testingTarget, "Can't create success object Rotation.");
+            Assert.IsInstanceOf<Rotation>(testingTarget, "Should return Rotation instance.");
```
myroot, line 92: May I request test with negative angle?
for example

``` c#
            var expectPitchRadian = new Radian(new Degree(30.0f));
            var expectYawRadian = new Radian(new Degree(-45.0f));
            var expectRollRadian = new Radian(new Degree(60.0f));
            var testingTarget = new Rotation(expectPitchRadian, expectYawRadian, expectRollRadian);
            Assert.IsNotNull(testingTarget, "Can't create success object Rotation.");
            Assert.IsInstanceOf<Rotation>(testingTarget, "Should return Rotation instance.");
```

hinohie: Negative angle also works well. The TC what you suggested also passed.

PR #5172, opened 2023-04-12
[Camera] Add new camera setting APIs

hsgwon:
### Description of Change ###

Add new camera setting APIs like the followings.
- Saturation
- Sharpness
- Gain
- WhiteBalanceTemperature

### API Changes ###

 - TCSACR-538


src/Tizen.Multimedia.Camera/Camera/CameraSettings.cs
```diff
@@ -375,6 +383,214 @@ public Range ExposureRange
...
+            }
+            set
+            {
+                Native.SetSaturation(_camera.GetHandle(), value).
+                    ThrowIfFailed("Failed to set camera saturation value.");
+            }
+        }
+
+        /// <summary>
+        /// Gets the available saturation value.
```
wiertel, line 409: that follows the pattern of already existing properties, but "Gets the available saturation value range." or "Gets the available saturation range." or "Gets the range of allowed saturation value."  would be better.

src/Tizen.Multimedia.Camera/Camera/CameraSettings.cs
```diff
@@ -375,6 +383,214 @@ public Range ExposureRange
...
+        /// </summary>
+        /// <since_tizen> 11 </since_tizen>
+        /// <remarks>
+        /// If the minimum value is greater than the maximum value, it means this feature is not supported.
+        /// </remarks>
+        /// <exception cref="NotSupportedException">Saturation is not supported.</exception>
```
wiertel, line 415: lines 413 and 415 suggest 2 different outcomes when saturation is not supported. Which is it then - an empty range or exception?

wiertel (approved):

src/Tizen.Multimedia.Camera/Camera/CameraSettings.cs
hsgwon: I'll fix it.

src/Tizen.Multimedia.Camera/Camera/CameraSettings.cs
hsgwon: As described in remarks, If min value is greater than the max value, it means XXRange is not supported and in this case, NotSupportedException will be thrown. 
Those values are tested by `GetRange()` method in 64 line.

wiertel: The thing is that it's not clear to me what will happen when the feature is not supported - It cannot be both an exception and some values.  Will app developer get an exception or an empty Range object?

PR #5190, opened 2023-04-19
[NUI] Remove duplicate ?. operator and use using statement

rabbitfor:
### API Changes ###

 - ACR:


hinohie (approved):
Good point. I think we can do same thing and reduce code complexity of similar codes in future

PR #5200, opened 2023-04-20
[NUI][API10] Fix SVACE issues.

huayongxu:
### Description of Change ###

 [STATISTICAL] Value InternalSource, which is result of method invocation with possible null return value, is dereferenced in member access expression InternalSource.Count

 [STATISTICAL] Value AppBar, which is result of method invocation with possible null return value, is dereferenced in method call AppBar.PassFocusableViewInsideIfNeeded()

[STATISTICAL] Value temp, which is result of method invocation with possible null return value, is dereferenced in member access expression temp.Width

### API Changes ###

 - ACR:


xuelian-bai (approved):
looks good to me

dongsug-song (approved):
approved

PR #5203, opened 2023-04-20
Implement libsessiond wrapper

adaszkiewicza:
## Description of Change ##
Implemented a wrapper for sessiond.  Changes placed into `src/Tizen.System.Session`.

### ACR ###:
Not applicable
#### static ####
- `int Session.CurrentUID` - Field
- `int Session.MaxUserLength` - Field
- `string Session.EmptyUser` - Field
- `Session Session.GetInstance(int sessionUID)`
#### requiring an object (Session sessionObj) ####
- `int sessionObj.SessionUID` - Property
- `IReadOnlyList<string> sessionObj.GetUsers()`
- `string sessionObj.GetCurrentUser()`
- `Task sessionObj.SubsessionAddUserAsync(string userName)`
- `Task sessionObj.SubsessionRemoveUserAsync(string userName)`
- `Task sessionObj.SubsessionSwitchUserAsync(string userName)`
- `void sessionObj.SubsessionEventMarkAsDone(SubsessionEventArgs subsessionEventArgs)`
- `event sessionObj.EventHandler<AddUserEventArgs> AddUserWait`
- `event sessionObj.EventHandler<RemoveUserEventArgs> RemoveUserWait`
- `event sessionObj.EventHandler<SwitchUserWaitEventArgs> SwitchUserWait`
- `event sessionObj.EventHandler<SwitchUserCompletionEventArgs> SwitchUserCompletion`

### Changed ###
None

### Removed ###
None

### DIFF
--- a/src/Tizen.System.Session/Session/Session.cs
+++ b/src/Tizen.System.Session/Session/Session.cs
@@ -80,7 +80,8 @@ namespace Tizen.System
         [EditorBrowsable(EditorBrowsableState.Never)]
         public static Session GetInstance(int sessionUID)
         {
-            s_sessionInstances.TryAdd(sessionUID, new Session(sessionUID));
+            if (!s_sessionInstances.ContainsKey(sessionUID))
+                s_sessionInstances.TryAdd(sessionUID, new Session(sessionUID));
             return s_sessionInstances[sessionUID];
         }
 
@@ -213,19 +214,21 @@ namespace Tizen.System
 
             _replyMap[taskID] = (int result, IntPtr data) =>
             {
-                _replyMap.Remove((int)data);
 
                 try
                 {
                     CheckError((SessionError)result, "Interop failed to remove a subsession user");
+                    task.SetResult(true);
                 }
                 catch (Exception exception)
                 {
                     task.SetException(exception);
                     return;
                 }
-
-                task.SetResult(true);
+                finally
+                {
+                    _replyMap.Remove((int)data);
+                }
             };
 
             SessionError ret = Interop.Session.SubsessionRemoveUser(SessionUID, userName, _replyMap[taskID], (IntPtr)taskID);
@@ -258,19 +261,20 @@ namespace Tizen.System
 
             _replyMap[taskID] = (int result, IntPtr data) =>
             {
-                _replyMap.Remove((int)data);
-
                 try
                 {
                     CheckError((SessionError)result, "Interop failed to switch to a different subsession user");
+                    task.SetResult(true);
                 }
                 catch (Exception exception)
                 {
                     task.SetException(exception);
                     return;
                 }
-
-                task.SetResult(true);
+                finally
+                {
+                    _replyMap.Remove((int)data);
+                }
             };
 
             SessionError ret = Interop.Session.SubsessionSwitchUser(SessionUID, userName, _replyMap[taskID], (IntPtr)taskID);
@@ -443,7 +447,7 @@ namespace Tizen.System
         /// <exception cref="UnauthorizedAccessException">Not permitted</exception>
         /// <exception cref="NotSupportedException">Not supported</exception>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public event EventHandler<SwitchUserCompletionEventArgs> SwitchUserCompletion
+        public event EventHandler<SwitchUserCompletionEventArgs> SwitchUserCompleted
         {
             add
             {
@@ -471,7 +475,14 @@ namespace Tizen.System
                 return;
 
             Log.Error(SessionErrorFactory.LogTag, msg);
-            SessionErrorFactory.ThrowException(ret);
+            Exception ex = SessionErrorFactory.CreateException(ret);
+            if (ex == null)
+            {
+                Log.Error(SessionErrorFactory.LogTag,
+                    "Unexpected exception type for SessionError: " + Enum.GetName(typeof(SessionError), ret));
+                throw new InvalidOperationException("Unrecognized error");
+            }
+            throw ex;
         }
 
         private void IntPtrToStringArray(IntPtr unmanagedArray, int size, out string[] managedArray)
--- a/src/Tizen.System.Session/Session/Session.cs
+++ b/src/Tizen.System.Session/Session/Session.cs
@@ -215,7 +215,6 @@ namespace Tizen.System
 
             _replyMap[taskID] = (int result, IntPtr data) =>
             {
-
                 try
                 {
                     CheckError((SessionError)result, "Interop failed to remove a subsession user");
--- a/src/Tizen.System.Session/Session/Session.cs
+++ b/src/Tizen.System.Session/Session/Session.cs
@@ -1,0 +1,509 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Tizen.System
+{
+    /// <summary>
+    /// Provides methods to manage subsession users. Allows to register for events triggered by operations on subsession users.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public sealed class Session
+    {
+        /// <summary>
+        /// UID of current system user.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static readonly int CurrentUID;
+
+        /// <summary>
+        /// Maximum length of any given user ID.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public const int MaxUserLength = 20;
+
+        /// <summary>
+        /// Special subsession ID, which is always present and does not represent any user.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public const string EmptyUser = "";
+
+        private static ConcurrentDictionary<int, Session> s_sessionInstances = new ConcurrentDictionary<int, Session>();
+
+        private readonly object _replyLock = new object();
+
+        private IDictionary<int, Interop.Session.SubsessionReplyCallback> _replyMap = new ConcurrentDictionary<int, Interop.Session.SubsessionReplyCallback>();
+
+        private int _replyID = 0;
+
+        private delegate void EventDelegate(SubsessionEventInfoNative infoNative, IntPtr data);
+
+        static Session()
+        {
+            CurrentUID = Interop.Session.GetUID();
+        }
+
+        private Session(int sessionUID)
+        {
+            SessionUID = sessionUID;
+        }
+
+        /// <summary>
+        /// Gets a Session object instance for a given sessionUID.
+        /// </summary>
+        /// <param name="sessionUID">Session UID of a requested Session object instance</param>
+        /// <returns>Returns requested Session object</returns>
+        /// <remarks>
+        /// To ensure thread safety, expilicit creation of Session object is not allowed.
+        /// </remarks>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static Session GetInstance(int sessionUID)
+        {
+            s_sessionInstances.TryAdd(sessionUID, new Session(sessionUID));
+            return s_sessionInstances[sessionUID];
+        }
+
+        /// <summary>
+        /// Gets session UID of this session object.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public int SessionUID { get; private set; }
+
+        /// <summary>
+        /// Gets a list of all availible subsession user IDs for this session.
+        /// </summary>
+        /// <remarks>
+        /// The list of users depends on whether the session UID for this session object exists or not. If it
+        /// doesn't, the user list is empty (in particular this is not an error).
+        /// However if the session UID exists, the user list will contain the subsession
+        /// IDs (if they exist), but also the default value, which is "" (empty string, see EmptyUser field).
+        /// This doesn't mean that "" is a subsession ID in the same way as others; it is just a marker meaning that no subsession is
+        /// enabled.
+        /// </remarks>
+        /// <exception cref="ArgumentException">Session UID of this object is invalid</exception>
+        /// <exception cref="OutOfMemoryException">Out of memory</exception>
+        /// <exception cref="IOException">Internal error</exception>
+        /// <exception cref="UnauthorizedAccessException">Not permitted</exception>
+        /// <exception cref="NotSupportedException">Not supported</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public IReadOnlyList<string> GetUsers()
+        {
+
+            IntPtr ptr;
+            int count;
+
+            SessionError ret = Interop.Session.SubsessionGetUserList(SessionUID, out ptr, out count);
+            CheckError(ret, "Interop failed to get user list");
+
+            string[] users;
+            IntPtrToStringArray(ptr, count, out users);
+            Interop.Session.SubsessionFreeUserList(ptr);
+
+            return new List<string>(users);
+        }
+
+        /// <summary>
+        /// Gets a currently active subession user ID for this session.
+        /// </summary>
+        /// <remarks>
+        /// When no subsession is enabled, "" (empty string, see EmptyUser field) is returned.
+        /// This doesn't mean that "" is a subsession ID in the same way as others; it is just a marker meaning
+        /// that no subsession is enabled.
+        /// </remarks>
+        /// <exception cref="ArgumentException">Session UID of this object is invalid</exception>
+        /// <exception cref="OutOfMemoryException">Out of memory</exception>
+        /// <exception cref="IOException">Internal error</exception>
+        /// <exception cref="UnauthorizedAccessException">Not permitted</exception>
+        /// <exception cref="NotSupportedException">Not supported</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public string GetCurrentUser()
+        {
+            StringBuilder user = new StringBuilder(MaxUserLength);
+            SessionError ret = Interop.Session.SubsessionGetCurrentUser(SessionUID, user);
+            CheckError(ret, "Interop failed to get current subsession user");
+
+            return user.ToString();
+        }
+
+        /// <summary>
+        /// Request new subsession to be created.
+        /// </summary>
+        /// <param name="userName">Subesssion user ID to be created</param>
+        /// <remarks>
+        /// Subsession ID must not start with a dot or have slashes.
+        /// </remarks>
+        /// <exception cref="ArgumentException">Session UID of this object is invalid, or user ID is not a valid subession ID</exception>
+        /// <exception cref="InvalidOperationException"> Provided subsession user ID already exists</exception>
+        /// <exception cref="OutOfMemoryException">Out of memory</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Task SubsessionAddUserAsync(string userName)
+        {
+            var task = new TaskCompletionSource<bool>();
+            int taskID = 0;
+
+            lock (_replyLock)
+            {
+                taskID = _replyID++;
+            }
+
+            _replyMap[taskID] = (int result, IntPtr data) =>
+            {
+                _replyMap.Remove((int)data);
+                try
+                {
+                    CheckError((SessionError)result, "Interop failed to complete adding a new subsession user");
+                }
+                catch (Exception exception)
+                {
+                    task.SetException(exception);
+                    return;
+                }
+
+                task.SetResult(true);
+            };
+
+            SessionError ret = Interop.Session.SubsessionAddUser(SessionUID, userName, _replyMap[taskID], (IntPtr)taskID);
+            CheckError(ret, "Interop failed to register a reply for adding a user");
+            return task.Task;
+        }
+
+        /// <summary>
+        /// Request an existing subsession to be removed.
+        /// </summary>
+        /// <param name="userName">Existing subesssion user ID to be removed</param>
+        /// <remarks>
+        /// Subsession ID must not start with a dot or have slashes.
+        /// Only inactive session ID can be removed. In order remove currently used session ID first switch to special
+        /// session ID "" (empty string, see EmptyUser), and only after switch completes, remove previously active session ID.
+        /// </remarks>
+        /// <exception cref="ArgumentException">Session UID of this object is invalid, or user ID is not a valid subession ID</exception>
+        /// <exception cref="InvalidOperationException">Provided subsession user ID does not exist</exception>
+        /// <exception cref="OutOfMemoryException">Out of memory</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Task SubsessionRemoveUserAsync(string userName)
+        {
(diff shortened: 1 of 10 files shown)


src/Tizen.System.Session/Session/Session.cs
```diff
@@ -0,0 +1,509 @@
...
+        /// </summary>
+        /// <param name="sessionUID">Session UID of a requested Session object instance</param>
+        /// <returns>Returns requested Session object</returns>
+        /// <remarks>
+        /// To ensure thread safety, expilicit creation of Session object is not allowed.
+        /// </remarks>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static Session GetInstance(int sessionUID)
+        {
+            s_sessionInstances.TryAdd(sessionUID, new Session(sessionUID));
```
myroot, line 83: It seems to wrong, 
It will temperory create Session always
```suggestion
            if (!s_sessionInstances.ContainsKey(sessionUID))
                  s_sessionInstances.TryAdd(sessionUID, new Session(sessionUID));
```

src/Tizen.System.Session/Session/SessionStructs.cs
```diff
@@ -0,0 +1,234 @@
...
+        public byte userNamePrev_11;
+        public byte userNamePrev_12;
+        public byte userNamePrev_13;
+        public byte userNamePrev_14;
+        public byte userNamePrev_15;
+        public byte userNamePrev_16;
+        public byte userNamePrev_17;
+        public byte userNamePrev_18;
+        public byte userNamePrev_19;
+
```
myroot, line 162: You don't need it
Please check it
https://learn.microsoft.com/en-us/dotnet/standard/native-interop/customize-struct-marshalling#customizing-string-field-marshalling

https://github.com/Samsung/TizenFX/blob/d586fc850c80f9db2004caef3bedc5b939bfed85/src/Tizen.Multimedia.Camera/Interop/Interop.Camera.cs#L361-L362

src/Tizen.System.Session/Session/SessionErrorFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+        PermissionDenied = ErrorCode.PermissionDenied,
+        NotSupported = ErrorCode.NotSupported,
+    }
+
+    internal static class SessionErrorFactory
+    {
+        internal const string LogTag = "Tizen.System.Session";
+
+        internal static void ThrowException(SessionError err)
+        {
```
myroot, line 41: You do better just return exception object instead throw, 
and throw exception should be done by caller
because of callstack

src/Tizen.System.Session/Session/Session.cs
```diff
@@ -0,0 +1,509 @@
...
+                    CheckError((SessionError)result, "Interop failed to switch to a different subsession user");
+                }
+                catch (Exception exception)
+                {
+                    task.SetException(exception);
+                    return;
+                }
+
+                task.SetResult(true);
+            };
```
myroot, line 274:
```suggestion
            _replyMap[taskID] = (int result, IntPtr data) =>
            {
                try
                {
                    CheckError((SessionError)result, "Interop failed to switch to a different subsession user");
                }
                catch (Exception exception)
                {
                    task.SetException(exception);
                    return;
                }

                task.SetResult(true);
                _replyMap.Remove((int)data);
            };
```
You should remove the callback at the last of callback. if not native part of callback was GCed and it can't return to native

src/Tizen.System.Session/Session/SessionErrorFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                throw new ArgumentException("Invalid parameter");
+            }
+            else if (error == SessionError.Io)
+            {
+                throw new IOException("I/O error");
+            }
+            else if (error == SessionError.OutOfMemory)
+            {
+                throw new OutOfMemoryException("Out of memory");
```
myroot, line 53: We don't throw OOM exception, because it only for CLR.

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types#outofmemoryexception

src/Tizen.System.Session/Session/Session.cs
```diff
@@ -0,0 +1,509 @@
...
+        /// <summary>
+        /// Event to be invoked when an existing subession user is successfully switched to an acvite state.
+        /// </summary>
+        /// <exception cref="ArgumentException">Session UID of this object is invalid</exception>
+        /// <exception cref="OutOfMemoryException">Out of memory</exception>
+        /// <exception cref="IOException">Internal error</exception>
+        /// <exception cref="UnauthorizedAccessException">Not permitted</exception>
+        /// <exception cref="NotSupportedException">Not supported</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public event EventHandler<SwitchUserCompletionEventArgs> SwitchUserCompletion
```
myroot, line 446: What about `SwitchUserCompleted` ? 

Please check it
https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-events

src/Tizen.System.Session/Session/Session.cs
```diff
@@ -0,0 +1,509 @@
...
+            byte[] tmpArray = new byte[MaxUserLength];
+
+            Marshal.Copy(unmanagedArray, byteArray, 0, MaxUserLength * size);
+
+            for (int i = 0; i < size; i++)
+            {
+                Array.Copy(byteArray, MaxUserLength * i, tmpArray, 0, MaxUserLength);
+                managedArray[i] = Encoding.UTF8.GetString(tmpArray);
+            }
+        }
```
myroot, line 491: It seems to be wrong
Please explain your intent

What is native type of `unmanagedArray`. We usually don't copy native buffer to get string

src/Tizen.System.Session/Session/SessionEnums.cs
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+using System.ComponentModel;
+
+namespace Tizen.System
+{
+    internal enum SessionEventType : int
```
myroot, line 21:
```suggestion
    [Flags]
    internal enum SessionEventType : int
```
https://learn.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=net-7.0

src/Tizen.System.Session/Session/Session.cs
```diff
@@ -0,0 +1,509 @@
...
+        /// <exception cref="NotSupportedException">Not supported</exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public event EventHandler<AddUserEventArgs> AddUserWait
+        {
+            add
+            {
+                lock (_addUserWaitLock)
+                {
+                    if (_addUserWaitHandler == null)
+                        RegisterCallbackForEvent(SessionEventType.AddUserWait, _addUserWaitCB, OnAddUserWait);
```
myroot, line 327: _addUserWaitCB is null
nowhere set content of _addUserWaitCB
I can't find it

src/Tizen.System.Session/Session/Session.cs
```diff
@@ -0,0 +1,509 @@
...
+        }
+
+        private void RegisterCallbackForEvent(SessionEventType eventType, Interop.Session.SubsessionEventCallback eventCallback,
+            EventDelegate delegateToSet)
+        {
+            eventCallback = new Interop.Session.SubsessionEventCallback(delegateToSet);
+            SessionError ret = Interop.Session.SubesssionRegisterEventCallback(SessionUID, eventType,
+                eventCallback, IntPtr.Zero);
+            CheckError(ret, $"Interop failed to register a callback for an event of type {eventType}");
+        }
```
myroot, line 500: It's wrong
`eventCallback` is not ref argument

src/Tizen.System.Session/Session/Session.cs
a-michalski2: This is set in #495 in `RegisterCallbackForEvent` function:

    eventCallback = new Interop.Session.SubsessionEventCallback(delegateToSet);

src/Tizen.System.Session/Session/Session.cs
a-michalski2: This may be a bit misleading, I admit. The original array definition looks like this:

    typedef char subsession_user_t[SUBSESSION_USER_MAXLEN];
    ...
    EXPORT_API int subsession_get_user_list(int session_uid, subsession_user_t **user_list, int *user_count)
    {
        ....
        subsession_user_t *const list = calloc(elems, sizeof *list);
        ....
        *user_list = list;
    }
    
which means it's actually an array of fixed size (SUBSESSION_USER_MAXLEN) arrays. Unfortunately, a simple IntPtrToStringArray implementation:

        static void IntPtrToStringArray(IntPtr unmanagedArray, int size, out string[] managedArray)
        {
            managedArray = new string[size];
            IntPtr[] IntPtrArray = new IntPtr[size];

            Marshal.Copy(unmanagedArray, IntPtrArray, 0, size);

            for (int iterator = 0; iterator < size; iterator++)
            {
                managedArray[iterator] = Marshal.PtrToStringAnsi(IntPtrArray[iterator]);
            }
        }

doesn't work in this case.

src/Tizen.System.Session/Session/Session.cs
a-michalski2: Yes. Looks like it needs to be `ref`. Will fix.

src/Tizen.System.Session/Session/SessionEnums.cs
a-michalski2: Thanks for hints. Unfortunately this case is more complicated, as the structure actually contains unions, which in turn constist of another structures:

	subsession_event_type_e event;
	int session_uid;

	union {
		struct {
			subsession_user_t user;
		} add_user;

		struct {
			subsession_user_t user;
		} remove_user;

		struct {
			uint64_t switch_id;
			subsession_user_t prev_user;
			subsession_user_t next_user;
		} switch_user;
	};
	
We've tried to use LayoutKind Explicit for unions and LayoutKind.Sequential for structs, but this unfortunately didn't work properly, event with a proper explicitly stated alignments.

src/Tizen.System.Session/Session/SessionErrorFactory.cs
a-michalski2: Could you explain what do you mean by saying "throw exception should be done by caller"?

src/Tizen.System.Session/Session/SessionStructs.cs
a-michalski2: Thanks for hints. Unfortunately this case is more complicated, as the structure actually contains unions, which in turn constist of another structures:

    typedef struct subsession_event_info {
    
        subsession_event_type_e event;
        int session_uid;
    
        union {
            struct {
                subsession_user_t user;
            } add_user;
    
            struct {
                subsession_user_t user;
            } remove_user;
        
            struct {
                uint64_t switch_id;
                subsession_user_t prev_user;
                subsession_user_t next_user;
            } switch_user;
        };
    } subsession_event_info;

We've tried to use LayoutKind Explicit for unions and LayoutKind.Sequential for structs, but this unfortunately didn't work properly, event with a proper explicitly stated alignments.

src/Tizen.System.Session/Session/SessionErrorFactory.cs
myroot: If we use this method, 
All exceptions call stack was ended with `ThrowException`

So, I recommend like it
``` c#
   if (err != SessionError.None)
        throw SessionErrorFactory.CreateException(err);
```

src/Tizen.System.Session/Session/SessionStructs.cs
myroot: here union sample code
https://learn.microsoft.com/en-us/dotnet/framework/interop/marshalling-classes-structures-and-unions#unions-sample

src/Tizen.System.Session/Session/SessionErrorFactory.cs
a-michalski2: What would you recommend instead of the OutOfMemoryException? Original `libsessiond`can return an out of memory exception sometimes (though it's not very probable, but it's possible anyway).

src/Tizen.System.Session/Session/SessionStructs.cs
a-michalski2: A union/several unions are not a problem if they occur separately. The problem is when they are members of a struct. It's the situation that we have with `libsessiond` (please see the attached definition of the C struct in the above comment). It leads to a runtime error: "object-field at offset 4 is incorrectly aligned or overlapped by a non-object field" when marshalled with StructLayout(LayoutKind.Explicit) and FieldOffset. A very similar situation was described here:

https://stackoverflow.com/questions/71426822/marshaling-complex-c-structs-with-c-unions-for-c-sharp

src/Tizen.System.Session/Session/Session.cs
myroot: so, unmanagedArray is sequence of char and length is 20 * elm_no
in this case, 
You can marshal like it
``` c#
    static void IntPtrToStringArray(IntPtr unmanagedArray, int size, out string[] managedArray)
    {
        managedArray = new string[size]; 
       var curr = unmanagedArray;
        for (int iterator = 0; iterator < size; iterator++)
        {
            managedArray[iterator] = Marshal.PtrToStringAnsi(curr, 20);
            curr = IntPtr.Add(curr, 20);
        }
    }

```

I didn't test this code, it could be contained error carefully check it

a-michalski2: Thanks for suggestion. Will investigate it further and check it.

src/Tizen.System.Session/Session/SessionErrorFactory.cs
myroot: We recommend InvalidOperationException

a-michalski2: Ok. Thanks.

src/Tizen.System.Session/Session/SessionErrorFactory.cs
a-michalski2: Ok.

src/Tizen.System.Session/Session/SessionStructs.cs
myroot: Maybe you can marshal with it
```
    internal struct SubsessionEventInfoNative
    {
        [FieldOffset(0)]
        public SessionEventType eventType;
        [FieldOffset(4)]
        public int sessionUID;
        [FieldOffset(8)]
        [MarshalAs(UnmanagedType.LPArray, SizeConst = 20)]
        public string AddUser;
        [FieldOffset(8)]
        [MarshalAs(UnmanagedType.LPArray, SizeConst = 20)]
        public string RemoveUser;
        [FieldOffset(8)]
        public Int64 switchID;
        [MarshalAs(UnmanagedType.LPArray, SizeConst = 20)]
        public string PrevUser;
        [MarshalAs(UnmanagedType.LPArray, SizeConst = 20)]
        public string NextUser;
    }

```

myroot: Anyway, .net marshal api can convert any struct/union on unmanaged, I hope to you check more marshal API

myroot:
``` c#
    internal struct SubsessionEventInfoNative
    {
        [FieldOffset(0)]
        public SessionEventType eventType;
        [FieldOffset(4)]
        public int sessionUID;
        [FieldOffset(8)]
        public IntPtr AddUserPtr;
        [FieldOffset(8)]
        public IntPtr RemoveUserPtr;
        [FieldOffset(8)]
        public Int64 switchID;
        [FieldOffset(16)]
        public IntPtr PrevUserPtr;
        [FieldOffset(36)]
        public IntPtr NextUserPtr;
    }
```
And It also works like it
``` c#
  var eventInfo = (SubsessionEventInfoNative)Marhsal.PtrToStructure( ...)
  var adduser = Marhsal.PtrToStringAnsi(eventInfo.AddUserPtr, 20);
...
  var prevUser = Marhsal.PtrToStringAnsi(eventInfo.PrevUserPtr, 20);

```

HJLeee: @a-michalski2 Please take a look if you can access CODE.
https://github.sec.samsung.net/dotnet/home/issues/421#issuecomment-1686515
Camera API marshals way more complicated struct than this one.

src/Tizen.System.Session/Session/Session.cs
```diff
@@ -0,0 +1,520 @@
...
+                    CheckError((SessionError)result, "Interop failed to complete adding a new subsession user");
+                }
+                catch (Exception exception)
+                {
+                    task.SetException(exception);
+                    return;
+                }
+
+                task.SetResult(true);
+            };
```
myroot, line 185:
```suggestion
            _replyMap[taskID] = (int result, IntPtr data) =>
            {

                try
                {
                    CheckError((SessionError)result, "Interop failed to complete adding a new subsession user");
                    task.SetResult(true);
                }
                catch (Exception exception)
                {
                    task.SetException(exception);
                    return;
                }
                finally
                {
                    _replyMap.Remove((int)data);
                }
            };
```

src/Tizen.System.Session/Session/SessionStructs.cs
a-michalski2: Thanks for suggestions. Unfortunately the following doesn't work (PrevUserPtr always contains null). 

``` C#
    internal struct SubsessionEventInfoNative
    {
        [FieldOffset(0)]
        public SessionEventType eventType;
        [FieldOffset(4)]
        public int sessionUID;

        [FieldOffset(8)]
        public IntPtr AddUserPtr;

        [FieldOffset(8)]
        public IntPtr RemoveUserPtr;

        [FieldOffset(8)]
        public Int64 switchID;
        [FieldOffset(16)]
        public IntPtr PrevUserPtr;
        [FieldOffset(36)]
        public IntPtr NextUserPtr;
    }
```

We've come up with something though:

``` C#
    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
    internal struct AddUserStruct
    {
        [MarshalAs(UnmanagedType.LPArray, SizeConst = 20)]
        internal string userName;
    }
    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
    internal struct RemoveUserStruct
    {
        [MarshalAs(UnmanagedType.LPArray, SizeConst = 20)]
        internal userName;
    }
    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
    internal struct SwitchUserStruct
    {
        internal Int64 switchID;
        [MarshalAs(UnmanagedType.LPArray, SizeConst = 20)]
        internal string prevUser;
        [MarshalAs(UnmanagedType.LPArray, SizeConst = 20)]
        internal string nextUser;
    }

    [StructLayout(LayoutKind.Explicit)]
    internal struct UserDataStruct
    {
        [FieldOffsetAttribute(0)]
        internal AddUserStruct AddUserData;
        [FieldOffsetAttribute(0)]
        internal RemoveUserStruct RemoveUserData;
        [FieldOffsetAttribute(0)]
        internal SwitchUserStruct SwitchUserData;
    }

    [StructLayout(LayoutKind.Sequential)]
    internal struct SessionEventInfo
    {
        internal SessionEventType eventType;
        internal int SessionUID;
        internal UserDataStruct userData;
    }
```

We'll be testing this today to see if it works.

src/Tizen.System.Session/Session/Session.cs
a-michalski2: Thank you. Yes, using ```finally``` is definitely a good idea here.

src/Tizen.System.Session/Session/Session.cs
a-michalski2: We tested it yesterday and it works. We'll include it in our changes. Thank you.

src/Tizen.System.Session/Session/SessionStructs.cs
a-michalski2: Unfortunately, this solution doesn't work either.

It truncates first 8 bytes of the `AddUserStruct.userName` and `RemoveUserStruct.userName` (`SwitchUserStruct.prevUser` and `SwitchUserStruct.nextUser` seem to be properly marshalled):
```
    Adding test users...
    Added - �
    Added - e
    Added - rf
    Added - murf
    Added - murf
    Added - murf
    Added - smurf
    Added - urf
    Added - urf
    Added - rf
    Users added
    Curr 
    Curr baby_smurf
    Curr ㅋㅌㅍ
    Curr smurfette
    Curr brainy_smurf
    Curr clumsy_smurf
    Curr greedy_smurf
    Curr grouchy_smurf
    Curr hefty_smurf
    Curr jokey_smurf
    Curr lazy_smurf
    Switching to each user...
    Switched from 
    Switched from ㅋㅌㅍ
    Switched from smurfette
    Switched from baby_smurf
    Switched from brainy_smurf
    Switched from clumsy_smurf
    Switched from greedy_smurf
    Switched from grouchy_smurf
    Switched from hefty_smurf
    Switched from jokey_smurf
    Switching finished.
```
Another attempt:
``` C#
 [NativeStruct("subsession_event_info", Include="sessiond.h", PkgConfig="libsessiond")]
    [StructLayout(LayoutKind.Explicit)]
    internal struct SubsessionEventInfoNative
    {
        [FieldOffset(0)]
        public SessionEventType eventType;
        [FieldOffset(4)]
        public int sessionUID;

        [FieldOffset(8)]
        [MarshalAs(UnmanagedType.ByValArray, SizeConst = 20)]
        public char[] AddUser;

        [FieldOffset(8)]
        [MarshalAs(UnmanagedType.ByValArray, SizeConst = 20)]
        public char[] RemoveUser;

        [FieldOffset(8)]
        public Int64 SwitchID;

        [FieldOffset(16)]
        [MarshalAs(UnmanagedType.ByValArray, SizeConst = 20)]
        public char[] SwitchUserPrev;
        [FieldOffset(36)]
        [MarshalAs(UnmanagedType.ByValArray, SizeConst = 20)]
        public char[] SwitchUserNext;
    }
```
gives the following runtime error:
```
Unhandled exception. System.TypeLoadException: Could not load type 'Tizen.System.SubsessionEventInfoNative' from assembly 'Tizen.System.Session, Version=4.0.0.0, Culture=neutral, PublicKeyToken=eba370b203a2e242' because it contains an object field at offset 8 that is incorrectly aligned or overlapped by a non-object field.
```

We've tried several other possible solutions but we end up with either the runtime error or a truncation of the first 4 or 8 characters of the `userName` for `AddUser` and `RemoveUser` union.

myroot: Try it also
``` c#
IntPtr nativePtr = ...
string addUser = Marshal.PtrToStringAnsi(IntPtr.Add(nativePtr, 8), 20)
...
string removeUser = Marshal.PtrToStringAnsi(IntPtr.Add(nativePtr, 8), 20)
...
string prevUser = Marshal.PtrToStringAnsi(IntPtr.Add(nativePtr, 16), 20)
string nextUser = Marshal.PtrToStringAnsi(IntPtr.Add(nativePtr, 36), 20)
```

a-michalski2: Thanks for further suggestions. We'll be trying to test more possible solutions. We're now moving the code to produce a MRE consisting only of the `.c` file with structures and a mocked function filling it and a simple C# project with a single `.cs` file to be able to build and check it quickly on Ubuntu without having to build the whole `TizenFX` which takes a lot of time.

adaszkiewicza: We finally managed to get this struct working without using unrolled arrays. Thanks for your help!

src/Tizen.System.Session/Session/SessionStructs.cs
```diff
@@ -0,0 +1,48 @@
...
+    [StructLayout(LayoutKind.Explicit)]
+    internal struct SubsessionEventInfoNative
+    {
+        [FieldOffset(0)]
+        public SessionEventType eventType;
+        [FieldOffset(4)]
+        public int sessionUID;
+
+        [FieldOffset(8)]
+        public IntPtr AddUserPtr;
```
myroot, line 35: I hope to add some comment here,
because, Actually it is not used and not correct, 
so I concern someone use it.

a-michalski2: What about marking the fields as private and adding the comment that the actual work is done via `Marshal.PtrToStructure` and `Marshal.PtrToStringAnsi` calls? Removing those fields won't work as gbs performs additional native structure validation step.

adaszkiewicza:
I think we fixed everything

myroot (approved):

PR #5204, opened 2023-04-21
[NUI] Remove unused signal : Window.ViewAdded

hinohie:
Due to the performance issue, we do not allow to use Window.ViewAdded signal.
Instead, We need to guide app to use View.AddedToWindow event


dongsug-song (approved):
approved

PR #5205, opened 2023-04-21
[NUI] Remove default parameter of Animation.Stop(EndActions)

hinohie:
Since we already have empty-parameter function named Stop(), Let we remove Stop(EndActions)'s default value so compiler don't feel confuse.


dongsug-song (approved):
approved

bshsqa (approved):

PR #5206, opened 2023-04-24
[NUI] Fix some comments of Animation class

hinohie:
1. Change mismatched comment format
2. Change default value of DisconnectAction
3. Add some remarks at EndAction property


dongsug-song (approved):
approved

PR #5214, opened 2023-04-26
[NUI] Add SetTapRecognizerTime

JoogabYun:
### Description of Change ###

This is the time from touch down to touch up to recognize a tap gesture.

If set to 300ms, a touch up after a touch down must occur within 300ms to be recognized as a tap gesture.

### API Changes ###

 - ACR:


hinohie (approved):
Approve.

PR #5217, opened 2023-04-26
[NUI] Change the call to GetRootLayer() so that the layer index can be updated.

JoogabYun:
### Description of Change ###

If the root layer index is not updated and another layer is added first, a problem occurs in calculating the layer index. Something like LowerBelow may not work.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
LGTM!!! 
Already applied to DevelNUI branch.

PR #5223, opened 2023-04-27
[NUI] Fix empty item case crash. wrong bracket conditions.

everLEEst:
when failed flag is true,
other condition should be ingored.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

taehyub (approved):
LGTM

PR #5227, opened 2023-04-28
[NUI.Scene3D] Make API to apply ModelMotion

hinohie:
Make a prototype of MotionData settor & Animation generator

Let we can use two type of motion as index
- MotionTransformIndex
- BlendShapeIndex

MotionTransformIndex can control only Position / Orientation / Scale.
and, BlendShape can control only BlendShape.

Currently a lots of case are not supported.
But we can test MotionData usecase by "string" ModelNodeId + "int" blendshape


src/Tizen.NUI.Scene3D/src/public/ModelMotion/MotionIndex.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// All other cases are invalid.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class MotionIndex : IDisposable
+    {
+        /// <summary>
+        /// TODO : Explaine me.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public enum PropertySlotType
```
bshsqa, line 47: This patch is for Motion like character animation or BlendShape morphing.
But this looks to allow all of the ViewProperties those are animatable, right?
So, The role goes beyond motion, and it seems like another new NUI.Animation for View property.

And, to set SlotType field by user is slightly unfriendly api.
We may can use these way if you OK.

MotionIndex drawio

User can simply select Motion type by creating proper child class.
And in the InternalGenerateMotionDataAnimation, we can simply check the class type.
And, GetPropertyName() can be used in Internal Model Class.

TransformType is just an example. If such a enum is used, the functionality is limited, but I think it will be a little easier to use.

src/Tizen.NUI.Scene3D/src/public/ModelMotion/MotionData.cs
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright(c) 2022 Samsung Electronics Co., Ltd.
```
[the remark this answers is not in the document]
hinohie: Oh... Thanks a lot. I'll fix it

src/Tizen.NUI.Scene3D/src/public/ModelMotion/MotionIndex.cs
hinohie: I think this architect is more useful then current `PropertySlotType`. Currently, user 'might' try to change only the `SlotType` value.  I'll apply this kind of class system.

bshsqa (approved):

hinohie:
All api are internal. If there is some fixed point exist, we'd better update it another patch.

PR #5259, opened 2023-05-09
[NUI] Deprecated Animation.Stop(EndActions)

hinohie:
That API change the 'EndAction' property.
And also, EndAction input parameter will not works well if

 - It's value is EndActions.Discard
 - Already finished animation

So, we'd better guide to app developer to use EndAction property instead of Stop(EndActions) API.

### API Changes ###

 - ACR: ACR-540 https://code.sec.samsung.net/jira/browse/TCSACR-540

Deprecated:
 - void Stop(EndActions action);

### DIFF
--- a/src/Tizen.NUI/src/public/Animation/Animation.cs
+++ b/src/Tizen.NUI/src/public/Animation/Animation.cs
@@ -702,12 +702,16 @@ namespace Tizen.NUI
         /// <summary>
         /// Stops the animation. It will change this animation's EndAction property.
         /// </summary>
-        /// <remark>
-        /// Change the value from EndActions.Discard, or to EndActions.Discard during animation is playing / paused will not works well.<br />
-        /// If you want to stop by EndActions.Discard, EndAction property also should be EndActions.Discard before Play API called.
-        /// </remark>
+        /// <remarks>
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Animation/Animation.cs
```diff
@@ -679,6 +679,10 @@ internal static Animation DownCast(BaseHandle handle)
         /// </remark>
         /// <param name="action">The end action can be set.</param>
         /// <since_tizen> 3 </since_tizen>
+        /// <remarks>
```
wiertel, line 682: I don't think we need to repeat that information in remarks. BTW, the `<remark>` is above is not a valid tag.
Please fix lines 676 and 679: remark -> remarks.
If you want to keep that note: add it to existing remarks section, leaving empty comment line or inserting `<br/>` to separate paragraphs/break the line. And the information should be changed a little into something like "The method was rarely used and was deprecated since API11." or "The method is deprecated since API11 because it was not used."

hinohie: Hi. Thank you for your review point. Now I change the tag name as `remarks`, and also add more informations the another way to do the same behaviour.

hyunjushin (approved):

wiertel (approved):

hinohie:
ACR Accept

PR #5262, opened 2023-05-10
[NUI] Add FontPreLoad API to FontClient

wonrst:
This patch adds a new FontPreLoad API that preloads font faces.
this can prevents delays on main thread during the inital call of FT_New_Face.

using memoryFontPathList involves loading font buffers into memory. 
And using FT_New_Memory_Face at runtime, which uses more memory but can lead to greater performance improvements.

https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-adaptor/+/292224/
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/292534/


hinohie (approved):
API looks fine to me. Please merge it if dali side patch merged.

PR #5263, opened 2023-05-10
[NUI.Scene3D] Change the API name for Samsung json morphing

hinohie:
Since json file can change not only facial, The name of API is not good.

Instead, Let we change the name of facial API as 'SamsungJsonMorphing'. Now user can notify that only samsung defined json formats allowed


src/Tizen.NUI.Scene3D/src/internal/Interop/Interop.Model.cs
```diff
@@ -96,10 +96,10 @@ internal static partial class Model
             public static extern global::System.IntPtr LoadBvhAnimationFromBuffer(global::System.Runtime.InteropServices.HandleRef model, string bvhBuffer, int bvhBufferLength, global::System.Runtime.InteropServices.HandleRef scale, bool translateRootFromModelNode);
 
             [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_Model_LoadFacialAnimation_1")]
-            public static extern global::System.IntPtr LoadFacialAnimation(global::System.Runtime.InteropServices.HandleRef model, string facialFilename);
+            public static extern global::System.IntPtr LoadSamsungJsonMorphingAnimation(global::System.Runtime.InteropServices.HandleRef model, string jsonFilename);
```
bshsqa, line 99: Actually, this is a kind of amazing technicolor dream naming.
We need to talk about this more....

hyunjushin: why don't you use LoadBlendShapeAnimation()?

hinohie: @bshsqa @hyunjushin  Thanks to both of you. I make the API name as `LoadBlendShapeAnimation`, and write some comment that we use json format file.

bshsqa (approved):

PR #5267, opened 2023-05-12
[Notifications] Add event for do not show again popup

sukhyungkang:
### Description of Change ###
Add event to set and get Do not show again check box

### API Changes ###

 - ACR: Non-ACR


sukhyungkang:
it's necessary to merge after native code is merged.

PR #5268, opened 2023-05-12
[NUI.Scene3D] Change the API name for BlendShape morphing

hinohie:
Since json file can change not only facial, The name of API is not good.

Instead, Let we change the name of facial API as 'BlendShapeAnimation'. Now user can notify that this API can control all kinds of blendshape.


bshsqa (approved):
LGTM+

PR #5270, opened 2023-05-12
[NUI] fix font size scale issue in fhub

wonrst:
Currently, it is impossible to differentiate between the mobile and common profiles.
Until profile differentiation becomes possible, set the default scale for each profile to 1.

* default font size scale of PROFILE_MOBILE = FontSizeScaleLarge
* default font size scale of PROFILE_COMMON = FontSizeScaleNormal


hinohie (approved):
We should not forget this patch, and revive in future.

PR #5276, opened 2023-05-15
[NUI] Binds GetCurrentProperty to Animatable class

bshsqa:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Animation/Animatable.cs
+++ b/src/Tizen.NUI/src/public/Animation/Animatable.cs
@@ -200,6 +200,18 @@ namespace Tizen.NUI
         }
 
         /// <summary>
+        /// Retrieves the latest rendered frame value of the property.
+        /// </summary>
+        /// <param name="index">The index of the property.</param>
+        /// <returns>The property value.</returns>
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/Animation/Animatable.cs
```diff
@@ -199,6 +199,18 @@ public PropertyValue GetProperty(int index)
             return ret;
         }
 
+        /// <summary>
+        /// Retrieves a property value.
```
hinohie, line 203:
```suggestion
        /// Retrieves the latest rendered frame value of the property.
```

I think we'd better match the comment with dali-core side comments. Actually, should be difference with `GetProperty` API comments.

https://github.com/dalihub/dali-core/blob/master/dali/public-api/object/handle.h#L480

hinohie (commented):
Change the `CaptureTest.cs` is what you intented?

src/Tizen.NUI/src/public/Animation/Animatable.cs
bshsqa: Oh. good good. I fixed.

hinohie (approved):
Looks good to me

PR #5280, opened 2023-05-16
[NUI] Reduce get/set overhead for some LottieAnimationView properties

hinohie:
Let's believe the cached properties what LottieAnimationView already cached.

Since ImageView.Image is heavy operation, we should call this thing less times.


dongsug-song (approved):
approved

hinohie:
Test done at real world VD app works well

PR #5281, opened 2023-05-16
[NUI][API10] Reduce get/set overhead for some LottieAnimationView properties

hinohie:
Let's beleve the cached properties what LottieAnimationVIew already cached.

Since ImageView.Image is heavy operation, we should call this thing less times.


dongsug-song (approved):
approved

PR #5284, opened 2023-05-16
[NUI][API10] Let LottieAnimationView use InternalVisual API

hinohie:
To avoid useless PropertyValue class creation, let we change to visual use InternalVisual APIs.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #5285, opened 2023-05-16
[NUI] Cache url when we call SetImage + Make way to skip visual creation

hinohie:
Previously, we don't cache Url infomation at cached property map when we use SetImage API instead of ResourceUrl property.

Now, let we keep this cache information.

And also, add parameter that we dont want to create new visual when we change this property.


dongsug-song (approved):
approved

PR #5287, opened 2023-05-18
[NUI][AT-SPI] Add AccessibilityDynamicAttributes

aswigon:
### Description of Change ###

Currently, the public API only allows static attribute values, for example:

```c#
// Initial value:
view.AccessibilityAttributes["current_page"] = CurrentPage;

// Need to update in some other place in code when CurrentPage changes:
view.AccessibilityAttributes["current_page"] = CurrentPage;
```

Due to this difficulty, `Pagination` uses a hack with a static attribute `AccessibilityAttributes["style"] = "pagecontrolbyvalue"` and unnecessarily implementing the `IAtspiValue` interface and passing `SelectedIndex` via `IAtspiValue.AccessibilityGetCurrent`.

However, the attributes are actually already evaluated dynamically on demand in the helper method `AccessibilityGetAttributes` called by the AT-SPI bridge. Therefore, it makes sense to allow associating a key with a functor as well as a string value, for example:

```c#
// Always up-to-date thanks to dynamic evaluation:
view.AccessibilityDynamicAttributes["current_page"] = () => CurrentPage;
```

The `Pagination` hack will be removed (and this new API will be used instead) in a separate PR (this will need to be synchronized with a matching change in Screen Reader).

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
@@ -61,6 +61,12 @@ namespace Tizen.NUI.BaseComponents
         [EditorBrowsable(EditorBrowsableState.Never)]
         public Dictionary<string, string> AccessibilityAttributes { get; } = new Dictionary<string, string>();
 
+        /// <summary>
+        /// Dictionary of dynamically-evaluated accessibility attributes (key-value pairs of strings).
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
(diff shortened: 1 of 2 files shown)


Seoyeon2Kim (approved):
LGTM.
BTW, if this attribute support FLUX components, then we should apply this PR to API10 branch as well.

Seoyeon2Kim:
> > LGTM. BTW, if this attribute support FLUX components, then we should apply this PR to API10 branch as well.
> 
> Sure, I'll backport it as soon as this PR is merged (in case there are any requests for changes in the process).

Okay~ I see. Thank you for your quick reply!

aswigon:
API10 backport: https://github.com/Samsung/TizenFX/pull/5299

PR #5290, opened 2023-05-19
Add DestroySignal Function

huiyueun:
### Description of Change ###

- DestroySignal을 넣어서 EventCallback종료시 호출
- Signal Interface를 추가해서 시그널이 상속받도록 변경

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Application/Application.cs
+++ b/src/Tizen.NUI/src/internal/Application/Application.cs
@@ -401,6 +401,26 @@ namespace Tizen.NUI
             DestroySignal(taskDeviceOrientationChangedSignal, applicationTaskDeviceOrientationChangedEventCallback);
             DestroySignal(taskAppControlSignal, applicationTaskAppControlEventCallbackDelegate);
 
+            initSignal = null;
+            terminateSignal = null;
+            pauseSignal = null;
+            resumeSignal = null;
+            resetSignal = null;
+            languageChangedSignal = null;
+            regionChangedSignal = null;
+            batteryLowSignal = null;
+            memoryLowSignal = null;
+            deviceOrientationChangedSignal = null;
+            appControlSignal = null;
+            taskInitSignal = null;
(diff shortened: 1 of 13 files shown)


src/Tizen.NUI/src/internal/Application/Application.cs
```diff
@@ -1957,5 +1846,14 @@ internal DeviceOrientationChangedSignalType TaskDeviceOrientationChangedSignal()
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
             return ret;
         }
+
+        private void DestroySignal(Signal signal, Delegate callbackDelegate)
```
rabbitfor, line 1850: The repeated pattern includes `signal = null`.
How about adding `ref` keyword to make signal to null?
```C#
private void DestorySignal(ref Signal signal, Delegate callbackDelegate)
{
    if (signal != null)
    {
        if (callbackDelegate != null)
        {
            signal.Disconnect(callbackDelegate);
        }
        signal.Dispose();
        signal = null;
    }
}
```

src/Tizen.NUI/src/internal/Common/Signal.cs
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+namespace Tizen.NUI
+{
+    internal interface Signal
```
rabbitfor, line 20: IMHO, it would be nice to add `I` prefix for the interface name :)
https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces

> DO prefix interface names with the letter I, to indicate that the type is an interface.

PR #5293, opened 2023-05-22
[NUI] Add DispatchParentTouchEvent

JoogabYun:
### Description of Change ###

If a View's DispatchParentTouchEvents is set to false, then parents will not receive a touch event signal either.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
LGTM

PR #5299, opened 2023-05-23
[NUI][AT-SPI] Add AccessibilityDynamicAttributes

aswigon:
### Description of Change ###

Note: This is a backport of https://github.com/Samsung/TizenFX/pull/5287.

Currently, the public API only allows static attribute values, for example:

```c#
// Initial value:
view.AccessibilityAttributes["current_page"] = CurrentPage;

// Need to update in some other place in code when CurrentPage changes:
view.AccessibilityAttributes["current_page"] = CurrentPage;
```

However, the attributes are actually already evaluated dynamically on demand in the helper method `AccessibilityGetAttributes` called by the AT-SPI bridge. Therefore, it makes sense to allow associating a key with a functor as well as a string value, for example:

```c#
// Always up-to-date thanks to dynamic evaluation:
view.AccessibilityDynamicAttributes["current_page"] = () => CurrentPage;
```

### API Changes ###

 - ACR: None


hinohie (approved):
Looks good to me.

Seoyeon2Kim (approved):
LGTM
Thank you for adding it to API10 branch.

PR #5301, opened 2023-05-25
[Applications.Common] Fix file not found exception

upple:
When ini file does not exist, LibIniParser.Load() makes error logs. This patch is to remove this case.

### DIFF
--- a/src/Tizen.Applications.Common/Tizen.Applications/CultureInfoHelper.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications/CultureInfoHelper.cs
@@ -20,58 +20,34 @@ using System.Runtime.InteropServices;
 
 namespace Tizen.Applications
 {
-
     internal static class CultureInfoHelper
     {
-        private class CultureInfoManager
-        {
-            private static CultureInfoManager _inst;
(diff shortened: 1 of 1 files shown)


src/Tizen.Applications.Common/Tizen.Applications/CultureInfoHelper.cs
```diff
@@ -15,32 +15,62 @@
...
+                _dictionary = Interop.LibIniParser.Load(_pathCultureInfoIni);
+            }
+        }
+
+        ~CultureInfoManager()
+        {
+            if (_dictionary != IntPtr.Zero)
+            {
+                Interop.LibIniParser.FreeDict(_dictionary);
+            }
```
hjhun, line 47: Please do not make using singleton pattern.
In C# language, static class is recommended instead of singleton pattern.

upple: I modified this implementation.
Thanks for the good suggestion 👍

hjhun (approved):

hjhun (approved):

PR #5306, opened 2023-05-30
[NUI] Double buffered ProcessorOnceEvent so we allow attach event dur…

hinohie:
…ing invoke

When we load cached image, `ResourceLoaded` signal invoked immediate. So, `ResourceLoaded` can be called during `UpdateImage(object, EventArgs)`.

But since we make null after invoke + C# didn't invoke event during invoking. That mean, attached `UpdateImage()` information disapeared.

So we need to allow `UpdateImage()` event invoke during `ProcessorOnceEvent` invoking.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/ProcessorController.cs
+++ b/src/Tizen.NUI/src/internal/Common/ProcessorController.cs
@@ -41,6 +41,11 @@ namespace Tizen.NUI
 
         internal ProcessorController(global::System.IntPtr cPtr, bool cMemoryOwn) : base(cPtr, cMemoryOwn)
         {
+            onceEventIndex = 0u;
+            internalProcessorOnceEvent = new EventHandler[2];
+            internalProcessorOnceEvent[0] = null;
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/internal/Common/ProcessorController.cs
```diff
@@ -68,15 +87,20 @@ public static ProcessorController Instance
 
         public void Process()
         {
-            ProcessorOnceEvent?.Invoke(this, null);
-            ProcessorOnceEvent = null;
+            // Let we add once event into 1 index during 0 event invoke
```
Seoyeon2Kim, line 90: Very minor :)
```suggestion
            // Let us add once event into 1 index during 0 event invoke
```

Seoyeon2Kim (approved):
Please update descriptions. And regarding others, LGTM :)

BTW, shouldn't we apply this commit to API10 branch as well?

PR #5307, opened 2023-05-30
[NUI][API10]Double buffered ProcessorOnceEvent so we allow attach event during in…

hinohie:
…voke

When we load cached image, `ResourceLoaded` signal invoked immediate. So, `ResourceLoaded` can be called during `UpdateImage(object, EventArgs)`.

But since we make null after invoke + C# didn't invoke event during invoking. That mean, attached `UpdateImage()` information disapeared.

So we need to allow `UpdateImage()` event invoke during `ProcessorOnceEvent` invoking.


Seoyeon2Kim (approved):
LGTM~

PR #5316, opened 2023-06-01
[NUI] Add disable hover event option

JoogabYun:
### Description of Change ###

Add system option to disable hoverEvent system-wide.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Globalization;
 using System.Runtime.InteropServices;
 using Tizen.NUI.Binding;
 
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -224,6 +226,16 @@ internal View(global::System.IntPtr cPtr, bool cMemoryOwn, bool shown = true) :
 
             GrabTouchAfterLeave = defaultGrabTouchAfterLeave;
             AllowOnlyOwnTouch = defaultAllowOnlyOwnTouch;
+
+            var hoverEnv = System.Environment.GetEnvironmentVariable("DISABLE_HOVERING");
```
rabbitfor, line 230: IMHO, I don't think it have to get environment variable and parse the value every time a View is created. How about making the `disableHoverEvent` to a static variable and calcuate once when it assigned in the first time?

PR #5323, opened 2023-06-09
[NUI] Fix incorrect calculation of configGestureCount.

JoogabYun:
### Description of Change ###

[NUI] Fix incorrect calculation of configGestureCount.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #5326, opened 2023-06-12
[NUI] Add Directional Light on Scene3D

bshsqa:
### Description of Change ###

Directional light is one of light option in 3D Graphics.
All fragments receive light from the same direction.
One of the example of Directional Light is sun light.
In this patch, includes Light class that control one of light source, color and direction.
This light class object will be added on SceneView, and every object in the SceneView affect by this light.
SceneView can support up-to 5 activated lights in the same time.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Scene3D/src/internal/Interop/Interop.Light.cs
+++ b/src/Tizen.NUI.Scene3D/src/internal/Interop/Interop.Light.cs
@@ -1,0 +1,50 @@
+﻿/*
+ * Copyright(c) 2023 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+namespace Tizen.NUI.Scene3D
+{
+    internal static partial class Interop
+    {
+        internal static partial class Light
+        {
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_Light_New_SWIG_0")]
+            public static extern global::System.IntPtr New();
+
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_new_Light_SWIG_0")]
+            public static extern global::System.IntPtr NewLight();
+
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_new_Light_SWIG_1")]
+            public static extern global::System.IntPtr NewLight(global::System.Runtime.InteropServices.HandleRef light);
+
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_delete_Light")]
+            public static extern void DeleteLight(global::System.Runtime.InteropServices.HandleRef light);
+
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_Light_DownCast")]
+            public static extern global::System.IntPtr LightDownCast(global::System.Runtime.InteropServices.HandleRef light);
+
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_Light_Enable")]
+            public static extern void Enable(global::System.Runtime.InteropServices.HandleRef light, bool enable);
+
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_Light_IsEnabled")]
+            [return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.U1)]
+            public static extern bool IsEnabled(global::System.Runtime.InteropServices.HandleRef light);
+
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_SceneView_GetMaximumEnabledLightCount")]
+            public static extern uint GetMaximumEnabledLightCount();
+        }
+    }
+}
--- a/src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
+++ b/src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
@@ -1,0 +1,122 @@
+/*
+ * Copyright(c) 2023 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+using System;
+using System.Runtime.InteropServices;
+using System.ComponentModel;
+using Tizen.NUI;
+using Tizen.NUI.Binding;
+using Tizen.NUI.BaseComponents;
+
+namespace Tizen.NUI.Scene3D
+{
+    /// <summary>
+    /// This class is to define 3D Light source.
+    /// Currently this Light class supports Directional Light that lights every position from the same direction. (e.g, Sun light)
+    /// If a Light object is added on SceneView, the 3D objects in the SceneView are shined the Light.
+    /// NUI Scene3D limits the maximum enabled light count per each SceneView.
+    /// Currently the maximum number is set to 5, and it can be retrieved by using <see cref="MaximumActivatedLightCount"/>.
+    /// If more than 5 enabled Light objects are added on SceneView, SceneView turns on only 5 lights in the order the lights were added.
+    /// This Light can be added to SceneView directly but also it can be added on other View.
+    /// When a parent actor is added to a SceneView, its Light behaves in the SceneView the same as if it were added directly to the SceneView.
+    /// </summary>
+    /// <remarks>
+    /// Light inherits View, so Light color and direction can be controlled by setting View's <see cref="View.Color"/> and <see cref="View.Orientation"/> Property.
+    /// <see cref="View.LookAt(Vector3, Vector3, Vector3, Vector3)"/> method can be used to set light direction easily.
+    /// </remarks>
+    /// <remarks>
+    /// Default light direction is to Z-axis
+    /// </remarks>
+    /// <example>
+    /// <code>
+    /// SceneView sceneView = new SceneView();
+    /// Light light = new Light();
+    /// light.Color = Color.Brown;
+    /// light.LookAt(new Vector3(1.0f, 1.0f, 1.0f));
+    /// sceneView.Add(light);
+    /// </code>
+    /// </example>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class Light : View
+    {
+        internal Light(global::System.IntPtr cPtr, bool cMemoryOwn) : base(cPtr, cMemoryOwn)
+        {
+        }
+
+        /// <summary>
+        /// Create an initialized Light.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Light() : this(Interop.Light.New(), true)
+        {
+            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+        }
+
+        /// <summary>
+        /// Copy constructor.
+        /// </summary>
+        /// <param name="light">Source object to copy.</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Light(Light light) : this(Interop.Light.NewLight(Light.getCPtr(light)), true)
+        {
+            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+        }
+
+        /// <summary>
+        /// Maximum Activated Light Count.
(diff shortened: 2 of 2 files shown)


src/Tizen.NUI.Scene3D/src/internal/Interop/Interop.Light.cs
```diff
@@ -0,0 +1,52 @@
...
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_Light_Assign")]
+            public static extern global::System.IntPtr LightAssign(global::System.Runtime.InteropServices.HandleRef light, global::System.Runtime.InteropServices.HandleRef sourceLight);
+
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_Light_DownCast")]
+            public static extern global::System.IntPtr LightDownCast(global::System.Runtime.InteropServices.HandleRef light);
+
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_Light_Enable")]
+            public static extern void Enable(global::System.Runtime.InteropServices.HandleRef light, bool enable);
+
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_Light_IsEnabled")]
```
rabbitfor, line 45: I think it needs to use `MarshalAs` to treat boolean typs as U1 size
```C#
[return: MarshalAs(UnmanagedType.U1)]
```

rabbitfor, line 45: Binding for return type `bool` need to care about the size.
If it is returning boolean type in the csharp binder, the c# code need to make it 1byte size using `MarshalAs`.
```C#
[return: MarshalAs(UnmanagedType.U1)]
```
If it is returning 4byte type in the csharp binder, the current c# code has no problem.

Please refer the official guide [here](https://learn.microsoft.com/en-us/dotnet/api/system.boolean?view=net-7.0#booleans-and-interop)

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Runtime.InteropServices;
+using System.ComponentModel;
+using Tizen.NUI;
+using Tizen.NUI.Binding;
+using Tizen.NUI.BaseComponents;
+
+namespace Tizen.NUI.Scene3D
+{
+    /// <summary>
+    /// </summary>
```
rabbitfor, line 28: Could you fill the description?

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// <summary>
+        /// The Light is turned on when the Light object is added on SceneView and enabled.
+        /// </summary>
+        /// <remarks>
+        /// SceneView can turn on only up to maximum enabled light count that can be retrieved by GetMaximumEnabledLightCount().
+        /// </remarks>
+        /// <param name="enable">True to make this Light enable.</param>
+        // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public void Enable(bool enable)
```
rabbitfor, line 100: In many cases, a C# class often provide a `IsEnabled` property instead of method to setting enabled value.
For example,
```C#
public bool IsEnabled
{
    get => Interop.Light.IsEnabled(SwigCPtr);
    set => Interop.Light.Enable(SwigCPtr, value);
}
```
How about combine `Enable()` and `IsEnabled()` into a property `IsEnabled` like the example?

test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/CaptureTest.cs
```diff
@@ -11,66 +12,127 @@ public class CaptureTest : IExample
...
+            textButton.Position = new Position(300, 0);
+            textButton.TextLabel.Text = "NUI text button";
+            window.Add(textButton);
+
+            //Capture
+            capture = new Capture();
+            capture.Finished += onCaptureFinished;
+
+
+//            log.Debug(tag, $"Activate(): start \n");
```
rabbitfor, line 48: Could you clean up the commented out code?

rabbitfor:
Could you check the sample code `CaptureTest.cs` you made change in this patch? It seems that it does not include `Light` related code.

src/Tizen.NUI.Scene3D/src/internal/Interop/Interop.Light.cs
bshsqa: Thank you master of NUI.

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
bshsqa: Of course, I missed to fill it.

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
bshsqa: In our case, Light inherits View. And View already has IsEnabled property.
So, in the new patch, we changed the name to EnableLight and IsEnabledLight.

However I think this not the best way.
How do you think How we can fix it?

bshsqa: IsLightEnabled property?
or 
current EnableLight and IsEnabledLight methods?
or
new other solution..?

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+        }
+
+        /// <summary>
+        /// Assignment operator.
+        /// </summary>
+        /// <param name="light">Source object to be assigned.</param>
+        /// <returns>Reference to this.</returns>
+        internal Light Assign(Light light)
```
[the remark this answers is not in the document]
bshsqa: Thank you for your detail.
I removed it.

Seoyeon2Kim (commented):
There are some commented code in the sample, `CaptureTest.cs`.
Could you add a reason why you commented them and write down a description of this PR?

rabbitfor:
> Could you check the sample code `CaptureTest.cs` you made change in this patch? It seems that it does not include `Light` related code.

Please check this comment. It seems that `CaptureTest.cs` is uploaded by mistake.

src/Tizen.NUI.Scene3D/src/internal/Interop/Interop.Light.cs
rabbitfor: :)

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
rabbitfor: Great! The description includes details that can help users to understand quickly.

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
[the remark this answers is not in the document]
rabbitfor: Thank you

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
rabbitfor: IMHO the class name itself is `Light`, so the `Light.IsLightEnabled` feels a bit redundant.

By the way, if the word `enable` here means that the on/off state of the light, what is the difference with belows?
(They are inherited from the `View`)
* Light.Visibility
* Light.Show(), Light.Hide()
* Light.Color = Color.Transparent;

bshsqa: Yes Enable of light is to turn on/off.
However...
Light is not a bulb. So Light has not concept of visibility.
In NUI/DALi, Visibility, Show, and Hide are used to define the View will be rendered or not.
But, the Light is not renderable object. And it is just a conceptual item to change other object's visual result.
And also Light is not renderable object it's color is not includes transparency.
To be honest, to use new property of LightColor is more precise, but because we need to use View to control transform so we don't want to make the redundant too.
And, if we set Color to Black (same as 24 bit transparent), it looks same result with IsLightEnable is false.
But for the case App need to store original color value to re-turn on the light. and it looks unhappy.
We can simply imagine real world's bulb color and turn on/off switch.


So........ Which naming is better?ㅠㅠㅠㅠㅠ

bshsqa:
> There are some commented code in the sample, `CaptureTest.cs`. Could you add a reason why you commented them and write down a description of this PR?

I need to remove the file from this PR

src/Tizen.NUI.Scene3D/src/internal/Interop/Interop.Light.cs
```diff
@@ -0,0 +1,50 @@
...
+    internal static partial class Interop
+    {
+        internal static partial class Light
+        {
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_Light_New_SWIG_0")]
+            public static extern global::System.IntPtr LightNew();
+
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_new_Light_SWIG_0")]
+            public static extern global::System.IntPtr NewLight();
+
```
Seoyeon2Kim, line 29: For me, `LightNew()` and `NewLight()` make a little confusing.
Could you explain them shortly?
Or how about changing `LightNew()` to just `New()` because the class name is already `Light`.

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
rabbitfor: It seems that the `View` does not perfect fit to be a base class of the `Light` but I guess it would have been your best option in the current NUI struture since there is no class to express transformable. 
Anyway, if the `IsLightEnabled` is the name from DALi, it would be the best :)

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.ComponentModel;
+using Tizen.NUI;
+using Tizen.NUI.Binding;
+using Tizen.NUI.BaseComponents;
+
+namespace Tizen.NUI.Scene3D
+{
+    /// <summary>
+    /// This class is to define 3D Light source.
+    /// Currently this Light class supports Directional Light that lits every position from the same direction. (e.g, Sun light)
```
Seoyeon2Kim, line 29: Minor :
If the verb, `light` is present,
```suggestion
    /// Currently this Light class supports Directional Light that lights every position from the same direction. (e.g, Sun light)
```
Or if the `light` is past,
```suggestion
    /// Currently this Light class supports Directional Light that lit every position from the same direction. (e.g, Sun light)
```

src/Tizen.NUI.Scene3D/src/internal/Interop/Interop.Light.cs
bshsqa: Very good for me.
It was one of my confusing too, because every other binder uses same manner.
I will fix it only in this patch.
But we can make better modification for all the other class in next patch.

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
bshsqa: You have hawk eye.
Because of my short english skill, how do you think which is clear for the users?

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
```diff
@@ -0,0 +1,127 @@
...
+    /// Default light direction is to Z-axis
+    /// </remarks>
+    /// <example>
+    /// <code>
+    /// SceneView sceneView = new SceneView();
+    /// Light light = new Light();
+    /// light.Color = Color.Brown;
+    /// light.LookAt(new Vector3(1.0f, 1.0f, 1.0f));
+    /// sceneView.Add(light);
+    /// </code>
```
Seoyeon2Kim, line 51: Adding code is great for other developers :)

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        /// <summary>
+        /// Checks whether this light is enabled or not.
+        /// </summary>
+        /// <returns>
+        /// True if this light is enabled.
+        /// </returns>
+        // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool IsLightEnabled()
```
Seoyeon2Kim, line 120: Just my curiosity :
Why do you not make Enabled property for LightEnabled, but separate them into the functions?
I think they can be
```
public bool Enabled
{ get; set; }
```

Seoyeon2Kim: Oh regarding this, @rabbitfor  already checked it ;)
Ignore my comment, please.

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
Seoyeon2Kim: Agree. :)
`IsLightEnabled` looks fine.

src/Tizen.NUI.Scene3D/src/internal/Interop/Interop.Light.cs
Seoyeon2Kim: OH :(
I'm so sorry other binder files are used like this...
Yes, we can make them in other patch!

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
bshsqa: I fixed it to Property

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
Seoyeon2Kim: Hawk eye kkk :smiley_cat: 
I think all descriptions of this class are written down as present tense, so `lights` is more suitable for me :)

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
bshsqa: Thank you I changed it to the property.

rabbitfor (approved):

Seoyeon2Kim (approved):
LGTM~

src/Tizen.NUI.Scene3D/src/public/Light/Light.cs
bshsqa: Thank you English expert.

PR #5340, opened 2023-06-16
[NUI] Skip when the View.BackgroundImage is set with the same value as previous one

dongsug-song:
### Description of Change ###
[NUI] Skip when the View.BackgroundImage is set with the same value as previous one

### API Changes ###
none


src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
```diff
@@ -215,6 +215,12 @@ public partial class View
         public static readonly BindableProperty BackgroundImageProperty = BindableProperty.Create(nameof(BackgroundImage), typeof(string), typeof(View), default(string),
             propertyChanged: (bindable, oldValue, newValue) =>
             {
+                if (oldValue == newValue || String.Equals(oldValue, newValue))
```
jaehyun0cho, line 218: it seems that String.Equals(oldValue, newValue) is enough

dongsug-song: thank you for review. I have changed as following your comment.

PR #5343, opened 2023-06-19
[NUI] Make subclass of ImageView can use ResourceUrl normally.

hinohie:
If we skip to set ImageVisualType, or set ImageVisualType as Image, we will try to find best visual by it's file format codes. But previously, we fix the type of visual.
So, instead of some special case like URLs, just make we use Image type.

And also, make _resourceUrl latest value if we success to get url from visual.
Now we can store _resourceUrl even we update the `Image` property.


dongsug-song (approved):
approved but other members's reviews seem to be required

PR #5344, opened 2023-06-20
[NUI] Fix some svace issues.

huayongxu:
### Description of Change ###

WID:56638093 An overflow in the arithmetic expression sizeof(IntPtr) * textures.Count which is used in Marshal.AllocHGlobal(sizeof(IntPtr) * textures.Count) may occur. Please use checked arithmetic to throw IntegerOverflowException in case of overflow instead of possible memory damage

WID:56646651 The iteration number in loop with condition i < _parts.Count - 1 is off by one

WID:56638091 An overflow in the arithmetic expression structSize * vertices.Length which is used in Marshal.AllocHGlobal(structSize * vertices.Length) may occur. Please use checked arithmetic to throw IntegerOverflowException in case of overflow instead of possible memory damage

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Rendering/VertexBuffer.cs
+++ b/src/Tizen.NUI/src/public/Rendering/VertexBuffer.cs
@@ -49,12 +49,12 @@ namespace Tizen.NUI
         /// This function expects an array of structures with the same format that was given in the construction.
         /// </summary>
         /// <param name="vertices">The vertex data that will be copied to the buffer.</param>
-        /// <exception cref="ArgumentNullException"> Thrown when vertices is null. </exception>
+        /// <exception cref="ArgumentNullException"> Thrown when vertices is null or length of the vertices is 0. </exception>
         /// <since_tizen> 8 </since_tizen>
 
         public void SetData<VertexType>(VertexType[] vertices) where VertexType : struct
         {
-            if (null == vertices)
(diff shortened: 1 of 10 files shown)


wb0716 (approved):

xuelian-bai (approved):
looks good to me

elishateng (approved):
LGTM

hyue7 (approved):
looks good

src/Tizen.NUI/src/public/Rendering/VertexBuffer.cs
```diff
@@ -54,7 +54,7 @@ internal VertexBuffer(global::System.IntPtr cPtr, bool cMemoryOwn) : base(cPtr,
 
         public void SetData<VertexType>(VertexType[] vertices) where VertexType : struct
         {
-            if (null == vertices)
+            if (null == vertices || vertices.Length == 0)
```
hinohie, line 57: Minor. Could you please add comments about exception (line 52) that we can throw exception if certices length is zero?

hinohie (approved):

src/Tizen.NUI/src/public/Rendering/VertexBuffer.cs
huayongxu: Done.
Thanks.

PR #5349, opened 2023-06-21
[NUI][API10] Fix a SVACE issue.

huayongxu:
### Description of Change ###

WID:61196553 [STATISTICAL] Value root.Layout, which is result of method invocation with possible null return value, is dereferenced in member access expression root.Layout.MeasuredHeight

### API Changes ###

 - ACR:


zg2nets (approved):
lgtm

dongsug-song (approved):
approved but other members' reviews are required

PR #5351, opened 2023-06-22
[NUI] Clear old resources cached marker data

hinohie:
After overhead reducetion patch #5280 merged, Some side effect occured.

Previously we always ask some properties to dali side visual. But now we return cached value in NUI.

Since URL property clear-up all properties, to fit previous logic, let we also clean-up cached properties. (Actually, mark only for this case)


dongsug-song (approved):
approved

Seoyeon2Kim (approved):
LGTM~
It's really good at gathering them together.

PR #5352, opened 2023-06-22
[NUI] Fix application argv issue

sparrow74:
Fix Application Argv issue in UI Thread.
The input param is missed in this case. This PR is to fix it.


hinohie:
Amazing bug.. o0o... Looks good to me.

PR #5356, opened 2023-06-26
[NUI] Exception if someone change disposed ImageView's property

hinohie:
Since we don't check validation of ImageView itself during change ImageView properties, it will not throw any exceptions. and it will be crashend when we ProcessOnceEvent invoked.

Normal View property directly access to DALi side. But ImageView property access to DALi side lazy.

Now this throw make app developer easy to find error codes.


dongsug-song (approved):
approved

PR #5359, opened 2023-06-26
[MediaPlayer] Add new API to set video codec type

hsgwon:
### Description of Change ###

Add new API to set video codec type

### API Changes ###

 - TCSACR-541


src/Tizen.Multimedia.MediaPlayer/Player/Player.Properties.cs
```diff
@@ -740,6 +736,48 @@ public CodecType AudioCodecType
             }
         }
 
+        /// <summary>
+        /// Gets or sets the default codec type of the video decoder.
```
wiertel, line 740: I'm not sure what "default" means here, why it is added. But it's copied from AudioCodecType property, so let's keep it. Still, do you know in what sense it is a "default" codec? Why not just "the (current) codec of decoder"?

hsgwon: It means, when user get `VideoCodecType` property for the first time without setting `VideoCodectype`, user will get `default` video codec type set by system. But after set `VideoCodecType`, user will get video codec type set by user. 
So I'll remove the word `default` from here.

PR #5366, opened 2023-06-30
[NUI] Disable ThemeManager preload

rabbitfor:
### Description of Change ###
Because of the dbus issue from tizen theme manager, this patch removed NUI ThemeManager from the preload.

### API Changes ###

 - ACR:


dongsug-song:
@rabbitfor 
Hello,
please add the description. 
and I am just curious why the code itself is not deleted.
I thought if the theme manager is removed in the preload scheme then it seemed to be OK to be removed.
could you explain why?
thank you.

JoogabYun (approved):

dongsug-song (approved):
thank you for explanation.
approved

PR #5367, opened 2023-07-04
[NUI] Partial update feature enable/disable for window

hinohie:
Sets whether the scene will update partial area or full area.

If `PartialUpdate` is true, window will update & render partial area.
If false, full area updated.

Relative dali patches :

https://review.tizen.org/gerrit/c/platform/core/uifw/dali-core/+/294634
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-adaptor/+/294635
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/294637

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.Window.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.Window.cs
@@ -176,6 +176,13 @@ namespace Tizen.NUI
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Window_GetPosition")]
             public static extern global::System.IntPtr GetPosition(global::System.Runtime.InteropServices.HandleRef jarg1);
 
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Window_SetPartialUpdateEnabled")]
+            public static extern void SetPartialUpdateEnabled(global::System.Runtime.InteropServices.HandleRef window, bool enabled);
+
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Window_IsPartialUpdateEnabled")]
+            [return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.U1)]
+            public static extern bool IsPartialUpdateEnabled(global::System.Runtime.InteropServices.HandleRef window);
+
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Window_SetTransparency")]
             public static extern global::System.IntPtr SetTransparency(global::System.Runtime.InteropServices.HandleRef jarg1, bool jarg2);
 
--- a/src/Tizen.NUI/src/public/Window/Window.cs
+++ b/src/Tizen.NUI/src/public/Window/Window.cs
@@ -566,6 +566,24 @@ namespace Tizen.NUI
             }
         }
 
+        /// <summary>
+        /// Gets or sets whether the window will update partial area or full area.
+        /// If this value is true, window will update and render partial area.
+        /// If false, full area updated.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool PartialUpdate
+        {
+            get
+            {
+                return IsPartialUpdate();
+            }
+            set
+            {
+                SetPartialUpdate(value);
+            }
+        }
+
         internal static Vector4 DEFAULT_BACKGROUND_COLOR
         {
             get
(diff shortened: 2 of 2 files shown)


rabbitfor:
Could you share related csharp-binder patch?

src/Tizen.NUI/src/internal/Interop/Interop.Window.cs
```diff
@@ -311,6 +311,13 @@ internal static partial class Window
             [return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.U1)]
             public static extern bool IsWindowRotating(global::System.Runtime.InteropServices.HandleRef window);
 
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Window_SetPartialUpdateEnabled")]
+            public static extern bool SetPartialUpdateEnabled(global::System.Runtime.InteropServices.HandleRef window, bool enabled);
```
dongsug-song, line 315: please add ' [return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.U1)]' this like the code below.

src/Tizen.NUI/src/public/Window/Window.cs
```diff
@@ -1892,6 +1908,29 @@ public bool IsWindowRotating()
             return ret;
         }
 
+        /// <summary>
+        /// Set the window use partial update or not.
+        /// </summary>
+        /// <param name="enabled">If window enable partial update or disable.</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public void SetPartialUpdateEnabled(bool enabled)
```
rabbitfor, line 1916: You already defined a property for this : `bool PartialUpdate`
Is there any reason to define it as methods too? It seems redundant.

dongsug-song (commented):
please add description~

src/Tizen.NUI/src/internal/Interop/Interop.Window.cs
rabbitfor, line 315: I'm afraid it's a `void` type method in csharp-binder.

hinohie: That was my mistake. I make this setter as void type.

hinohie: Oh.. It was totally my mistake!! Thank you for your points. Fix done.

src/Tizen.NUI/src/public/Window/Window.cs
hinohie: Hi. There are several reason about this API.

1. Most of Windows API already implements similar patterns
2. I just want to 'hide' the Interop API using for `Property` level.
3. For product release sight, `hidden`-to-`public` is more easier than `internal`-to-`public` .

hinohie:
I write more descriptions on code + this PR.

And also write relative dali patches about it.

src/Tizen.NUI/src/public/Window/Window.cs
rabbitfor: I agree with 2&3.

For 1, it seems not common to have both public property and methods for the same functionality.
And even if it is true, I think they have at least same API names
```C#
bool PartiaUpdateEnabled;
void SetPartiaUpdateEnabled(bool);
bool IsPartiaUpdateEnabled();
```
Or
```C#
bool PartiaUpdate;
void SetPartiaUpdate(bool);
bool IsPartiaUpdate();
```

dongsug-song (approved):
I checked the patch had been fixed normally.
approved.

src/Tizen.NUI/src/public/Window/Window.cs
rabbitfor: Let me add one more thing.
For property name, it is better to have a prefix `Is` for a suffix `Enabled` as a set.
`bool IsPartialUpdateEnabled`
For example,
https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.visualelement.isenabled?view=net-maui-7.0
And NUI has simalr property name patterns.

hinohie: Append keyword 'Enabled' at Property feel bad to me. I make API name as `SetPartialUpdate`.

rabbitfor: I understand your feelings to a specific name for a property.  :smile:
NUI already have properties named `IsXXXEnabled` and also MAUI has some. You can give more chance to `Enabled` suffix.

rabbitfor: Anyway, It is not a big deal but It's still hard to find a case having both public property and method in NUI for me.
Even if there is a case I believe it was not intentional.
I found a few cases but their methods are `internal` normally. In that case, it seems that they don't have a plan to be a `public`. (And I think they need to be a private)

Do you really need both `property` and `method`?

JoogabYun (approved):

src/Tizen.NUI/src/public/Window/Window.cs
rabbitfor: If you don't want a property, you can remove property and provide only methods anyway.

PR #5368, opened 2023-07-04
[NUI] Add InterceptWheelEvent

JoogabYun:
### Description of Change ###

The Wheel event calls the WheelEvent callback by going back from the last child actor to the parent via hitTest. InterceptWheelEvent checks the wheel event in the parent first. Returning false from interceptWheelEvent allows child actors to receive WheelEvents. If it returns true, the actor will receive a WheelEvent.

for example
```c++
   View parent = new View();
   View child = new View();
   parent.Add(child);
   child.WheelEvent += childFunctor;
   parent.WheelEvent += parentFunctor;
```

The callbacks are called in the order childFunctor -> parentFunctor.

If you connect InterceptWheelEvent to parentActor.
```c++
   parent.InterceptWheelEvent += interceptFunctor;
```

When interceptFunctor returns false, it is called in the same order childFunctor -> parentFunctor. If intereptFunctor returns true, it means that the WheelEvent was intercepted. So the child actor will not be able to receive wheel events. Only the parentFunctor is called.

refer :
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-core/+/295232/
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/295233/

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.ActorSignal.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.ActorSignal.cs
@@ -52,6 +52,12 @@ namespace Tizen.NUI
             [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Actor_WheelEventSignal_Disconnect")]
             public static extern void WheelEventDisconnect(HandleRef actor, HandleRef handler);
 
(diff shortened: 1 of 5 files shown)


hinohie (approved):
I love that this sample use LazyFeedHover();

PR #5374, opened 2023-07-05
[NUI.Scene3D] Make MotionData use Native API, instead of C#

hinohie:
Since we allow to use MotionData feature in dali-side, let we use it.

Now we can load bvh and blendshape animation from file asyncronously.

required dali patch : https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/295288


src/Tizen.NUI.Scene3D/src/public/ModelMotion/MotionData.cs
```diff
@@ -16,109 +16,384 @@
...
+        {
+            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+        }
+
+        /// <summary>
+        /// Create an initialized motion data with duration.
+        /// </summary>
+        /// <param name="durationSeconds">Duration of this motion data when it be generated as Animation.</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public MotionData(float durationSeconds) : this(Interop.MotionData.MotionDataNew(durationSeconds), true)
```
bshsqa, line 107: NUI handles animation duration with integer value denotes milliseconds.
Please changes float duration to integer.

src/Tizen.NUI.Scene3D/src/public/ModelMotion/MotionData.cs
```diff
@@ -16,109 +16,384 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Get or set the duration of this motion data as seconds.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        [Obsolete("Do not use this MotionValues property directly. Use Add and GetIndex, GetValue instead.")]
-        public List<(MotionIndex, MotionValue)> MotionValues
+        public float Duration
```
bshsqa, line 142: How do you think which is better Duration or DurationMillisecond?
NUI usually uses later naming in animation.

src/Tizen.NUI.Scene3D/src/public/ModelMotion/MotionData.cs
```diff
@@ -16,109 +16,384 @@
...
+        /// If scale is null, default use as Vector3.ONE
+        /// After load completed, <see cref="LoadCompleted"/> event will be invoked.
         /// </summary>
+        /// <param name="bvhFilename">Name of bvh format file.</param>
+        /// <param name="scale">Scale value of bvh animation match with model.</param>
+        /// <param name="synchronousLoad">Load synchronously or not. Default is async load.</param>
+        // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public void Clear()
+        public void LoadBvhAnimation(string bvhFilename, Vector3 scale = null, bool synchronousLoad = false)
```
bshsqa, line 241: This naming is quite good because Bvh is very common format.
But bvh is one of format type for motion capture data.
So, when we support other motion capture format in any future than we need to make new method for the format.
So how about to use LoadMotionCaptureAnimation as name?
Then we can reuse the method when we support other formats.

bshsqa:
I think you spend big effort for this patch.
Overall, this patch is great. So I leave only some of minor comments.

src/Tizen.NUI.Scene3D/src/public/ModelMotion/MotionData.cs
hinohie: Oh... That was good point!
Match the experience with Tizen.NUI.Animation is important. Fix done.

src/Tizen.NUI.Scene3D/src/public/ModelMotion/MotionData.cs
hinohie: The name of property use still `Duration`. I'll keep current logic.

src/Tizen.NUI.Scene3D/src/public/ModelMotion/MotionData.cs
hinohie: Your point looks good to me! `LoadBvh` was too file specific API so it will be hard ti apend more jobs ib future. I make API name as `LoadMotionCaptureAnimation' niw.

bshsqa (approved):
Looks good to me!

src/Tizen.NUI.Scene3D/src/internal/Interop/Interop.MotionData.cs
```diff
@@ -0,0 +1,87 @@
...
+ *
+ */
+
+namespace Tizen.NUI.Scene3D
+{
+    internal static partial class Interop
+    {
+        internal static partial class MotionData
+        {
+            [global::System.Runtime.InteropServices.DllImport(Libraries.Scene3D, EntryPoint = "CSharp_Dali_MotionData_New_SWIG_0")]
```
dongsug-song, line 24: very great patch! two thumbs up!
just minor comment for next modification.

https://github.com/Samsung/TizenFX/blob/master/src/Tizen.NUI/src/internal/Interop/Interop.Accessibility.cs
like above, by `using global::System;  using System.Runtime.InteropServices;` we can make it simpler. ^^;

you can fix it from next time.~

dongsug-song (approved):
very good for description added, sample code added. 
approved.

PR #5375, opened 2023-07-06
[NUI] Delete the used window in widget

tscholb:
Delete the used window so that the window can be deleted normally when deleting the widget. this is for memory leak issue.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Widget/WidgetImpl.cs
+++ b/src/Tizen.NUI/src/internal/Widget/WidgetImpl.cs
@@ -35,6 +35,21 @@ namespace Tizen.NUI
             //throw new global::System.MethodAccessException("C++ destructor does not have public access");
         }
 
+        protected override void Dispose(DisposeTypes type)
+        {
+            if (disposed)
+            {
+                return;
+            }
+
+            if (type == DisposeTypes.Explicit)
+            {
+                SwigDirectorDisconnect();
+            }
+
+            base.Dispose(type);
+        }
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI/src/internal/Widget/WidgetImpl.cs
```diff
@@ -312,7 +312,10 @@ private void SwigDirectorOnCreate(string contentInfo, global::System.IntPtr wind
             {
                 ret = new Window(window, true);
             }
+
             OnCreate(contentInfo, ret);
+            ret.Dispose();
```
hinohie, line 317: To safety way, we must 'not' call Dispose here. I don't know detail logic for it but.. If the `ret` Window created by user-side, it can make unknown disposed.

Please check what I did

https://github.com/Samsung/TizenFX/pull/5374/files#diff-8cd06cf3f9b6472b4b266dd90e5e950851d14d7b85054d4f227bd5e1601e871fR196

hinohie: Or, see `Tizen.NUI.Extensions.GetInstanceSafely<Window>` API usage. It will be helpful to catch the memory leak.

hinohie:
I think #4123 patch is helpful to you.

hinohie (approved):
Basic flow looks good to me. I hope youd better check the change of sample is intent or not

PR #5380, opened 2023-07-11
[NUI] Fix not to propagate gesture from scrim

jaehyun0cho:
Scrim should block all touch and gesture events.
Therefore, gesture is also not propagated from scrim.

### API Changes ###

 - ACR:


dongsug-song (approved):
same as https://github.com/Samsung/TizenFX/pull/5381
looks good to me

PR #5386, opened 2023-07-11
[NUI][AT-SPI] Set focus in OnAccessibilityActivated()

aswigon:
### Description of Change ###

This aligns the behaviour of `Components` (derived from `CustomView`), and other controls, with that of `BaseComponents` (and other controls with `ViewAccessibilityMode.Default`, backed by the C++ accessibility implementation), which set the keyboard focus on activation. Please note that in the latter case, the implementation of `View.OnAccessibilityActivated` is unused, and the unrelated method (albeit bearing the same name) `Control::OnAccessibilityActivated` is called:

```cpp
bool Control::OnAccessibilityActivated()
{
  if(Toolkit::KeyboardFocusManager::Get().SetCurrentFocusActor(Self()))
  {
    return OnKeyboardEnter();
  }
  return false;
}
```

So, in other words, this patch makes C# controls behave in a similar manner, i.e. set the keyboard focus on accessibility activation.

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/CustomView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/CustomView.cs
@@ -501,6 +501,11 @@ namespace Tizen.NUI.BaseComponents
         [EditorBrowsable(EditorBrowsableState.Never)]
         protected override bool OnAccessibilityActivated()
         {
+            if (!base.OnAccessibilityActivated())
+            {
+                return false;
+            }
+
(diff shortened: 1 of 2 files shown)


loleksak (approved):
LGTM, now its consistent with other toolkits

Seoyeon2Kim (approved):
I think it's Okay because the similar implementation is already applied to dali-toolkit one. :)

JoogabYun (approved):

PR #5391, opened 2023-07-12
[NUI] Switch camera includes current camera's Unparent()

jmmhappy:
Synchronizing DALi and NUI cameras' parent.

DALi's SelectCamera() includes unparenting current camera component, and NUI's SelectCamera() simply calls the function.
However, since they manage different objects on each side, NUI's camera is still attached to its parent.

This can be a problem in a situation such as:

// dummy is a View
dummy.Add(defaultCamera);
sceneView.Add(dummy);

sceneView.SelectCamera(1u);

sceneView.SelectCamera(0u);
dummy.Add(defaultCamera);

In NUI side, default camera was never detached from dummy but DALi's SceneView now has both dummy and default camera as its children. (The last line is ignored on NUI side.)

### DIFF
--- a/src/Tizen.NUI.Scene3D/src/public/Controls/SceneView.cs
+++ b/src/Tizen.NUI.Scene3D/src/public/Controls/SceneView.cs
@@ -354,6 +354,7 @@ namespace Tizen.NUI.Scene3D
             {
                 return;
             }
+            this.GetSelectedCamera()?.Unparent();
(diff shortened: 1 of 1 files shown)


bshsqa (approved):
Simple and good solution.

hinohie (approved):
Maybe there will have some another sad problem about `GetParent()` API. But Anywayt, This PR it self looks good to me.

PR #5397, opened 2023-07-18
[WebRTC] Add Mute API for audio track

hsgwon:
### Description of Change ###

Add Mute API for audio track

### API Changes ###

 - TCSACR-542


wiertel (approved):
could be even better if the default value was mentioned in the description

wiertel (approved):

PR #5398, opened 2023-07-18
[NUI] Add VisibilityChanged event to Layer

jaehyun0cho:
To notice the visibility changes of Layer, VisibilityChanged event is added to Layer.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Common/Layer.cs
+++ b/src/Tizen.NUI/src/public/Common/Layer.cs
@@ -17,6 +17,7 @@
 using System;
 using Tizen.NUI.BaseComponents;
 using System.ComponentModel;
+using System.Runtime.InteropServices;
 
 namespace Tizen.NUI
 {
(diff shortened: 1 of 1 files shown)


hinohie (approved):
Should'nt we have to Disconnect at `Dispose()`??

src/Tizen.NUI/src/public/Common/Layer.cs
```diff
@@ -611,6 +616,89 @@ public void SetHoverConsumed(bool consume)
...
+            }
+        }
+
+        /// <summary>
+        /// Event arguments of visibility changed.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public class VisibilityChangedEventArgs : EventArgs
+        {
+            private Layer _layer;
```
huiyueun, line 658: NUI 네이밍 룰과 맞지 않는 사용인거 같아요 '_layer'

jaehyun0cho: Thank you for your comment!
NUI naming rule is applied.

PR #5399, opened 2023-07-18
[NUI] Add VisibilityChanged event to Layer

jaehyun0cho:
To notice the visibility changes of Layer, VisibilityChanged event is added to Layer.

### API Changes ###

 - ACR:


hinohie (approved):

hinohie:
Should'nt we have to Disconnect at `Dispose()`??

src/Tizen.NUI/src/public/Common/Layer.cs
```diff
@@ -609,6 +614,78 @@ public void SetHoverConsumed(bool consume)
...
+            }
+        }
+
+        /// <summary>
+        /// Event arguments of visibility changed.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public class VisibilityChangedEventArgs : EventArgs
+        {
+            private Layer _layer;
```
huiyueun, line 653: NUI 네이밍 룰과 맞지 않는 사용인거 같아요 '_layer'

jaehyun0cho:
> Should'nt we have to Disconnect at `Dispose()`??

Thank you for your comment!
The disconnect has been added in Dispose().

src/Tizen.NUI/src/public/Common/Layer.cs
jaehyun0cho: Thank you for your comment!
NUI naming rule is applied.

huiyueun (approved):

huiyueun (approved):

PR #5411, opened 2023-07-25
[NUI] Add logic to check if GetLastTouchEvent() is null.

JoogabYun:
### Description of Change ###

[NUI] Add logic to check if GetLastTouchEvent() is null.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window/Window.cs
+++ b/src/Tizen.NUI/src/public/Window/Window.cs
@@ -1168,7 +1168,13 @@ namespace Tizen.NUI
                 internalHoverTimer = new Timer(time);
                 internalHoverTimer.Tick += (s, e) =>
                 {
-                    FeedHover();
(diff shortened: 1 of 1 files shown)


hinohie (approved):

src/Tizen.NUI/src/public/Window/Window.cs
```diff
@@ -1168,7 +1168,13 @@ public void LazyFeedHover(uint time = 48)
                 internalHoverTimer = new Timer(time);
                 internalHoverTimer.Tick += (s, e) =>
                 {
-                    FeedHover();
+                    using Touch touch = GetLastTouchEvent();
+                    if(touch != null && touch.GetPointCount() > 0)
```
huiyueun, line 1172: merge되었지만, if사이에 띄어쓰기가 안되어있어요...

PR #5417, opened 2023-07-25
[NUI] Make Action id as internal static readonly

hinohie:
Since we share the same action id for each Play / Pause / etc, we don't need to consider overwrite the action id now.


huiyueun (approved):

huiyueun:
내부에서 같은 Index를 사용하고 있었군요. 더 깔끔하게 정리된거 같습니다!

PR #5419, opened 2023-07-27
[Tizen.Log] Fix null reference exception

ppdl:
When a null string comes, replace it with empty string so that not to throw null reference exception.

### Description of Change ###

If a string parameter is null, then s.Length fails, throwing null reference exception. In addition, Encoding.GetBytes() will throw ArgumentNullException when it receives null char *. To fix it, replace all null string with empty string, which can be handled as an object.

### API Changes ###

 - ACR:


chanwoochoi (approved):
Looks good to me

PR #5437, opened 2023-08-01
[NUI] Lazy Update CornerRadius/Borderline

hinohie:
Prototype codes for Borderline / CornerRadius feature update lazy.

Since we don't need to apply background extra data multiple times, Let we collect all changeness informations + Apply it once.


dongsug-song (approved):
approved

PR #5443, opened 2023-08-02
[Camera] Add new APIs for manual focus level, product id and vendor id

hsgwon:
### Description of Change ###

Add new APIs for manual focus level, product id and vendor id

### API Changes ###

 - TCSACR-545

### DIFF
--- a/src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
+++ b/src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
@@ -114,7 +114,7 @@ namespace Tizen.Multimedia
         }
 
         /// <summary>
-        /// Gets the product id for given <param name="device"/>.
+        /// Gets the product ID for given <param name="device"/>.
         /// </summary>
         /// <exception cref="ArgumentException">Invalid enumeration.</exception>
         /// <exception cref="ObjectDisposedException">The CameraDeviceManager already has been disposed.</exception>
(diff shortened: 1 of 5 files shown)


src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
```diff
@@ -113,6 +113,36 @@ public IEnumerable<CameraDeviceInformation> SupportedDevices
             }
         }
 
+        /// <summary>
+        /// Gets the product id for given <param name="device"/>.
```
wiertel, line 117: please write "ID" in the text, same for line 132,
lines 126 and 141 can stay unchanged

src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
```diff
@@ -113,6 +113,36 @@ public IEnumerable<CameraDeviceInformation> SupportedDevices
...
+        /// </summary>
+        /// <exception cref="ArgumentException">Invalid enumeration.</exception>
+        /// <exception cref="ObjectDisposedException">The CameraDeviceManager already has been disposed.</exception>
+        /// <since_tizen> 11 </since_tizen>
+        public ushort GetProductId(CameraDevice device)
```
wiertel, line 122: please consider changing `ushort` -> `UInt16`, those are exactly the same type (one is an alias for another) so nothing needs to change apart form return type, but would be more explicit which is better for public API
same in line 137

wiertel (changes requested):

src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
hsgwon: ushort is exactly same with UInt16. I just want to keep this like other alias types.

src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
hsgwon: Fixed

src/Tizen.Multimedia.Camera/Camera/CameraDeviceManager.cs
wiertel: ok, `ushort` is not used very often, but both styles are used across TizenFX, so I will not object

wiertel (approved):

PR #5444, opened 2023-08-02
[MediaVision] Add Timestamp API for MediaVisionSource

hsgwon:
### Description of Change ###

Add Timestamp API for MediaVisionSource

### API Changes ###

 - TCSACR-546

### DIFF
--- a/src/Tizen.Multimedia.Vision/MediaVision/MediaVisionSource.cs
+++ b/src/Tizen.Multimedia.Vision/MediaVision/MediaVisionSource.cs
@@ -218,6 +218,26 @@ namespace Tizen.Multimedia.Vision
         }
 
         /// <summary>
+        /// Gets or sets the timestamp of the media source.
+        /// </summary>
+        /// <exception cref="ObjectDisposedException">The <see cref="MediaVisionSource"/> has already been disposed.</exception>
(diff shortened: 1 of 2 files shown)


src/Tizen.Multimedia.Vision/MediaVision/MediaVisionSource.cs
```diff
@@ -217,6 +217,26 @@ public ColorSpace Colorspace
             }
         }
 
+        /// <summary>
+        /// Gets or sets the timestamp of the media source.
+        /// </summary>
+        /// <exception cref="ObjectDisposedException">The <see cref="MediaVisionSource"/> has already been disposed.</exception>
+        /// <since_tizen> 11 </since_tizen>
+        public ulong Timestamp
```
wiertel, line 225: please consider changing `ulong` to .Net type `UInt64`, those are aliases of the same type but the UInt64 is more explicit which is better for public API

wiertel (approved):

src/Tizen.Multimedia.Vision/MediaVision/MediaVisionSource.cs
hsgwon: It's same type so I just want to keep alias type - uint.

PR #5446, opened 2023-08-02
[NUI] Block lazy Background

hinohie:
Since #5437 PR make visual lazy, we got some error when we try to make animation with "CornerRadius" property.

Until resolve VisualAnimation issue. let we just update visual synchronously.


taehyub (approved):
LGTM

jaehyun0cho (approved):
LGTM

PR #5455, opened 2023-08-08
[NUI.Gadget] Add a new method to load assembly

hjhun:
### Description of Change ###

The Load() method is added to load the assembly of the NUIGadget.


dongsug-song (approved):
looks good to me

PR #5460, opened 2023-08-08
[NUI] Fix issue when LottieAnimationView use DesiredSize

hinohie:
Since #5350 patch will reset previous whole informations when we use ResourceUrl.

So, If user set ResourceUrl set DesiredSize, the desired size informatins disapeared.


tscholb (approved):
LGTM :)

PR #5461, opened 2023-08-08
[NUI] Remove NUI dotnet 6.0 build warnings

hinohie:
Remove liteweights build warnings when build latest NUI

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityEvent.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityEvent.cs
@@ -94,7 +94,7 @@ namespace Tizen.NUI.BaseComponents
         /// <returns>True if equal GestureInfoType, otherwise false</returns>
         public bool Equals(GestureInfoType other)
         {
-            if ((other == null) || !this.GetType().Equals(other.GetType()))
+            if (!this.GetType().Equals(other.GetType()))
             {
                 return false;
             }
--- a/src/Tizen.NUI/src/public/Window/DefaultBorder.cs
+++ b/src/Tizen.NUI/src/public/Window/DefaultBorder.cs
@@ -67,11 +67,7 @@ namespace Tizen.NUI
         private float preScale = 0;
 
         private View windowView = null;
-        private bool isWinGestures = false;
-        private Timer timer;
         private Timer overlayTimer;
-
-        private CurrentGesture currentGesture = CurrentGesture.None;
 
         private uint borderLineThickness;
         private Size2D minSize;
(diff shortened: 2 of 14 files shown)


src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityEvent.cs
```diff
@@ -94,7 +94,7 @@ public override bool Equals(System.Object obj)
         /// <returns>True if equal GestureInfoType, otherwise false</returns>
         public bool Equals(GestureInfoType other)
         {
-            if ((other == null) || !this.GetType().Equals(other.GetType()))
```
Seoyeon2Kim, line 97: Shouldn't we check `(other == null)` condition here anymore?
The original code is that the function returns false when `other` parameter is null value.

src/Tizen.NUI/src/public/Window/DefaultBorder.cs
```diff
@@ -85,13 +81,6 @@ public class DefaultBorder : BindableObject, IDisposable, IBorderInterface
         #endregion //Events
 
         #region Enums
-        private enum CurrentGesture
```
Seoyeon2Kim, line 88: Could you explain the reason this enumeration is removed?

Seoyeon2Kim (commented):
From dotnet 6.0, it looks like that `IntPtr` should be compared with `IntPtr.Zero`, not `null` value!

src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityEvent.cs
hinohie: Since `GestureInfoType` is struct, and ot `INullable`, it cannot be null. (This is what compiler said to me). I just agree with it so remove this null check.

src/Tizen.NUI/src/public/Window/DefaultBorder.cs
hinohie: This is private enum. And the private member value `geustureState` never be used in this code. (Compiler said to me)
It is private so we dont need to worry about some other stranger use this enum anywhere. So I think we can remove it.

src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityEvent.cs
Seoyeon2Kim: Got it. Thank you for explaining it :)

src/Tizen.NUI/src/public/Window/DefaultBorder.cs
Seoyeon2Kim: Agree. 
@JoogabYun , could you check it once? Can we remove this private enum value because it's not used yet?
주갑님! 이 enum값 아직 쓰이는 곳은 없다는데 지워도 될까요?

JoogabYun: 네 초기에 들어갔다가 필요없어진  enum인데 아직 남아있었네요 지워도 됩니다.

JoogabYun (approved):

huiyueun (approved):

Seoyeon2Kim (approved):
LGTM!

PR #5475, opened 2023-08-11
[NUI] Change all CallingConvention to `Cdecl`

Seoyeon2Kim:
### Description of Change ###
- Change all the CallingConvention from `Stdcall` to `Cdecl`.
- `stdcall` is mainly used for Win32, and there's a default calling convention that `cdecl` should be used in native C/C++ programs.
   NUI is the C# GUI library, which is binding C++ library. So, it should use `cdecl` for the programs.
- In terms of security, the `cdecl` convention is "safer" because it is the caller that needs to deallocate the stack.

### API Changes ###
- N/A

### DIFF
--- a/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
+++ b/src/Tizen.NUI.Components/Controls/ScrollableBase.cs
@@ -873,7 +873,7 @@ namespace Tizen.NUI.Components
 
         private Animation scrollAnimation;
         // Declare user alpha function delegate
-        [UnmanagedFunctionPointer(CallingConvention.StdCall)]
+        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
         private delegate float UserAlphaFunctionDelegate(float progress);
         private UserAlphaFunctionDelegate customScrollAlphaFunction;
         private float velocityOfLastPan = 0.0f;
--- a/src/Tizen.NUI/src/internal/Application/ComponentApplication.cs
+++ b/src/Tizen.NUI/src/internal/Application/ComponentApplication.cs
@@ -85,7 +85,7 @@ namespace Tizen.NUI
             return ret;
         }
 
-        [UnmanagedFunctionPointer(CallingConvention.StdCall)]
+        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
         private delegate IntPtr NUIComponentApplicationCreatenativeEventCallbackDelegate();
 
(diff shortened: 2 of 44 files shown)


tscholb (approved):
LGTM!

dongsug-song:
@Seoyeon2Kim 
Thank you very much for the fix. 
I could see this fix resolves the emulator crash, but could you explain more about the difference of stdcall and cdecl?
thank you.

Seoyeon2Kim:
> @Seoyeon2Kim Thank you very much for the fix. I could see this fix resolves the emulator crash, but could you explain more about the difference of stdcall and cdecl? thank you.

Hello! Thank you for checking the emulator crash.
Normally, the biggest difference in the two calling conventions is who cleans the stack.
`StdCall` is that the **callee** cleans the stack. On the other hand, `Cdecl` is that the **caller** cleans the stack. This enables calling functions with varargs, which makes it appropriate to use for methods that accept a variable number of parameters. 
(https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.callingconvention?view=net-7.0)

Furthermore, there's a big reason we have to fix these.
`stdcall` is mainly used for Win32, and there's a default calling convention that `cdecl` should be used in native C/C++ programs.
You can see the differences in the pages below !
English) https://learn.microsoft.com/en-us/cpp/cpp/stdcall?view=msvc-170
              https://learn.microsoft.com/en-us/cpp/cpp/cdecl?view=msvc-170
Korean) https://m.blog.naver.com/work1989/221275066623

I'll add more descriptions about this in this PR!

Seoyeon2Kim:
I've checked this a little bit more, and I found that there's a code that converts `stdcall` in dali-csharp-binder inside.
https://review.tizen.org/gerrit/gitweb?p=platform/core/uifw/dali-csharp-binder.git;a=blob;f=dali-csharp-binder/common/dali-wrap.cpp;h=c311c500dd0c8de4360b2505b06b198d6b5e8925;hb=refs/heads/devel/master#l109

```
#ifndef SWIGSTDCALL
# if defined(_WIN32) || defined(__WIN32__) || defined(__CYGWIN__)
#   define SWIGSTDCALL __stdcall
# else
#   define SWIGSTDCALL
# endif
#endif
```

However, in TizenFX, it seems to have been auto-generated without considering it.

dongsug-song:
@Seoyeon2Kim 
Great!
Thank you very much for detailed explanation to make the reason of this patch clear.
thank you

PR #5484, opened 2023-08-17
[Tts] Add new API for TTS synthesized PCM

stom-hwang:
### API Changes ###

 - ACR:


urmain (approved):
LGTM~!!

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
```diff
@@ -230,6 +230,67 @@ public enum ServiceState
...
+    };
+
+    /// <summary>
+    /// Enumeration for the synthesized pcm event.
+    /// </summary>
+    /// <since_tizen> 11 </since_tizen>
+    public enum SynthesizedPcmEvent
+    {
+        /// <summary>
+        /// Event when the synthesized PCM is failed.
```
wiertel, line 274: I'm not sure if I understood the implemented behavior, but I would like the description to be changed. I assume that a series of event will be fired in response to a utterance requested. Maybe like this:
``` cs
/// The PCM synthesis failed.
/// Received initial synthesized PCM data (optionally add a word like: portion, block, fragment) for an utterance.
/// Received synthesized PCM data, not the first and not last in the stream.
/// Received the final synthesized PCM data.

**Review by wiertel — CHANGES_REQUESTED** [review]

File: src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs [hunk 2]
```diff
@@ -230,6 +230,67 @@ public enum ServiceState
         Unavailable
     };
 
+    /// <summary>
+    /// Enumeration for the playing mode of TTS.
```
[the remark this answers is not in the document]
stom-hwang: Thank you. I fixed it.

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
```diff
@@ -230,6 +230,67 @@ public enum ServiceState
...
+        RawS16 = 0,
+
+        /// <summary>
+        /// Unsigned 8-bit audio type.
+        /// </summary>
+        RawU8 = 1
+    };
+
+    /// <summary>
+    /// Enumeration for the synthesized pcm event.
```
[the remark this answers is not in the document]
stom-hwang: Thank you. I fixed it.

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
```diff
@@ -848,6 +959,34 @@ public bool IsScreenReaderOn
...
+        /// </value>
+        /// <feature>
+        /// http://tizen.org/feature/speech.synthesis
+        /// </feature>
+        /// <exception cref="InvalidOperationException">This exception can be due to an invalid state.</exception>
+        /// <exception cref="NotSupportedException">This exception can be due to TTS not supported.</exception>
+        /// <pre>
+        /// The Client must be in the <see cref="State.Created"/> state.
+        /// </pre>
+        public PlayingMode CurrentPlayingMode
```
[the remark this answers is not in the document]
stom-hwang: Thank you. I think that It is better to follow the common way. I fixed it.

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
```diff
@@ -664,6 +728,53 @@ public TtsClient()
             }
         }
 
+        /// <summary>
+        /// Event to be invoked when the synthesized pcm data is come from the engine.
```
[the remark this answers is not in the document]
stom-hwang: Thank you. I fixed it.

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
stom-hwang: Thank you. I think you understand exactly, so I fixed it as you mentioned.

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
```diff
@@ -848,6 +959,34 @@ public bool IsScreenReaderOn
...
+        /// </value>
+        /// <feature>
+        /// http://tizen.org/feature/speech.synthesis
+        /// </feature>
+        /// <exception cref="InvalidOperationException">This exception can be due to an invalid state.</exception>
+        /// <exception cref="NotSupportedException">This exception can be due to TTS not supported.</exception>
+        /// <pre>
+        /// The Client must be in the <see cref="State.Created"/> state.
+        /// </pre>
+        public PlayingMode PlayingMode
```
stom-hwang, line 977: Dear, @wiertel .
I have a question about this property.
There is only a setter function for playing mode in native API, so I implemented only a setter for this property.
Is it okay to remain it or better to implement a getter?

wiertel: Having a write-only property is possible, but unusual. More fitting to cases when value is a secret, so should not be accessible. On one hand we prefer limit implementing logic in C# layer, but on the other hand it should not be problematic to implement getter for that property. If the default playing mode is defined in C API we could initialize the property to that value, if there is no default defined we would need to choose one. It's even more complicated if C API requires that the playing mode is set before other operations or there are some additional limitations/requirements.
My suggestion is to implement the getter if it is not complicated.

src/Tizen.Uix.Tts/Tizen.Uix.Tts/SynthesizedPcmEventArgs.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            get;
+            internal set;
+        }
+
+        /// <summary>
+        /// The event of the PCM data.
+        /// </summary>
+        /// <since_tizen> 11 </since_tizen>
+        public SynthesizedPcmEvent SynthesizedPcmEvent
```
wiertel, line 49: I like the name changes for other properties - removing prefix that is already contained in class name is a good direction. This name could have stayed as it was, short "PcmEvent". Or could even be simplified further into something like "Kind" / "EventType". It's just a suggestion to consider.

src/Tizen.Uix.Tts/Tizen.Uix.Tts/TtsClient.cs
stom-hwang: Thank you for sharing your oppinion.
The default playing mode is defined as PlayingMode.ByServer in C API, so I think it is not hard to make getter for this property.
I will request review again after adding getter.

src/Tizen.Uix.Tts/Tizen.Uix.Tts/SynthesizedPcmEventArgs.cs
stom-hwang: I agree you. I think `EventType` is good. I will fix it and request review again.
Thank you.

PR #5522, opened 2023-08-30
[NUI] Add missing code

wonrst:
ref : 5ca58e74d71e2e91a9b0a8b549b3cffad232f6f9

Make sure that Interop return type is bool


hinohie (approved):
Looks good to me

huiyueun (approved):

PR #5526, opened 2023-09-01
[Tts] Add new internal API for TTS synthesized PCM

stom-hwang:
### Description of Change ###

Add internal method for playing mode and synthesized pcm event.

### DIFF
--- a/src/Tizen.Uix.Tts/Tizen.Uix.Tts/SynthesizedPcmEventArgs.cs
+++ b/src/Tizen.Uix.Tts/Tizen.Uix.Tts/SynthesizedPcmEventArgs.cs
@@ -1,0 +1,88 @@
+﻿/*
+* Copyright (c) 2023 Samsung Electronics Co., Ltd All Rights Reserved
+*
+* Licensed under the Apache License, Version 2.0 (the License);
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


src/Tizen.Uix.Tts/Tizen.Uix.Tts/SynthesizedPcmEventArgs.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.ComponentModel;
+
+namespace Tizen.Uix.Tts
+{
+    /// <summary>
+    /// This class holds information related to the TTS SynthesizedPcm event.
+    /// </summary>
+    /// <since_tizen> 11 </since_tizen>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class SynthesizedPcmEventArgs
```
myroot, line 27: It should inherit from System.EventArgs

stom-hwang: Thank you for checking. I think other event args in Tts do not inherit the System.EventAgrs.
So Is it okay to fix it all other classes without ACR?

urmain (approved):
LGTM~!!

syeonholic (approved):
It looks good to me.

PR #5527, opened 2023-09-04
[Applications.Common] Add TimeZoneChanged Event

hjhun:
### Description of Change ###

The TimeZoneChangedEventArgs class is added for Time Zone changed event.

### API Changes ###

 - ACR: TCSACR-555

Added:
1. CoreApplication.cs
+ void OnTimeZoneChanged(TimeZoneChangedEventArgs);
+ EventHandler<TimeZoneChangedEventArgs> TimeZoneChanged;

2. TimeZoneChangedEventArgs
+ string TimeZone { get; } // Property
+ string TimeZoneId { get; } // Property
+ void TimeZoneChangedEventArgs(string timeZone, string, timeZoneId); // Constructor

### DIFF
--- a/src/Tizen.Applications.Common/Tizen.Applications.CoreBackend/EventType.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications.CoreBackend/EventType.cs
@@ -94,6 +94,13 @@ namespace Tizen.Applications.CoreBackend
         /// <since_tizen> 6 </since_tizen>
         public static readonly EventType SuspendedStateChanged = "SuspendedStateChanged";
 
+        /// <summary>
+        /// Pre-defined event type "TimeZoneChanged".
+        /// The TimeZoneChangedEventArgs class is an event argument class for this EventType.
+        /// </summary>
+        /// <since_tizen> 11 </since_tizen>
+        public static readonly EventType TimeZoneChanged = "TimeZoneChanged";
+
         private string _typeName;
 
         /// <summary>
--- a/src/Tizen.Applications.Common/Tizen.Applications.CoreBackend/EventType.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications.CoreBackend/EventType.cs
@@ -96,7 +96,7 @@ namespace Tizen.Applications.CoreBackend
 
         /// <summary>
         /// Pre-defined event type "TimeZoneChanged".
-        /// The TimeZoneChangedEventArgs class is an event argument class for this EventType.
+        /// The <see cref="TimeZoneChangedEventArgs"/> class is an event argument class for this EventType.
         /// </summary>
         /// <since_tizen> 11 </since_tizen>
         public static readonly EventType TimeZoneChanged = "TimeZoneChanged";
--- a/src/Tizen.Applications.Common/Tizen.Applications/CoreApplication.cs
+++ b/src/Tizen.Applications.Common/Tizen.Applications/CoreApplication.cs
@@ -110,6 +110,12 @@ namespace Tizen.Applications
         public event EventHandler<DeviceOrientationEventArgs> DeviceOrientationChanged;
 
         /// <summary>
+        /// Occurs when the time zone is changed.
+        /// </summary>
+        /// <since_tizen> 11 </since_tizen>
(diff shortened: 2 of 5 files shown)


pjh9216 (approved):

src/Tizen.Applications.Common/Tizen.Applications.CoreBackend/EventType.cs
```diff
@@ -94,6 +94,13 @@ public class EventType
         /// <since_tizen> 6 </since_tizen>
         public static readonly EventType SuspendedStateChanged = "SuspendedStateChanged";
 
+        /// <summary>
+        /// Pre-defined event type "TimeZoneChanged".
+        /// The TimeZoneChangedEventArgs class is an event argument class for this EventType.
```
wiertel, line 99: consider:
```suggestion
        /// The <see cref="TimeZoneChangedEventArgs"/> class is an event argument class for this EventType.
```

src/Tizen.Applications.Common/Tizen.Applications/CoreApplication.cs
```diff
@@ -277,6 +284,17 @@ protected virtual void OnDeviceOrientationChanged(DeviceOrientationEventArgs e)
             DeviceOrientationChanged?.Invoke(this, e);
         }
 
+        /// <summary>
+        /// Overrides this method if want to handle behavior when the time zone is changed.
```
wiertel, line 288: Overrides -> Override, "if want" -> "if you want"
 I see that the same is in other On* methods

src/Tizen.Applications.Common/Tizen.Applications/TimeZoneChangedEventArgs.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// The class for the argument of the TimeZoneChanged EventHandler.
+    /// </summary>
+    /// <since_tizen> 11 </since_tizen>
+    public class TimeZoneChangedEventArgs : EventArgs
+    {
+
+        /// <summary>
+        /// Initializes TimeZoneChangedEventArgs class
+        /// </summary>
+        /// <param name="timeZone">The information of the Locale</param>
```
wiertel, line 31: missing `<param` for the second parameter, I suspect that it would be ok if the constructor was not public

src/Tizen.Applications.Common/Tizen.Applications/TimeZoneChangedEventArgs.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <since_tizen> 11 </since_tizen>
+        public TimeZoneChangedEventArgs(string timeZone, string timeZoneId)
+        {
+            TimeZone = timeZone;
+            TimeZoneId = timeZoneId;
+        }
+
+        /// <summary>
+        /// The property to get the information of the TimeZone.
+        /// </summary>
```
wiertel, line 41: please consider adding `<value>` tag in both properties with an example showing what data/format is used

wiertel (changes requested):

src/Tizen.Applications.Common/Tizen.Applications/CoreApplication.cs
hjhun: Done

src/Tizen.Applications.Common/Tizen.Applications/TimeZoneChangedEventArgs.cs
hjhun: Done

src/Tizen.Applications.Common/Tizen.Applications/TimeZoneChangedEventArgs.cs
hjhun: Done. I've added the <value> tag.

src/Tizen.Applications.Common/Tizen.Applications.CoreBackend/EventType.cs
hjhun: > consider:

Done.

wiertel (approved):

PR #5533, opened 2023-09-06
[NUI] Fix the SVACE issue.

zhouhao02:
### Description of Change ###

fix the issue : Value e.PropertyName, which is result of method invocation with possible null return value, is dereferenced in method call e.PropertyName.Equals()

### API Changes ###

 - ACR:


hyue7 (approved):
looks good

wanchao-xu (approved):
Looks good to me

wb0716 (approved):
Good for me

dongsug-song (approved):
approved

PR #5572, opened 2023-09-18
DragAndDrop : hide window event handler

taehyub:
DragAndDrop : hide window event handler

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/DragAndDrop/DragAndDrop.cs
```diff
@@ -33,6 +33,7 @@ public class DragAndDrop : BaseHandle
         public delegate void SourceEventHandler(DragSourceEventType sourceEventType);
```
Seoyeon2Kim, line 33: Could you add `[EditorBrowsable(EditorBrowsableState.Never)]` tag here and 35 line, too?

Seoyeon2Kim: Please ignore my comment :)

PR #5589, opened 2023-09-26
[NUI] Implement defaut key construct

hinohie:
Let we use default key creation API if we only need to use empty key handle.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #5590, opened 2023-09-26
[NUI] Let us not dispose cached LastEvents

hinohie:
Since we create new DALi handle internally, we also delete them.
But if we mark cMemOwn as false, we dont call deletr autimatically.

The GetLast~~Events are special cases that have there own handle, but not have memory ownership
So we need to call Delete function hardly, by our selfs.

TODO : We need to find more good way to fix this Dispose issue. Currently just remark to user that dont call Dispose for it l.


everLEEst (approved):
LGTM.

JoogabYun (approved):

PR #5595, opened 2023-09-27
[NUI] Let registry throw exception at error case

hinohie:
Let we throw exception if same object try to register.


dongsug-song (approved):
approved

huiyueun (approved):

PR #5603, opened 2023-10-08
[NUI] BaseHandle without Registry but own memory

hinohie:
There are some usecase s.t. we have to call ReleaseSwigCPtr but don't want register to Registry.

For example, PixelBuffer and PixelData dont need to be create only for main thread.


dongsug-song (approved):
approved

PR #5604, opened 2023-10-08
[NUI] Consider WeakReference resurrection at Registry.

hinohie:
Since NUI use DisposeQueue system instead of GC directly, the lifecycle of WeakReference may need to be alive more times.

If we make trackResurrection as false, WeakReference.Target become null when the item is in DisposeQueue.


dongsug-song (approved):
you solve very difficult problem which has been hardly reproduced!
very good!
approved

PR #5611, opened 2023-10-10
[NUI] Add TextFitArray

wonrst:
Add a new function to satisfy the UX that
operates TextFit by considering the PointSize and MinLineSize of the text.

For example, TextFit should be able to find the one that fits among [PointSize 24 + MinLineSize 40] or [PointSize 28 + MinLineSize 44].

But the previous TextFit only considers PointSize.

TextFitArray can solve this problem,
and this implemented it to respond to additional requirements without modifying the API.

https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-toolkit/+/299557/ 
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/299690/

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/TextConstants.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/TextConstants.cs
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Samsung Electronics Co., Ltd.
+// Copyright (c) 2023 Samsung Electronics Co., Ltd.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 
@@ -544,6 +545,77 @@ namespace Tizen.NUI.Text
         /// <returns>The hash code.</returns>
         [EditorBrowsable(EditorBrowsableState.Never)]
         public override int GetHashCode() => (Enable, MinSize, MaxSize, StepSize, FontSizeType, FontSize).GetHashCode();
+    }
+
+    /// <summary>
+    /// A struct to pass data of TextFitArray's OptionList.
+    /// </summary>
+    /// <remarks>
+    /// The TextFitArrayOption struct is used as an item to TextFitArray's OptionList. <br />
+    /// See <see cref="Tizen.NUI.Text.TextFitArray"/>.
+    /// </remarks>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1815: Override equals and operator equals on value types")]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public struct TextFitArrayOption
+    {
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="pointSize">The PointSize for TextFitArrayOption</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public TextFitArrayOption(float pointSize)
+        {
+            PointSize = pointSize;
+            MinLineSize = null;
+        }
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="pointSize">The PointSize for TextFitArrayOption</param>
+        /// <param name="minLineSize">The MinLineSize for TextFitArrayOption</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public TextFitArrayOption(float pointSize, float? minLineSize)
+        {
+            PointSize = pointSize;
+            MinLineSize = minLineSize;
+        }
+
+        /// <summary>
+        /// Point size for text fit array.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public float PointSize { get; set; }
+
+        /// <summary>
+        /// Min line size for text fit array.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public float? MinLineSize { get; set; }
+    }
+
+    /// <summary>
+    /// A struct to pass data of SetTextFitArray and GetTextFitArray methods.
+    /// </summary>
+    /// <remarks>
+    /// The TextFitArray struct is used as an argument to SetTextFitArray and GetTextFitArray methods. <br />
+    /// See <see cref="Tizen.NUI.BaseComponents.TextLabel.SetTextFitArray"/> and <see cref="Tizen.NUI.BaseComponents.TextLabel.GetTextFitArray"/>.
+    /// </remarks>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1815: Override equals and operator equals on value types")]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public struct TextFitArray
+    {
+        /// <summary>
+        /// True to enable the text fit array or false to disable (the default value is false).
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool Enable { get; set; }
+
+        /// <summary>
+        /// A List of TextFitArrayOptions.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public List<TextFitArrayOption> OptionList { get; set; }
     }
 
     /// <summary>
--- a/src/Tizen.NUI/src/public/BaseComponents/TextLabel.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/TextLabel.cs
@@ -1,5 +1,5 @@
 /*
- * Copyright(c) 2021 Samsung Electronics Co., Ltd.
+ * Copyright(c) 2023 Samsung Electronics Co., Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


dongsug-song (approved):
very good for new approach using raw array instead of PropertyMap nor PropertyArray!
approved!
however for next time, please add some sample app or sample code for users to get the way how to use easily

src/Tizen.NUI/src/public/BaseComponents/TextConstants.cs
```diff
@@ -546,6 +547,74 @@ public struct TextFit : IEquatable<TextFit>
...
+    /// The TextFitArrayOption struct is used as an item to TextFitArray's OptionList. <br />
+    /// See <see cref="Tizen.NUI.Text.TextFitArray"/>.
+    /// </remarks>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1815: Override equals and operator equals on value types")]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public struct TextFitArrayOption
+    {
+        /// <summary>
+        /// Constructor.
+        /// </summary>
```
Seoyeon2Kim, line 563: Shouldn't we add the description for the parameter?
```suggestion
        /// </summary>
        /// <param name="pointSize">The PointSize for TextFitArrayOption</param>
```

src/Tizen.NUI/src/public/BaseComponents/TextConstants.cs
```diff
@@ -546,6 +547,74 @@ public struct TextFit : IEquatable<TextFit>
...
+        {
+            PointSize = pointSize;
+            MinLineSize = null;
+        }
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public TextFitArrayOption(float pointSize, float? minLineSize)
```
Seoyeon2Kim, line 575: Just my curiosity :
Is there any reason to check null only for the type of `minLineSize`?
`pointSize` is not checked for Null?

src/Tizen.NUI/src/public/BaseComponents/TextConstants.cs
```diff
@@ -546,6 +547,74 @@ public struct TextFit : IEquatable<TextFit>
...
+    /// </remarks>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1815: Override equals and operator equals on value types")]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public struct TextFitArray
+    {
+        /// <summary>
+        /// True to enable the text fit array or false to disable (the default value is false).
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool Enable { get; set; }
```
Seoyeon2Kim, line 609: If the default value of `Enable` is false, then you can set like this :
```suggestion
        public bool Enable { get; set; } = false;
```
It's an another way to tell the default value. :)

src/Tizen.NUI/src/public/BaseComponents/TextLabel.cs
```diff
@@ -1404,6 +1404,105 @@ public TextFit GetTextFit()
...
+        /// Set TextFitArray to TextLabel. <br />
+        /// TextFitArray finds and applies the largest PointSize that fits among OptionList.
+        /// </summary>
+        /// <param name="textFitArray">The TextFitArray</param>
+        /// <remarks>
+        /// TextFitArray tries binary search by default. <br />
+        /// The precondition for TextFitArray to perform binary search is sorting in ascending order of MinLineSize. <br />
+        /// Because if MinLineSize is not sorted in ascending order, <br />
+        /// binary search cannot guarantee that it will always find the best value. <br />
+        /// In this case, search sequentially starting from the largest PointSize. <br />
```
Seoyeon2Kim, line 1417: Probably,
```suggestion
        /// In this case, the search sequentially starts from the largest PointSize. <br />
```

src/Tizen.NUI/src/public/BaseComponents/TextLabel.cs
```diff
@@ -1404,6 +1404,105 @@ public TextFit GetTextFit()
...
+        /// If TextFitArrayOption's MinLineSize is set to null or 0, <br />
+        /// TextFitArray is calculated without applying MinLineSize. <br />
+        /// If TextFitArray is enabled, TextLabel's MinLineSize property is ignored. <br />
+        /// See <see cref="Tizen.NUI.Text.TextFitArray"/> and <see cref="Tizen.NUI.Text.TextFitArrayOption"/>.
+        /// </remarks>
+        /// <example>
```
Seoyeon2Kim, line 1423: It's really great to describe the functions in detail and add the example code :+1:

src/Tizen.NUI/src/public/BaseComponents/TextLabel.cs
```diff
@@ -1404,6 +1404,105 @@ public TextFit GetTextFit()
...
+        ///
+        /// [Binary search not possible]
+        /// |            | List index  |  0 |  1 |  2 |  3 |
+        /// | OptionList | PointSize   | 24 | 28 | 32 | 48 |
+        /// |            | MinLineSize | 40 | 48 | 38 | 62 | << MinLineSize is not sorted in ascending order
+        ///                                         ^
+        /// </code>
+        /// </example>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public void SetTextFitArray(TextFitArray textFitArray)
```
Seoyeon2Kim, line 1456: Just my curiosity ! 
Could you tell me why you don't provide these features as one property but as separate Set/Get methods?

src/Tizen.NUI/src/public/BaseComponents/TextLabel.cs
```diff
@@ -1404,6 +1404,105 @@ public TextFit GetTextFit()
...
+        ///     new Tizen.NUI.Text.TextFitArrayOption(24, 40),
+        ///     new Tizen.NUI.Text.TextFitArrayOption(28, 48),
+        ///     new Tizen.NUI.Text.TextFitArrayOption(32, 56),
+        ///     new Tizen.NUI.Text.TextFitArrayOption(50, 72),
+        /// };
+        /// label.SetTextFitArray(textFitArray);
+        /// </code>
+        /// <br />
+        /// The table below shows cases where binary search is possible and where it is not possible. <br />
+        /// <code>
```
Seoyeon2Kim, line 1440: The table makes it easier for users to use and understand the function :) :bowing_woman:

Seoyeon2Kim (commented):
Overall great. :)
But, can you add an example for these features?
This PR is to add new features, so it would be nice to add a sample of how to use it.
(Of course, I know you added it to the description of `SetTextFitArray`, but.. it's up to you :+1: )

src/Tizen.NUI/src/public/BaseComponents/TextConstants.cs
wonrst: I added description for the parameter, thanks

src/Tizen.NUI/src/public/BaseComponents/TextConstants.cs
wonrst: In TextFitOption,
`PointSize` is essential, but `MinLineSize` is optional,so I used nullable.
Is there a better way?

src/Tizen.NUI/src/public/BaseComponents/TextConstants.cs
wonrst: I tried it and got this error, 
`error CS8400: Feature 'struct field initializers' is not available in C# 8.0. Please use language version 10.0 or greater.`

Do you know how to solve thiS?

src/Tizen.NUI/src/public/BaseComponents/TextLabel.cs
wonrst: Done, thanks :)

src/Tizen.NUI/src/public/BaseComponents/TextLabel.cs
wonrst: There is no problem implementing it as a property at the C# level.
But, other text styles also use the set/get method. (Although due to the usability of PropertyMap)
So I used method for consistency.

src/Tizen.NUI/src/public/BaseComponents/TextConstants.cs
Seoyeon2Kim: Oh.. I think we need to update the language version or remove it.
But, I'm not sure the reason we added it before.

```
// TizenFX/src/Tizen.NUI/Tizen.NUI.csproj
<LangVersion>8.0</LangVersion>
```

@dongsug-song (동석님) Can we update the version like this?
`<LangVersion>10.0</LangVersion>`
or
`<LangVersion>latest</LangVersion>`

src/Tizen.NUI/src/public/BaseComponents/TextConstants.cs
Seoyeon2Kim: Fast update :+1:

src/Tizen.NUI/src/public/BaseComponents/TextConstants.cs
Seoyeon2Kim: Understood.
As you mentioned, the constructor with the essential parameter exists above.
And no, I think you can go the way you created it just like now :)
Thank you for answering it.

src/Tizen.NUI/src/public/BaseComponents/TextLabel.cs
Seoyeon2Kim: I got it.
Like other methods of Text style, we'd be better go to the Set/Get method for consistency in this file.
Thanks for explaining it.

test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/TextFitChangedSample.cs
```diff
@@ -34,6 +38,39 @@ public void Activate()
                 TextFit textfit = label.GetTextFit();
                 Tizen.Log.Error("NUI", $"FontSize : {textfit.FontSize}\n");
             };
+
```
Seoyeon2Kim, line 41: You've added the sample! :smiley: 
Thank you, Text master !!!

src/Tizen.NUI/src/public/BaseComponents/TextConstants.cs
dongsug-song: good suggestion!!!
as I know net6 supports LangVersion 10, so it seems to be OK, (tizen8.0 is on net6)
and I think it would be better to write the 10.0 explicitly (not latest) for readability.

Seoyeon2Kim: Okay! I'll update the version in another PR.
Thank you for checking my question!

PR #5612, opened 2023-10-10
[NUI][AT-SPI] Fix Accessibility.Say interop

aswigon:
### Description of Change ###

This is one of the oldest interops and was actually problematic for multiple reasons; the return type didn't match and the third parameter was expected to have exactly the same value in every invocation. While fixing these issues, parsing the status string was moved from C++ to C# for readability.

Dependency: https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/299825/

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
+++ b/src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
@@ -16,10 +16,11 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.InteropServices;
 using Tizen.NUI.BaseComponents;
-using System.Diagnostics.CodeAnalysis;
 
 namespace Tizen.NUI.Accessibility
 {
@@ -105,12 +106,10 @@ namespace Tizen.NUI.Accessibility
         /// <returns></returns>
         // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public static bool Say(string sentence, bool discardable)
+        public static void Say(string sentence, bool discardable)
         {
-            bool ret = Interop.Accessibility.Say(sentence, discardable, Marshal.GetFunctionPointerForDelegate<Delegate>(callback));
-
+            Interop.Accessibility.Say(sentence, discardable, SayFinishedEventCallback);
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
-            return ret;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
```diff
@@ -105,12 +106,10 @@ public static bool IsScreenReaderEnabled
         /// <returns></returns>
         // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public static bool Say(string sentence, bool discardable)
+        public static void Say(string sentence, bool discardable)
```
Seoyeon2Kim, line 109: Oh.. the return type is changed.
AFAIK, this feature was handled by Mr.Song, so I don't know the history.

@dongsug-song , Could you check this feature?
1. Are there any reason to add `Say` and other features previously?
2. Is it okay to change the return type? + VD / DA team doesn't use these features anymore?

Seoyeon2Kim:
Oh.. Probably this change is based on this PR... (Support Accessibility on Lite UI)
https://github.sec.samsung.net/dotnet/flux/pull/218#issuecomment-2269604
Am I right, @aswigon ?

src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
aswigon: The problem with the return type is that the interop (`Interop.Accessibility.Say`) was declared in C# as having a `bool` return type, but it returned `void` from C++, so the value would be indeterminate (meaningless).

Therefore I also changed the public API from `bool` to `void`, but I can keep `bool` and just add `return true` at the end if needed for compatibility with existing code. (But I doubt that anyone used the return value since it was basically meaningless.)

aswigon:
> Oh.. Probably this change is based on this PR... (Support Accessibility on Lite UI) https://github.sec.samsung.net/dotnet/flux/pull/218#issuecomment-2269604 Am I right, @aswigon ?

Yes @Seoyeon2Kim, `Interop.Accessibility.Say` (aka. `csharp_dali_accessibility_say`) was broken because it only supported having one callback. This worked in NUI because NUI passed the same callback every time to `Interop.Accessibility.Say`. However, LiteUI will use a different callback every time. Rather than adding a second interop specifically for LiteUI, I fixed the existing one in order to avoid code duplication.

src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
dongsug-song: @Seoyeon2Kim 
I think Mr. aswigon's opinion is reasonable. 
as you pointed out, it had been added for GBM, but actually it is not used currenly as far as I know, and it is hidden API so that it can be changed without ACR. 
thank you for your careful comments but I think it would be OK to follow Mr. aswigon's idea.

Seoyeon2Kim: Understood.
As aswigon mentioned, the return value was meaningless before.
Also, this hidden API is not used anywhere now, so it can be changed. 
Thank you for checking it once. :)

Seoyeon2Kim:
> > Oh.. Probably this change is based on this PR... (Support Accessibility on Lite UI) https://github.sec.samsung.net/dotnet/flux/pull/218#issuecomment-2269604 Am I right, @aswigon ?
> 
> Yes @Seoyeon2Kim, `Interop.Accessibility.Say` (aka. `csharp_dali_accessibility_say`) was broken because it only supported having one callback. This worked in NUI because NUI passed the same callback every time to `Interop.Accessibility.Say`. However, LiteUI will use a different callback every time. Rather than adding a second interop specifically for LiteUI, I fixed the existing one in order to avoid code duplication.

I got it.
To avoid the duplication, the PR should be applied to the latest Tizen branch.

Seoyeon2Kim (approved):
LGTM.
The patch should be merged with dali-csharp-binder patch together.

dongsug-song (approved):
looks good to me

PR #5621, opened 2023-10-13
[NUI] Resolve registry issue for CanvasView Gradient

hinohie:
Since we always try to register to Registry for Gradient case, there can make some error for Shape use case.


JoogabYun (approved):

JoogabYun:
https://github.com/Samsung/TizenFX/pull/5595와 연관이 있는 패치입니다.

PR #5622, opened 2023-10-15
[NUI.Wearable] Remove build warning

hinohie:
Remove dotnet build warnings for Tizen.NUI.Wearable.

- Add comment for public API
- Remove some unused private value/method

### DIFF
--- a/src/Tizen.NUI.Wearable/src/public/CircularPagination.cs
+++ b/src/Tizen.NUI.Wearable/src/public/CircularPagination.cs
@@ -47,8 +47,6 @@ namespace Tizen.NUI.Wearable
         private bool isCurrentIndicatorCentered = false; // When the current indicator is the center one, this variable becomes true.
         private bool isOddNumber = true;
         private bool uninitializedLeftIndicator = true; // Need it when the indicators are asymmetry and the right indicator count is set earlier than left one.
-        private Animation selectAnimation = null;
-        private bool isNeedAnimation = false; // TODO : Animation will support using override function later.
 
         Position2D[] oddArray = new Position2D[] { new Position2D(36, 74), new Position2D(47, 60), new Position2D(60, 47), new Position2D(74, 36),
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI.Wearable/src/public/CircularPagination.cs
```diff
@@ -48,7 +48,7 @@ public class CircularPagination : Control
         private bool isOddNumber = true;
         private bool uninitializedLeftIndicator = true; // Need it when the indicators are asymmetry and the right indicator count is set earlier than left one.
         private Animation selectAnimation = null;
-        private bool isNeedAnimation = false; // TODO : Animation will support using override function later.
+        //private bool isNeedAnimation = false; // TODO : Animation will support using override function later.
```
Seoyeon2Kim, line 51: It's not used anywhere, then can we remove it at all?

Seoyeon2Kim (commented):
Thank you for fixing the warning, but could you update the description of this PR?
(even though Wearable repo is not used anymore)

src/Tizen.NUI.Wearable/src/public/CircularPagination.cs
hinohie: I'm not sure what TODO comment writer want to do in future. So I just want to keep this comment. That's why I dont remove this line. What do you think about it? Could I remove this line?

hinohie: There is also an unused API in this file. If we can remove unused values, Ill also try to clean up soon

Seoyeon2Kim: From what I've remembered, Circular Pagination had a requirement to select whether animation is working or not.
That's why this variable was added.
However, `NUI.Wearable` is meaningless now :/ and we can add the unused variables later if this class will be used someday.

So, Please remove the unused values in this file

Seoyeon2Kim (approved):
LGTM!!!!!!!!!
Thank you for updating the files.

PR #5625, opened 2023-10-16
[Connection] Fix ASAN crash issue

SeonahMoon:
### Description of Change ###
Use g_free() instead of free() to avoid ASAN crash.


dh79pyun (approved):
Looks good

zhouhao02:
Looks good

PR #5628, opened 2023-10-17
[Applications.Common] Use Marshal.FreeHGlobal() instead of Libc.Free()

hjhun:
### Description of Change ###

This patch changes using free() to using Marshal.FreeHGlobal().
To avoid crash issues in the asan binary, the library should not use Libc.Free() function directly as possible.


pjh9216 (approved):

upple (approved):
LGTM 👍

PR #5629, opened 2023-10-17
[Applications.WidgetControl] Use Marshal.FreeHGlobal() instead of Lib…

kilig:
### Description of Change ###
This patch changes using free() to using Marshal.FreeHGlobal().
To avoid crash issues in the asan binary, the library should not use Libc.Free() function directly as possible.


devusr-sw-kim (approved):
With this change, no crash issue on asan built environment for both Tizen.Applications.WidgetControl and Tizen.Applications.Cion. LGTM.

PR #5637, opened 2023-10-18
[NUI] Make AddFrameUpdateCallback with root view

hinohie:
Previously, we can only add the root view of FrameUpdateCallback as root layer of window.

Now let we add some API that user can set specific view for given FrameUpdateCallback will use.

This API will be used when we want to optimize callback behaviour.


dongsug-song (approved):
looks good to me

JoogabYun (approved):

PR #5640, opened 2023-10-18
[NUI] Add an API to set offset for drag window

taehyub:
### Description of Change ###
[NUI] Add an API to set offset for drag window

### API Changes ###

 - ACR: N/A

### DIFF
--- a/src/Tizen.NUI/src/public/DragAndDrop/DragAndDrop.cs
+++ b/src/Tizen.NUI/src/public/DragAndDrop/DragAndDrop.cs
@@ -43,6 +43,8 @@ namespace Tizen.NUI
         private Window mDragWindow;
         private int shadowWidth;
         private int shadowHeight;
+        private int dragWindowOffsetX = 0;
+        private int dragWindowOffsetY = 0;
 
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/DragAndDrop/DragAndDrop.cs
```diff
@@ -123,7 +125,7 @@ public void StartDragAndDrop(View sourceView, View shadowView, DragData dragData
                 shadowHeight = MinDragWindowHeight;
             }
 
-            mDragWindow = new Window("DragWindow", new Rectangle(-shadowWidth, -shadowHeight, shadowWidth, shadowHeight), true)
+            mDragWindow = new Window("DragWindow", new Rectangle(dragWindowOffsetX, dragWindowOffsetY, shadowWidth, shadowHeight), true)
```
hinohie, line 128: Minor question. After this patch, the default offset changed from "-shadowSize" to "0". Mean, default behaviour changed. Is it okay??

hinohie (approved):
If my commented point was resolved already, looks good to me.

PR #5646, opened 2023-10-19
[NUI][AT-SPI] Detach NUIViewAccessible in View.Dispose()

aswigon:
### Description of Change ###

The application may crash if the View is disposed but the Accessibility infrastructure calls one of the View methods. Detaching the NUIViewAccessible proxy object in View.Dispose() should prevent that.

Dependency: https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/300277/

### API Changes ###

 - ACR: None

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityWrappers.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityWrappers.cs
@@ -53,7 +53,7 @@ namespace Tizen.NUI.BaseComponents
 
             if (view is null)
             {
-                NUILog.Error($"RefObject 0x{refObjectPtr:x} is not a View");
+                throw new ArgumentException($"RefObject 0x{refObjectPtr:x} is not a View", nameof(refObjectPtr));
             }
 
             return view;
@@ -70,9 +70,7 @@ namespace Tizen.NUI.BaseComponents
                 return atspiInterface;
             }
 
-            NUILog.Error($"RefObject 0x{refObjectPtr:x} is not a {typeof(T).FullName}");
-
-            return default(T);
+            throw new ArgumentException($"RefObject 0x{refObjectPtr:x} is not a {typeof(T).FullName}", nameof(refObjectPtr));
         }
 
         private static IntPtr DuplicateString(string value)
@@ -113,8 +111,6 @@ namespace Tizen.NUI.BaseComponents
         private static ulong AccessibilityCalculateStatesWrapper(IntPtr self, ulong initialStates)
         {
             View view = GetViewFromRefObject(self);
-            if (view == null)
-                return 0UL;
 
             ulong bitMask = 0UL;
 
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityWrappers.cs
```diff
@@ -113,8 +113,6 @@ private static void InitializeAccessibilityDelegateAccessibleInterface()
         private static ulong AccessibilityCalculateStatesWrapper(IntPtr self, ulong initialStates)
         {
             View view = GetViewFromRefObject(self);
-            if (view == null)
```
jaehyun0cho, line 116: If view is null, then the following code is not accessible?
Or view cannot be null?
```
bitMask = view.AccessibilityCalculateStates().BitMask;
```

rabbitfor, line 116: `GetViewFromRefObject`는 로직상 null 을 리턴할 수 있는데, null 리턴이 절대 일어나지 않아야 한다면 debug assert나 exception 또는 최소한 로그 정도는 남겨두어야 문제 상황을 추적하기 쉬울 것 같은데 어떻게 생각하시나요?

aswigon: It cannot be null, see below.

aswigon: There is already an error log reported in `GetViewFromRefObject()`. It should never be `null`.

I deleted the null check in `AccessibilityCalculateStatesWrapper()` because it is the only one, and there are 61 more `Accessibility...Wrapper()` functions in this file **without** a null check.

The null check in only one of 62 wrapper methods was added in https://github.com/Samsung/TizenFX/pull/4370, as a workaround / hack. At that time it was not understood what the underlying cause was... This PR should address the underlying cause, so the workaround here should not longer be necessary.

hinohie: Could you please throw exception than? If so, it might be helpful when some unknown error occured in future.

aswigon: Okay, added throwing `ArgumentException` in both `GetViewFromRefObject` and `GetInterfaceFromRefObject`.

src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityWrappers.cs
```diff
@@ -53,7 +53,7 @@ private static View GetViewFromRefObject(IntPtr refObjectPtr)
 
             if (view is null)
             {
-                NUILog.Error($"RefObject 0x{refObjectPtr:x} is not a View");
+                throw new ArgumentException($"RefObject 0x{refObjectPtr:x} is not a View", nameof(refObjectPtr));
```
hinohie, line 56: I just worryed about some case of View's that NUI didn't created. (For example, `Tizen.NUI.Scene3D.Model` create `Tizen.NUI.Scene3D.ModelNode` internally, but they are automatically created only at dali-side. So NUI (actually, Registry) don't know that View. But `ModelNode` is also kind of View.

hinohie (approved):
But anyway, looks find to me now.

src/Tizen.NUI/src/public/BaseComponents/ViewAccessibilityWrappers.cs
aswigon: This shouldn't be a problem. The methods in `ViewAccessibilityWrappers.cs` are entry points for `NUIViewAccessible` methods (a C++ subclass of `ControlAccessible`) and `NUIViewAccessible` is only created for (a subset of) C# `View`s.

jaehyun0cho (approved):
LGTM

PR #5648, opened 2023-10-23
[NUI.Scene3D] Bug fix when we get ModelNode from Model

hinohie:
Since View constructor automatically change 'PositionUsesPivotPoint' value as false, some logic might changed when we Register ModelNode newly.


hinohie:
TODO : Need to do same thing at 'GetModelRoot()' API.

huiyueun (approved):

PR #5649, opened 2023-10-24
[NUI] Preload some static values

hinohie:
Since C# initialize static values when that namespace / class access first time, We need to access some static values at static Preload() timing.

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -105,6 +105,16 @@ namespace Tizen.NUI.BaseComponents
             RegisterAccessibilityDelegate();
         }
 
+        static internal new void Preload()
+        {
+            Container.Preload();
+
+            // Do nothing. Just call for load static values.
+            var temporalPositionPropertyGroup = positionPropertyGroup;
+            var temporalSizePropertyGroup = sizePropertyGroup;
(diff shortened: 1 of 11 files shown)


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -105,6 +105,16 @@ static View()
             RegisterAccessibilityDelegate();
         }
 
+        static internal new void Preload()
+        {
+            Container.Preload();
+
+            // Do nothing. Just call for load static values.
+            var object0 = positionPropertyGroup;
```
Seoyeon2Kim, line 113: I know this PR is to fix the urgent VD issues.
But, is it okay for the variable name???

```suggestion
            var positionBbject = positionPropertyGroup;
```

dongsug-song (approved):
Looks good to me! 
Great patch!
According to VD's local test, it works. Therefore, preloading does not necessarily require creating a new instance of the class (e.g., View).

src/Tizen.NUI/src/public/BaseComponents/View.cs
hinohie: Thank you for your comments. Now I make them as 'temporalXXX', instead of meaningless variable name :)

Seoyeon2Kim: Thank you for the quick update!

PR #5653, opened 2023-10-24
[NUI] Supports set/get full screen window

sparrow74:
Supports set/get full screen window.

The full screen window means the window's size is same of full screen size.
The related DALi's patches are the belows.
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-adaptor/+/300078/
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/300079/

Recently, Taskbar application was added to platform applications.
The Taskbar application is always floating on the home application. The application's window is the highest of the other window.
If the Taskbar application works, the other application window should be resized excepted Taskbar window's size.
The other requirement, some application want their windows are resized with full screen although Taskbar works.
To do that, the window uses full screen option.

The guide is 
--------------------------------------------------------------------------------------
                        if(mainWin.GetFullScreen() == false)
                        {
                            mainWin.SetFullScreen(true);
                        }
                        else
                        {
                            mainWin.SetFullScreen(false);
                        }
--------------------------------------------------------------------------------------
If you want to use the related sample, refer the example, please
TizenFX/test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/WindowTest.cs


dongsug-song (approved):
looks good to me.
very good to add TC and sample.

PR #5654, opened 2023-10-24
[NUI.Scene3D][API10] Bug fix when we get ModelNode from Model

hinohie:
Since View constructor automatically change 'PositionUsesPivotPoint' value as false, some logic might changed when we Register ModelNode newly.


dongsug-song (approved):
looks good to me

PR #5655, opened 2023-10-24
[NUI] Change Accessibility.cs not to use net6.0 dependency function

hinohie:
Let us minimize the .net core dependency

### DIFF
--- a/src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
+++ b/src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
@@ -332,7 +332,11 @@ namespace Tizen.NUI.Accessibility
 
         private static void SayFinishedEventCallback(string status)
         {
-            var result = sayFinishedStateDictionary.GetValueOrDefault(status, SayFinishedState.Invalid);
+            SayFinishedState result;
+            if (!sayFinishedStateDictionary.TryGetValue(status, out result))
+            {
+                result = SayFinishedState.Invalid;
+            }
             NUILog.Debug($"sayFinishedEventCallback(res={result}) called!");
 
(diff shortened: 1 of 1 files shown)


dongsug-song:
@Seoyeon2Kim @lol-github @aswigon 
Hello, 
please review this, it is atspi related.
thank you.

src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
```diff
@@ -332,7 +332,8 @@ public enum SayFinishedState
 
         private static void SayFinishedEventCallback(string status)
         {
-            var result = sayFinishedStateDictionary.GetValueOrDefault(status, SayFinishedState.Invalid);
+            SayFinishedState result = SayFinishedState.Invalid;
+            sayFinishedStateDictionary.TryGetValue(status, out result);
```
aswigon, line 336: I think that `result` would be set by `TryGetValue` in all cases, so perhaps the assignment of `SayFinishedState.Invalid` should be done after `TryGetValue`.

The [docs](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue) say:

> When this method returns, contains the value associated with the specified key, if the key is found; otherwise, the default value for the type of the value parameter. This parameter is passed uninitialized.

hinohie: Wow! I didn't know detail of that! I fix it now. Thank you for your comments.

dongsug-song (approved):
approved

hinohie:
Cool. more nice PR name. Thank you for you fixs and reviews. I merge it.

PR #5660, opened 2023-10-26
[NUI] Add KeepWidgetSize for control widget's size

tscholb:
WidgetView keep widget instance's size and don't resize even if WidgetView is resized.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Widget/WidgetView.cs
+++ b/src/Tizen.NUI/src/public/Widget/WidgetView.cs
@@ -138,6 +138,25 @@ namespace Tizen.NUI
             return instance.InternalEffect;
         });
 
+        /// <summary>
+        /// UpdateWidgetSizeProperty
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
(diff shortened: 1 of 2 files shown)


jaehyun0cho (approved):
LGTM

src/Tizen.NUI/src/public/Widget/WidgetView.cs
```diff
@@ -685,6 +704,44 @@ private PropertyMap InternalEffect
             }
         }
 
+        /// <summary>
+        /// Gets or sets KeepWidgetSize
+        ///
+        /// if this value is true, WidgetView keep widget instance's size and don't resize even if WidgetView is resized.
+        /// </summary>
+        /// <since_tizen> 8 </since_tizen>
```
Seoyeon2Kim, line 712: I think we can remove this since_tizen tag.
And as you know, the current latest version is 12 :crying_cat_face:

tscholb: Thank you. i will remove it!

PR #5662, opened 2023-10-26
[NUI] Make NativeImageQueue generate valid url

hinohie:
Previously NativeImageQueue.GenerateUrl() binded to invalid csharp API.

It will make some crash. So let we use valid bind API instead.

Required dali patches:
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/300505
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/300513


dongsug-song (approved):
looks good to me

PR #5663, opened 2023-10-27
[API10][NUI][AT-SPI] Detach NUIViewAccessible in View.Dispose()

aswigon:
### Description of Change ###

The application may crash if the View is disposed but the Accessibility infrastructure calls one of the View methods. Detaching the NUIViewAccessible proxy object in View.Dispose() should prevent that.

Dependency: https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/300567/

### API Changes ###

 - ACR: None


jaehyun0cho (approved):
LGTM

PR #5665, opened 2023-10-27
[NUI] Resolve DisposeRecursively crashed when we dispose scrollbase

hinohie:
Due to some app's implements, ContentView can be disposed before ScrollBase itself disposed. In this case, ContentView is not null.

So, when we try to call RemovePropertyNotification, it become crashed.

To avoid these cases, let we make DisposeRecursively API makes more safety.

And also, fix some crash issue in PerformanceTest1Page.xaml.cs what can be problem in future.


dongsug-song (approved):
looks good to me

PR #5666, opened 2023-10-27
[NUI][Devel.Tests.Ubuntu] Make it works on net6.0

hinohie:
Let we make Devel.Tests.Ubuntu works well under net6.0 version.

Also, add some logics to debug deadlock issue when we try to terminate TCT app hardly.


dongsug-song (approved):
looks good to me

PR #5672, opened 2023-10-30
[MediaController] Add new API to check Response request

hsgwon:
### Description of Change ###

Add new API to check whether sender ask Respose for command or not.
Fix crash issue when request_id is null.

### API Changes ###

 - TCSACR-560


src/Tizen.Multimedia.Remoting/MediaController/MediaControlCommand.cs
```diff
@@ -59,10 +60,22 @@ internal void SetRequestInformation(string receiverId)
...
+            _requestId = requestId;
+
+            if (_requestId == null)
+            {
+                Log.Info(GetType().FullName, "request_id is null. Response() should not be called.");
+            }
         }
 
+        /// <summary>
+        /// Gets the status of response.
```
wiertel, line 73: how it sounds suggests that it can change, like when Response is supplied it would be set to false then,
suggestion: "Indicates whether a response is expected for this command." or "Indicates if the command should be responded to" or shorter brief "Indicates if the response is needed."

src/Tizen.Multimedia.Remoting/MediaController/MediaControlCommand.cs
```diff
@@ -59,10 +60,22 @@ internal void SetRequestInformation(string receiverId)
...
+        /// </summary>
+        /// <remarks>If false, the receiver should not response for the received command.</remarks>
```
wiertel, line 75:
```suggestion
        /// <remarks>If false, the receiver should not respond to the received command.</remarks>
```

src/Tizen.Multimedia.Remoting/MediaController/MediaControlCommand.cs
```diff
@@ -59,10 +60,22 @@ internal void SetRequestInformation(string receiverId)
...
+        /// <since_tizen> 12 </since_tizen>
+        public bool NeedToResponse => _requestId != null;
```
wiertel, line 77: please change the name of this property, like "IsResponseNeeded", "ResponseNeeded", "NeedToRespond", it cannot be [to + noun]

src/Tizen.Multimedia.Remoting/MediaController/MediaControlCommand.cs
```diff
@@ -90,6 +103,11 @@ internal void SetResponseInformation(string receiverId, string requestId)
         /// <param name="bundle">The extra data.</param>
```
wiertel, line 103: Could you change the description at line 99 (and for the next method)? Like "Sends command response to the client" or "Responds to the clients command"
Unfortunately we cannot change the name of the method, "Response" is a noun and method names should be verb (or centered around verb), so it should be Respond() or SendResponse()

wiertel (changes requested):

src/Tizen.Multimedia.Remoting/MediaController/MediaControlCommand.cs
hsgwon: I'll change it to 'IsResponseNeeded'. But 'NeedToResponse' is already used in VD, so I'll keep it internal api only.

src/Tizen.Multimedia.Remoting/MediaController/MediaControlCommand.cs
hsgwon: I'll change it to 'Indicates if the response is needed.'

src/Tizen.Multimedia.Remoting/MediaController/MediaControlCommand.cs
hsgwon: Done

src/Tizen.Multimedia.Remoting/MediaController/MediaControlCommand.cs
hsgwon: I'll change it to `Responds to the clients command.`.

wiertel (approved):

PR #5673, opened 2023-10-30
[NUI] Do not register when we try to copy instance

hinohie:
Since we register to registry by BaseObjectPtr(), it is impossible to register duplicated handle.

So, actually, we have to deplicate copy BaseHandle constructor. But for now, just make we don't register into registry now.


dongsug-song (approved):
very good fix!
look good to me

huiyueun (approved):

hinohie:
I'll merge it after domain test at VD binary, which is after integrated.

hinohie:
Sorry.. It was some kind of big problem. I think we have to merge this PR as soon as we can.

PR #5674, opened 2023-10-30
[NUI] Make DaliAccessibilityDetachAccessibleObject called at main thread

hinohie:
void Dispose(bool) can be called from worker thread, when disposing value is false.

To ensure we detach AccessibileObject in main thread, let we remove it at Dispose(DisposeTypes), instead of Dispose(bool)


dongsug-song (approved):
looks good to me

Seoyeon2Kim:
@lol-github @aswigon @m-bialota 
Could you check this PR even if this PR is merged soon?
If the View is disposed by GC, `Interop.ControlDevel.DaliAccessibilityDetachAccessibleObject` function could be called from GC thread, which could cause a crash.
This PR makes it to operate only on the main thread.

PR #5675, opened 2023-10-30
[NUI] Remove basehandle usage at Accessibility static

hinohie:
Since we try to resolve some unusual behavior of BaseHandle, the dummyHandle in Accessibility can give us some confuse logs.

Actually, we don't need to hold it as BaseHandle type. we only need this as object type. So, let we make it as object, instead of BaseHandle.

### DIFF
--- a/src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
+++ b/src/Tizen.NUI/src/public/Accessibility/Accessibility.cs
@@ -212,7 +212,7 @@ namespace Tizen.NUI.Accessibility
 
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
 
-            return dummyHandle.GetInstanceSafely<View>(ptr);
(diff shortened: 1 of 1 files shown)


hinohie:
@Seoyeon2Kim  @lol-github @aswigon @m-bialota
Could you check this PR please? It will remove some useless BaseHandle creation before NUI Application initialized.

aswigon (approved):
LGTM. I've just checked that `GetInstanceSafely<T>` only requires `T` to be a subclass of `BaseHandle`, but its parameter can simply be `object`, so there should be no semantic change with this patch.

```c#
public static T GetInstanceSafely<T>(this object wrapper, IntPtr cPtr) where T : BaseHandle
```

loleksak (approved):
LGTM

dongsug-song:
The TizenFX master branch has been changed for API12(tizen_9.0).
So the DevelNUI branch also did a hard reset with the master branch to sync this.
I think you need to rebase this PR or re-upload the new PR.
I am sorry for the inconvenience.

dongsug-song (approved):
looks good to me

PR #5679, opened 2023-10-30
[Build] Create vconf related to .NET as a file in RO location

JongHeonChoi:
Guided by Appfw, vconf daemon does not run at mic stage. So, ```vconf_get_*``` api does not work normally. Therefore, it provides a second chance to read the file when ```vconf_get_*``` is null. So, Create vconf related to .NET as a file in RO location.


HJLeee (approved):

JoonghyunCho:
Do we still need vconf set for the legacy?

sung-su (approved):
LGTM

PR #5683, opened 2023-10-31
[NUI] Resolve many problems relative with BaseHandle

hinohie:
- Ensure to get 'already registered, or create new one' what we get item from DALi engine
 - Let we also ensure create only one items if we use static ::Get() API
 - Do not register to Registry if we copy the reference.
  - Remove some useless copy constructor.
 - Do not register to Registry if we create empty object.


dongsug-song (approved):
approved

PR #5684, opened 2023-10-31
[NUI] Fix some SVACE issues.

zhouhao02:
### Description of Change ###

1. Value rootLayout is the result of a method call with a possible null retum value, which is dereferenced in the member access expression root.Layout. MeasuredWidth
  Value rootLayout is the result of a method call with a possible null retum value, which is dereferenced in the member access expression root.Layout. MeasuredHeight
2. Nesting level does not match indentation

### API Changes ###

 - ACR:


huayongxu (approved):
lgtm

zg2nets (approved):
lgtm

hinohie (approved):

elishateng (approved):
LGTM

wanchao-xu (commented):
Looks good to me

PR #5685, opened 2023-10-31
[Build] Modify build error for vconf of runtime_version

JongHeonChoi:
The name of the ```dotnet_resolving.info``` file is changed. And the vconf creation of runtime_version is incorrect.


JoonghyunCho (approved):
Please make sure to find a better way to get the version

PR #5703, opened 2023-11-02
[Build] Modify the script that generates the TFM list

JongHeonChoi:
Modify the script that generates the TFM list. And remove the use of vconf related to dotnet.


sung-su (approved):
LGTM

packaging/maketfm.py
```diff
@@ -20,12 +21,21 @@
                             tfm_list1.append(tfm.replace(".", "").strip())
                         else:
                             tfm_list2.append(tfm.strip())
+                    elif tfm.startswith("net"):
+                        if len(tfm) < 8:
```
JoonghyunCho, line 25: What is this `8` means? Isn't it better to match with `net6.0`?

JongHeonChoi: Because the length from ```net6.0``` to ```net99.0``` is up to ```7```.
```
len(net6.0) == 6
len(net10.0) == 7
len(net99.0) == 7
```

JoonghyunCho: understood.
Just the suggestion for later, I think we can think about using regex match rather than checking the length.

PR #5707, opened 2023-11-02
[NUI] Do not register when BaseHandle doesn't have body

hinohie:
If BaseHandle doesn't have body, the refCPtr value is IntPtr.Zero.

If than, the various cases have wrong registration in future.

To avoid this case, let we ignore to register to registry, and make BaseHandle.registerMe value as false.


dongsug-song (approved):
approved

PR #5736, opened 2023-11-10
[NUI] DetachAccessibleObject at Dispose time well

hinohie:
There was several problem when we call DaliAccessibilityDetachAccessibleObject at Dispose(bool) function

 - It will detach accessibility even if view doesn't owned the handle memory
 - Dispose(bool) can be called at worker thread (during GC) so we need to check thread here

I make to DaliAccessibilityDetachAccessibleObject function at Dispose(DisposeTypes) that we can ensure it called at main thread clearly.

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
@@ -397,12 +397,6 @@ namespace Tizen.NUI.BaseComponents
 
             internalName = null;
 
-            if (SwigCPtr.Handle != IntPtr.Zero && global::System.Threading.Thread.CurrentThread.ManagedThreadId == Registry.Instance.SavedApplicationThread.ManagedThreadId)
-            {
-                Interop.ControlDevel.DaliAccessibilityDetachAccessibleObject(SwigCPtr);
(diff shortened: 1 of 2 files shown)


aswigon (approved):

m-bialota (approved):

rabbitfor (approved):
LGTM! I'd like to recommend you to use this when check native errors.
https://github.com/Samsung/TizenFX/blob/96db81b7c85cb4858444428ed3011f6fc03c06fb/src/Tizen.NUI/src/internal/Interop/NDalicPINVOKE.cs#L152-L155

Seoyeon2Kim:
This PR is related to DALi patches below :
https://github.sec.samsung.net/Tizen-DA/da-nui-components/pull/785

hinohie:
> LGTM! I'd like to recommend you to use this when check native errors.
> 
> https://github.com/Samsung/TizenFX/blob/96db81b7c85cb4858444428ed3011f6fc03c06fb/src/Tizen.NUI/src/internal/Interop/NDalicPINVOKE.cs#L152-L155

Cool! I update the code line as what you suggest to me.

hinohie:
Actually, this patch don't have dali dependency. So I'll merge it. @Seoyeon2Kim

PR #5737, opened 2023-11-10
[NUI] Fix TotalFrame getter bug when async load

hinohie:
There was some bug when we try to get TotalFrame when async load json file.

DALi engine side default total frame number is 0, and NUI is -1. So, if we try to get TotalFrame from dali which is still loading, it will cache the internal total frame as 0. After than, we can never get valid value.

To avoid this problem, we have 2 choice

1. Wait until async load finished, and then get total frame at dali side.
2. Just get 0 for now, and try again when total frame cached as 0.

Solution 1 need some time to resolve it. So now just make 2.


bshsqa (approved):
GTM

PR #5743, opened 2023-11-14
[NUI] Support EncodedImageBuffer ImageType property + Upgrade Encoded…

hinohie:
…ImageBuffer demo

Let we allow to give the type of buffer s.t. what buffer is mean. We can load svg and lottie image as encoded image buffer now.

+

Let we make EncodedImageBuffer can be created at worker thread. And make the demo to load buffer asynchronously.

This patch have dependency with below dali patch :

https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/301314


dongsug-song (approved):
very good to add test sample!
approved

PR #5744, opened 2023-11-14
[NUI.Scene3D] Fix exception when ModelNode don't have ModelNode as child

hinohie:
If ModelNode don't have ModelNod as child, it might have some null handle exception when we DownCast to ModelNode. (For example, user try to get child ModelNode, but failed.)

Currently, a lots of app don't consider ModelNode now. But in future we might need to add something under each ModelNode.

To make Scene3D more safety, let we check null or invalidate, and then get/set the value internally.


dongsug-song (approved):
approved

PR #5745, opened 2023-11-14
[NUI] Change SetMinAndMaxFrameByMarker behavior + Fix some async cases

hinohie:
Let we allow to call SetMinAndMaxFrameByMarker API even if load is not finished. Instead, let we just call some log if the file is not loaded yet.

Plus, Some API (GetContentInfo, GetMarkerInfo) might cache wrong values when we use async load

Before try to get correct value, let we check whether the resource is loaded or not.


dongsug-song (approved):
approved

JoogabYun (approved):

PR #5746, opened 2023-11-14
[NUI] Add AliveCount property to get currently alived View number

dongsug-song:
### Description of Change ###
[NUI] Add AliveCount property to get currently living View number
- add only getter as hidden api
- consider the two cases of explicitly disposed case by user and implicitly case by GC.

### API Changes ###
none


src/Tizen.NUI/src/public/BaseComponents/ViewAccessibility.cs
```diff
@@ -403,6 +403,7 @@ protected override void Dispose(bool disposing)
                 {
                     // at this case, implicit nor explicit dispose is not required. No native object is made.
                     disposed = true;
+                    aliveCount--;
```
hinohie, line 406: Wow.. A lots of people will miss this cases! Looks good to me

hinohie (approved):

PR #5747, opened 2023-11-14
[NUI] Remove old frame range if we set marker + Minor fix up when marker invalid

hinohie:
LottieAnimationView cache both frame range as int, and marker. But two API set same property - PlayRange.

So if some code like below thing was not working well

```
  // Let "marker" frame range is 10~20

  lottieView.SetMinMaxFrame(30, 40); // play range become 30~40
  lottieView.SetMinMaxFrameByMarker(marker); // play range become 10~20
  lottieView.SetMinMaxFrame(30, 40); // play range need to be 30~40 again, but it didn't due to the cache.
```

To avoid like above case, let we clean cache explicit parameter.

Also, Let we add some more error check if app use invalid marker (Lottie file not loaded, or marker is invalid.)

In this case, let we just follow minMaxSetTypes.NotSetByUser logic then.


dongsug-song (approved):
approved

PR #5791, opened 2023-12-04
Fix defect that is detected by static analisys tool

urmain:
### Description of Change ###

There is a potential problem with the code involving the variable 'e.OldItems'. This variable is the result of a method invocation, and the tool is warning that the method might return a null value, which can cause dereferencing a null reference. It might cause it to crash or behave unpredictably.

Therefore, to prevent this, the variable 'e.OldItems' should be checked if it is null or not before use.

### API Changes ###

 - No ACR:


stom-hwang (approved):
LGTM 👍

PR #5792, opened 2023-12-04
[NUI] Support RootTranslationOnly option for motion capture data

bshsqa:
### Description of Change ###
This patch allow an option to use only root translation for motion capture data.

### API Changes ###

 - ACR:


hinohie (approved):
Wow! Thank you for your support!

PR #5795, opened 2023-12-05
[NUI] Add log for ClearFocus

JoogabYun:
### Description of Change ###

Add log for ClearFocus

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
@@ -1146,6 +1146,7 @@ namespace Tizen.NUI.BaseComponents
             // If the view had focus, it clears focus.
             if (child == FocusManager.Instance.GetCurrentFocusView())
             {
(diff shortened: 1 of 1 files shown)


hinohie (approved):

src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
```diff
@@ -1146,6 +1146,7 @@ internal void RemoveChild(View child)
             // If the view had focus, it clears focus.
             if (child == FocusManager.Instance.GetCurrentFocusView())
             {
+                Tizen.Log.Debug("NUI", $"ClearFocus id:({child.ID})\n");
```
hinohie, line 1149:
```suggestion
                Tizen.Log.Debug("NUI", $"ClearFocus due to View id:({child.ID}) removed from scene.\n");
```

Could you write some more informations at this line? Currently, we don't know whether app call ClearFocus itself, or not.

PR #5800, opened 2023-12-05
[NoACR][MachineLearning.Inference] Fix bug using wrong handle

niley7464:
### Description of Change ###
This patch fixes the bug which use wrong handle.
GetTensorsCount should use TensorsInfo handle instead of TensorsData handle.


jaeyun-jung (approved):

again4you (approved):
LGTM!

PR #5810, opened 2023-12-07
[NUI] Make ensure ImageView._resourceUrl is latest

hinohie:
Previous code didn't assume that _resourceUrl value of ImageView was not a latest.

 - If we set ImageView by ```Image``` property, the _resourceUrl was updated when cached map updating.
 - When we try to get ResourceUrl gettor, we ask to cached map always.

But PropertyValue.Get(out string) might spend long time if the URL length is long. It will make useless string copy, what we can reduce.

Now We make every logic make ensure that _resourceUrl stored valid resource url. If this is validated, we can reduce some C# - Native networking.


dongsug-song (approved):
approved

PR #5811, opened 2023-12-08
[NUI] Skip text getter when Text never setted before

hinohie:
If we never set text into TextLabel before, the result of Text propert getter is always empty. To avoid useless C# - native network, let we just ignore text getter function if we can assume that previous text was empty.


hinohie:
Hi. I check `Tizen.TV.FLUX.Component`'s `TextBox` Implementation. There is no dependency with this implements. So I think we can ignore it.

PR #5813, opened 2023-12-11
[NUI] Make we flush DisposeQueue before app start

hinohie:
If we don't call DisposeQueue.Instance.Initialize(), it might not flush the dispose queue.

Before #5806 and until currently, we didn't allow to call ProcessDisposables() before Initialize(). It might occure memory leak if app try to create & dispose unlimitly before Application start.

To make ensure it, let we make ProcessDisposables() API can be called even if DisposeQueue is not initialized.


dongsug-song (approved):
approved

PR #5814, opened 2023-12-11
[NUI] Add API for ProcessController without Initialize

hinohie:
To keep the backword stability, let we use previous API (CSharp_Dali_new_ProcessorController) call Initialize internally and new API (CSharp_Dali_new_ProcessorController_Without_Initialize) will not call this.

Since that #5806 change the logic of ProcessorController. So some library change the logical flow.
This patch make the logical flow keep same as previous if required.

required dali patch :  https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/302596


dongsug-song (approved):
approved

PR #5824, opened 2023-12-14
[NUI] Initialize C# items for ProcessorController if required

hinohie:
We have create ProcessorController constructor without initalize as default.

If NUI core want to change this policy, we can initialize at constructor time whenever we want.

But this case code has some error. It didn't create C# side objects and only create native side.

Let we make constructor with initialize works well now.

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/ProcessorController.cs
+++ b/src/Tizen.NUI/src/internal/Common/ProcessorController.cs
@@ -40,8 +40,12 @@ namespace Tizen.NUI
         {
         }
 
-        private ProcessorController(bool initializeOnConstructor) : this(initializeOnConstructor ? Interop.ProcessorController.New() : Interop.ProcessorController.NewWithoutInitialize(), true)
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/internal/Common/ProcessorController.cs
```diff
@@ -40,8 +40,12 @@ private ProcessorController() : this(true)
         {
         }
 
-        private ProcessorController(bool initializeOnConstructor) : this(initializeOnConstructor ? Interop.ProcessorController.New() : Interop.ProcessorController.NewWithoutInitialize(), true)
+        private ProcessorController(bool initializeOnConstructor) : this(Interop.ProcessorController.NewWithoutInitialize(), true)
```
dongsug-song, line 43: It seems to me that it always initializes Dali regardless of the bool argument.
I don't see any problem.

dongsug-song (approved):
looks good to me

dongsug-song:
@jaehyun0cho
@rabbitfor 
@everLEEst 
@taehyub 
@Seoyeon2Kim 

I request your review.

jaehyun0cho (approved):
LGTM

PR #5827, opened 2023-12-18
[NUI] Revert dispose synchronously logic when APP is not started

hinohie:
We add some logic to dispose items when application created & disposed before application initialized.

But some app just have some issue when some thread dependency items disposed before application started.

To avoid this cases, let we block dispose sync logic and revert we always dispose queue ensure in main thread.


jaehyun0cho (approved):
LGTM

dongsug-song (approved):
approved

PR #5837, opened 2023-12-20
[Metadata] Deprecate MetadataEditor module

hsgwon:
### Description of Change ###

Deprecate MetadataEditor module

### API Changes ###

 - TCSACR-569


src/Tizen.Multimedia.Metadata/MetadataEditor/MetadataEditor.cs
```diff
@@ -439,6 +454,7 @@ public string UnsyncLyrics
         /// Internal error.<br/>
         /// </exception>
         /// <exception cref="ObjectDisposedException">The <see cref="MetadataEditor"/> has already been disposed of.</exception>
+        [Obsolete("Deprecated in API12; Will be removed in API14")]
```
wiertel, line 457: please move that line down, so it not be in the middle of /// comment block

wiertel (changes requested):

src/Tizen.Multimedia.Metadata/MetadataEditor/MetadataEditor.cs
hsgwon: OK I'll fix it.

wiertel (approved):

PR #5851, opened 2023-12-31
[NUI] Support NativeImageQueue creation with queue count

hinohie:
Let we allow to create NativeImageQueue with queue count. Previously, we only allow to create the NativeImageQueue based on the environment value "DALI_TBM_SURFACE_QUEUE_SIZE". (This value is 3 as default)

It will be useful if some user want to reduce the queue count for some cases, not for whole queue.

relative DALi patches :

https://review.tizen.org/gerrit/c/platform/core/uifw/dali-adaptor/+/303523 https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/303524


JoogabYun (approved):

dongsug-song (approved):
approved

PR #5854, opened 2024-01-02
[NUI.Scene3D] Add AddLight API

hinohie:
Let we make new API for Light class for SceneView.

Note : To avoid the changness of native side code minimalize, let we use internal API of NUI.


bshsqa (approved):
Good

PR #5856, opened 2024-01-03
[NUI] Do not recreate visual when we change ImageColor only

hinohie:
Since we don't need to re-create the visual if we only chage the ImageColor property, let we use UpdateProperty Action.


JoogabYun (approved):

JoogabYun (approved):

Seoyeon2Kim (approved):
LGTM

PR #5858, opened 2024-01-04
[NUI] Fix text padding issue

jaehyun0cho:
This patch solves the problem that text with padding does not work in nui layout.

For text components, the control's padding is used to calculatie size in native, so it should not be set to zero value.

This avoids setting padding to zero when it comes to text layout.

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
LGTM~

PR #5863, opened 2024-01-06
[NUI] Do not call default creator when we set value

hinohie:
"ALL" default BindableProperty.PropertyChanged events what NUI using, dont use 'oldValue'. But due to that usecase, we always call DefaultValueCreator when we try to set value.

To make NUI platform faster, let we ignore that property getter when we only use setter.

For XMAL case, support for backward stabilize, let we don't use this feature.

Performance result for app `Tizen.NUI.PerformanceTest`

(dali version : 2.3.5. Release, Execute on WSL server based on Ubuntu22.04, i7 dual core)

|Creation Type|Before PR (ms)|After PR (ms)|Reduce Rate (1 - after / before)| 
|-------------|-------------|-------------|-------------| 
| 40 COLOR | 2.6657574074074066 | 2.19945 | 17.492492232 % | 
| 40 IMAGE | 1.9131259259259266 | 1.56895 | 17.99023897 % | 
| 40 TEXT | 7.013750000000001 | 6.103418518518519 | 12.979240513 % |
| 40 ROUNDED COLOR | 3.342475925925925 | 2.8528203703703694 | 14.649486381 % |
 | 40 BORDER COLOR | 4.039824074074074 | 3.8768833333333332 | 4.033362289 % |
 | 40 ROUNDED BORDER COLOR | 4.641790740740741 | 4.230518518518518 | 8.860206011 % | 
| 40 BLUR COLOR | 2.7265814814814813 | 2.329972222222222 | 14.546026295 % | 
| 40 ROUNDED BLUR COLOR | 2.817055555555555 | 2.56367962962963 | 8.994353179 % |

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ImageViewBindableProperty.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ImageViewBindableProperty.cs
@@ -209,15 +209,6 @@ namespace Tizen.NUI.BaseComponents
             var imageView = (ImageView)bindable;
             if (newValue != null)
             {
-                if (oldValue != null)
-                {
-                    bool oldBool = (bool)oldValue;
-                    bool newBool = (bool)newValue;
-                    if (oldBool == newBool)
-                    {
-                        return;
-                    }
-                }
                 imageView.UpdateImage(NpatchImageVisualProperty.BorderOnly, new PropertyValue((bool)newValue));
             }
         },
@@ -238,15 +229,6 @@ namespace Tizen.NUI.BaseComponents
             var imageView = (ImageView)bindable;
             if (newValue != null)
             {
-                if (oldValue != null)
-                {
-                    bool oldBool = (bool)oldValue;
-                    bool newBool = (bool)newValue;
-                    if (oldBool == newBool)
-                    {
-                        return;
-                    }
-                }
                 // Note : We need to create new visual if previous visual was async, and now we set value as sync.
                 imageView.UpdateImage(ImageVisualProperty.SynchronousLoading, new PropertyValue((bool)newValue), (bool)newValue);
             }
@@ -268,15 +250,6 @@ namespace Tizen.NUI.BaseComponents
             var imageView = (ImageView)bindable;
             if (newValue != null)
             {
-                if (oldValue != null)
-                {
-                    bool oldBool = (bool)oldValue;
-                    bool newBool = (bool)newValue;
-                    if (oldBool == newBool)
-                    {
-                        return;
-                    }
-                }
                 // Note : We need to create new visual if previous visual was async, and now we set value as sync.
                 imageView.UpdateImage(ImageVisualProperty.SynchronousLoading, new PropertyValue((bool)newValue), (bool)newValue);
             }
@@ -298,15 +271,6 @@ namespace Tizen.NUI.BaseComponents
             var imageView = (ImageView)bindable;
             if (newValue != null)
             {
-                if (oldValue != null)
-                {
-                    bool oldBool = (bool)oldValue;
-                    bool newBool = (bool)newValue;
-                    if (oldBool == newBool)
-                    {
-                        return;
-                    }
-                }
                 imageView.UpdateImage(ImageVisualProperty.OrientationCorrection, new PropertyValue((bool)newValue));
             }
         },
--- a/src/Tizen.NUI/src/public/XamlBinding/BindableObject.cs
+++ b/src/Tizen.NUI/src/public/XamlBinding/BindableObject.cs
@@ -310,7 +310,10 @@ namespace Tizen.NUI.Binding
                 }
 
                 object oldvalue = null;
-                BindablePropertyContext context = GetOrCreateContext(property);
+                // Let we don't call default value creater if it never be created.
+                // It will reduce many cases when PropertyChanging / PropertyChanged don't use oldValue.
+                // 2024-01-06 Implement by eunkiki.hong@samsung.com.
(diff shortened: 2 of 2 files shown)


hinohie:
Hello @elishateng @AdunFang.

I'd like to check whether this way to fix is looks okay.

To support it correctly, we'd better add some flag something like `IgnoreDefaultValueCreatorIfSetValueFirstTime` (Default is false) at `BindableProperty` class, and add that flags for all `BindableProperty` what NUI use default. But this class is relative with `Tizen.NUI.XamlBuild` repository, so I think add new flag / property is quite dangerous.

Could you check whether this PR has no problem even if we use XAML cases?
+ Could you check that add the property at `BindableProperty`?

elishateng:
> Hello @elishateng @AdunFang.
> 
> I'd like to check whether this way to fix is looks okay.
> 
> To support it correctly, we'd better add some flag something like `IgnoreDefaultValueCreatorIfSetValueFirstTime` (Default is false) at `BindableProperty` class, and add that flags for all `BindableProperty` what NUI use default. But this class is relative with `Tizen.NUI.XamlBuild` repository, so I think add new flag / property is quite dangerous.
> 
> Could you check whether this PR has no problem even if we use XAML cases?
> 
> * Could you check that add the property at `BindableProperty`?


Hi Mr. Hong,
Here is a sample of Text BindableProperty of TextLabel:

```.cs
        public static readonly BindableProperty TextProperty = BindableProperty.Create(nameof(Text), typeof(string), typeof(TextLabel), string.Empty, propertyChanged: (BindableProperty.BindingPropertyChangedDelegate)((bindable, oldValue, newValue) =>
        {
            var textLabel = (TextLabel)bindable;

            if (newValue is Selector<string> selector)
            {
                textLabel.TextSelector = selector;
            }
            else
            {
                textLabel.selectorData?.Text?.Reset(textLabel);
                textLabel.SetText((string)newValue);
            }
        }),
        defaultValueCreator: (BindableProperty.CreateDefaultValueDelegate)((bindable) =>
        {
            var textLabel = (TextLabel)bindable;

            // Do not try to get string if we know that previous text was empty.
            return textLabel.textIsEmpty ? "" : Object.InternalGetPropertyString(textLabel.SwigCPtr, TextLabel.Property.TEXT);
        }));
```


This is the result of my talk with Mr. Fang, let’s take TextProperty as an example: 
 
1. Initially the property.DefaultValue of Text is string.Empty,  we call SetValue(TextProperty), then the property.DefaultValue assigns to oldvalue, it may not the default value in Dali side.  //new TextLabel("Test")

2. After calling Setvalue again, if the property changed from dali side, like we do animation for the Text property, then the property value may different from previous value of SetValue.

3. In xaml binding, the property value also can changed from dali. It's same with above 2th item.

4. Above all, we recommand use DefaultValueCreator to get the real value from dali side.

hinohie:
@elishateng 
This patch only consider "SetValue" cases. (We will use DefaultCreator for GetValue cases)

Let we think about only `SetValue` cases.
`BindableObject` Try to get value from context, by `property.ValueGetter(this)` at `BindablePropertyContext GetOrCreateContext` line 903 or 907. And then, `context.Value` become sync with DALi side.

But... If we call `InternalSetValue` when `IsBound = false`. We set `context.Value` and send this value as `BindableProperty.PropertyChanged`'s oldValue.

The review point of this patch is, we can ignore that `oldValue` since NUI don't use it now.
(For xaml binding case, IsBound might be true so It will go to `SetValueCore`, where I make `ignoreDefaultValueCreator : false` now.)

This is the sample code for example,

```[code=c#]
            control.Position = new Position(0.0f, 0.0f);
            var position = control.Position;
            Tizen.Log.Error("NUI", $"Positoin : {position.X} {position.Y}\n");
            position = control.Position;
            Tizen.Log.Error("NUI", $"Positoin : {position.X} {position.Y}\n");
            control.Position = new Position(2.0f, 2.0f);
            position = control.Position;
            Tizen.Log.Error("NUI", $"Positoin : {position.X} {position.Y}\n");
```

Before this patch :
```
Before

ERROR: NUI: ViewBindableProperty.cs: PositionProperty(1018) > GetPosition
ERROR: NUI: ViewBindableProperty.cs: PositionProperty(1009) > SetPosition
ERROR: NUI: ViewBindableProperty.cs: PositionProperty(1018) > GetPosition
ERROR: NUI: ControlSample.cs: Activate(40) > Positoin : 0 0
ERROR: NUI: ViewBindableProperty.cs: PositionProperty(1018) > GetPosition
ERROR: NUI: ControlSample.cs: Activate(42) > Positoin : 0 0
ERROR: NUI: ViewBindableProperty.cs: PositionProperty(1018) > GetPosition
ERROR: NUI: ViewBindableProperty.cs: PositionProperty(1009) > SetPosition
ERROR: NUI: ViewBindableProperty.cs: PositionProperty(1018) > GetPosition
ERROR: NUI: ControlSample.cs: Activate(45) > Positoin : 2 2
```

After this patch :
```
ERROR: NUI: ViewBindableProperty.cs: PositionProperty(1009) > SetPosition
ERROR: NUI: ViewBindableProperty.cs: PositionProperty(1018) > GetPosition
ERROR: NUI: ControlSample.cs: Activate(40) > Positoin : 0 0
ERROR: NUI: ViewBindableProperty.cs: PositionProperty(1018) > GetPosition
ERROR: NUI: ControlSample.cs: Activate(42) > Positoin : 0 0
ERROR: NUI: ViewBindableProperty.cs: PositionProperty(1009) > SetPosition
ERROR: NUI: ViewBindableProperty.cs: PositionProperty(1018) > GetPosition
ERROR: NUI: ControlSample.cs: Activate(45) > Positoin : 2 2
```



Please check this PR only for `SetValue` cases.

PR #5866, opened 2024-01-08
[NUI] Allow to call FrameUpdate callback return value

hinohie:
DALi's FrameUpdateCallback have interface whether we need to update scene or not. Previously, we always return it as false.
It might make some render / update timing issue when App is paused.

To make ensure whether we need to update scene or not, Let we override the new API that app can control the return value.

Relative dali patch : https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/303947


JoogabYun (approved):

dongsug-song (approved):
approved

PR #5881, opened 2024-01-11
[StreamRecorder] Remove module

hsgwon:
### Description of Change ###

Remove deprecated StreamRecorder module

### API Changes ###

 - TCSACR-573


wiertel (approved):
every element of this module was marked with `Obsolete("Deprecated in API10; Will be removed in API12")` - removing it now in accordance with the deprecation policy

PR #5886, opened 2024-01-16
[Packagemanager] Throw an exception when failed to add event listener

jeremy-jang:
### Description of Change ###

Throw an exception when failed to add event listener

### API Changes ###

 - ACR: TCSACR-574


src/Tizen.Applications.PackageManager/Tizen.Applications/PackageManager.cs
```diff
@@ -72,6 +72,7 @@ public static class PackageManager
         /// <summary>
         /// InstallProgressChanged event. This event occurs when a package is getting installed and the progress of the request to the package manager is changed.
         /// </summary>
+        /// <exception cref="IOException">Thrown when failed to subscribe package manager event.</exception>
```
wiertel, line 75: Please rephrase all of the new `<exception` lines: Thrown when subscribing to package manager event failed.

wiertel (changes requested):
The `IOException` does not feel to be the best choice, but it usually is difficult to match native error codes perfectly to C# exception and it's pretty consistent with rest of the module. I assume that current implementation throws exactly that error and You've checked that. In that case it's justified.

src/Tizen.Applications.PackageManager/Tizen.Applications/PackageManager.cs
jeremy-jang: Done. Thanks.

wiertel (approved):

jeremy-jang:
> Generally event handler accessors should not throw, but if it can fail then the error should be reported and exception might be the best option to do that, so OK. The `IOException` does not feel to be the best choice, but it usually is difficult to match native error codes perfectly to C# exception and it's pretty consistent with rest of the module. I assume that current implementation throws exactly that error and You've checked that. In that case it's justified.

Yes. The native API returns this error in this case. So i choosed IOException.

gogo9216 (approved):

pjh9216 (approved):

PR #5889, opened 2024-01-18
[NUI] Do not check theme applied flag when set ThemeChangeSensitive

rabbitfor:
New theme can affect to the views that were not affected by previous theme.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

xiaowei-guan (commented):
Good

PR #5893, opened 2024-01-19
[NUI] Add some remarks at BaseHandle class about Dispose

hinohie:
Add some comments about BaseHandle behaviour.

Since NUI BaseHandle is totally depend on Dali side BaseHandle, We shoild notify ti NUI developer s.t. The mean of Dispose is not matched with usual case. So just add some remarks comment.


myroot (approved):

dongsug-song (approved):
approved

xiaowei-guan (commented):
Good

PR #5895, opened 2024-01-21
[NUI.Sample] Resolve memory leak for DisposeTest sample

hinohie:
There was leak issue since we don't call FreeHGlobal after AllocHGlobal.

And also, let we Dispose Renderer / Geometry / Shader / TextureSet what NUI create. Those classes memory might give confuse when we test the memory.


dongsug-song (approved):
approved

jaehyun0cho (approved):
LGTM

PR #5901, opened 2024-01-23
[NUI][API10] DetachAccessibleObject at Dispose time well

Seoyeon2Kim:
### Description of Change ###
There were several problems when we call DaliAccessibilityDetachAccessibleObject at Dispose(bool) function

 - It will detach accessibility even if view doesn't own the handle memory
 - Dispose(bool) can be called at worker thread (during GC) so we need to check thread here

Make to `DaliAccessibilityDetachAccessibleObject` function at Dispose(DisposeTypes) so that we can ensure it called at main thread clearly.

### API Changes ###
- N/A


Seoyeon2Kim:
This PR is based on the original commits :
https://github.com/Samsung/TizenFX/pull/5736
https://github.com/Samsung/TizenFX/pull/5136

hinohie (approved):
Looks god to me :)

PR #5902, opened 2024-01-23
[NUI] Call Mashal.FreeHGlobal() what we allocated

hinohie:
Let we free allocated heap memory by Mashal.

It might cause memory leak issue if user try to use VertexBuffer.

And also, let we also fix FrameBrokerBase implementation that might be the guideline who want to use custom Geometry class.


jaehyun0cho (approved):
LGTM

dongsug-song (approved):
approved

PR #5907, opened 2024-01-24
[NUI] Fix LinearLayout to update its child's MeasuredSizeHeight 0 correctly

jaehyun0cho:
LinearLayout updates its child's MeasuredSizeWidth 0 correctly but it does not update its child's MeasuredSizeHeight 0 due to the range typo.

The range typo has been fixed so child's MeasuredSizeHeight 0 is also handled correctly.

### API Changes ###

 - ACR:


hinohie (approved):

dongsug-song (approved):
approved

PR #5920, opened 2024-01-26
[NUI.WindowSystem] Add new API GetSourceMimetypes of KVMService

juns-kim:
KVMService.GetSourceMimetypes() returns mimetype list that current drag source offered.

## Description of Change ##
Introduce new API to get mimetype list that current drag source offered.

### Internal API Changes ###
Added:
 - List<string> GetSourceMimetypes(void)
   - method to get mimetype list that current drag source offered.


dongsug-song (approved):
approved

PR #5927, opened 2024-01-30
[NUI] Support FrameUpdateCallback get world transform

hinohie:
Let we bind some more codes that UpdateProxy in dali side already supported.

relative dali patch :
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/305266


dongsug-song (approved):
approved

PR #5934, opened 2024-02-01
[NUI] Make view-wrapper signal input as weak handle

hinohie:
Since we don't need to control the ownership of native memory at callback side, let we just keep whole input value memory ownership as false.

relative dali patch :
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/305409


jaehyun0cho (approved):
LGTM

PR #5936, opened 2024-02-01
[NUI][API11] Make view-wrapper signal input as weak handle

hinohie:
Since we don't need to control the ownership of native memory at callback side, let we just keep whole input value memory ownership as false.

relative dali patch :
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/305416


Seoyeon2Kim (approved):
LGTM. Same as DevelNUI branch one.

hinohie:
Close this PR due to backword stability. After merge this PR, we cannot use previous version dll.

PR #5939, opened 2024-02-06
[MediaContent] Deprecate meaningless media group enum

hsgwon:
### Description of Change ###

Deprecate meaningless media group enum

### API Changes ###

 - TCSACR-576


src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoColumnKey.cs
```diff
@@ -25,82 +27,117 @@ public enum MediaInfoColumnKey
         /// <summary>
         /// Display name.
         /// </summary>
+        [Obsolete("Deprecated since API12; Will be removed in API14.")]
```
wiertel, line 30: I would like a dot "." instead of colon ";" better. But it's not an issue.

wiertel (approved):

src/Tizen.Content.MediaContent/Tizen.Content.MediaContent/MediaInfoColumnKey.cs
hsgwon: OK. I'll change all the same statement in MediaContent module.

PR #5940, opened 2024-02-06
[NUI] Enable relayout only when layout is set for the layer.

JoogabYun:
### Description of Change ###

If you do not check this, all views on layers that do not have a layout set will be searched unnecessarily.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

Seoyeon2Kim (approved):
LGTM
This PR is to check `if(layer?.LayoutCount > 0)` :)

PR #5952, opened 2024-02-15
[NUI][API11] Fix error if LottieAnimationView set BackgroundData befo…

hinohie:
…re set URL

Since LottieAnimationView.URL property change whole Visual property by itself, CornerRadius / Borderline properties become ignored when we set URL.

```
lav = new LottieAnimationView();
lav.CornerRadius = new Vector4(100.0f, 100.0f, 100.0f, 100.0f);
lav.URL = "~~~.json"; ///< CornerRadius information ignored.
```

To avoid this case, let we ensure to set CornerRadius / Borderline if we set URL

Same PR with #5951

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/LottieAnimationView.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/LottieAnimationView.cs
@@ -198,6 +198,18 @@ namespace Tizen.NUI.BaseComponents
 
                 Image = map;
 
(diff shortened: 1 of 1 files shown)


dongsug-song (approved):
approved

jaehyun0cho (approved):
LGTM

huayongxu (commented):
lgtm

PR #5963, opened 2024-02-20
[NUI] Unified native singletone class instance getter

hinohie:
There was 2 way to get sigletone class - `Get()` method and `Instance` proeprty.

After merge #5683 PR, some singletone class' `Get()` method register registry at that moment. If someone call `Instance` after `Get()` call, the singletone class become GC, due to unknown reason.

To avoid this cases, let we make `Get()` method also use `Instance` property internally, and deprecate that method.

For watch the trends after change the singletone pattern, let we just print error log if some NUI developer implement wrong way in future.

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/KeyInputFocusManager.cs
+++ b/src/Tizen.NUI/src/internal/Common/KeyInputFocusManager.cs
@@ -21,6 +21,7 @@ namespace Tizen.NUI
 {
     internal class KeyInputFocusManager : BaseHandle
     {
+        private static readonly KeyInputFocusManager instance = KeyInputFocusManager.GetInternal();
         internal KeyInputFocusManager(global::System.IntPtr cPtr, bool cMemoryOwn) : base(cPtr, cMemoryOwn)
         {
         }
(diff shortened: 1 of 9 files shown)


dongsug-song (approved):
approved

src/Tizen.NUI/src/internal/Common/KeyInputFocusManager.cs
```diff
@@ -35,13 +36,30 @@ private KeyInputFocusManager() : this(Interop.KeyInputFocusManager.NewKeyInputFo
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
         }
 
-        public static KeyInputFocusManager Get()
```
myroot, line 38: 다른 클래스는 `Get()` 구현을 삭제하지 않고 제공하는데, 
여기서는 삭제하네요, 사용하는 곳이 없어서 그런걸까요? internal도 public 처럼 접근하는 TV.NUI나 FLUX등에서도 사용하지 않은지 체크는 해야할것 같네요

hinohie: 검색해본 결과, `KeyInputFocusManager` 를 사용하는 코드는 없는 것으로 확인되었습니다.
다만, 다른 class 와의 통일성을 위해서 `Get()` API 를 다시 살려두는 방향으로 구현하도록 하겠습니다.

Seoyeon2Kim (approved):
I have one question.
So, now we should give a guide to use `Instance` more than `Get()`, right?
After your reply, you can merge this PR. :)

hinohie:
@Seoyeon2Kim Yes. Since NUI use `Instance` keyword internally, we should guide to use Instance property, instead of Get(). The only exception case is TtsPlayer with create mode, but I think it is not a big deal :D

PR #5964, opened 2024-02-21
[NUI] Make ProcessorOnceEvent as WeakEvent

hinohie:
We use double buffered ProcessorOnceEvent, to avoid newly added event become ignored. But this patch might make some another issue if event removal case.

Most of case, event removal doing on Dispose cases. So if we remove the events only for 'next event' side, the disposed eventhanderl might be invoked.


dongsug-song (approved):
approved

PR #5965, opened 2024-02-21
[NUI] Resolve the potential issue when DisposeQueued handle get

hinohie:
It is possible that DisposeQueued handle be used as input of GetManagedBaseHandleFromRefObject.

In this case, even if app hold that BaseHandle, it will be Disposed by DisposeQueue. If then the returned item become disposed, and it might breakdown app logics.

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/Registry.cs
+++ b/src/Tizen.NUI/src/internal/Common/Registry.cs
@@ -82,14 +82,14 @@ namespace Tizen.NUI
                     var target = weakReference.Target;
 
                     BaseHandle ret = target as BaseHandle;
-                    if (target == null)
+                    if ((ret == null) || ret.Disposed || ret.IsDisposeQueued)
                     {
-                        // Special case. If WeakReference.Target is null, it might be disposed by GC. just Add forcely.
+                        // Special case. If WeakReference.Target is null or it might be disposed by GC, just remove previous item forcibly.
                         if (Instance._controlMap.TryRemove(refCptr, out weakReference) != true)
                         {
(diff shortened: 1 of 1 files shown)


dongsug-song (approved):
approved

src/Tizen.NUI/src/internal/Common/Registry.cs
```diff
@@ -82,9 +82,9 @@ internal static bool Register(BaseHandle baseHandle)
                     var target = weakReference.Target;
 
                     BaseHandle ret = target as BaseHandle;
-                    if (target == null)
+                    if (target == null || ret.Disposed || ret.IsDisposeQueued)
                     {
-                        // Special case. If WeakReference.Target is null, it might be disposed by GC. just Add forcely.
+                        // Special case. If WeakReference.Target is null, or it might be disposed by GC. just Add forcely.
```
Seoyeon2Kim, line 87: How about changing this description as below so as not to be confused?
```suggestion
                        // Special case. If WeakReference.Target is null or it might be disposed by GC, just add {something?} forcibly.
```
I think it'll be easier to understand if you just get rid of the dot in the middle for one sentence.
And please add something for `add` + {what} .

hinohie: Good suggestion! I change the comment as `remove previous item` instead of `add forcely`. The behavior under the comment remove the disposed item. So the keyword `remove` is more good for this cases.

huiyueun (approved):

Seoyeon2Kim (approved):
LGTM :)

PR #5966, opened 2024-02-21
[NUI][API11] Make ProcessorOnceEvent as WeakEvent

hinohie:
We use double buffered ProcessorOnceEvent, to avoid newly added event become ignored.

Due to this double buffered system, we dont need to worry about add / remove during invoke.
So ProcessorOnceEvent is good point to change as WeakEventHandler.

Since WeakEvent has bug for now but, it didn't change previous behavior.


dongsug-song (approved):
approved

PR #5969, opened 2024-02-21
[NUI] Make DisposeRecursively() full search of children list even if Disposed

hinohie:
There was guard code when we traversal children list. There was some bugs when we traversal disposed item. But now, this bug is not happend.

To make the behavior match with comment, let we recursive the tree even if self is disposed.


dongsug-song (approved):
approved

huiyueun (approved):

PR #5970, opened 2024-02-22
[NUI.Components] Guard more cases when Container disposed

hinohie:
It might be possible if Container is disposed before ScrollBase disposed.

If then, Several logics breakdown.

Note : Since ScrollBase is time critical logics. So it is heavy to ensure the Container exist at every Container usage.

Instead, let we just check Container is null at `OnScrollAnimationEnded` only, which user can call immediatly by `Dispose()` API.


dongsug-song (approved):
approved

Seoyeon2Kim (approved):
LGTM

JoogabYun (approved):

everLEEst (approved):
LGTM.

PR #5977, opened 2024-02-27
[NUI] Do not remove once event when ImageView is dispose

hinohie:
Remove event from global static handler is heavy operation.

We can detach whether the view is disposed or not.
So just check disposed is more chip rather than remove event callback.


dongsug-song (approved):
approved

huiyueun (approved):

PR #5981, opened 2024-03-03
[NUI] Supports update area for frame update

sparrow74:
Add updated area function to update-proxy for partial rendering.
It is to improve the performance of update area hint.

### DIFF
--- a/test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/FrameUpdateCallbackTest.cs
+++ b/test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/FrameUpdateCallbackTest.cs
@@ -74,6 +74,7 @@ namespace Tizen.NUI.Samples
             private uint containerId;
             private List<uint> viewId; ///< View ID in the container.
             private List<float> viewPosition;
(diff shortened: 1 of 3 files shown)


hinohie (approved):

test/Tizen.NUI.Samples/Tizen.NUI.Samples/Samples/FrameUpdateCallbackTest.cs
```diff
@@ -620,6 +645,9 @@ private bool OnTouch(object source, View.TouchEventArgs e)
                     SetFrameUpdateCallback(position.X);
                 }
 
+                Vector4 updatedArea = new Vector4(position.X - 10.0f, position.Y - 10.0f, 20.0f, 20.0f);
```
hinohie, line 648: I just give approve but, is this code valid item for this sample?

JoogabYun (approved):

sparrow74:
Hi All,
I got offline review with Mrs Shin. After that, this way is incorrect for user convenience.
So, It will be abandon and I will update new PR.

PR #5982, opened 2024-03-05
[NUI] Fix SVACE issue

dongsug-song:
### Description of Change ###

[NUI] Fix SVACE issue
https://analysishub.sec.samsung.net/dm/tizen/sb2/main/review#PRJID=1284&WGID=711497

### API Changes ###
nothing

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyle.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/Style/ViewStyle.cs
@@ -566,13 +566,13 @@ namespace Tizen.NUI.BaseComponents
         {
             var source = other as ViewStyle;
 
-            IncludeDefaultStyle = source.IncludeDefaultStyle;
-            SolidNull = source.SolidNull;
-
             if (source == null || source.DirtyProperties == null || source.DirtyProperties.Count == 0)
             {
                 return;
             }
+
+            IncludeDefaultStyle = source.IncludeDefaultStyle;
+            SolidNull = source.SolidNull;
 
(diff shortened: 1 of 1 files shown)


taehyub (approved):
LGTM

jaehyun0cho (commented):
I think that source? is not required because it is not null.

dongsug-song:
> I think that source? is not required because it is not null.

Thank you for review.
could you teach me why the `source` never become null?

jaehyun0cho:
> > I think that source? is not required because it is not null.
> 
> Thank you for review. could you teach me why the `source` never become null?

In the following case, if source == null is true, then it returns.
So source is not null when source?.DirtyProperties == null and source?.DirtyProperties?.Count == 0 is checked.
In the same way, source?.DirtyProperties is not null when source?.DirtyProperties?.Count == 0 is checked.
```
if (source == null || source?.DirtyProperties == null || source?.DirtyProperties?.Count == 0)
```

dongsug-song:
> > > I think that source? is not required because it is not null.
> > 
> > 
> > Thank you for review. could you teach me why the `source` never become null?
> 
> In the following case, if source == null is true, then it returns. So source is not null when source?.DirtyProperties == null and source?.DirtyProperties?.Count == 0 is checked. In the same way, source?.DirtyProperties is not null when source?.DirtyProperties?.Count == 0 is checked.
> 
> ```
> if (source == null || source?.DirtyProperties == null || source?.DirtyProperties?.Count == 0)
> ```

thank you for kind explanation.
I have fixed as you mentioned.
thank you again

jaehyun0cho (approved):
LGTM

PR #5983, opened 2024-03-06
[NoACR][MachineLearning.Inference] Fix bug using wrong handle

niley7464:
### Description of Change ###
This patch fixes the bug which use wrong handle.
GetTensorsCount should use TensorsInfo handle instead of TensorsData handle.

See also : #5800


again4you (approved):
Good Job!

PR #5996, opened 2024-03-13
[NUI.Scene3D] Add Rotation constructor using pitch,yaw,roll

huiyueun:
### Description of Change ###
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/307648/

Add Rotation constructor using pitch,yaw,roll

### API Changes ###

 - ACR:


Seoyeon2Kim (approved):
Now it's good.
But, later could you add an example about how to use this constructor and some use cases?

huiyueun:
> Now it's good. But, later could you add an example about how to use this constructor and some use cases?

Thank you, I'll add example with this rotation constructor.

PR #5892, opened 2024-01-19
[NUI] Fix some incorrect implements when we get Texture relative items

hinohie:
Since now we have a unified pattern to avoid memory leak problem between NUI and Native side handle count problem,
Let we change some mis-implementated codes relative with Texture.

PR #6003, opened 2024-03-18
[NUI][API11] Fix some memory leak codes what already fixed at API12

hinohie:
Let we backport some minor memory leak resolve PR what we already merged.

This PR squash 4 different patches.

---

Fix some incorrect implements when we get Texture relative items

Since now we have a unified pattern to avoid memory leak problem between NUI and Native side handle count problem,
Let we change some mis-implementated codes relative with Texture.

(Follow PR #5892)

Add some remarks at BaseHandle class about Dispose

(Follow PR #5893)

Resolve memory leak for DisposeTest sample

There was leak issue since we don't call FreeHGlobal after AllocHGlobal.

And also, let we Dispose Renderer / Geometry / Shader / TextureSet what NUI create. Those classes memory might give confuse when we test the memory.

(Follow PR #5895)

Call Mashal.FreeHGlobal() what we allocated

Let we free allocated heap memory by Mashal.

It might cause memory leak issue if user try to use VertexBuffer.

And also, let we also fix FrameBrokerBase implementation that might be the guideline who want to use custom Geometry class.

(Follow PR #5902)

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Wearable/src/public/Title.cs
+++ b/src/Tizen.NUI.Wearable/src/public/Title.cs
@@ -326,17 +326,24 @@ namespace Tizen.NUI.Components
             vertex3.position = new Vec2(0.5f, -0.5f);
             vertex4.position = new Vec2(0.5f, 0.5f);
 
-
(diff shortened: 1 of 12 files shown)


dongsug-song (approved):
approved

PR #6023, opened 2024-03-27
[Tizen.System.Storage] Register eventhandler if it is already existed

ul24:
### Description of Change ###

In the previous implementation, no event handler was registered within class Storage when it was registered by storage type.
As a result of that, when the USB (or SD card) was removed, it attempted to obtain the state from outside and provided incorrect information to the application because there is no storage already removed.
It registers event handler within class Storage when event handler was registered by storage type.

### DIFF
--- a/src/Tizen.System.Storage/Storage/Storage.cs
+++ b/src/Tizen.System.Storage/Storage/Storage.cs
@@ -59,7 +59,10 @@ namespace Tizen.System
             };
         }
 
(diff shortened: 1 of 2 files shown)


chanwoochoi (commented):
Before merging, must test csharp TCT for storaged API (Tizen.System.Storage). Please test TCT.

chanwoochoi (approved):

PR #6034, opened 2024-03-29
[NUI] Fix LottieAnimationView not to reference disposed delegate

jaehyun0cho:
Delegate rootCallback of LottieAnimationView is called by multi threads for lottie dynamic properties.

This causes the following issues.
- Delegate rootCallback of disposed LottieAnimationView can be referenced.
- Local variables in RootCallback is disposed by multi threads.

To resolve the above issues, the followings are done.
- Delegate rootCallback of LottieAnimationView is changed to be static.
- Local variables in RootCallback is not disposed.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #6056, opened 2024-04-09
[NUI] Add property NotifyAfterRasterization + Fix EnableFrameCache works well

hinohie:
Let we add some flag that notify to event/main thread every frames when given lottie image rasterize finished.

Until before, there was no way to notify render thread that lottie surface updated. So we make render thread always make rendered if lottie image used.

But sometimes, if some application want to render 30fps or less frame rate lottie image, 60fps rendering is not a neccessary behavior.

To give them a choice, we give the property, `NotifyAfterRasterization`.

If we set this flag, some 30fps lottie file will awake render thread every 30fps.

Instead, the awake message might give overhead. So if lottie file is 60fps, we'd better set this flag as false.

Below dali patch required:

* https://review.tizen.org/gerrit/c/platform/core/uifw/dali-adaptor/+/309120
* https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/309127
* https://review.tizen.org/gerrit/c/platform/core/uifw/dali-extension/+/309128


dongsug-song (approved):
looks good to me

Seoyeon2Kim (approved):

PR #6067, opened 2024-04-17
[NUI] Fix svace issue on Popup

rabbitfor:
### Description of Change ###
Add nullity check for `buttonStyle` when `ApplyStyle()` for `Popup`.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI.Components/Controls/Popup.cs
+++ b/src/Tizen.NUI.Components/Controls/Popup.cs
@@ -738,17 +738,21 @@ namespace Tizen.NUI.Components
                     ButtonHeight = (int)ppStyle.Buttons.SizeHeight;
                 }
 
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI.Components/Controls/Popup.cs
```diff
@@ -738,17 +738,19 @@ public override void ApplyStyle(ViewStyle viewStyle)
                     ButtonHeight = (int)ppStyle.Buttons.SizeHeight;
                 }
 
-                buttonStyle = (ButtonStyle)ppStyle.Buttons?.Clone();
```
jaehyun0cho, line 741: It seems that this variable is not local variable.

rabbitfor: Thank you. I renamed the variable!

hinohie (approved):
Looks good to me.

PR #6076, opened 2024-04-18
[NUI] Change LottieAnimationView's property and renaming summary at NUIApplication

Clarik:
### Description of Change ###
[NUI] Change LottieAnimationView's property and renaming summary at NUIApplication

### API Changes ###
Nothing

PR #6077, opened 2024-04-19
[NUI] Add IsUsingXaml flag in properties of View and AnimatedVectorImageView

dongsug-song:
### Description of Change ###
[NUI] Add IsUsingXaml flag in properties of View and AnimatedVectorImageView
 - this kind of fix will be done for other classes also step by step
 - this is for speed up the property usage performance for the APP that is not using bindable properties.
 
### API Changes ###
should be nothing

### DIFF
--- a/src/Tizen.NUI/Tizen.NUI.csproj
+++ b/src/Tizen.NUI/Tizen.NUI.csproj
@@ -5,7 +5,7 @@
         <LangVersion>8.0</LangVersion>
     </PropertyGroup>
     <PropertyGroup>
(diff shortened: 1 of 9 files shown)


hinohie (approved):
Looks good to me but.. Isnt it have any dependency with #6076 ?

dongsug-song:
> Looks good to me but.. Isnt it have any dependency with #6076 ?

no, I don't think so.

everLEEst (approved):
LGTM :)

PR #6127, opened 2024-05-08
[NUI] Clean BackgroundImage setter logic

hinohie:
Usually we load image load asynchronously as default.

(The only one except case is lottie.
But most of cases we don't need to consider
set BackgroundImage as Lottie file.)

So, for usual cases, we don't need to apply the informations whether we need to load image synchronously or not.

But previous implementation set `BackgroundImageSynchronousLoading` forcibly.

It might required `BackgroundImage` getter, what we don't need to know.

And also, it doesn't skip this getter/setter logic when the value is same as previous.

To avoid this useless visual creation, let we consider for backgroundImageSynchronousLoading property works well and works faster for usual cases, who don't set this property.

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -4780,15 +4780,18 @@ namespace Tizen.NUI.BaseComponents
             }
             set
             {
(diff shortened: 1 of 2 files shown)


dongsug-song (approved):
looks good to me

bshsqa (approved):
Go.

PR #6128, opened 2024-05-08
[NUI] Do not ask DALi to get BackgroundImage url

hinohie:
We don't need to ask DALi side always when we want to get the url of image.

Instead, just keep it at NUI View side, and use it.


dongsug-song (approved):
looks good to me

bshsqa (approved):
Go

PR #6129, opened 2024-05-09
[NUI] Add SetTransitionEffectOption for ImageView's transition effect

tscholb:
This method allows users to configure the blending of two images(previous and currnet) using alpha values. \
if only the transition effect is set to true, the default values are used. 
however, if options are configured through this API, the effect can be applied with the values desired by the user.

https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-toolkit/+/310863/
https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/310864/

### API Changes ###

 - ACR:


hinohie (approved):

taehyub (approved):
LGTM

PR #6150, opened 2024-05-31
[NUI] Unregister disposed / disposequeued BaseHandle and register again if need

hinohie:
It is possible that dispose queued basehandle try to get / register again due to some reason. Like, Native side object is alived but C# side has no reference.

We already to to detect that cases. But if BaseHandle added at DisposeQueue, we cannot avoid to call the Dispose(); after some event loops later, and then Unregister the handle.

To avoid some error cases, let we Unregister the basehandle synchronously and then re-register base on the new BaseHandle.

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/Registry.cs
+++ b/src/Tizen.NUI/src/internal/Common/Registry.cs
@@ -82,13 +82,13 @@ namespace Tizen.NUI
                     var target = weakReference.Target;
 
                     BaseHandle ret = target as BaseHandle;
-                    if ((ret == null) || ret.Disposed || ret.IsDisposeQueued)
+                    // Note : Do not use ret == null because BaseHandle override operator ==.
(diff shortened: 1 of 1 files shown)


dongsug-song (approved):
approved

src/Tizen.NUI/src/internal/Common/Registry.cs
```diff
@@ -82,13 +82,13 @@ internal static bool Register(BaseHandle baseHandle)
                     var target = weakReference.Target;
 
                     BaseHandle ret = target as BaseHandle;
-                    if ((ret == null) || ret.Disposed || ret.IsDisposeQueued)
+                    // Note : Do not use ret == null because BaseHandle override operator ==.
```
Seoyeon2Kim, line 85: It's really important for the future!
Thanks to add it and I hope everyone know this note !

Seoyeon2Kim (approved):
LGTM~~

PR #6152, opened 2024-06-03
[TCSACR-584][NUI] Window.Instance is deprecated and Default is newly added

dongsug-song:
### Description of Change ###
[NUI] Window.Instance is deprecated and Default is newly added

### API Changes ###
https://jira.sec.samsung.net/browse/TCSACR-584

### DIFF
--- a/src/Tizen.NUI/src/public/Window/Window.cs
+++ b/src/Tizen.NUI/src/public/Window/Window.cs
@@ -416,13 +416,20 @@ namespace Tizen.NUI
             BottomRight = 8,
         }
 
-
         /// <summary>
         /// The stage instance property (read-only).<br />
         /// Gets the current window.<br />
         /// </summary>
-        /// <since_tizen> 3 </since_tizen>
+        [Obsolete("This has been deprecated in API12, please use Default instead")]
         public static Window Instance { get; internal set; }
+
+        /// <summary>
+        /// Gets the default window.
+        /// The main window or default window is automatically created when the application is launched, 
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Window/Window.cs
```diff
@@ -424,6 +424,13 @@ public enum ResizeDirection
         /// <since_tizen> 3 </since_tizen>
         public static Window Instance { get; internal set; }
 
+
```
wiertel, line 427: un

src/Tizen.NUI/src/public/Window/Window.cs
```diff
@@ -424,6 +424,13 @@ public enum ResizeDirection
...
+        /// <summary>
+        /// Gets the default window.
```
wiertel, line 429: you might consider adding sth more, like "Gets the default window - the initial window of the application.", maybe terms like "main window", "top window" do not fit or maybe the do. I just would prefer to have some explanation added about how it is default, which one is it, does it change, it is always available? On the other hand most users will find that they should just use this "Window.Default" and that will be all they need to know.

also, the extra empty line 427 is not needed

dongsug-song: @wiertel 
Thank you for reviewing. I have changed it according to your comments.

wiertel (approved):

jaehyun0cho (approved):
LGTM

PR #6159, opened 2024-06-10
[NUI][API10] Fix memory leak when DragAndDrop try to get Position

hinohie:
Since DragEvent's GetPosition return new Vector2 class internally, we should keep the positoin with memory ownership.


JoogabYun (approved):

Seoyeon2Kim (approved):
LGTM! for API10

PR #6167, opened 2024-06-11
[NUI][API11] Fix worker thread handle removal issue

hinohie:
Since `ControlHandle` could be deleted at GC thread, we should use DisposeQueue here.


dongsug-song (approved):
great job!
approved

PR #6168, opened 2024-06-11
[NUI] Fix worker thread handle removal issue

hinohie:
Since `ControlHandle` could be deleted at GC thread, we should use DisposeQueue here.


dongsug-song (approved):
approved

JoogabYun (approved):

Seoyeon2Kim:
Lukasz( @lol-github ) and Maria (@m-bialota ),
Could you check this PR?
It solves the crash issue!

PR #6169, opened 2024-06-12
[NUI.WindowSystem] Add new API PerformDrop(target) of KVMService

juns-kim:
## Description of Change ##
Introduce new API PerformDrop(DropTarget target) and enum DropTarget
PerformDrop receives target parameter to specify the drop target.
drop target can be following enum valuse :
 - KVMServiceWin : Drops directly onto the KVM Service Window.
 - UnderPointer: Drops onto the window located immediately under the mouse pointer.

### Internal API Changes ###

Added:
 - enum DropTarget.KVMServiceWin
 - enum DropTarget.UnderPointer
 - void PerformDrop(DropTarget target);


dongsug-song (approved):
approved

PR #6188, opened 2024-06-27
[NUI] Remove build warning

hinohie:
1. Remove XML parsing error
2. Let we use SafeHandle.Dispose(bool), instead of Dispose() API directly.
3. Add 'new' keyword what we re-define properties intention.
4. Remove unmatched property name with it's proeprty.


jaehyun0cho (approved):
LGTM

Seoyeon2Kim (approved):
LGTM
OH... `/// </sumary>`.......

PR #6199, opened 2024-07-02
[NUI] Remove unused macro

hinohie:
Their was mis-removal at PR https://github.com/Samsung/TizenFX/pull/4764.

NUI_ANIMATION_PROPERTY_CHANGE_1 macro already works well for every cases.

Now we can always enable them.


dongsug-song (approved):
approved

PR #6204, opened 2024-07-03
[NUI] Make ProcessController.Initialized check as static + Initialize at Instance getter

hinohie:
Let we don't get the instance of ProcessController at DisposeQueue.

Since DisposeQueue could be works at worker thread, or execute before NUIApplication initialized, we need to check whether ProcessController was initialized or not.

### DIFF
--- a/src/Tizen.NUI/src/internal/Application/Application.cs
+++ b/src/Tizen.NUI/src/internal/Application/Application.cs
@@ -684,12 +684,6 @@ namespace Tizen.NUI
             DisposeQueue.Instance.Initialize();
             Tizen.Tracer.End();
 
-            Log.Info("NUI", "[NUI] OnApplicationInit: ProcessorController Initialize");
-            Tizen.Tracer.Begin("[NUI] OnApplicationInit: ProcessorController Initialize");
-            // Initialize ProcessorController Singleton class. This is also required to create ProcessorController on main thread.
-            ProcessorController.Instance.Initialize();
-            Tizen.Tracer.End();
-
             Log.Info("NUI", "[NUI] OnApplicationInit: GetWindow");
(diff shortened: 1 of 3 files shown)


jaehyun0cho (approved):
LGTM

rookiejava (approved):

src/Tizen.NUI/src/internal/Application/Application.cs
```diff
@@ -684,12 +684,6 @@ private void OnApplicationInit(IntPtr data)
             DisposeQueue.Instance.Initialize();
             Tizen.Tracer.End();
 
-            Log.Info("NUI", "[NUI] OnApplicationInit: ProcessorController Initialize");
-            Tizen.Tracer.Begin("[NUI] OnApplicationInit: ProcessorController Initialize");
-            // Initialize ProcessorController Singleton class. This is also required to create ProcessorController on main thread.
-            ProcessorController.Instance.Initialize();
```
jaehyun0cho, line 690: Is it ok if ProcessorController is not initialized here?

hinohie: Yes. We don't need to initialize it now. Previously this API was exist due to we need to assume that ProcessController callback registered at main thread.

But mostly it was unneccessary worries, and if someone create it before application initialized, applicaton would be crashed, since API9. So I think we can remove it.

dongsug-song (approved):
approved

dongsug-song:
@hinohie 
Thank you for this patch. 
by the way, can you apply this change into tizen_8.0 with the dali-csharp-binder patch?
because, as you know, this is related to NUI2.
thank you.

hinohie:
@dongsug-song  Hello world.

dali-csharp-binder is not a mandantory. And I already merge same patch into API11. That single PR #6205 is enough :D

PR #6205, opened 2024-07-03
[NUI][API11] Make ProcessController.Initialized check as static + Initialize at Instance getter

hinohie:
Let we don't get the instance of ProcessController at DisposeQueue.

Since DisposeQueue could be works at worker thread, or execute before NUIApplication initialized, we need to check whether ProcessController was initialized or not.


jaehyun0cho (approved):
LGTM

PR #6209, opened 2024-07-11
[NUI] Support Shadow and ColorVisual cutout policy

hinohie:
Let we support to BoxShadow and ColorVisual, to ignore rendering inside of View area.

- None is default, same as previous logics.

- CutoutView is cutout render area as bounding box of view. It is faster than CutoutViewWithCornerRadius.

- CutoutWithCornerRadius is cutout render area consider the corner radius of view.

Since BoxShadow use ColorVisual internally and it doesn't consider View's property,
their might be some limitations.

Limitation:

- CutoutViewWithCornerRadius result not be matched if View use BorderlineOffset > -1.0f

- We don't support this property for xaml yet.

Required dali patch :
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/314255

Sample result :

Animation7-ezgif com-optimize (1)

### DIFF
--- a/src/Tizen.NUI/src/public/ViewProperty/Shadow.cs
+++ b/src/Tizen.NUI/src/public/ViewProperty/Shadow.cs
@@ -39,6 +39,7 @@ namespace Tizen.NUI
         public Shadow() : base()
         {
             BlurRadius = 0;
+            CutoutPolicy = ColorVisualCutoutPolicyType.None;
             Color = defaultColor;
         }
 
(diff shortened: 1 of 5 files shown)


dongsug-song:
Hello, There was a diff and conflict between DevelNUI and master, causing difficulties in integrating NUI. Therefore, a hard reset was performed on DevelNUI. As a result, the pull requests (https://github.com/Samsung/TizenFX/pulls) in DevelNUI are now conflicting. To resolve this, you can go to the specific PR, where you will see an error indicating a conflict. You can then use the command line to resolve the conflict.
or Here is another example. This one updates everything below the topmost patch, leaving only the topmost patch unchanged.

git fetch upstream
git rebase --onto=upstream/DevelNUI HEAD~1
git push {current working branch} -f

thank you.

dongsug-song (approved):
approved

jaehyun0cho (approved):
LGTM

PR #6248, opened 2024-08-01
[NUI] Make VisibilityChangedEventArgs.View return visibility changed view

hinohie:
Until now, VisibilityChangedEventArgs.View return sender itself.

It is useless information for user.

Follow up some other platform's behavior,
( reference : https://developer.android.com/reference/android/view/View#onVisibilityChanged(android.view.View,%20int) ) Let we make VisibilityChangedEventArgs.View return visibility changed view.

Relative dali patch :

- https://review.tizen.org/gerrit/c/platform/core/uifw/dali-core/+/315475
- https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/315484


dongsug-song (approved):
approved

bshsqa (approved):
Good To Go

PR #6269, opened 2024-08-14
[NUI] Fix SVACE

dongsug-song:
[NUI] Fix SVACE

- WID:52739702 Value eXamlData, which has null value, is dereferenced in member access expression eXamlData.Root

### Description of Change ###
nothing


hinohie (approved):
Looks good to me

PR #6272, opened 2024-08-16
[NUI.Devel.Tests] Do not use ColorFormat.RGBA, which might the behavor will be changed

hinohie:
Let we use BGRA8888 format to keep legacy behavior.


jaehyun0cho (approved):
LGTM

dongsug-song (approved):
approved

PR #6275, opened 2024-08-19
[TCSACR-586][NUI] Add IWindowProvider to be used as a common window interface across different modules

dongsug-song:
### Description of Change ###
[NUI] Add IWindowProvider to be used as a common window interface across different modules

- move Tizen.Multimedia.Common.IWindowProvider to Tizen.Common.IWindowProvider.
- The Window is used not only for Multimedia but also in other places (such as WindowSystem), so this change seems to be right.
- This will be review requested to Multimedia team and Window System team.
- ACR will be done and the link will be updated soon.
- The QuickPanel sample could not be tested because there is no mobile target for the tizen_9.0 version (e.g., no TM1, no FHUB).

### API Changes ###
https://jira.sec.samsung.net/browse/TCSACR-586

### DIFF
--- a/src/Tizen.Multimedia/Common/Display.cs
+++ b/src/Tizen.Multimedia/Common/Display.cs
@@ -15,6 +15,7 @@
  */
 using System;
 using ElmSharp;
+using Tizen.Common;
 
 namespace Tizen.Multimedia
 {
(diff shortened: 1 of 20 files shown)


jaehyun0cho (approved):
LGTM

PR #6276, opened 2024-08-19
[NUI] Fix lottie dynamic property crash

rabbitfor:
### Description of Change ###
NUI should not dispose the return value of dynamic property callback.

### API Changes ###

 - ACR:


hinohie (approved):
Looks good to me.. but why it makes crash? looks strange

jaehyun0cho (approved):
LGTM

rabbitfor:
In case the application creates only one property value and return the reference in the dynamic property callback, it will crash in second time becuase it is disposed in the first time.

hinohie:
I understand! Thank you for your comments. Double approved

JoogabYun (approved):

PR #6278, opened 2024-08-20
[Multimedia] Add MediaPacket TBM surface APIs

hsgwon:
### Description of Change ###

- Add TbmSurface property on MediaPacket to get native tbm surface handle
- Add HasTbmSurface property to check whether the packet have TBM surface or not

### API Changes ###

 - [TCSACR-590](https://jira.sec.samsung.net/browse/TCSACR-590)


src/Tizen.Multimedia/MediaTool/MediaPacket.cs
```diff
@@ -15,6 +15,7 @@
  */
 
 using System;
+using System.ComponentModel;
```
myroot, line 18: 이 using은 `[EditorBrowsable(EditorBrowsableState.Never)]`를 넣기 위해서 들어간 using으로 보여서, 이제 필요하지 않을듯합니다.

myroot (approved):
PR감사합니다.

src/Tizen.Multimedia/MediaTool/MediaPacket.cs
hsgwon: 넵 수정하겠습니다.

rookiejava (approved):

myroot (approved):

wiertel (commented):
The API is simple, the description is good, but IntPtr should be avoided in public API. Some solutions would be: wrap the pointer in SafeHandle (or derivative) - it is still a pointer, but this way it would at least have a type (like [SafeNativeWindowHandle](https://samsung.github.io/TizenFX/API9/api/Tizen.NUI.Window.SafeNativeWindowHandle.html)),  another idea is to expose an interface/class that would provide the pointer internally and the other side would consume that pointer internally so that the user would not touch the pointer - that one is possible only if the consuming part is somewhere in TizenFX.
Please share how the pointer is used / what API consumes it / what the app developer will need to do with it. I'm wondering if the usage scenario justifies having pointer in the API.

myroot:
> The API is simple, the description is good, but IntPtr should be avoided in public API. Some solutions would be: wrap the pointer in SafeHandle (or derivative) - it is still a pointer, but this way it would at least have a type (like [SafeNativeWindowHandle](https://samsung.github.io/TizenFX/API9/api/Tizen.NUI.Window.SafeNativeWindowHandle.html)), another idea is to expose an interface/class that would provide the pointer internally and the other side would consume that pointer internally so that the user would not touch the pointer - that one is possible only if the consuming part is somewhere in TizenFX. Please share how the pointer is used / what API consumes it / what the app developer will need to do with it. I'm wondering if the usage scenario justifies having pointer in the API.

Yes, I understand your concern, but It can't wrapping with managed type. because it directly used on Interop calls
We already have `IntPtr` type in public API 
for example, [EvasObject.Handle](https://github.com/Samsung/TizenFX/blob/d9b4f92c88eb73c7fa3afc93d85c55dc5d60cc19/src/ElmSharp/ElmSharp/EvasObject.cs#L207), [IWindowProvider.WindowHandle](https://github.com/Samsung/TizenFX/blob/d9b4f92c88eb73c7fa3afc93d85c55dc5d60cc19/src/Tizen/Tizen.Common/IWindowProvider.cs#L33)
It is same reason why IntPtr API was provided

We can consider, A managed TbmSurface class if we providing functionality of TbmSurface on C# API, but not now.

myroot:
We plan to provide a view that supporing tbm surface. [Here](https://github.sec.samsung.net/dotnet/nui2/pull/233)

We can update content of view with tbm_surface handle with below API
``` c#
        /// <summary>
        /// Sets the source of the TbmSurfaceView to the given tbm_surface_h.
        /// </summary>
        /// <param name="tbmSurface">The handle to the tbm_surface_h to set as the source.</param>
        /// <remarks>
        ///  Note that this method must be called on the UI thread only during the first call. Subsequent calls can be made from any thread.
        /// </remarks>
        public void SetSource(IntPtr tbmSurface);
```

wiertel:
Wanted to ensure that alternatives are considered. The chosen solution seems justified.

wiertel (approved):

PR #6292, opened 2024-08-27
[MetadataExtractor] Add a StitchedContent360 metadata API

hsgwon:
### Description of Change ###

Add a StitchedContent360 metadata API

### API Changes ###

 - TCSACR-594


src/Tizen.Multimedia.Metadata/MetadataExtractor/Metadata.cs
```diff
@@ -402,6 +403,16 @@ internal Metadata(MetadataExtractor extractor)
...
         public string Content360 { get; }
 
+        /// <summary>
+        /// Gets the information for stitched 360 content of the media.
+        /// </summary>
+        /// <value>
+        /// 0:None(There's no stitched 360 content metadata), 1:Non-stitched content, 2:stitched content
+        /// </value>
+        /// <since_tizen> 12 </since_tizen>
+        /// <value>A string representing the information for stitched 360 content, or null if the information does not exist.</value>
```
wiertel, line 413: The `<value>` tag has been added twice, please use only one.

src/Tizen.Multimedia.Metadata/MetadataExtractor/Metadata.cs
```diff
@@ -402,6 +403,16 @@ internal Metadata(MetadataExtractor extractor)
         /// <value>A string representing the information for 360 content, or null if the information does not exist.</value>
         public string Content360 { get; }
 
+        /// <summary>
+        /// Gets the information for stitched 360 content of the media.
+        /// </summary>
+        /// <value>
+        /// 0:None(There's no stitched 360 content metadata), 1:Non-stitched content, 2:stitched content
```
wiertel, line 410: is it 0 or None or maybe both for cases when content does not specify that information?
Could we have an enum with the listed possible values? That would make the description easier.  Or maybe nullable int (`int?`) could be used? That would also depend on underlying metadata stored.

src/Tizen.Multimedia.Metadata/MetadataExtractor/Metadata.cs
```diff
@@ -402,6 +403,16 @@ internal Metadata(MetadataExtractor extractor)
         /// <value>A string representing the information for 360 content, or null if the information does not exist.</value>
         public string Content360 { get; }
 
+        /// <summary>
+        /// Gets the information for stitched 360 content of the media.
```
wiertel, line 407: How does it work? A 360 video can be stitched or not, and this does not apply to any other media, right? Maybe a description like "Gets the information about 360 video content being stitched or not." or "Provides information whether a 360 video content is stitched."

wiertel (changes requested):

src/Tizen.Multimedia.Metadata/MetadataExtractor/Metadata.cs
hsgwon: I'll fix it.

src/Tizen.Multimedia.Metadata/MetadataExtractor/Metadata.cs
hsgwon: I'll change it as "Gets the information about 360 video content being stitched or not."

src/Tizen.Multimedia.Metadata/MetadataExtractor/Metadata.cs
hsgwon: I'll change it as `int?` type.

wiertel (approved):

PR #6297, opened 2024-08-29
[NUI] Support FrameSpeedFactor for AnimatedImage relative classes

hinohie:
Let we make some property to control the speed of animated image.

If we want to play animation 2x speed, Set this value as 2.0f.

If we want to play animation 0.5x speed, Set this value as 0.5f.

Note : For the platform safety reason, let we constraint the input value as 0.01f to 100.0f

Relative dali patch :

https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/316847


dongsug-song (approved):
approved

JoogabYun (approved):

PR #6303, opened 2024-09-04
[NUI] Deprecate Dali::Property constructor with componentwise

hinohie:
Since those usecase break some logic, let we decide to deprecate the componentewise logic.

Instead just create new property index - like COLOR_RED, SIZE_HEIGHT.

Relative dali patches (Not mandatory)
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-core/+/317090
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/317091
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/317087


dongsug-song (approved):
approved

everLEEst (approved):
Looks good to me and those API seem not ACR done so it looks safe to remove. but please make sure it won't used anywhere. thank you:)

PR #6342, opened 2024-09-23
[NUI] Add Outline Offset and BlurRadius at struct Outline

ANZ1217:
### Description of Change ###
Fixed a problem where the Offset and BlurRadius options of Outline did not exist in the member variable of Text.Outline Struct.


hinohie (approved):
Looks good to me.

wonrst (approved):

PR #6351, opened 2024-09-25
[NUI.Scene3D] Fix miss used method in SceneView

bshsqa:
- DeleteCameraProperty has been used for the case that should use DeleteCamera.

### API Changes ###

 - ACR:


AnglerLee (approved):
The issue of using DeleteCameraProperty instead of DeleteCamera in SceneView has been fixed. Thank you for your hard work.

hinohie (approved):
Looks good to me. Thank you for find it. It would be hard to found it in future :D

tscholb (approved):
LGTM :)

PR #6381, opened 2024-09-27
[NUI] Support Tbm dependency image source

hinohie:
Let we make some class that could create NativeImageSource / NativeImageQueue with raw-buffer-handle of tbm_surface_h / tbm_queue_h.

It will be used for some advanced tizen platform developer.

Relative dali patch :
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/318356


dongsug-song (approved):
looks good to me

JoogabYun (approved):

PR #6393, opened 2024-10-04
[MachineLearning.Train] Change internal logic to resolve a memory leak

SeoHyungjun:
### Description of Change ###
There were issues with heap use after free and heap overflow occurring (TNINE-4152, 4153, 4154).

In order to solve this problem, modified the internal logic of the API.

### API Changes ###
No API change.


devusr-sw-kim (approved):
Verified in ASAN built image C# test for those 3 ASAN issues. Not anymore those asan issues are repoted.

PR #6399, opened 2024-10-10
[Tizen.System] Enhance API description

germonado:
### Description of Change ###

Add more detailed API descriptions to the modules below.
- Tizen.System.Device
- Tizen.System.Feedback

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.System.Feedback/Feedback/Feedback.cs
+++ b/src/Tizen.System.Feedback/Feedback/Feedback.cs
@@ -413,6 +413,7 @@ namespace Tizen.System
         /// bool res = feedback.IsSupportedPattern(FeedbackType.Vibration, "Tap");
         /// </code>
         /// </example>
+        /// <seealso cref="FeedbackType"/>
         public bool IsSupportedPattern(FeedbackType type, String pattern)
         {
             bool supported = false;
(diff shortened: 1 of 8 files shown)


src/Tizen.System.Feedback/Feedback/Feedback.cs
```diff
@@ -596,7 +601,7 @@ public uint GetCountOfThemeInternal(FeedbackType type)
         }
 
         /// <summary>
-        /// Gets the id of theme selected.
+        /// Gets the current ID of the theme selected from available themes described in the conf file.
```
ppdl, line 604:
```suggestion
        /// Gets the current id of the theme selected from available themes described in the conf file.
```

Keep consistency, id or ID?

src/Tizen.System/Device/Battery.cs
```diff
@@ -76,11 +76,20 @@ public static class Battery
     {
         private static readonly object s_lock = new object();
         /// <summary>
-        /// Gets the battery charge percentage.
+        /// Gets the current device'sinvalid battery charge percentage as an interger value.
```
ppdl, line 79:
```suggestion
        /// Gets the current device's invalid battery charge percentage as an interger value.
```

chanwoochoi (approved):

src/Tizen.System.Feedback/Feedback/Feedback.cs
germonado: I changed as "id" to keep consistency thanks.

src/Tizen.System/Device/Battery.cs
germonado: I changed that description as you suggested.

ppdl (approved):

chanwoochoi (approved):

PR #6403, opened 2024-10-10
[NUI] Revert ApplyCornerRadius access level as internal

hinohie:
Since protected level make runtime error for TV FLUX project, let we revert it as internal again.


dongsug-song (approved):
approved

tscholb (approved):

PR #6413, opened 2024-10-16
[NUI] Fix Animation issue for Background and BoxShadow color

hinohie:
Since dali_2.3.41, MixColor property animate by Vector4 type. But NUI animation property helper doesn't apply those changeness.

To match the previous behavior, let we make BackgroundColor and BoxShadow.Color animate by Color class well.


dongsug-song (approved):
approved

jaehyun0cho (approved):
LGTM

PR #6436, opened 2024-11-01
[NUI] Remove build warnings at DragAndDrop + TextConstants

hinohie:
1. Fix typo error (navtive -> native)
2. Vector2 is already nullable class. We don't need to make it as Vector2? type.


Seoyeon2Kim (approved):
LGTM :) Good point

jaehyun0cho (approved):
LGTM

PR #6456, opened 2024-11-19
[NUI] Block native callback if view is disposed or disposing

hinohie:
For now, let we just check for View, ImageView, TextLabel, TextField, TextEditor


jaehyun0cho (approved):
LGTM

Seoyeon2Kim (approved):
LGTM..!!!

PR #6419, opened 2024-10-22
[NUI] Remove the legacy code of AccessibilityManager

Seoyeon2Kim:
### Description of Change ###
- `AccessibilityManager` hasn't been used anymore after Tizen 7.0 version.
- Now, all the legacy codes of AccessibilityManager should be removed in order to prevent the confusion.
- All Accessibility APIs are not public ones because they have never done ACR before, and that's why we can remove the codes like this.

* dali related patch : https://review.tizen.org/gerrit/#/c/platform/core/uifw/dali-csharp-binder/+/319388/

### API Changes ###
- APIs related to `AccessibilityManager` class
- Some APIs are removed, but they're not public ones on the platform.

PR #6475, opened 2024-11-27
[NUI] Match ViewImpl Connector parameter with csharp binder

hinohie:
After PR #6419 merged, some director callback parameter not matched with dali-csharp-binder.

We also need to remove delegater 26, 28, and 29 at binder level.


JoogabYun (approved):

Seoyeon2Kim (approved):
Thank you ㅠㅠㅠㅠㅠ

PR #6482, opened 2024-11-27
Update actions for main branch 

sung-su:
### Description of Change ###

- Update Action according to the change of the default branch name from `master` to `main`.
  - This should be merged after the default branch name is changed to `main`.
- Removed `Download Base API Spec` step on build-PR actions.

### DIFF
--- a/.github/workflows/build-pull-request.yml
+++ b/.github/workflows/build-pull-request.yml
@@ -3,7 +3,7 @@ name: "Build Pull Request"
 on:
   pull_request:
     branches:
-    - master
+    - main
     - DevelNUI
     - API*
   workflow_dispatch:
--- a/.github/workflows/build-pull-request.yml
+++ b/.github/workflows/build-pull-request.yml
@@ -78,44 +78,25 @@ jobs:
         name: reports
         path: Artifacts
 
-    # - name: Download Base API Spec
-    #   id: base-spec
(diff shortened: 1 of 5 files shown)


JoonghyunCho:
Any reason you left `Download Base API Spec` as commented ?

sung-su:
Update done.

.github/workflows/build-pull-request.yml
```diff
@@ -78,37 +78,25 @@ jobs:
...
-      uses: TizenAPI/tizenfx-build-actions/s3-download-public@master
-      with:
-        region: ap-northeast-2
-        bucket: tizenfx-api-specs
-        key: ${{ github.event.pull_request.base.sha }}.json
-        path: Artifacts/base-api.json
-
-    - uses: actions/checkout@v3
-      if: ${{ steps.base-spec.outputs.code != 200 }}
+    - name: PR event trigger
```
JoonghyunCho, line 81: Maybe it is better to name it something like "Checkout to base spec"? "Checkout to base  API spec"?

sung-su: OK `Checkout to base spec` is better

JoonghyunCho: Thanks!

PR #6486, opened 2024-12-02
[NUI] Block native callback if window is disposed or disposing

JoogabYun:
### Description of Change ###

Block native callback if window is disposed or disposing

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #6489, opened 2024-12-02
[NUI] Block native callback if InputMethodContext is disposed

jaehyun0cho:
Block native callback if InputMethodContext is disposed

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #6501, opened 2024-12-06
[NUI] Make Window events safe while window is being collected

rabbitfor:
### Description of Change ###
Ref : https://github.sec.samsung.net/NUI/OneUIComponents/issues/6

* Make Windows's native event callbacks not collected before native window is released.
* Add guard for window callbacks to check disposal.

### API Changes ###

 - ACR:

PR #6503, opened 2024-12-09
[NUI] Make Window events safe while window is being collected

rabbitfor:
### Description of Change ###
Ref : https://github.sec.samsung.net/NUI/OneUIComponents/issues/6

Make Windows's native event callbacks not collected before native window is released.
Add guard for window callbacks to check disposal.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window/WindowEvent.cs
+++ b/src/Tizen.NUI/src/public/Window/WindowEvent.cs
@@ -20,6 +20,7 @@ using System.ComponentModel;
 using System.Runtime.InteropServices;
 using Tizen.NUI.BaseComponents;
 using System.Collections.Generic;
+using System.Diagnostics;
 
 namespace Tizen.NUI
 {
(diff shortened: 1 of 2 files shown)


everLEEst (approved):
Looks good to me.

src/Tizen.NUI/src/public/Window/WindowEvent.cs
```diff
@@ -2021,10 +2173,24 @@ public event EventHandler<AccessibilityHighlightEventArgs> AccessibilityHighligh
                 {
                     using WindowAccessibilityHighlightEvent signal = new WindowAccessibilityHighlightEvent(Interop.WindowAccessibilityHighlightSignal.GetSignal(SwigCPtr), false);
                     signal.Ensure()?.Disconnect(AccessibilityHighlightEventCallback);
-                    AccessibilityHighlightEventCallback = null;
+                    ReleaseSafeCallback(ref AccessibilityHighlightEventCallback);
                 }
             }
         }
+
+        void CreateSafeCallback<T>(T method, out T safeCallback) where T : Delegate
```
Seoyeon2Kim, line 2181: To maintain consistency with NUI coding rule, could you add `internal` or `private` to these functions here?

Seoyeon2Kim (commented):
One minor comment

src/Tizen.NUI/src/public/Window/WindowEvent.cs
rabbitfor: Thank you for your comment. I will make another commit to add `private` keyword because this patch is to sync with [this](https://github.com/Samsung/TizenFX/pull/6501)  patch in `DevelNUI` branch so it would not be good to have any changes with it.

Seoyeon2Kim: Oh Okay!

PR #6502, opened 2024-12-06
[NUI] Access managed fields only when explicit view disposal

rabbitfor:
### Description of Change ###

Ref : https://github.sec.samsung.net/NUI/OneUIComponents/issues/5

When a View becomes a target for GC and enters the dispose queue, the signal reference it had as a member may not be valid anymore.

#### Test
```C#
void StartTest()
{
    for (var i = 0; i < 1000; i++)
    {
        var view = new View();
        view.AccessibilityGestureInfoReceived += (s, e) => {};
        view.AccessibilityDescriptionRequested += (s, e) => {};
        view.AccessibilityNameRequested += (s, e) => {};
        view.AccessibilityReadingSkipped += (s, e) => {};
        view.AccessibilityReadingPaused += (s, e) => {};
        view.AccessibilityReadingResumed += (s, e) => {};
        view.AccessibilityReadingCancelled += (s, e) => {};
        view.AccessibilityReadingStopped += (s, e) => {};
    }

    Timer gctimer = new Timer(100);
    gctimer.Tick += (s, e) =>
    {
        FullGC();
        return false;
    };
    gctimer.Start();
}
```

### API Changes ###

 - ACR:

PR #6505, opened 2024-12-10
[NUI] Access managed fields only when explicit view disposal

rabbitfor:
### Description of Change ###
Ref : https://github.sec.samsung.net/NUI/OneUIComponents/issues/5
Ref : https://github.com/Samsung/TizenFX/pull/6502

When a View becomes a target for GC and enters the dispose queue, the signal reference it had as a member may not be valid anymore.

### API Changes ###

 - ACR:


hinohie (approved):
Looks good to me. I think it will help a lots of unknown crash issue.

JoogabYun (approved):

PR #6513, opened 2024-12-11
[NUI] Support PropertyValue set/get without create new Vector2/3/4

hinohie:
Since NUI Vector2 / Vector3 / Vector4 are managed classes, create and dispose them might give overhead sightly.

If some memory/performance critical path don't want to create useless temperal Vector234 classes when we get/set PropertyValue, we need to support it.

Relative dali patch:

https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/316151


JoogabYun (approved):

dongsug-song (approved):
very good!
approved!

PR #6514, opened 2024-12-11
[NUI][API12] Support PropertyValue set/get without create new Vector2/3/4

hinohie:
This is cherry-pick patch of #6513

Since NUI Vector2 / Vector3 / Vector4 are managed classes, create and dispose them might give overhead sightly.

If some memory/performance critical path don't want to create useless temperal Vector234 classes when we get/set PropertyValue, we need to support it.

Relative dali patch:

https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/316158


dongsug-song (approved):
approved

PR #6524, opened 2024-12-18
[NUI][API12] Support RemoveIdle API for NUIApplication + Use unified idler callback

hinohie:
Let we keep Idler callback list as membery of internal Application class, and allow to remove them during idler callback execute.

# Changed Behavior
- We don't garentee that `Delegate` function emit time matched with `AddIdle` function call.
- We don't allow to add same `Delegate` function multiple times.  Until now, `Delegate` emitted as the number of `AddIdle` function call.


taehyub (approved):
LGTM

jaehyun0cho (approved):
LGTM

PR #6530, opened 2024-12-27
[NUI] Refactoring ControlState to use bitmask

everLEEst:
ControlState was implemented inefficently on the memory and the performance. This patch purposed to reduce inefficency by using bitflags on the state instead of string list. [https://github.sec.samsung.net/NUI/OneUIComponents/issues/15]

long type bitmask will be represent each states,
1 1 1 1 1
O S D P F

Normal   : 0L
Focused  : 1L
Pressed  : 2L
Disabled : 4L
Selected : 8L
Other    : 16L
and All  : 31L

This concept is based on VisualState of NUI2,
https://github.sec.samsung.net/dotnet/nui2/blob/main/src/Tizen.NUI2.Components/Base/ViewState.cs but we had to modified few states to keep backward compatibility of NUI ControlState.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/ControlStateUtility.cs
+++ b/src/Tizen.NUI/src/internal/Common/ControlStateUtility.cs
@@ -1,0 +1,73 @@
+/*
+ * Copyright(c) 2025 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+using System;
+using System.Collections.Generic;
+
+namespace Tizen.NUI
+{
+    /// <summary>
+    /// Manages state name and bit mask.
+    /// </summary>
+    internal static class ControlStateUtility
+    {
+        private const int MaxBitWidth = 62;
+        private static readonly Dictionary<string, ulong> registeredStates = new Dictionary<string, ulong>();
+        private static int nextBitPosition = 0;
+
+        /// <summary>
+        /// </summary>
+        public static ulong FullMask => (1UL << MaxBitWidth) - 1UL;
+
+        /// <summary>
+        /// </summary>
+        public static IEnumerable<(string, ulong)> RegisteredStates()
+        {
+
+            foreach (var (key, value) in registeredStates)
+            {
+                yield return (key, value);
+            }
+        }
+
+        public static ulong Register(string stateName)
+        {
+            if (stateName == null)
+                throw new ArgumentNullException($"{nameof(stateName)} cannot be null.", nameof(stateName));
+
+            if (string.IsNullOrWhiteSpace(stateName))
+                throw new ArgumentException($"{nameof(stateName)} cannot be whitespace.", nameof(stateName));
+
+            string trimmed = stateName.Trim();
+            ulong bitMask = 0UL;
+
+            if (trimmed != "Normal" && !registeredStates.TryGetValue(trimmed, out bitMask))
+            {
+                if (nextBitPosition + 1 > MaxBitWidth)
+                {
+                    throw new ArgumentException($"The given state name '{stateName}' is not acceptable since there is no more room to register a new state.");
+                }
+
+                bitMask = 1UL << nextBitPosition;
+                registeredStates.Add(trimmed, bitMask);
+                nextBitPosition++;
+            }
+
+            return bitMask;
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


everLEEst:
I've tested this on NUITizenGallery and works with no issue,
but we need to test more various cases including TCT and real applications.

Please check the state and bit max is good enough including Other states.

src/Tizen.NUI/src/internal/Common/ControlStateUtility.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public static ulong Register(string stateName)
+        {
+            if (stateName == null)
+                throw new ArgumentNullException($"{nameof(stateName)} cannot be null.", nameof(stateName));
+
+            if (string.IsNullOrWhiteSpace(stateName))
+                throw new ArgumentException($"{nameof(stateName)} cannot be whitespace.", nameof(stateName));
+
+            string trimmed = stateName.Trim().ToLowerInvariant();
```
rabbitfor, line 54: `ToLowerInvariant` would break the result of `ControlState.ToString()`.
Previsouly, the `ControlState` stores the trimmed string name without `ToLowerInvariant`.

https://github.com/Samsung/TizenFX/blob/dd3a51e968800a0b8f245eb7d635754e0a0c4a3d/src/Tizen.NUI/src/public/BaseComponents/ControlState.cs#L112-L128

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -212,18 +170,24 @@ public bool Equals(ControlState other)
...
-        public override int GetHashCode() => (name.GetHashCode() * 397) ^ IsCombined.GetHashCode();
+        public override int GetHashCode() => bitFlags.GetHashCode();
 
         ///  <inheritdoc/>
         [EditorBrowsable(EditorBrowsableState.Never)]
         public override string ToString()
         {
-            string name = "";
-            for (int i = 0; i < stateList.Count; i++)
+            var sbuilder = new StringBuilder();
```
rabbitfor, line 179: For the `bitflag = 0UL`, the method shoud returns "Normal" string.

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -86,20 +84,24 @@ public class ControlState : IEquatable<ControlState>
...
 
         /// <summary>
         /// Gets or sets a value indicating whether it has combined states.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public bool IsCombined => stateList.Count > 1;
+        public bool IsCombined => bitFlags != 0UL && Math.Ceiling(Math.Log2(bitFlags)) != Math.Floor(Math.Log2(bitFlags));
 
-        private ControlState() { }
+        ControlState(ulong bitMask)
```
rabbitfor, line 97: It'd be better to have explicit `private` keyword like other methods in this file.

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -138,7 +127,8 @@ public static ControlState Create(params ControlState[] states)
             if (states.Length == 1)
                 return states[0];
 
-            ControlState newState = new ControlState();
+            var newState = new ControlState(0UL);
```
rabbitfor, line 130: Unlike to NUI2, the `ControlState` is class type so it creates temporary `ControlState` object on heap when operating `+=`.
How about this?
```C#
public static ControlState Create(params ControlState[] states)
{
    ulong newbits = 0UL;

    for (int i = 0; i < states.Length; i++)
    {
        newbits |= states[i].bitFlags;
    }

    return new ControlState(newbits);
}
```

rabbitfor (approved):

src/Tizen.NUI/src/public/BaseComponents/ControlState.cs
```diff
@@ -86,20 +85,25 @@ public class ControlState : IEquatable<ControlState>
...
 
-        private List<ControlState> stateList = new List<ControlState>();
-        private readonly string name = "";
 
         /// <summary>
         /// Gets or sets a value indicating whether it has combined states.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public bool IsCombined => stateList.Count > 1;
+        public bool IsCombined => bitFlags != 0UL && Math.Ceiling(Math.Log2(bitFlags)) != Math.Floor(Math.Log2(bitFlags));
```
hinohie, line 97:
```suggestion
        public bool IsCombined => bitFlags != 0UL && ((bitFlags & (bitFlags - 1UL) != 0UL);
```

I dont like float point operation.

As I understand, this bool used if `bitFlags` has two or more bits. If than, we can detect that case by simple bitwise operation.

everLEEst: 이야... 안그래도 지윤님이랑 뭔가 더 쉬운 방법 있지 않을까 했는데 역시!!

dongsug-song (approved):
approved

everLEEst:
This patch is ready to review and merge.
with many tests of real components, it works as same as the previous implementation.
There are a few known issues,

1. Strictly speaking, it breaks backward compatibilities.
   In previous implementations, custom state can be added infinitely as the size of memory allows(as they use a list)
   In this implementation, we have restrictions of less than 57 states for customs as the bit flag size cannot be bigger than long.
   but 57 is big enough for custom states as we thought, so won't be an issue.

2. Some components have custom state-losing issues regardless of this patch applicant.
   this is caused by the wrong implementation of UpdateState method, so will be fixed another patch set.

PR #6537, opened 2024-12-31
[NUI] Fix wrong state update

everLEEst:
### Description of Change ###

State can be set individually with some custom states.
so this patch let state not assign directly, but add or remove from original state.

WARNING : this patch can be change the action of component or visual effect.
before we configure it clear, do not merge.

### API Changes ###

 - ACR:


everLEEst:
After discuss regarding this issue with @rabbitfor. we decided to keep current action for old component.
we need to make the new component in the right usage.

PR #6539, opened 2025-01-03
[Build] Modified the RIDs sorted in descending order

JongHeonChoi:
Before :
```
9.0.0:8.0.0:7.0.0:6.5.0:6.0.0:5.5.0:5.0.0:4.0.0:10.0.0
```

After :
```
10.0.0:9.0.0:8.0.0:7.0.0:6.5.0:6.0.0:5.5.0:5.0.0:4.0.0
```

### DIFF
--- a/packaging/makerid.py
+++ b/packaging/makerid.py
@@ -5,6 +5,9 @@ scrpit_dir = os.path.dirname(os.path.abspath(__file__))
 runtime_dir = os.path.join(scrpit_dir, "../pkg/Tizen.NET/runtime.json")
 spec_dir = os.path.join(scrpit_dir, "csapi-tizenfx.spec")
 
+def parse_rid_as_int(rid):
+    return [int(part) for part in rid.split(".")]
+
 with open(runtime_dir) as json_file:
     json_data = json.load(json_file)
     json_string = json_data["runtimes"]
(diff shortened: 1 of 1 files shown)


packaging/makerid.py
```diff
@@ -20,10 +21,16 @@
         key = key.replace("tizen.", "")
         key = key.replace("tizen", "")
         if key.strip():
-            rid_list.append(key)
-
-    rid_list = list(set(rid_list))
-    rid_list.sort(reverse=True)
+            if len(key) == 5:
```
JoonghyunCho, line 24: It doesn't seem like good idea to use the string count of rid(`9.0.0`) to sort `10.0.0`.
How about adding a function to make rid  as integer group(`[10, 0, 0]`) and compare rid as integer.
This will help increase maintainability for future rid, for example 11.0.0, 12.0.0, etc.
```python
def parse_rid_as_int(rid):
    return [int(part) for part in rid.split(".")]
...
rid_list = sorted(rid_list, key=parse_rid_as_int, reverse=True)
```

JongHeonChoi: Thank you for your good idea.

HJLeee (approved):

JoonghyunCho (approved):
Thanks for update :)

PR #6542, opened 2025-01-06
[NUI] Do not re-create visual if we change only transform

hinohie:
Since transform data is kind of renderer property internally, we don't need to re-create them.

Let we use UpdateProperty action for this case.

Required dali patch (merged) :
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/316868/4/dali-toolkit/internal/visuals/visual-base-impl.cpp

### DIFF
--- a/src/Tizen.NUI/src/public/Visuals/VisualObject/VisualBase.cs
+++ b/src/Tizen.NUI/src/public/Visuals/VisualObject/VisualBase.cs
@@ -92,12 +92,13 @@ namespace Tizen.NUI.Visuals
                 extraWidth = 0.0f;
                 extraHeight = 0.0f;
 
+                cachedVisualTransformPropertyMap?.Dispose();
                 cachedVisualTransformPropertyMap = null;
 
                 changed = true;
             }
 
-            internal void ConvertToPropertyMap()
+            internal PropertyMap ConvertToPropertyMap()
             {
                 if (cachedVisualTransformPropertyMap == null)
                 {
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Visuals/VisualObject/VisualBase.cs
```diff
@@ -967,6 +967,20 @@ internal void UpdateVisualPropertyMap()
...
+                transformInfo.changed = false;
+                cachedVisualPropertyMap?.Remove((int)Tizen.NUI.Visual.Property.Transform);
+
+                transformInfo.ConvertToPropertyMap();
+
+                if (transformInfo.cachedVisualTransformPropertyMap != null)
+                {
+                    using var transformValue = new PropertyValue(transformInfo.cachedVisualTransformPropertyMap);
+                    UpdateVisualProperty((int)Tizen.NUI.Visual.Property.Transform, transformValue, false);
+                }
```
rabbitfor, line 981: This comment is minor and is not related with the current patch but I'd like to suggest one for the future update.
`transformInfo.ConvertToPropertyMap` has dependancy with `transformInfo.cachedVisualTransformPropertyMap`. Which means once `ConvertToPropertyMap()` is called `cachedVisualTransformPropertyMap` can not be null. 
Their dependancy is not easly visible unless the developer sees the code of `VisualTransformInfo`.

In my opinion, rather than exposing a `cachedVisualTransformPropertyMap`, it would be better to return the map itself, whether it is cached one or not.
```C#
using var transformValue = new PropertyValue(transformInfo.GetPropertyMap());
```

rabbitfor (approved):

src/Tizen.NUI/src/public/Visuals/VisualObject/VisualBase.cs
rabbitfor: I gave the approvement anyway!

hinohie: Thank you for your comments! I agree with your opinion. So let I make `ConvertToPropertyMap` return `cachedVisualTransformPropertyMap` now. So nobody now don't touch `transformInfo.cachedVisualTransformPropertyMap` now.

everLEEst (approved):
Looks good to me.

PR #6543, opened 2025-01-06
[NUI] Remove unnecessary PInvoke calls

lwc0917:
### Description of Change ###

Code related to preload() will be updated later.

### API Changes ###
No Changes

### DIFF
--- a/src/Tizen.NUI/src/public/Common/NUIConstants.cs
+++ b/src/Tizen.NUI/src/public/Common/NUIConstants.cs
@@ -1323,210 +1323,99 @@ namespace Tizen.NUI
     /// <since_tizen> 3 </since_tizen>
     public struct ParentOrigin
     {
+        private static readonly Position topLeft = new Position(Top, Left, Middle);
+        private static readonly Position topCenter = new Position(Top, Middle, Middle);
+        private static readonly Position topRight = new Position(Top, Right, Middle);
+        private static readonly Position centerLeft = new Position(Middle, Left, Middle);
+        private static readonly Position center = new Position(Middle, Middle, Middle);
+        private static readonly Position centerRight = new Position(Middle, Right, Middle);
+        private static readonly Position bottomLeft = new Position(Bottom, Left, Middle);
+        private static readonly Position bottomCenter = new Position(Bottom, Middle, Middle);
+        private static readonly Position bottomRight = new Position(Bottom, Right, Middle);
+
         /// <summary>
         /// Top
         /// </summary>
         /// <since_tizen> 3 </since_tizen>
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Common/NUIConstants.cs
```diff
@@ -1327,206 +1327,85 @@ public struct ParentOrigin
...
-        public static float Top
-        {
-            get
-            {
-                float ret = Interop.NDalicParentOrigin.ParentOriginTopGet();
-                if (NDalicPINVOKE.SWIGPendingException.Pending) throw new InvalidOperationException("FATAL: get Exception", NDalicPINVOKE.SWIGPendingException.Retrieve());
-                return ret;
-            }
-        }
+        public const float Top = 0.0f;
```
rabbitfor, line 1330: This changes `Property` to `Field`. ACR required;

How about this? This way maintains the API as a `Property` and returns a constant value.
```C#
public static float Top => 0.0f;
```

src/Tizen.NUI/src/public/Common/NUIConstants.cs
```diff
@@ -1327,206 +1327,85 @@ public struct ParentOrigin
...
-        {
-            get
-            {
-                global::System.IntPtr cPtr = Interop.NDalicParentOrigin.ParentOriginTopLeftGet();
-                Position ret = (cPtr == global::System.IntPtr.Zero) ? null : new Position(cPtr, false);
-                if (NDalicPINVOKE.SWIGPendingException.Pending) throw new InvalidOperationException("FATAL: get Exception", NDalicPINVOKE.SWIGPendingException.Retrieve());
-                return ret;
-            }
-        }
+        public static readonly Position TopLeft = new Position(Top, Left, Middle);
```
rabbitfor, line 1360: To avoid ACR for the non-constant value, I think it will be better to create private field and return it.
```C#
private static readonly Position topLeft = new Position(Top, Left, Middle);
```
```C#
public static readonly Position TopLeft => topLeft;
```

rabbitfor: And it will be nice to add it to `Preload` sequence to initialize static private fields if possible.

lwc0917:
@rabbitfor 

I committed a pull request to get feedback on allowing what seems like a non-issue ABI break, as mentioned in the PR description.

Thank you for the great idea!

hinohie (approved):
Looks good to me. I think we can doing similar progress for `PivotPoint` somdays.

everLEEst (approved):
Looks good to me. nice work :)

PR #6544, opened 2025-01-06
[NUI] Add value type: color, vector2, shadow

rabbitfor:
### Description of Change ###
Introduce the value type of color, vector2 and shadow. They will replace existing reference types gradually.

### API Changes ###

 - ACR:


everLEEst:
As I locally tested, it works fine when only use Color with using Tizen.NUI.
and also works very well with Color and L.Color.
I suggest one more thing to put them in devel/LiteValue or something similar so all Structed Type gathered in same place we could move them in same place in the public or internal after all.

everLEEst (approved):
Looks good to me

jaehyun0cho (approved):
LGTM

PR #6546, opened 2025-01-06
[NUI] Add value type: control state

rabbitfor:
### Description of Change ###
Introduce the value type of ControlState. It replaces ControlState implementation.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
@@ -1575,6 +1575,8 @@ namespace Tizen.NUI.BaseComponents
                     ControlState += ControlState.Disabled;
                 }
             }
+
+            EnabledChanged?.Invoke(this, EventArgs.Empty);
         }
 
 
--- a/src/Tizen.NUI/src/devel/Base/DevelView.cs
+++ b/src/Tizen.NUI/src/devel/Base/DevelView.cs
@@ -1,0 +1,33 @@
+/*
+ * Copyright(c) 2025 Samsung Electronics Co., Ltd.
+ *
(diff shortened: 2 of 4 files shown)


src/Tizen.NUI/src/devel/LControlState.cs
```diff
@@ -0,0 +1,194 @@
...
+using System.Text;
+using System.ComponentModel;
+
+namespace Tizen.NUI.BaseComponents
+{
+    /// <summary>
+    /// Defines a value type of control state.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public readonly struct LControlState
```
everLEEst, line 28: As we discussed on offline,
It is slightly different with other L-Struct so I think it could be renamed something else. ControlState has this struct(has-a), so State, or StateValue could be the option.

rabbitfor: Please review the updated code. Thank you :)

src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
```diff
@@ -1556,6 +1556,8 @@ protected virtual bool HandleControlStateOnTouch(Touch touch)
         [EditorBrowsable(EditorBrowsableState.Never)]
         protected virtual void OnEnabled(bool enabled)
         {
+            EnabledChanged?.Invoke(this, EventArgs.Empty);
```
everLEEst, line 1559: I think this event need to be invoked at the end of funtion as we assign the ControlState in the line.

everLEEst (approved):
Looks good to me

lwc0917 (approved):
LGTM

PR #6548, opened 2025-01-06
[NUI] Reduce managed PropertyValue creation for VisualObject

hinohie:
Let we add 'using' mark for temperally used `PropertyValue`.

Optimize PropertyMap.Add() API if it is prepared

It is kind of next-step. So current PR could be keep going on.


rabbitfor (approved):

everLEEst (approved):
Looks good to me :) :+1:

PR #6550, opened 2025-01-07
[NUI.Devel.Tests] Fix mis-implements for NativeImageSource TCT

hinohie:
- We MUST call ReleaseBuffer after call AcquireBuffer
- We MUST call AcquireBUffer before call ReleaseBuffer

If not, exception or deadlock will occured.

### DIFF
--- a/test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/public/Images/TSNativeImageSource.cs
+++ b/test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/public/Images/TSNativeImageSource.cs
@@ -88,6 +88,7 @@ namespace Tizen.NUI.Devel.Tests
             try
             {
                 testingTarget.AcquireBuffer(ref width, ref height, ref stride);
+                testingTarget.ReleaseBuffer(); ///< Must call after AcquireBuffer
             }
             catch (Exception e)
             {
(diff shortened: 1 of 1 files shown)


jaehyun0cho (approved):
LGTM

test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/public/Images/TSNativeImageSource.cs
```diff
@@ -88,6 +88,7 @@ public void NativeImageSourceAcquireBuffer()
             try
             {
                 testingTarget.AcquireBuffer(ref width, ref height, ref stride);
+                testingTarget.ReleaseBuffer(); ///< Must call after AcquireBuffer
```
JoogabYun, line 91: This is a minor opinion.

If ReleaseBuffer is needed, in this TC it seems better to place ReleaseBuffer before Dispose() outside of try~catch.

try
{
  testingTarget.AcquireBuffer(ref width, ref height, ref stride);
} catch (Exception e) { ... }

testingTarget.ReleaseBuffer(); ///< Must call after AcquireBuffer
testingTarget.Dispose();

JoogabYun (approved):

test/Tizen.NUI.Tests/Tizen.NUI.Devel.Tests/testcase/public/Images/TSNativeImageSource.cs
JoogabYun: If AcquireBuffer succeeds and ReleaseBuffer fails, problems may arise. It would be better to keep it inside a try-catch like now. 

Please ignore my previous comment

hinohie: I though same thing. But also, if their was some exception occured, internal mutex "might" not be locked. In this case, we should not call ReleaseBuffer();

So I make it inside of try.

PR #6560, opened 2025-01-08
[NUI] Remove unnecessary PInvoke calls (PivotPoint, AnchorPoint, Position)

lwc0917:
### Description of Change ###
- PivotPoint, AnchorPoint, Position 

It related with below issue

### API Changes ###
No Changes

### DIFF
--- a/src/Tizen.NUI/src/public/Common/NUIConstants.cs
+++ b/src/Tizen.NUI/src/public/Common/NUIConstants.cs
@@ -1428,194 +1428,87 @@ namespace Tizen.NUI
         /// Top
         /// </summary>
         /// <since_tizen> 3 </since_tizen>
-        public static float Top
-        {
-            get
-            {
-                float ret = Interop.NDalicAnchorPoint.AnchorPointTopGet();
-                if (NDalicPINVOKE.SWIGPendingException.Pending) throw new InvalidOperationException("FATAL: get Exception", NDalicPINVOKE.SWIGPendingException.Retrieve());
-                return ret;
-            }
-        }
+        public static float Top => ParentOrigin.Top;
+
         /// <summary>
         /// Bottom
         /// </summary>
         /// <since_tizen> 3 </since_tizen>
-        public static float Bottom
-        {
-            get
-            {
-                float ret = Interop.NDalicAnchorPoint.AnchorPointBottomGet();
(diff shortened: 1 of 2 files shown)


lwc0917:
Please dont review yet :)

src/Tizen.NUI/src/public/Common/NUIConstants.cs
```diff
@@ -1428,194 +1428,87 @@ public struct PivotPoint
...
-        public static float Top
-        {
-            get
-            {
-                float ret = Interop.NDalicAnchorPoint.AnchorPointTopGet();
-                if (NDalicPINVOKE.SWIGPendingException.Pending) throw new InvalidOperationException("FATAL: get Exception", NDalicPINVOKE.SWIGPendingException.Retrieve());
-                return ret;
-            }
-        }
+        public static float Top => ParentOrigin.Top;
```
hinohie, line 1431: We cannot always assuemt that `PivotPoint.Top` is same as `ParentOrigin.Top`. But for not, they are eqaully 0.0f. So looks find for these days.

src/Tizen.NUI/src/public/Common/NUIConstants.cs
```diff
@@ -1428,194 +1428,87 @@ public struct PivotPoint
...
-        public static float Left
-        {
-            get
-            {
-                float ret = Interop.NDalicAnchorPoint.AnchorPointLeftGet();
-                if (NDalicPINVOKE.SWIGPendingException.Pending) throw new InvalidOperationException("FATAL: get Exception", NDalicPINVOKE.SWIGPendingException.Retrieve());
-                return ret;
-            }
-        }
+        public static float Left => ParentOrigin.Left;
```
hinohie, line 1443: We cannot always assuemt that `PivotPoint.Left` is same as `ParentOrigin.Left`. But for not, they are eqaully 0.0f. So looks find for these days.

hinohie (approved):
For now, `ParentOrigin.Top` and `ParentOrigin.Left` have same value with `PivotPoint.Top` and PivotPoint.Left`. So it might be okay. But don't miss if native side `AnchorPoint` constant be changed in future

rabbitfor (approved):
LGTM

PR #6561, opened 2025-01-08
[NUI] Remove unnecessary PInvoke calls related with axis

lwc0917:
### Description of Change ###
It related with below issue.

### API Changes ###
No Changes

### DIFF
--- a/src/Tizen.NUI/src/public/Common/Position.cs
+++ b/src/Tizen.NUI/src/public/Common/Position.cs
@@ -29,6 +29,8 @@ namespace Tizen.NUI
     [Tizen.NUI.Binding.TypeConverter(typeof(PositionTypeConverter))]
     public class Position : Disposable, ICloneable
     {
+        private static readonly Position one = new Position(1.0f, 1.0f, 1.0f);
+        private static readonly Position zero = new Position(0.0f, 0.0f, 0.0f);
 
         /// <summary>
         /// The constructor.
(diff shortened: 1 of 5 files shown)


src/Tizen.NUI/src/public/Common/Position.cs
```diff
@@ -29,6 +29,8 @@ namespace Tizen.NUI
     [Tizen.NUI.Binding.TypeConverter(typeof(PositionTypeConverter))]
     public class Position : Disposable, ICloneable
     {
+        private static readonly Vector3 one = new Vector3(1.0f, 1.0f, 1.0f);
+        private static readonly Vector3 zero = new Vector3(0.0f, 0.0f, 0.0f);
```
hinohie, line 33:
```suggestion
        private static readonly Position one = new Position(1.0f, 1.0f, 1.0f);
        private static readonly Position zero = new Position(0.0f, 0.0f, 0.0f);
```

Don't we need to use `Position` class for here, instead `Vector3`?

lwc0917: Thank you for suggestion

src/Tizen.NUI/src/public/Common/Vector3.cs
```diff
@@ -96,122 +99,50 @@ internal Vector3(Vector3ChangedCallback cb, float x, float y, float z) : this(In
...
-        {
-            get
-            {
-                global::System.IntPtr cPtr = Interop.Vector3.XaxisGet();
-                Vector3 ret = (cPtr == global::System.IntPtr.Zero) ? null : new Vector3(cPtr, false);
-                if (NDalicPINVOKE.SWIGPendingException.Pending) throw new InvalidOperationException("FATAL: get Exception", NDalicPINVOKE.SWIGPendingException.Retrieve());
-                return ret;
-            }
-        }
+        public static Vector3 XAxis => PositionAxis.X;
```
hinohie, line 108:
```suggestion
        private static readonly xaxis = new Vector3(1.0f, 0.0f, 0.0f);
        public static Vector3 XAxis => xaxis;
```

Can we freely convert from `Vector3` to `Position`? I think we'd better newly define `xaxis` for `Vector3` class

rabbitfor (approved):

everLEEst (approved):
Looks good to me

PR #6568, opened 2025-01-09
[NUI] Add markup for some view properties using value type

rabbitfor:
### Description of Change ###
Add markup for some view properties using value type.

This also includes PropertyMap optimizations to prevent value type being conveted to reference type again.

These markup extensions will be used in new component library.

image

```C#
NUIApplication.GetDefaultWindow().Add(new View()
    .BackgroundColor(0xFF00FF)
    .Size(100, 100)
    .Position(200, 200)
    .CornerRadius(0.5f, true)
    .BoxShadow(12, L.Color.Black.WithAlpha(0.5f), -10, 10)
    .Self(out var view));
```

* BackgroundColor(uint hex) : set background color with value types
* BackgroundColor(float r, float g, float b, float a) : set background color with value types
* Size(float w, float h) : set size with value types
* SizeWidth(float w) : set size with value types
* SizeHeight(float h) : set size with value types
* Position(float x, float y) : set position with value types
* PositionX(float x) : set position with value types
* PositionY(float y) : set position with value types
* CornerRadius(float uniform, bool isRelative = false) : set corner radius and policy at once
* CornerRadius(float tl, float tr, float bl, float br, bool isRelative = false) : set corner radius and policy at once
* BoxShadow(...) : set box shadow with value types
* Self(out View) : assign current view to local variable

⚠️ depends https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/317875

Please note that the main purpose of providing markup methods for view is to reduce creating heap objects such as Color, Size, Position and Shadow.
For the `CornerRadius`, this patch does not meet the main purpose because it is not easy to do it. I hope it will not create reference type of vector4 internally in the future.

### API Changes ###

 - ACR:


src/Tizen.NUI/src/devel/Lite/L.Vector2.cs
```diff
@@ -53,6 +53,8 @@ public float Y
             get;
         }
 
+        public readonly bool IsZero => X == 0 && Y == 0;
+
```
everLEEst, line 57: are there any plan to add operator overridings?
```suggestion
        public readonly bool IsZero => X == 0 && Y == 0;

        /// <summary>
        /// Compares two vectors to determine if they are equal.
        /// </summary>
        /// <param name="a">The first vector to compare.</
        /// <param name="b">The second vector to compare.</param>
        /// <returns>True if the vectors are equal, false otherwise.</returns>
        public static bool operator ==(Vector2 a, Vector2 b)
        {
            return (a.X == b.X) && (a.Y == b.Y);
        }

        /// <summary>
        /// Compares two vectors to determine if they are not equal.
        /// </summary>
        /// <param name="a">The first vector to compare.</param>
        /// <param name="b">The second vector to compare.</param>
        /// <returns>True if the vectors are not equal, false otherwise.</returns>
        public static bool operator !=(Vector2 a, Vector2 b)
        {
            return (a.X != b.X) || (a.Y != b.Y);
        }

        /// <summary>
        /// Returns true if the X and Y values of this are exactly equal to those in the argument.
        /// </summary>
        /// <param name="o">Another Vector2.</param>
        /// <returns>true if the X and Y values are equal to those in o. Returns false if o is not a Vector2.</returns>
        public override bool Equals(object o)
        {
            if (!(o is Vector2))
                return false;

            return this == (Vector2)o;
        }

        /// <summary>
        /// Returns a hash value for the Vector2.
        /// </summary>
        /// <returns>A value intended for efficient insertion and lookup in hashtable-based data structures.</returns>
        public override int GetHashCode()
        {
            return X.GetHashCode() ^ (Y.GetHashCode() * 397);
```

src/Tizen.NUI/src/devel/Lite/L.Color.cs
```diff
@@ -46,6 +49,28 @@ public Color(float r, float g, float b, float a)
             A = a;
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Color"/> struct.
+        /// </summary>
+        /// <param name="value">The hex value of 0xRRGGBB format.</param>
```
hinohie, line 55:
```suggestion
        /// <param name="value">The hex value of 0xRRGGBBAA format.</param>
```

Why RGB? Is RGBA format is more familier to developer?

src/Tizen.NUI/src/internal/Interop/InteropLite.Actor.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using System.Runtime.InteropServices;
+
+namespace Tizen.NUI
+{
+    internal static partial class InteropLite
+    {
+        internal static partial class Actor
+        {
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "ActorSetPropertyVector4", CallingConvention = CallingConvention.Cdecl)]
```
hinohie, line 26:
```suggestion
            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Handle_SetPropertyLiteVector4", CallingConvention = CallingConvention.Cdecl)]
```

Could you please add prefix of API something like this, so everyone can know that Input argument is `Dali::Handle`?

hinohie (approved):

src/Tizen.NUI/src/devel/Lite/L.Corner.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <summary>
+        /// The radius of the bottom left corner of the rectangle.
+        /// </summary>
+        public float BottomLeft { get; }
+
+        /// <summary>
+        /// The radius of the bottom right corner of the rectangle.
+        /// </summary>
+        public float BottomRight { get; }
```
hinohie, line 69:
```suggestion
        /// <summary>
        /// The radius of the bottom right corner of the rectangle.
        /// </summary>
        public float BottomRight { get; }

        /// <summary>
        /// The radius of the bottom left corner of the rectangle.
        /// </summary>
        public float BottomLeft { get; }
```

The order of property define is important! (Since it is `LayoutKind.Sequential`)

rabbitfor:
I will reopen this after big change done!

src/Tizen.NUI/src/devel/Lite/L.Vector2.cs
rabbitfor: Current version including basic structure and simple functionalities. The operators could be added anytime later :) Thank you

rabbitfor: document done

src/Tizen.NUI/src/devel/Lite/L.Color.cs
rabbitfor: I had same concern about it 👍 

GUI toolkits tend to use alpha value as percentage rather than hex format.
e.g. figma
image

More importantly, the typical in-house application GUI guide (e.g. FLUX, NUI) also represents alpha as percentage.

image
image



Previously the NUI application developers with those guide had to choose one of followings,
1. Converts hex rgb to percentage value : `new Color(RR/255, GG/255, BB/255, a)`
2. Convects alpha to hex value: `new Color("#RRGGBB" + ((uint)(255*a)).ToString("X2"))`
3. Creates color with rgb and then sets the alpha : `new Color("#RRGGBB") { A = a };`

I wanted to make it simple -> new L.Color(0xRRGGBB, a);
(FYI, the constructor with string parameter will support "#RGBA" format later like `NUI.Color`)

src/Tizen.NUI/src/internal/Interop/InteropLite.Actor.cs
rabbitfor: I decided not to use this bindings. Thank you!

src/Tizen.NUI/src/internal/Interop/InteropLite.Actor.cs
```diff
@@ -0,0 +1,34 @@
...
+            public static extern void SetColorProperty(HandleRef actor, int propertyType, L.Color color);
+
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Actor_InternalRetrievingVisualPropertyVector4", CallingConvention = CallingConvention.Cdecl)]
+            [return: MarshalAs(UnmanagedType.U1)]
```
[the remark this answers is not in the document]
rabbitfor: I decided not to use this bindings. Thank you!

src/Tizen.NUI/src/devel/Lite/L.Corner.cs
rabbitfor: Good point. It's a big mistake! Even though I decided not to use sequential thing, I fixed it to align with constructor parameter sequence.

src/Tizen.NUI/src/devel/Lite/L.Color.cs
hinohie: Okay. If it is trend of these days, it looks okay to me.

src/Tizen.NUI/src/internal/Common/ReusablePool.cs
```diff
@@ -0,0 +1,163 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Tizen.NUI
+{
+    internal static class ReusablePool<T> where T : Disposable, new()
+    {
+        private const int MaxPoolSize = 3;
+        private static readonly Stack<T> pool = new ();
```
hinohie, line 26:
```suggestion
        private static readonly ConcurrentStack<T> pool = new ();
```

Minor. Can't we use `ConcurrentStack` instead of `Stack`? Since `Tizen.NUI.Disposable` classes are thread safe.

rabbitfor: Great idea! I also thought about it but
(1) since this reusable pool should be used in event thread, (It is used when manipulating View's property which should be done in event thread)
(2) and the `ConcurrentStack` is less-than-ideal performance compared with simple stack
So I concluded not to use it for now. Instead I'll add descriptions to notify this it not thread-safe!

src/Tizen.NUI/src/internal/Interop/Interop.Vector4.cs
```diff
@@ -187,6 +188,9 @@ internal static partial class Vector4
 
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_delete_Vector4")]
             public static extern void DeleteVector4(global::System.Runtime.InteropServices.HandleRef jarg1);
+
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_Vector4_set_all")]
```
everLEEst, line 192: I personally prefer to match the usage of InteropServices, but also I understand why you get rid of this. maybe we can update other dllimport also in other time :)

everLEEst (approved):
Looks good to me.

PR #6569, opened 2025-01-09
[NUI] Check exception from native when get background color

rabbitfor:
### Description of Change ###
It needs to check exception when invoke native method.

### API Changes ###

 - ACR:


hinohie (approved):
We miss this case! Looks good to me and merge for clean code.

JoogabYun (approved):

PR #6575, opened 2025-01-13
[NUI] Rename State to ViewState

rabbitfor:
### Description of Change ###
Since `State` already preseved as property name of `View` so the struct class should be renamed : `ViewState`

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

everLEEst (approved):
looks good to me 👍

PR #6590, opened 2025-01-20
[NUI] Make FrameUpdateCallback could use UIVector2 and UIColor

hinohie:
Let we allow to use FrameUpdateCallbackInterface could use UIVector2 and UIColor so we can reduce GC call at render thread.

Require dali patch :

https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/318520

### DIFF
--- a/src/Tizen.NUI/src/devel/Common/DevelFrameUpdateCallbackInterface.cs
+++ b/src/Tizen.NUI/src/devel/Common/DevelFrameUpdateCallbackInterface.cs
@@ -1,0 +1,143 @@
+/*
+ * Copyright(c) 2025 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+using System;
+using System.ComponentModel;
+
+namespace Tizen.NUI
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI/src/devel/Common/DevelFrameUpdateCallbackInterface.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        /// This will be public opened after ACR done. Before ACR, need to be hidden as inhouse API.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected bool BakePosition(uint id, UIVector2 position)
+        {
+            if (proxyIntPtr == IntPtr.Zero)
+            {
+                return false;
+            }
+            bool ret = Interop.FrameUpdateCallbackInterface.FrameCallbackInterfaceBakePositionVector2Componentwise(proxyIntPtr, id, position.X, position.Y);
```
rabbitfor, line 49: FYI, there is a `ReusablePool<T>` that reuses object for PInvoke.
For example,
https://github.com/Samsung/TizenFX/blob/9258291d1f5257a8916399c5c8071a41d946daea/src/Tizen.NUI/src/internal/Common/Object.cs#L331-L344
Thic code get `Vector4` object from the pool and use it for PInvoke and return it back to the pool.
By using this, you can avoid to create bindings for every struct cases.

hinohie: As I know, `ReusablePool` is not thread safe. This API will be called at RenderThread, so I think we cannot use this pool here :(

rabbitfor: Oh my.. It need to be thread safe! since current resuable pool is not thread safe you would not use it.

rabbitfor (approved):

bshsqa (approved):
Approve.
And if it requires cost to use Vector2 in NUI.
How do you think about to deprecate previous apis. 
Because the FrameUpdateCallback is called in every frame so it looks be better to guide  for VD to use new API.

PR #6592, opened 2025-01-20
[NUI.Scene3D] Add Capture for SceneView

bshsqa:
### Description of Change ###
This PR adds Capture feature to SceneView.
This feature is backported thing from API 12.
Because API 10 hasn't enough features to support full spec of the Capture, this PR supports limited things.

One of thing is that the SceneView.Capture should be requested after the previous capture is finished.

### API Changes ###

 - ACR:


hyunjushin (approved):

src/Tizen.NUI.Scene3D/src/public/Controls/SceneView.cs
```diff
@@ -469,6 +531,34 @@ public void ResetResolution()
...
+        /// If capture is requested during previous capture, invalid index(-1) is returned.</returns>
+        // This will be public opened after ACR done. (Before ACR, need to be hidden as Inhouse API)
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public int Capture(Scene3D.Camera camera, Vector2 size)
+        {
+            if(inCapture)
+            {
+                Tizen.Log.Error("NUI", "The previous capture request is not finished yet.\n");
+                return -1;  // invalid index
+            }
```
hinohie, line 555:
```suggestion
            }
            if (camera == null)
            {
                Tizen.Log.Error("NUI", "Invalid Camera used.\n");
                return -1;
            }
```

Please add null check for camera

bshsqa: I fixed it.

PR #6602, opened 2025-01-22
[NUI] Fix GridLocations to preserve empty space for Stretch Expand

jaehyun0cho:
Previously, if a cell has multiple ColumnSpacing or RowSpacing with Stretch Expand, its column or row does not occupy Expand space if the column or row is empty.

e.g. Child2 begins the beginning of the GridLayout without preserving the empty space for Row 1, Column 0.
GridLayout Width 400.
Child1 Row 0, Column 0, ColumnSpan 2, HorizontalStretch Expand
Child2 Row 1, Column 1, ColumnSpan 1, HorizontalStretch None, Width 100

\-------------------------------------
|||||||||||||||Child1||||||||||||||||
\-------------------------------------
||Child2||                          |
\-------------------------------------

Now, it has been fixed to make the column or row occupy Expand space even though the column or row is empty.
To do so, virtual edges for the column or row have Stretch Expand. It works the same way as CSS grid.
e.g. Child2 begins after the empty space for Row 1, Column 0.
\-------------------------------------
|||||||||||||||Child1||||||||||||||||
\-------------------------------------
|                 ||Child2||        |
\-------------------------------------

### API Changes ###

 - ACR:

### DIFF
--- a/test/NUITizenGallery/Examples/GridTest/GridTest5.cs
+++ b/test/NUITizenGallery/Examples/GridTest/GridTest5.cs
@@ -1,0 +1,38 @@
+/*
+ * Copyright(c) 2025 Samsung Electronics Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


test/NUITizenGallery/Examples/GridTest/GridTest5.cs
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright(c) 2024 Samsung Electronics Co., Ltd.
```
rabbitfor, line 2: 2025

rabbitfor (approved):

rabbitfor (commented):
I think it should be same result for the `Child2` as there is an item in (row:1, col:0) with expand width.

rabbitfor (commented):
(Offline conversation record)
* The basic rule for this follows `css`
* Android does not work as same as `css`

jaehyun0cho:
> Looks good to me but please fix the copyright year :)

I have updated it. Thank you!

PR #6604, opened 2025-01-22
Remove duplicate calling of Invoke in NDalic Color

huayongxu:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/NativeBinding/NDalic.cs
+++ b/src/Tizen.NUI/src/internal/NativeBinding/NDalic.cs
@@ -218,1634 +218,2373 @@ namespace Tizen.NUI
             }
         }
 
+        private static Color aliceBlue = null;
         public static Color ALICE_BLUE
         {
             get
             {
-                global::System.IntPtr cPtr = Interop.NDalicColor.AliceBlueGet();
-                Color ret = (cPtr == global::System.IntPtr.Zero) ? null : new Color(cPtr, false);
-                if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
-                return ret;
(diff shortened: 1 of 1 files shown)


rabbitfor (commented):
Hi Mr.Xu
I think it will be useful if there are many places to call these color properties!
Currently it is only used when initializing static colors in `Color.cs`.
Considering that there is only one calling spot, caching it in a private field seems redundant.

dongsug-song:
@huayongxu 
@AdunFang 

Thank you for the extensive patch work. 
However, I apologize because I think I gave you the wrong guide. 
Actually, NDalic.XXX_COLOR is used in the following code:
```
public class Color : Disposable, ICloneable
{
    /// <summary>
    /// Gets the alice_blue colored Color class.
    /// </summary>
    public static readonly Color AliceBlue = NDalic.ALICE_BLUE;

    /// <summary>
    /// Gets the antique_white colored Color class.
    /// </summary>
    public static readonly Color AntiqueWhite = NDalic.ANTIQUE_WHITE;
```
Here, members like AliceBlue are public static readonly, so changing the interface is impossible. 
Also, NDalic.ALICE_BLUE is used only once in the code above. 
Therefore, adding a private variable as in the current patch seems like a waste of memory. 
We should either leave it as it is (since it's used only once, we can just leave the pinvoke call as it is) or modify it as in Tizen.Common's Color, where we do not make native pinvoke calls and directly find the RGB values in dali (dali-core/dali/public-api/common/constants.h) and write them using `new Color(r,g,b,a)`.

```
    /// <summary>
    /// Aqua color instance. Its RGB value is (0, 255, 255).
    /// </summary>
    public static readonly Color Aqua = FromRgb(0, 255, 255);
```
Please check this again.
thank you.

PR #6605, opened 2025-01-22
[NUI] Extract View variables for Layout to remove wasted memory

jaehyun0cho:
If an application does not use Layout, then View's variables used for Layout waste memory.
To remove those wasted memory, those variables are moved to the new internal class LayoutExtraData and it is not allocated if Layout is not used.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM.

rabbitfor (approved):
LGTM

PR #6614, opened 2025-02-03
[NUI] Fix several custom shader could be works at Vulkan backend

hinohie:
Let we change custom shader codes as unified-dali-style shader codes.


dongsug-song (approved):
approved

src/Tizen.NUI/src/internal/FrameBroker/FrameBrokerBase.cs
```diff
@@ -262,7 +263,7 @@ private Shader CreateShader()
                 "vTexCoord = aPosition + vec2(0.5);\n" +
                 "}\n";
 
-            string fragment_shader =
+            string fragmentShader =
```
everLEEst, line 266: nice!

everLEEst (approved):
good works!

PR #6621, opened 2025-02-04
[NUI] Remove unused APIs

dongsug-song:
### Description of Change ###
[NUI] Remove unused APIs
- this patch was tested with cosmos private build: https://vdcosmos-hub.sec.samsung.net/build/27231083
- also local smoke test was done in PontusM TV

### API Changes ###
nothing


lwc0917 (approved):
Thank you for the good PR.
it would be a good to delete functions that are no longer needed in csharp binder as well related with this.

dongsug-song:
> Thank you for the good PR. it would be a good to delete functions that are no longer needed in csharp binder as well related with this.

Thank you for review.
yes. that is very good point. 
I will make a patch next time.

Seoyeon2Kim (approved):

PR #6630, opened 2025-02-07
[NUI] Make InputMethodContext's signal callbacks not collected

jaehyun0cho:
This PR makes InputMethodContext's signal callbacks not collected before native InputMethodContext gone.

Unlike other classes, InputMethodContext's callbacks are not removed from the static dictionary until it is disposed.
Because eventReceivedEventCallback is called from unmanaged code although it is disconnected.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

everLEEst (approved):

PR #6647, opened 2025-02-10
[NUI] Add Margin and Padding to View markup methods

jaehyun0cho:
Margin is added to View markup method to be used by the layout to size and position the view within its parent layout container. Padding is added to View markup method to be used by the layout to size and position the child views within the layout container.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/ViewProperty/LayoutExtraData.cs
+++ b/src/Tizen.NUI/src/public/ViewProperty/LayoutExtraData.cs
@@ -77,6 +77,16 @@ namespace Tizen.NUI.BaseComponents
         /// </summary>
         public float MaximumHeight { get; set; } = int.MaxValue;
 
+        /// <summary>
+        /// Gets or sets the margin of the view.
+        /// </summary>
+        public UIExtents Margin { get; set; } = UIExtents.Zero;
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI/src/public/ViewProperty/LayoutExtraData.cs
```diff
@@ -77,6 +77,16 @@ internal sealed class LayoutExtraData : IDisposable
         /// </summary>
         public float MaximumHeight { get; set; } = int.MaxValue;
 
+        /// <summary>
+        /// Gets or sets the margin of the view.
+        /// </summary>
+        public UIExtents Margin { get; set; } = 0;
```
rabbitfor, line 83:
```suggestion
        public UIExtents Margin { get; set; } = UIExtents.Zero;
```

rabbitfor (approved):

src/Tizen.NUI/src/public/ViewProperty/LayoutExtraData.cs
jaehyun0cho: Thank you for your comment!
I have updated the code as you suggested.

lwc0917 (approved):
LGTM

PR #6649, opened 2025-02-11
[NUI] Fix a memory leak in Vector4.

huayongxu:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Common/Vector4.cs
+++ b/src/Tizen.NUI/src/public/Common/Vector4.cs
@@ -732,13 +732,6 @@ namespace Tizen.NUI
             NDalicPINVOKE.ThrowExceptionIfExists();
         }
 
-        internal static Vector4 GetVector4FromPtr(global::System.IntPtr cPtr)
-        {
-            Vector4 ret = new Vector4(cPtr, false);
-            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
-            return ret;
-        }
-
         internal float Dot(Vector3 other)
         {
             float ret = Interop.Vector4.DotWithVector3(SwigCPtr, Vector3.getCPtr(other));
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI/src/public/Common/Vector4.cs
```diff
@@ -732,13 +732,6 @@ internal void Reset(float x, float y, float z, float w)
             NDalicPINVOKE.ThrowExceptionIfExists();
         }
 
-        internal static Vector4 GetVector4FromPtr(global::System.IntPtr cPtr)
-        {
-            Vector4 ret = new Vector4(cPtr, false);
```
huayongxu, line 737: the parameter value 'false' in Vector4 constructor means that the memory in DALi is freed by DALi itself.
This might cause a memory leak in csharp-binder.

For example in src/Tizen.NUI/src/public/BaseComponents/VectorGraphics/Shape.cs:
global::System.IntPtr cPtr = Interop.Shape.GetStrokeColor(BaseHandle.getCPtr(this));

This interop API is related to the API in csharp-binder:
SWIGEXPORT void *SWIGSTDCALL CSharp_Dali_Shape_GetStrokeColor(char *pShape) {
  Dali::CanvasRenderer::Shape shape;
  Dali::Vector4 result;

  if (!pShape) {
    SWIG_CSharpSetPendingExceptionArgument(
        SWIG_CSharpArgumentNullException,
        "Attempt to dereference null Dali::CanvasRenderer::Shape", 0);
    return 0;
  }
  shape = *(Dali::CanvasRenderer::Shape *)pShape;
  {
    try {
      result = shape.GetStrokeColor();
    }
    CALL_CATCH_EXCEPTION(0);
  }

  return new Dali::Vector4((const Dali::Vector4 &)result);
}

Dali::Vector4 is allocated by new operator, it need be deallocated by delete operator.

So in C#, ReleaseSwigCPtr need be called, then CSharp_Dali_delete_Vector4 is called to free memory in csharp-binder.
If the second parameter is false in the Vector4 constructor, ReleaseSwigCPtr  will not be callled.

dongsug-song (approved):
thank you for the explanation about memory leak.
approved

hinohie (approved):
Looks good to me.

PR #6651, opened 2025-02-11
[NUI] Remove bindings in Vector4.

huayongxu:
Almost all bindings in Vector4 are removed except
Constructor/ReleaseSwigCPtr.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Common/Vector4.cs
+++ b/src/Tizen.NUI/src/public/Common/Vector4.cs
@@ -16,6 +16,7 @@
  */
 using System;
 using System.ComponentModel;
+using System.Runtime.InteropServices;
 using Tizen.NUI.Binding;
 
 namespace Tizen.NUI
@@ -34,6 +35,40 @@ namespace Tizen.NUI
         private static readonly Vector4 yaxis = new Vector4(0.0f, 1.0f, 0.0f, 0.0f);
         private static readonly Vector4 zaxis = new Vector4(0.0f, 0.0f, 1.0f, 0.0f);
 
+        private struct XYZW
+        {
+            public float x;
+            public float y;
+            public float z;
+            public float w;
+        }
+
+        private struct RGBA
+        {
+            public float r;
+            public float g;
+            public float b;
+            public float a;
+        }
+
+        private struct STPQ
+        {
+            public float s;
+            public float t;
+            public float p;
+            public float q;
+        }
+
+        [StructLayout(LayoutKind.Explicit)]
+        private struct Union
+        {
+            [FieldOffset(0)] public XYZW xyzw;
+            [FieldOffset(0)] public RGBA rgba;
+            [FieldOffset(0)] public STPQ stpq;
+        }
+
+        private Union union;
+
         internal static new void Preload()
         {
             // Do nothing. Just call for load static values.
@@ -45,7 +80,7 @@ namespace Tizen.NUI
         /// <since_tizen> 3 </since_tizen>
         public Vector4() : this(Interop.Vector4.NewVector4(), true)
         {
-            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
+            NDalicPINVOKE.ThrowExceptionIfExists();
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/Common/Vector4.cs
```diff
@@ -728,42 +797,44 @@ public void Clamp(Vector4 min, Vector4 max)
...
         {
-            Interop.Vector4.SetAll(SwigCPtr, x, y, z, w);
-            NDalicPINVOKE.ThrowExceptionIfExists();
+            union.xyzw.x = x;
+            union.xyzw.y = y;
+            union.xyzw.z = z;
+            union.xyzw.w = w;
         }
 
         internal static Vector4 GetVector4FromPtr(global::System.IntPtr cPtr)
```
huayongxu, line 806: This API will be removed in https://github.com/Samsung/TizenFX/pull/6649

dongsug-song: By adding a struct in this way, is it true storing the same value in both the nui and dali sides? Wouldn't this result in duplicate memory usage? Do you think using more memory is beneficial compared to the delay of the pinvoke call?

huayongxu: the struct you said means that
[StructLayout(LayoutKind.Explicit)]
private struct Union
{
[FieldOffset(0)] public XYZW xyzw;
[FieldOffset(0)] public RGBA rgba;
[FieldOffset(0)] public STPQ stpq;
}
is it right?

I followed the suggestions in https://github.sec.samsung.net/NUI/OneUIComponents/issues/26 ,
If Vector4.R / Add / Divide etc are calculated directly, a struct must be cached in c#.
I did not find other ways to realize the suggestions in the link above.
do I misunderstand the suggestions?

dongsug-song: @rabbitfor 
@jaehyun0cho 
@everLEEst 
@lwc0917 

Hello,
could you please check the above comments and share your opinion?
I think the point is that doubled memory VS vector operation pinvoke call.
in my private opinion, vector operation is not used frequently, so I prefer to maintain the operator pinvoke call.
please share any idea.
thank you.

huayongxu: It seems that current patch has some problem:
var v = new View();
var c = new Vector4(1.0f, 1.0f, 1.0f, 1.0f);
c.R = 0.5f;
v.BackgroundColor = c;

In this case, c.R value in DALi is 1.0f still, not 0.5f.

But in Vector4.cs:
public float R
{
[Obsolete("Do not use this setter, that is deprecated in API8 and will be removed in API10. Use new Vector4(...) constructor")]
set
{
union.rgba.r = value;
callback?.Invoke(R, G, B, A);
}
get
{
return union.rgba.r;
}
}

s.R is obsolete.
After set APIs are removed, current patch works.

Another way is that a struct like Vector4 is not created/destroyed in csharp-binader,
4 float parameters are transferred directly between C# and csharp-binader.

dongsug-song: > It seems that current patch has some problem: var v = new View(); var c = new Vector4(1.0f, 1.0f, 1.0f, 1.0f); c.R = 0.5f; v.BackgroundColor = c;
> 
> In this case, c.R value in DALi is 1.0f still, not 0.5f.
> 
> But in Vector4.cs: public float R { [Obsolete("Do not use this setter, that is deprecated in API8 and will be removed in API10. Use new Vector4(...) constructor")] set { union.rgba.r = value; callback?.Invoke(R, G, B, A); } get { return union.rgba.r; } }
> 
> s.R is obsolete. After setter APIs are removed, current patch works.
> 
> Another way is that a struct like Vector4 is not created/destroyed in csharp-binder, 4 float parameters are transferred directly between C# and csharp-binder.

@huayongxu 
Hello,
if you think this patch has a problem. please close this.
also please upload your explanation on https://github.sec.samsung.net/NUI/OneUIComponents/issues/26
thank you.

huayongxu: > > It seems that current patch has some problem: var v = new View(); var c = new Vector4(1.0f, 1.0f, 1.0f, 1.0f); c.R = 0.5f; v.BackgroundColor = c;
> > In this case, c.R value in DALi is 1.0f still, not 0.5f.
> > But in Vector4.cs: public float R { [Obsolete("Do not use this setter, that is deprecated in API8 and will be removed in API10. Use new Vector4(...) constructor")] set { union.rgba.r = value; callback?.Invoke(R, G, B, A); } get { return union.rgba.r; } }
> > s.R is obsolete. After setter APIs are removed, current patch works.
> > Another way is that a struct like Vector4 is not created/destroyed in csharp-binder, 4 float parameters are transferred directly between C# and csharp-binder.
> 
> @huayongxu Hello, if you think this patch has a problem. please close this. also please upload your explanation on https://github.sec.samsung.net/NUI/OneUIComponents/issues/26 thank you.

Done.

PR #6657, opened 2025-02-12
[NUI] Fix dictionary variables to support non UsingXaml case

jaehyun0cho:
In some codes, dictionary variables' getter does not check if the key exists in the dictionary.
This causes an exception by accessing key's value which does not exist.

This PR resolves the above cases.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/FlexLayout.cs
+++ b/src/Tizen.NUI/src/public/Layouting/FlexLayout.cs
@@ -136,7 +136,14 @@ namespace Tizen.NUI
             }
             else
             {
-                return flexAlignmentSelfMap[view];
+                if (flexAlignmentSelfMap.TryGetValue(view, out var flexAlignmentSelf))
+                {
+                    return flexAlignmentSelf;
+                }
+                else
+                {
+                    return AlignmentType.Auto;
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/Layouting/FlexLayout.cs
```diff
@@ -136,7 +136,15 @@ public static AlignmentType GetFlexAlignmentSelf(View view)
             }
             else
             {
-                return flexAlignmentSelfMap[view];
+                AlignmentType flexAlignmentSelf;
+                if (flexAlignmentSelfMap.TryGetValue(view, out flexAlignmentSelf))
```
rabbitfor, line 140:
```suggestion
                if (flexAlignmentSelfMap.TryGetValue(view, out var flexAlignmentSelf))
```

dongsug-song (approved):
I have tested local test app and it was ok.
approved

src/Tizen.NUI/src/public/Layouting/FlexLayout.cs
jaehyun0cho: Thank you very much for your comment!!
I have updated like your suggestion.

rabbitfor (approved):

PR #6660, opened 2025-02-12
[NUI] Introduce new binding style.

everLEEst:
The binding in NUI cause performance and memory issue, so we plan to deprecate them by setting IsUsingXaml disable.

This patch is aim to introduce new style of binding, which do not causing performance and memory seriously, and user can add binding property on existing view class, so that we do not need to add every property to bindable.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/devel/Binding/ViewBindings.cs
+++ b/src/Tizen.NUI/src/devel/Binding/ViewBindings.cs
@@ -1,0 +1,193 @@
+﻿using System;
+using System.ComponentModel;
+using System.Collections.Generic;
+using Tizen.NUI.BaseComponents;
+
+namespace Tizen.NUI.Binding
+{
+    /// <summary>
+    /// Provides a set of static properties that represent the binding properties of the <see cref="View"/> class.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static class ViewBindings
+    {
+        /// <summary>
+        /// Gets the binding property for the width of a <see cref="View"/>.
+        /// </summary>
+        public static BindingProperty<View, float> WidthProperty { get; } = new BindingProperty<View, float>
+        {
+            Setter = (v, value) => v.SizeWidth = value,
+        };
+
+        /// <summary>
+        /// Gets the binding property for the height of a <see cref="View"/>.
+        /// </summary>
+        public static BindingProperty<View, float> HeightProperty { get; } = new BindingProperty<View, float>
+        {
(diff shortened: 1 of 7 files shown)


dongsug-song:
@everLEEst 님
패치 감사합니다.

https://github.com/Samsung/TizenFX/pull/6660#issuecomment-2652835847

에 대해서 51개 Added ACR Required 가 떴는데요, hidden 으로 우선 추가를 해 주시기를 요청 드립니다.
필요하면 추후에 ACR 도 진행 부탁드립니다. 감사합니다.

src/Tizen.NUI/src/devel/Binding/ViewBindings.cs
```diff
@@ -0,0 +1,193 @@
...
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static class ViewBindings
+    {
+        /// <summary>
+        /// Gets the binding property for the width of a <see cref="View"/>.
+        /// </summary>
+        public static BindingProperty<View, float> WidthProperty { get; } = new BindingProperty<View, float>
+        {
+            Setter = (v, value) => v.SizeWidth = value,
```
huayongxu, line 19: I have a query. will all public properties in View be added into ViewBindings like WidthProperty here?

huayongxu: it seems that all public properties which have type BindableProperty in View need be added into ViewBindings.

huayongxu: Another query, when is only Setter (OneWay) added? when are both Setter & Getter (TwoWay) are added?
How to distinguish them?

everLEEst: 1. ViewBindings should have essential property only. as application can create their own BindingProperty in their code, we do not need to provide all property here.
2. TwoWay is only possible when the property notify their changes. TextField, TextEditor notify the Text property change in the TextChanged Event, so it is provided TwoWay, but Text property in TextLabel won't send any event so we can only provide OneWay. most cases, two way is not necessary and the case that real user can modify the property value (similar like editor... ) Two way would be necessary.

huayongxu: got it. thank you for your detailed explanation.

dongsug-song (approved):
approved

huayongxu (approved):
lgtm

PR #6662, opened 2025-02-12
[NUI] Fix AbsoluteLayout to consider Margin and Padding

jaehyun0cho:
AbsoluteLayout is fixed to consider Margin and Padding when it measures and layouts.

### API Changes ###

 - ACR:


jaehyun0cho:
This PR can be applied with AbsoluteLayout's LayoutBounds and LayoutFlags options.
i.e. Margin and Padding is considered in Measure and Layout only if LayoutBounds and LayoutFlags options are used.

jaehyun0cho:
This PR will be added to https://github.com/Samsung/TizenFX/pull/6663

PR #6670, opened 2025-02-18
[NUI.Samples] Fix compile error after unused class removed

hinohie:
Fix compile error issue after #6621 patch merged

- After `RulerPtr` removed, we cannot make Dali's ScrollView make stop fixed position.
- Remove unused class and API
- Make custom shader support on Vulkan backend


dongsug-song (approved):
approved

rabbitfor (approved):

PR #6684, opened 2025-02-24
[NUI.Scene3D] Check HasBody before Register at GeneratedMotionAnimation

hinohie:
It is possible that `Model.GenerateMotionAnimation` return null if input indecies are not exist in current model.

Until now, we just `Register` to Registry forcibly, and check `HasBody()` But in this case, `Tizen.Log.Error("NUI", $"We try to register BaseHandle dont have body! input type:{baseHandle.GetType()}\n");` will print error log, which might not be helpful to application user.

To avoid that useless error printing,
Let we check the Animation class's body first before register to Registry.


dongsug-song (approved):
approved

rabbitfor (approved):

PR #6694, opened 2025-02-25
[NUI] Minimize system settings API calls in Text

wonrst:
Each text component's call to the system API takes a lot of time.
Especially for FontType, initialization of FontConfig occurs natively and consumes most of the time.

TODO: manager should handle all events.

### DIFF
--- a/src/Tizen.NUI/src/public/Theme/DefaultThemeCommon.cs
+++ b/src/Tizen.NUI/src/public/Theme/DefaultThemeCommon.cs
@@ -37,7 +37,7 @@ namespace Tizen.NUI
             // TextLabel style.
             theme.AddStyleWithoutClone("Tizen.NUI.BaseComponents.TextLabel", new TextLabelStyle()
             {
-                FontFamily = Tizen.NUI.FontFamily.UseSystemSetting,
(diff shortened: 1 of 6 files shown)


src/Tizen.NUI/src/public/Theme/DefaultThemeCommon.cs
```diff
@@ -37,7 +37,7 @@ public Theme Create()
             // TextLabel style.
             theme.AddStyleWithoutClone("Tizen.NUI.BaseComponents.TextLabel", new TextLabelStyle()
             {
-                FontFamily = Tizen.NUI.FontFamily.UseSystemSetting,
+                FontFamily = "BreezeSans",
```
hinohie, line 40: Is it intended actions to change the default theme font family?

wonrst: Yes
Currently, the rpi image has only one System FontType, BreezeSans.

With this patch.. Settings API will be called only once, but that still introduces some overhead.

I think until multiple System FontTypes are available, it would be better to avoid the call entirely, even if it's just once.

PR #6696, opened 2025-02-26
[NUI] Minimize system settings API calls in Text

wonrst:
Each text component's call to the system API takes a lot of time. Especially for FontType, initialization of FontConfig occurs natively and consumes most of the time.

TODO: manager should handle all events.

### API Changes ###

 - ACR:


hinohie (approved):
Looks good to me

PR #6705, opened 2025-02-27
[NUI] Remove boxing/unboxing in TextEditor.

huayongxu:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/TextEditorBindableProperty.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/TextEditorBindableProperty.cs
@@ -32,21 +32,16 @@ namespace Tizen.NUI.BaseComponents
         
         internal static void SetInternalTextProperty(BindableObject bindable, object oldValue, object newValue)
         {
-            var textEditor = (TextEditor)bindable;
             if (newValue != null)
             {
-                textEditor.isSettingTextInCSharp = true;
-
-                Object.InternalSetPropertyString(textEditor.SwigCPtr, TextEditor.Property.TEXT, (string)newValue);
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/BaseComponents/TextEditorBindableProperty.cs
```diff
@@ -32,21 +32,16 @@ public partial class TextEditor
         
         internal static void SetInternalTextProperty(BindableObject bindable, object oldValue, object newValue)
```
dongsug-song, line 33: Hello, 
I am curious about why this 'SetInternalTextProperty` has been remained even though this is not used.
is this still used in some other place?

dongsug-song: anyway, it can be removed later, so it would be OK for me.

dongsug-song (approved):
approved

elishateng (approved):
LGTM

src/Tizen.NUI/src/public/BaseComponents/TextEditorBindableProperty.cs
huayongxu: > Hello, I am curious about why this 'SetInternalTextProperty` has been remained even though this is not used. is this still used in some other place?

It is used when Xaml is enabled:
if (NUIApplication.IsUsingXaml)
{
    TextProperty = BindableProperty.Create(nameof(Text), typeof(string), typeof(TextEditor), string.Empty, propertyChanged: SetInternalTextProperty, defaultValueCreator: GetInternalTextProperty);
   ....

PR #6707, opened 2025-03-03
[Tizen.Network.WiFi] Remove thread local support 

saksham9891:
**Description of Change**
Remove thread local support in Tizen.Network.WiFi

**Reason of change:**
Until Tizen 5.0, wifi-manager handles were supported only as thread locals, so the WiFi C# API was implemented so.
However, starting from Tizen 6.0, wifi-manager was modified to remove this restriction, so there is no longer need to use Post().
So, this is being removed.

**API Changes**
ACR: None

**Related**
#6659


akash1-kumar (approved):

jeik01 (approved):
Looks good to me.

PR #6721, opened 2025-03-07
[NUI](Rendering) Create Renderer constructor without Geometry and Shader

hinohie:
Until now, geometry and shader was mendentory.
But if developper try to derive the class, this give some bad experience.

So let we allow to create renderer without geometry or shader.


dongsug-song (approved):
approved

JoogabYun (approved):

PR #6722, opened 2025-03-07
[NUI](Rendering) Construct shader with name

hinohie:
Let we make the name of shader class at C# side.

Relative dali patch : https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/320746


dongsug-song (approved):
approved

JoogabYun (approved):

PR #6733, opened 2025-03-12
[NUI] Remove some private member variables in View.

huayongxu:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -66,7 +66,6 @@ namespace Tizen.NUI.BaseComponents
         private Position internalPosition = null;
         private Position2D internalPosition2D = null;
         private Vector3 internalScale = null;
-        private Size internalSize = null;
         private Size2D internalSize2D = null;
         private int layoutCount = 0;
         private ControlState propagatableControlStates = ControlState.All;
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -66,7 +66,6 @@ public partial class View : Container, IResourcesProvider
         private Position internalPosition = null;
         private Position2D internalPosition2D = null;
         private Vector3 internalScale = null;
-        private Size internalSize = null;
         private Size2D internalSize2D = null;
```
huayongxu, line 69: @dongsug-song 
private member variables beginning with 'internal '  in View could be removed.
it would reduce memory a little.
what do you think?

dongsug-song: yes. that's good idea. 
but could you add some verifying test code or test app?
after testing the verifying code, then we can merge this.
thank you.

myroot: `internalSize` is used to reduce temporary instance so if remove it, increased temporary memory. 
please profile with .net memory profiler

huayongxu: ok, I will do profile after removed.

huayongxu: after removing internalXXX,  memory is increased a little.
so I will close this PR.

PR #6735, opened 2025-03-13
[NUI] Fix LayoutLength not to take invalid float values

jaehyun0cho:
### Description of Change ###
LayoutLength is fixed not to take invalid float values such as non-finite values, float.MinValue, float.MaxValue. If user wants to use float.MinValue or float.MaxValue, then int.MinValue or int.MaxValue are suggested.

### API Changes ###
nothing


rabbitfor (approved):

dongsug-song (approved):
very good to add discription!
if feasible, please add some TC code from next time.
approved.

PR #6740, opened 2025-03-14
[NUI] Remove dali instance

AdunFang:
### Description of Change ###

The Color need SwigCPtr, because the DALi only provide the APIs to get/set the Vector4,
but we don't need to keep SwigCPtr for every Color object.

The new work flow is:
1. Create a singleton SwigCPtr in Vector4.
2. Remove SwigCPtr from Color.
3. When try to set Color to View (or other object), copy the value of Color to the singleton Vector4.SwigCPtr, and then set the Vector4.SwigCPtr to View.
4. When try to get Color from View, set the Vector4.SwigCPtr to view at first, then copy the value from Vecto4.SwigCPtr to Color.

This way can also apply to Vector/Position/Size,
so it can greatly reduce the number of DALi instances and the number of new deletions.

There are one point to note when doing this:
1. Because all Colors share the same singleton SwigCPtr, the following code is incorrect:
    var color1 = Color.Red;
    var color2 = Color.Green;
    var swigCPtr1 = color1.SwigCPtr;
    var swigCPtr2 = color2.SwigCPtr;

    Interop.ActorInternal.SetColor(view1.SwigCPtr, swigCPtr1);
    Interop.ActorInternal.SetColor(view2.SwigCPtr, swigCPtr2);

Because the SwigCPtr is internal used, so we can guarantee to avoid this problem.

### API Changes ###

 - ACR:


dongsug-song:
@rabbitfor 
리뷰 부탁드립니다.

dongsug-song:
@AdunFang 
Hello, please add detailed description, because this change seems to take large effects on existing NUI behaviors.
also please add some test code or test sample app to verify this PR.
thank you.

src/Tizen.NUI/src/public/Common/Vector4.cs
```diff
@@ -20,6 +20,32 @@
 
 namespace Tizen.NUI
 {
+    internal class DaliVector4 : Disposable
+    {
+        public DaliVector4() : base(Interop.Vector4.NewVector4(), true)
```
huayongxu, line 25: true is passed as a parameter.
maybe false is better here? 
if false, the Disposable above can be removed.
if false, a static variable need be added at csharp-binder side.

src/Tizen.NUI/src/public/Common/Vector4.cs
```diff
@@ -20,6 +20,32 @@
 
 namespace Tizen.NUI
 {
+    internal class DaliVector4 : Disposable
```
huayongxu, line 23: can Disposable be removed? or need ReleaseSwigCPtr from Disposable be overridden?

AdunFang:
> @rabbitfor 리뷰 부탁드립니다.



> @AdunFang Hello, please add detailed description, because this change seems to take large effects on existing NUI behaviors. also please add some test code or test sample app to verify this PR. thank you.

Hi Mr. Song
I have updated the description.

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
```diff
@@ -2294,7 +2294,7 @@ private void SetBackgroundColor(Color value)
 
             if (backgroundExtraData == null)
             {
-                Object.InternalSetPropertyVector4(SwigCPtr, View.Property.BACKGROUND, ((Color)value).SwigCPtr);
+                Object.InternalSetPropertyVector4(SwigCPtr, View.Property.BACKGROUND, value.SwigCPtr);
```
dongsug-song, line 2297: I don't know, so I am just asking...
here, is this code needed?
`value.CopyValuesFromDali();`

src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -6018,6 +6020,8 @@ private Color GetInternalColor()
                 internalColor = new Color(OnColorChanged, 0, 0, 0, 0);
             }
             Object.InternalRetrievingPropertyVector4(SwigCPtr, View.Property.COLOR, internalColor.SwigCPtr);
+            internalColor.CopyValuesFromDali();
```
dongsug-song, line 6023: must this kind of code `someColorTypeVariable.CopyValuesFromDali();` be added in all places where color values are set into dali side?
I just can see your changes only in Textxxx classes, so I am just asking...
is this kind of POC? so should this kind of fix be done in other places also?

src/Tizen.NUI/src/public/BaseComponents/ViewBindableProperty.cs
AdunFang: No, because when try to get the SwigCPtr of Color,

huayongxu (approved):
this patch is to assume that all Color instances are operated in main thread.
it looks good to me.

src/Tizen.NUI/src/public/Common/Color.cs
```diff
@@ -1484,100 +1506,76 @@ private static bool IsInvalidValue(ref float value)
 
         /// This will not be public opened.
         [EditorBrowsable(EditorBrowsableState.Never)]
-        protected override void ReleaseSwigCPtr(System.Runtime.InteropServices.HandleRef swigCPtr)
+        public new void Dispose()
```
dongsug-song, line 1509: could you check this Mr. Lee's comment again?
https://github.com/Samsung/TizenFX/pull/6727#discussion_r1988470751

src/Tizen.NUI/src/public/Common/Color.cs
```diff
@@ -1484,100 +1506,76 @@ private static bool IsInvalidValue(ref float value)
...
 
-        private Color Add(Color rhs)
+        /// This will not be public opened.
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected override void Dispose(bool disposing)
         {
-            Color ret = new Color(Interop.Vector4.Add(SwigCPtr, Color.getCPtr(rhs)), true);
-            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
-            return ret;
+            callback = null;
```
hinohie, line 1517: Please write some comments why we don't upcall the `base.Dispose(disposing);` here.

src/Tizen.NUI/src/public/Common/Color.cs
```diff
@@ -923,16 +928,30 @@ public class Color : Disposable, ICloneable
...
+            Vector4.SingletonDaliVector4.SetValues(r, g, b, a);
+            return Vector4.SingletonDaliVector4.SwigCPtr;
+        }
+
         /// <summary>
         /// Default constructor
         /// </summary>
         /// <since_tizen> 3 </since_tizen>
-        public Color() : this(Interop.Vector4.NewVector4(), true)
+        public Color()
```
hinohie, line 951:
```suggestion
        public Color() : this(0.0f, 0.0f, 0.0f, 0.0f)
```

Don't we need to initialize the local variables?

src/Tizen.NUI/src/public/Common/Color.cs
```diff
@@ -30,6 +30,11 @@ namespace Tizen.NUI
     [Tizen.NUI.Binding.TypeConverter(typeof(ColorTypeConverter))]
     public class Color : Disposable, ICloneable
```
hinohie, line 31: I think you have plan to remove 'Disposable' class inheritment in future.

Please write some TODO comments to `getCPtr`, `SwigCPtr`, and `CopyValueAndReturnSingletonHandleRef`.

BTW, Is it possible to remove inheritment of class? As I know it break ABI so it is impossible to change it.

src/Tizen.NUI/src/public/Common/Color.cs
```diff
@@ -923,16 +928,30 @@ public class Color : Disposable, ICloneable
             // Do nothing. Just call for load static values.
         }
 
+        internal static System.Runtime.InteropServices.HandleRef getCPtr(Color obj)
+        {
+            return (obj == null) ? new global::System.Runtime.InteropServices.HandleRef(null, global::System.IntPtr.Zero) : obj.SwigCPtr;
+        }
+
+        internal new System.Runtime.InteropServices.HandleRef SwigCPtr
```
hinohie, line 936: Add new keyword looks bad to me. We should found some way to remove this new keywords somedays :(

src/Tizen.NUI/src/public/Common/Vector4.cs
```diff
@@ -39,6 +75,21 @@ public class Vector4 : Disposable, ICloneable
             // Do nothing. Just call for load static values.
         }
 
+        private static DaliVector4 singletonObj;
+
+        internal static DaliVector4 SingletonDaliVector4
```
hinohie, line 80: Minor, but important questoin for this concept.

Since `Color` (and `Vector4` also I think) has no main thread dependency, we could make and change them at workerthread freely. (For example, user try to build `PropertyMap` at worker thread, and try to call `PropertyValue.Get(Color)`).

Is current implement thread safe enough? I hope we `must` consider multi thread cases for this `SingletonDaliVector4` using scenario.

dongsug-song:
@AdunFang 
Hello, 
please check the @hinohie 's reviews and answer the questions.
thank you.

src/Tizen.NUI/src/public/Common/Color.cs
AdunFang: This patch is using to remove the DALi instance from Color, so it don't need to call base.Dispose(disposing).

src/Tizen.NUI/src/public/Common/Color.cs
AdunFang: This patch is using to remove the DALi instance from Color, so it don't need to do this.

src/Tizen.NUI/src/public/Common/Color.cs
AdunFang: If we can't remove Disposable from Color, we can only don't create the SwigCPtr for Color.

src/Tizen.NUI/src/public/Common/Vector4.cs
AdunFang: This type is internal, so it can only be used by NUI,
I have checked the usage of Color, except the PropertyValue,
all value of Color will be set to some Properties of View,
so that it will be done in main thread.

About the PropertyValue, I will use the "NewPropertyValueVector4Componentwise" to instead of "NewPropertyValueVector4", so that is will not visit the SwigCPtr of Color.

src/Tizen.NUI/src/public/Common/Color.cs
AdunFang: In this patch, we will not do anything in the Dispose of Color, because the SwigCPtr in Color will not be created.
So it doesn't has issue of multi thread.

src/Tizen.NUI/src/public/Common/Color.cs
hinohie: I check may scenarios, and looks find now.

src/Tizen.NUI/src/public/Common/Vector4.cs
hinohie: Looks fine to me now. For now, I think their is no more usage to get / set Color class's value from Dali at workerthread, in Tizen.NUI side.

Now it looks work wel.

src/Tizen.NUI/src/public/BaseComponents/View.cs
hinohie: Please remain some comments for `CopyValuesFromDali` API usage by summary keywords.

src/Tizen.NUI/src/public/Common/Color.cs
```diff
@@ -1067,18 +1097,30 @@ public Color(global::System.Drawing.Color color) : this(Interop.Vector4.NewVecto
...
 
-        internal Color(ColorChangedCallback cb, float r, float g, float b, float a) : this(Interop.Vector4.NewVector4(ValueCheck(r), ValueCheck(g), ValueCheck(b), ValueCheck(a)), true)
+        internal Color(ColorChangedCallback cb, Color other) : this(cb, other.R, other.G, other.B, other.A)
         {
-            callback = cb;
-            if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
         }
 
-        internal Color(ColorChangedCallback cb, Color other) : this(cb, other.R, other.G, other.B, other.A)
+        internal void CopyValuesFromDali()
```
hinohie, line 1113:
```suggestion
        /// <summary>
        /// This API MUST be called when we update Color values from dali by SwigCPtr
        /// </summary>
        internal void CopyValuesFromDali()
```

We need to remain some comments when we *MUST* call this API.

hinohie (commented):
Please remove `new Dispose()`, and write some more comments what I remained.

dongsug-song:
@AdunFang 
Hello, 
please check the @hinohie 's reviews again?
thank you.

dongsug-song:
Hello, There was a diff and conflict between DevelNUI and main, causing difficulties in integrating NUI. Therefore, a hard reset was performed on DevelNUI. As a result, the pull requests (https://github.com/Samsung/TizenFX/pulls) in DevelNUI are now conflicting. To resolve this, you can go to the specific PR, where you will see an error indicating a conflict. You can then use the command line to resolve the conflict.
or Here is another example. This one updates everything below the topmost patch, leaving only the topmost patch unchanged.

git fetch upstream
git rebase --onto=upstream/DevelNUI HEAD~1
git push {current working branch} -f

thank you.

dongsug-song:
No response and no patch updates, so closing it now.

PR #6743, opened 2025-03-14
[Applications.UnitedService] Add a new module for services

hjhun:
### Description of Change ###

United Service can be run from Service application or UI Application. 
This module provides the ability to author a composition Service.
Services are deployed using Resource Package.


ilho159kim (approved):
LGTM

PR #6746, opened 2025-03-17
[NUI](Internal) Implement AsyncImageLoader ImageLoaded signal

hinohie:
Let we bind AsyncImageLoader's ImageLoaded signal, so we can make it hidden easly in future.

Relative dali patch :
- https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/320850


bshsqa (approved):
LGTM

PR #6747, opened 2025-03-17
[NUI] Add PaddingAppliedInNative to check if native applied padding

jaehyun0cho:
Padding of TextLabel, TextField and TextEditor is applied by native instead of layout.
To indicate this, PaddingAppliedInNative is added to LayoutItem.

### API Changes ###

 - ACR:


dongsug-song (approved):
very good to add detailed comment for newly added API.
approved

PR #6752, opened 2025-03-18
[NUI] Fix to apply text objects' padding after added to view

jaehyun0cho:
Previously, Padding of text objects such as TextLabel, TextField, TextEditor was not applied if Padding was set after those objects were added to Layout View.
e.g. The following did not work
var textLabel = new TextLabel();
absoluteLayoutView.Add(textLabel);
textLabel.Padding = padding;

Now, the above problem has been resolved by not clearing Padding of those text objects.
Padding of those text objects is managed by native.

### API Changes ###

 - ACR:


rabbitfor (approved):

rabbitfor:
Thank you for fast fix

PR #6754, opened 2025-03-18
[NUI](ImageView) Create new visual if DesiredWidth/DesiredHeight changed

hinohie:
Until now, we don't request image visual update if desired size changed.

The result of desired size applied and not have big difference. So we need to re-load the image if desired size changed


bshsqa (approved):
Good to merge

PR #6760, opened 2025-03-20
[NUI] Remove some unused codes that are related to Signal.

huayongxu:
### API Changes ###

 - ACR:


huayongxu:
In this patch, some unused codes about Signal will be removed.
please help review.
If it is ok, I will remove the codes in csharp-binder.

dongsug-song (approved):
approved

PR #6763, opened 2025-03-21
[NUI] Try to fix github action error

dongsug-song:
### Description of Change ###
[NUI] Try to fix github action error
it is said that the error was caused by github authentication access issue, but I don't know.
so I just try to use another script instead of `github.rest.issues.createComment()`

### API Changes ###
nothing


taehyub (approved):
LGTM

hinohie (approved):
Looks good to me

PR #6764, opened 2025-03-21
[NUI] Remove github action error part

dongsug-song:
### Description of Change ###
[NUI] Remove github action error part
error is still existed, so remove the whole error part again

### API Changes ###
nothing


hinohie (approved):
Looks good to me. I hope it will be fixed in future days.

PR #6766, opened 2025-03-21
[NUI] Fix Layout to respect MinimumSize and MaximumSize

jaehyun0cho:
Previously, MinimumSize and MaximumSize were not respected by Layout.
e.g. AbsoluteLayout, LinearLayout, RelativeLayout, etc.

Now, MinimumSize and MaximumSize are respected by Layout.

Since the priority of MinimumSize is higher than MaximumSize in DALi,
Layout also follows it.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #6770, opened 2025-03-24
[NUI] Minimize system LocaleLanguage API calls in Text

wonrst:
Every time we set Translatable text,
system LocaleLanguage API is called, which creates overhead.


rabbitfor (approved):
Applying #6762 to API11.

PR #6771, opened 2025-03-24
[NUI] Minimize system LocaleLanguage API calls in Text

wonrst:
Every time we set Translatable text,
system LocaleLanguage API is called, which creates overhead.

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/SystemLocaleLanguageChangedManager.cs
+++ b/src/Tizen.NUI/src/internal/Common/SystemLocaleLanguageChangedManager.cs
@@ -28,6 +28,9 @@ namespace Tizen.NUI
     /// </summary>
     internal static class SystemLocaleLanguageChangedManager
     {
+        private static string localeLanguage = string.Empty;
+        private static WeakEvent<EventHandler<LocaleLanguageChangedEventArgs>> proxy = new WeakEvent<EventHandler<LocaleLanguageChangedEventArgs>>();
+
         static SystemLocaleLanguageChangedManager()
         {
             SystemSettings.LocaleLanguageChanged += SystemLocaleLanguageChanged;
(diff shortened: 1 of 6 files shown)


src/Tizen.NUI/src/internal/Common/SystemLocaleLanguageChangedManager.cs
```diff
@@ -0,0 +1,81 @@
...
+                if (string.IsNullOrEmpty(localeLanguage))
+                {
+                    localeLanguage = SystemSettings.LocaleLanguage;
+                }
+                return localeLanguage;
+            }
+        }
+
+        private static string localeLanguage = string.Empty;
+        private static WeakEvent<EventHandler<LocaleLanguageChangedEventArgs>> proxy = new WeakEvent<EventHandler<LocaleLanguageChangedEventArgs>>();
```
bshsqa, line 79: Does NUI usually define variable and event handler in upper side of class?

hinohie: https://confluence.sec.samsung.net/pages/viewpage.action?pageId=191373542

Follow as NUI Coding Rule, we define static field at the top of classes, above the constructor.

rabbitfor: Nice comments but this patch is #6762 of master branch. I think It's better not changing it from the original.
Please fix things in other patch.

wonrst: 음.. DevelNUI 패치를 가져온거라 생각지못했군요 일단 수정은 했고,
이거 반영 이후 거꾸로 DevelNUI에도 반영하는걸로 가야겟네요

hinohie (approved):
looks good to me

PR #6773, opened 2025-03-25
[NUI] Add Insert method to View

rabbitfor:
### Description of Change ###
Provide `Insert` for view.
It enables to change child sequence in layouting.
For example, plevisouly only sequential adding was possible in LinearLayout.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM.
As we discussed, with this PR, we need to modify SiblingOrder, RaiseToTop, LowerToBottom not to change LayoutChildren order.

PR #6781, opened 2025-03-25
[NUI] Fix SetSize of ExcludeLayouting with LayoutWidth/Height

jaehyun0cho:
Previously, ExcludeLayouting case sets layout size only if Width/HeightSpecification is negative value.
If Width/HeightSpecification is negative value, then it is either MatchParent or WrapContent instead of fixed size value. Therefore, calculated layout size should be set to view size.

Now, LayoutWidth/Height is used instead of Width/HeightSpecification. Width/HeightSpecification sets view size if it is fixed size value. Unlike Width/HeightSpecification, LayoutWidth/Height does not set view size although it is fixed size value. View size is set by Layout.

Therefore, the condition that LayoutWidth/Height is not fixed size value is not the same with the condition that Width/HeightSpecification is not fixed size value since view size may not be set.

To make the condition same as before, additional condition is added.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
+++ b/src/Tizen.NUI/src/public/Layouting/LayoutItem.cs
@@ -601,7 +601,10 @@ namespace Tizen.NUI
                     {
                         // If height or width specification is not explicitly defined,
                         // the size of the owner view must be reset even the ExcludeLayouting is true.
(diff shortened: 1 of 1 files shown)


dongsug-song (approved):
approved

PR #6785, opened 2025-03-26
[NUI] Add background blur effect option: blur once

jmmhappy:
This enables CreateBackgroundBlurEffect(blurRadius) an additional argument, whether to blur once or always.
Default behavior is to refresh every frame(always).

Note: Since RenderEffect instances are immutable(no getter), the option is only applicable at creation.

Needs to be merged after following DALi patches:
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/321580
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/321582


dongsug-song (approved):
가독성을 위해서 **_다음부터는_** blurOnce = false 로 띄어쓰기를 해 주시면 좋겠습니다.
approved

PR #6786, opened 2025-03-26
[NUI] Fix to check IsPaddingHandledByNative for Padding getter

jaehyun0cho:
If a view.Layout.IsPaddingHandledByNative is true, then the view.Padding setter sets view.Padding instead of view.Layout.Padding. Because the padding is applied by DALi instead of NUI Layout.

In the same way, view.Padding getter gets view.Padding instead of view.Layout.Padding if view.Layout.IsPaddingHandledByNative is true.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #6796, opened 2025-03-28
[NUI] Remove UIColor constructor with vector4 and etc.

rabbitfor:
### Description of Change ###
* Remove internal UIColor constructor with vector4 type. (It caused wrong use of `new UIColor(0xFFFFFF)`)
* Add predefined colors: Yellow, Cyan, Magenta, Gray

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #6800, opened 2025-03-31
[NUI.Palette] Make Palette.GenerateAsync() thread safe

hinohie:
Until now, We use some static pointer at `ColorCutQuantizer` class.

Since we could create that class at worker thread freely, we should not access static pointer without any locking points.
And also, those static values should not be static.

To fix this issue, let we make `ColorCutQuantizer.Vbox` hold it's parent, and remove static keyword.


bshsqa (approved):
Looks good to merge.
More simple

PR #6801, opened 2025-03-31
[NUI] Define constant color values at seperated files

hinohie:
Seperate the definition of default constant colors with class itselfs.

### DIFF
--- a/src/Tizen.NUI/src/devel/Lite/UIColorCosntants.cs
+++ b/src/Tizen.NUI/src/devel/Lite/UIColorCosntants.cs
@@ -1,0 +1,80 @@
+/*
+ * Copyright(c) 2025 Samsung Electronics Co., Ltd.
+ *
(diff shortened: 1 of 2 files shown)


dongsug-song (approved):
approved

src/Tizen.NUI/src/devel/Lite/UIColorCosntants.cs
```diff
@@ -0,0 +1,80 @@
...
+using System;
+using System.ComponentModel;
+using Tizen.NUI.BaseComponents;
+
+namespace Tizen.NUI
+{
+    public partial struct UIColor
+    {
+        /// <summary>
+        /// The default color. (This is to distinguish from zero corners)
```
hinohie, line 26:
```suggestion
        /// The default color. (This is to distinguish from transparent)
```

Maybe copy & paste error?

PR #6820, opened 2025-04-03
[NUI] Bind corner radius properties

jmmhappy:
This enables control corner radius, control radius policy, control squareness to NUI.
The property sets corner values for:
- control's visuals(background, image, shadow)
- offscreen rendered result
- render effect result.

Note: This disables corner values for SceneView, etc  but Eunki will manage it soon.

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
@@ -34,20 +34,15 @@ namespace Tizen.NUI.BaseComponents
         [Flags]
         internal enum BackgroundExtraDataUpdatedFlag : byte
         {
-            BackgroundCornerRadius = 1 << 0,
             BackgroundBorderline = 1 << 1,
-            ShadowCornerRadius = 1 << 2,
-            ContentsCornerRadius = 1 << 3, /// Subclass cases.
             ContentsBorderline = 1 << 4, /// Subclass cases.
 
-            Background = BackgroundCornerRadius | BackgroundBorderline,
-            Shadow = ShadowCornerRadius,
+            Background = BackgroundBorderline,
 
-            CornerRadius = BackgroundCornerRadius | ShadowCornerRadius | ContentsCornerRadius,
             Borderline = BackgroundBorderline | ContentsBorderline,
 
             None = 0,
-            All = Background | Shadow,
+            All = Background,
         }
 
         internal BackgroundExtraDataUpdatedFlag backgroundExtraDataUpdatedFlag = BackgroundExtraDataUpdatedFlag.None;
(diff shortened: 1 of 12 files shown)


bshsqa (approved):
Looks good.
Can merge if it tested.

src/Tizen.NUI/src/public/ViewProperty/BackgroundExtraData.cs
```diff
@@ -50,17 +47,6 @@ internal Rectangle BackgroundImageBorder
...
-        internal Vector4 CornerRadius { get; set; }
-
-        /// <summary></summary>
-        internal Vector4 CornerSquareness { get; set; }
-
-        /// <summary>
-        /// Whether the CornerRadius value is relative (percentage [0.0f to 0.5f] of the view size) or absolute (in world units).
-        /// </summary>
-        internal VisualTransformPolicyType CornerRadiusPolicy { get; set; } = VisualTransformPolicyType.Absolute;
-
```
hinohie, line 63: Bad news.

Their are some codes who using those values internally. actually, FLUX.
Until those code changed, we need to keep those properties.

```
[Obsolete("Do not use this, that is deprecated in API13."]
internal Vector4 CornerRadius { get; set; }

[Obsolete("Do not use this, that is deprecated in API13."]
internal Vector4 CornerSquareness { get; set; }

[Obsolete("Do not use this, that is deprecated in API13."]
internal Vector4 CornerRadiusPolicy { get; set; } = VisualTransformPolicyType.Absolute;
```

src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
```diff
@@ -1237,43 +1228,8 @@ internal virtual void UpdateBackgroundExtraData()
...
             }
-            if ((backgroundExtraDataUpdatedFlag & BackgroundExtraDataUpdatedFlag.CornerRadius) != BackgroundExtraDataUpdatedFlag.None)
-            {
-                ApplyCornerRadius();
-            }
-            backgroundExtraDataUpdatedFlag = BackgroundExtraDataUpdatedFlag.None;
-        }
 
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        internal virtual void ApplyCornerRadius()
```
hinohie, line 1248: Bad news.

Their are some custom view who override this function.

Please mark as Obsolete it first, and delete it later.

```
[Obsolete("Do not use this, that is deprecated in API13."]
internal virtual void ApplyCornerRadius()
{
    Tizen.Log.Error("NUI", "ApplyCornerRadius() deprecated internally, Please don't use it.\n");
}
```

src/Tizen.NUI/src/public/ViewProperty/BackgroundExtraData.cs
jmmhappy: Do we have to set and get actual properties?
Or can I just return zero and leave only the form.

src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
jmmhappy: Done.

hinohie (approved):
Looks good to me. Let we try to remove Borderline feature too!

rabbitfor (commented):
Corner animation still uses visual property. Is it OK?

PR #6822, opened 2025-04-03
[NUI] Rewrite LowerFirstLetter

HJLeee:
Reduce string creation by `Tizen.NUI.Animation.AnimateTo()`

- Prune unnecessary case
- Unsafe char[] access to avoid string creation

Unit test done in the local but I wish it could be fully tested on various scenario.
**More desirable case is to implement this logic in the native side.**


HJLeee:
image-2025-2-13_17-29-29

dongsug-song (approved):
looks good to me, but other members' reviews seem to be required.

PR #6824, opened 2025-04-04
[Applications.Common] Add internal API to toggle watchdog

newb1e-kim:
### Description of Change ###
 Add internal APIs to handling watchdog timer
 Implement the static functions in Tizen.Applications.CoreApplication


hjhun:
Please separate implementation from the CoreApplication.cs.
Please add a new static class as below:
```
public static class WatchdogTimer {
  public static void Enable();
  public static void Disable();
  public static void Kick();
}
```

hjhun (approved):

upple (approved):

PR #6826, opened 2025-04-04
[NUI] Keep DRGLView delegate at least 1 frame rendering

hinohie:
Since DRGLView's rendering callback invoked at render thread, we need to keep those delegate at least 1 frame until those delegate will never be called at native side.

We have good internal class for this case, named `RenderThreadObjectHolder`.


JoogabYun (approved):

dongsug-song:
@hinohie 
안녕하세요.
"Rebase and merge" 버튼으로 직접 머지를 부탁드립니다.
(만약 안 되면 다시 말씀해 주시기 바랍니다.)
감사합니다.

PR #6838, opened 2025-04-10
[NUI] Bind Constraint with custom function

hinohie:
Let we bind custom constraints generate API.

* Basic demo app using Constraint

Animation3

Custom shader's effect following mouse cursor position. The effect applied animated position of source View.
And also, use view's `CornerRadius`, `CornerRadiusPolicy`, `CornerSquareness` value as constraint input.
And also, rotate view's orientation follow mouse curosr position.

For now, need to review API and CX level usage.

Relative dali patch :

https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/322672 (To make CornerRadius as constraint input)
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/322068

### DIFF
--- a/src/Tizen.NUI/src/internal/Animation/Constraint.cs
+++ b/src/Tizen.NUI/src/internal/Animation/Constraint.cs
@@ -15,84 +15,467 @@
  *
  */
 
-using System.ComponentModel;
+using global::System;
+using global::System.ComponentModel;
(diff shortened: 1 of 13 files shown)


hinohie:
Since all API are internal, we'd better trial-and-fix after mege it.

PR #6844, opened 2025-04-16
[NUI] Introduce MaskEffect

ANZ1217:
### Description of Change ###
Introduce MaskEffect
MaskEffect is a fragment shader that applies a masking operation using a source view.
It controls the visibility of the target view based on alpha or luminance values from the mask source.
The effect supports transform and scale for flexible mask positioning.

How to use:
targetView.SetRenderEffect(RenderEffect.CreateMaskEffect(sourceView));
targetView.ClearRenderEffect();

OR

// target view, mask mode, positionX, positionY, scaleX, scaleY
targetView.SetRenderEffect(RenderEffect.CreateMaskEffect(sourceView, RenderEffect.MaskMode.Luminance, 0.0f, 0.0f, 1.0f, 1.0f));
targetView.ClearRenderEffect();

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.WebView.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.WebView.cs
@@ -312,6 +312,9 @@ namespace Tizen.NUI
             [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_WebView_RegisterNavigationPolicyDecidedCallback")]
             public static extern void RegisterNavigationPolicyDecidedCallback(HandleRef jarg1, HandleRef jarg2);
 
+            [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_WebView_RegisterOverScrolledCallback")]
+            public static extern void RegisterOverScrolledCallback(HandleRef jarg1, HandleRef jarg2);
+
             [DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_Dali_WebView_RegisterNewWindowPolicyDecidedCallback")]
             public static extern void RegisterNewWindowPolicyDecidedCallback(HandleRef jarg1, HandleRef jarg2);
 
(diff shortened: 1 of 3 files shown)


src/Tizen.NUI/src/public/RenderEffects/RenderEffect.cs
```diff
@@ -33,6 +34,16 @@ namespace Tizen.NUI
     [EditorBrowsable(EditorBrowsableState.Never)]
     public class RenderEffect : BaseHandle
     {
+        /// <summary>
+        /// Enumeration for selecting how the mask source interprets pixel data
+        /// Alpha: Uses the alpha channel of the mask texture. (Default)
+        /// Luminance: Converts RGB to grayscale and uses the luminance as mask value.
+        /// </summary>
+        public enum MaskMode
```
hinohie, line 42:
```suggestion
        [EditorBrowsable(EditorBrowsableState.Never)]
        public enum MaskMode
```

Please add browsable state never here.

ANZ1217: Done!

rabbitfor: Please avoid to define nested enum. It will be better to place outside of the class with other name for example `MaskEffectMode`

src/Tizen.NUI/src/public/RenderEffects/RenderEffect.cs
```diff
@@ -51,5 +63,38 @@ public static RenderEffect CreateBackgroundBlurEffect(float blurRadius, bool blu
...
+
+        /// <summary>
+        /// Create a mask effect
+        /// </summary>
+        /// <remarks>
+        /// Created RenderEffect is immutable.
+        /// </remarks>
+        /// <param name="control">The mask source control.</param>
+        /// <returns>mask effect with given control.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
bshsqa, line 75: We are now in refactoring of RenderEffect to make it mutable object.
Please change this after the refactoring

PR #6847, opened 2025-04-16
[MediaContent] Remove .scan_ignore documentation

hsgwon:
### Description of Change ###

Remove documentation of useless folder scan ignore functionality

### API Changes ###

 - TCSACR-609


wiertel (commented):
Is it a change in behavior or documentation update to describe already existing behavior?
If in last version (API level 9) the `.scan_ignore` was used to disable scanning, then I would prefer note "(up to API Level 9)" to those statements instead of removing them. Otherwise removing is ok.

hsgwon:
> Is it a change in behavior or documentation update to describe already existing behavior? If in last version (API level 9) the `.scan_ignore` was used to disable scanning, then I would prefer note "(up to API Level 9)" to those statements instead of removing them. Otherwise removing is ok.

C# layer will remove only related document but its behavior will be changed in native layer. 
And I agree with your opinion. I'll revert and add "(up to API Level 9)"

PR #6851, opened 2025-04-16
[NUI] Add a cache for key event in View.

huayongxu:
### Description of Change ###

When KeyEvent is triggered frequently, a cache can reduce the peak value of memory usage.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #6859, opened 2025-04-17
[NUI] Fix ScrollableBase Layout to respect MaximumSize

jaehyun0cho:
Previously, ScrollableBase Layout did not respect MaximumSize so the Size property value might be bigger than actual rendered size on screen if MaximumSize is smaller than Size property value. Because actual rendered size respects MaximumSize in DALi.

Now, ScrollableBase Layout respects MaximumSize so the Size property value and actual rendered size on screen are the same.

### API Changes ###

 - ACR:


everLEEst (approved):
LGTM.

PR #6863, opened 2025-04-17
[NUI] Fix FlexLayout not to use Dictionary taking child views

jaehyun0cho:
Not to block GC of child views, Dictionary taking child views is removed from FlexLayout.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #6875, opened 2025-04-25
Hide UI state

rabbitfor:
### Description of Change ###
Hide `UIState` as internal and rename it `ControlStateImpl`.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #6840, opened 2025-04-14
Revert "[NUI] Rewrite LowerFirstLetter"

dongsug-song:
### Description of Change ###
This reverts commit b61e9111329191dc5ab0f28412e3fc21b2231134.
same with https://github.com/Samsung/TizenFX/pull/6839

### API Changes ###
nothing

PR #6879, opened 2025-04-28
[NUI] Implement LowerFirstLetter as Native side

hinohie:
Follow up the #6822 and #6840

Let we change the string at native side, not at c# with string builder.

Relative dali patches:
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/323372


dongsug-song (approved):
approved

PR #6889, opened 2025-04-30
[NUI] Add constructor flag to ignore accessibility relative data

hinohie:
Let we add flag for ignore accessibility relatve codes.

It will be used when we want to remove useless memories for accessibility if application know that they don't use accessibility feature.

Relative dali patches :

https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/323520

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/View.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/View.cs
@@ -472,6 +472,13 @@ namespace Tizen.NUI.BaseComponents
             /// </summary>
             [EditorBrowsable(EditorBrowsableState.Never)]
             Custom,
+            /// <summary>
+            /// Ignore all accessibility implementation. Cannot use View.Accessibility...()
+            /// will be necessary to reduce memory of the View.
+            /// Note that it will make invalidate it's all childrens and parent's accessibility features.
+            /// </summary>
+            [EditorBrowsable(EditorBrowsableState.Never)]
+            Ignore,
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/BaseComponents/View.cs
```diff
@@ -472,6 +472,13 @@ public enum ViewAccessibilityMode
             /// </summary>
             [EditorBrowsable(EditorBrowsableState.Never)]
             Custom,
+            /// <summary>
+            /// Ignore all accessibility implementation. Cannot use View.Accessibility...()
+            /// will be necessary to reduce memory of the View.
+            /// Note that it will make invalidate it's all childrens and parent's accessibility features.
+            /// </summary>
+            [EditorBrowsable(EditorBrowsableState.Never)]
+            Ignore,
```
Seoyeon2Kim, line 481: This is `View` class, so I'm wondering whether we should add `Accessibility` naming on this feature because it's only related to accessibility.
Furthermore, there is a function that makes disabled AT-SPI itself in `Accessibility` class of NUI :
https://github.com/Samsung/TizenFX/blob/main/src/Tizen.NUI/src/public/Accessibility/Accessibility.cs#L173
We shared one to other developers in order to use this API, so I think we can use this one probably?
If not, it would be better to think the naming again :)

hinohie: Hello world!

`Accessibility.BridgeDisableAutoInit()` is disable whole systemical AT-SPI system, not for specific `View`.
Also that API didn't ignore the memory allocation for control's accesibility relative items.

But also, I got some review that this creation feature looks equavalance with `View.AccessibilityHidden = true` for user side. So I need to refactorize behavior soon :(

hinohie:
We'd better consider more good way to reduce memory + Also synchronize the behavior when we use `View.AccessibilityHidden = true` at DALi engine side.

I'll re-open, or recreate if the solution occured.

PR #6897, opened 2025-05-08
[NUI] Add Gradient.StartOffset to upper case table

ANZ1217:
### Description of Change ###
Add Gradient.StartOffset to upper case table


hinohie (approved):
Looks good to me. Fix typo error

PR #6903, opened 2025-05-12
[NUI.Samples] Fix rendering bugs at RendererUpdateAreaTest

hinohie:
We should match shader uniform's value type and
registered value type.

Unfortunatly, UIVector2 become Vector3 when we make it input of PropertyValue.


jaehyun0cho (approved):
LGTM

PR #6906, opened 2025-05-13
[NUI] Introduce Conic Gradient

ANZ1217:
### Description of Change ###
You can use conic gradient using Center and StartAngle.

How to Use:

using(PropertyMap map = new PropertyMap())
{
    // Do some gradient thing
    map.Insert((int)GradientVisualProperty.Center, new PropertyValue(new Vector2(960, 540)));
    map.Insert((int)GradientVisualProperty.StartAngle, new PropertyValue(Math));
    view.Background = map;
}


jaehyun0cho (approved):
LGTM

PR #6909, opened 2025-05-13
[NUI] Implement InnerShadow + Bind Control Borderline and remove relative codes.

hinohie:
Let we make property for InnerShadow.
It will be used when we want to render the shadow inset the view.

Also, Let we make Borderline as top of views, not to touch visual directly. It will be rendered above the inner shadow now.

* How to use

UIExtents = Inset length of each direction's shadow.
BlurRadius = Radius of shadow. Same as BoxShadow usage.
Color = Color of shadow, Same as BoxShadow usage.

```cs
// InnerShadow for top-left
view.InnerShadow = new InnerShadow(new UIExtents(20.0f, 0.0f, 20.0f, 0.0f), blurRadius, color);

// InnerShadow for inset
view.InnerShadow = new InnerShadow(new UIExtents(20.0f), blurRadius, color);
```

* Sample code scene

Animation5

### DIFF
--- a/src/Tizen.NUI.Scene3D/src/public/Controls/SceneView.cs
+++ b/src/Tizen.NUI.Scene3D/src/public/Controls/SceneView.cs
@@ -835,34 +835,6 @@ namespace Tizen.NUI.Scene3D
         }
 
         /// <summary>
-        /// Callback when Borderline property changed.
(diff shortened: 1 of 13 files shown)


dongsug-song (approved):
approved

PR #6914, opened 2025-05-14
[NUI] Temporal revert for Animatable View corner radius

jmmhappy:
This PR is a temporal PR to revert corner radius bindings to re-enable animation.
We are planning to support Animatable corner radius property on DALi side very soon.

### DIFF
--- a/src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
+++ b/src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
@@ -31,6 +31,24 @@ namespace Tizen.NUI.BaseComponents
     {
         internal string styleName;
 
+        [Flags]
+        internal enum BackgroundExtraDataUpdatedFlag : byte
+        {
+            BackgroundCornerRadius = 1 << 0,
+            ShadowCornerRadius = 1 << 2,
+            ContentsCornerRadius = 1 << 3, /// Subclass cases.
+
+            Background = BackgroundCornerRadius,
+            Shadow = ShadowCornerRadius,
+
+            CornerRadius = BackgroundCornerRadius | ShadowCornerRadius | ContentsCornerRadius,
+
+            None = 0,
+            All = Background | Shadow,
+        }
+
+        internal BackgroundExtraDataUpdatedFlag backgroundExtraDataUpdatedFlag = BackgroundExtraDataUpdatedFlag.None;
+
         // TODO : Re-open this API when we resolve Animation issue.
         // private bool backgroundExtraDataUpdateProcessAttachedFlag = false;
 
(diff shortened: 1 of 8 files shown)


bshsqa (commented):
Where is code to block to animate cornerRadius when offscreenRendering is true?

jmmhappy:
> Where is code to block to animate cornerRadius when offscreenRendering is true?

We can animate when `OffScreenRendering = View.OffScreenRenderingType.RefreshAlways;` .
So I guess it would be fine without blocks.

bshsqa:
But then, both of Visuals and OffscreenRendering are animated in the same time.

bshsqa:
I means.. the block is to guard to play animation for each visual's corner radius.

src/Tizen.NUI.Scene3D/src/public/Controls/SceneView.cs
```diff
@@ -834,6 +834,36 @@ private Rotation GetSkyboxOrientation()
             return ret;
         }
 
+        /// <summary>
+        /// Callback when CornerRadius property changed.
+        /// </summary>
+        [Obsolete("Do not use this, that is deprecated in API13.")]
+        internal override void ApplyCornerRadius()
```
hinohie, line 841: Please do not revive this API. SceneView don't need `ApplyCornerRadius()` anymore. (Also, we cannot animate corner radius of sceneview before, actually)

src/Tizen.NUI/src/public/WebView/WebView.cs
```diff
@@ -3131,6 +3131,28 @@ internal WebView Assign(WebView webView)
             return ret;
         }
 
+        [Obsolete("Do not use this, that is deprecated in API13.")]
```
hinohie, line 3134:
```suggestion
        // [Obsolete("Do not use this, that is deprecated in API13.")]
```

Remove this Obsolete. It will show compile warning messages to application developer.

src/Tizen.NUI/src/public/ViewProperty/BackgroundExtraData.cs
```diff
@@ -32,6 +32,11 @@ internal BackgroundExtraData()
         internal BackgroundExtraData(BackgroundExtraData other)
         {
             BackgroundImageBorder = other.BackgroundImageBorder;
+
+            // Deprecated in API13
+            CornerRadius = other.CornerRadius;
```
hinohie, line 37: Also, Remove this Obsolete for these properties.

src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
```diff
@@ -1144,10 +1162,115 @@ internal ResourceLoadingStatusType GetBackgroundResourceStatus()
...
+                ApplyCornerRadius();
+            }
+
+            backgroundExtraDataUpdatedFlag = BackgroundExtraDataUpdatedFlag.None;
+        }
+
         [Obsolete("Do not use this, that is deprecated in API13.")]
         internal virtual void ApplyCornerRadius()
         {
             Tizen.Log.Error("NUI", "ApplyCornerRadius() deprecated internally, Please don't use it.\n");
```
hinohie, line 1245:
```suggestion
            // Tizen.Log.Error("NUI", "ApplyCornerRadius() deprecated internally, Please don't use it.\n");
```

Please do not print this error message.
Now we revive the API, so we should use it.

src/Tizen.NUI/src/public/BaseComponents/ViewInternal.cs
```diff
@@ -1144,10 +1162,115 @@ internal ResourceLoadingStatusType GetBackgroundResourceStatus()
...
+
+            if ((backgroundExtraDataUpdatedFlag & BackgroundExtraDataUpdatedFlag.CornerRadius) != BackgroundExtraDataUpdatedFlag.None)
+            {
+                ApplyCornerRadius();
+            }
+
+            backgroundExtraDataUpdatedFlag = BackgroundExtraDataUpdatedFlag.None;
+        }
+
         [Obsolete("Do not use this, that is deprecated in API13.")]
```
hinohie, line 1242:
```suggestion
        // [Obsolete("Do not use this, that is deprecated in API13.")]
```
Remove this Obsolete. It will show compile warning messages to application developer.

hinohie:
@bshsqa OffscreenRendering's corner radius didn't animatable now.

jmmhappy:
Oh... I just reproduced an error situation. View without image, shadow, or background might crash by corner animation.
(Checked with empty view + background blur effect)

As Eunki mentioned, I heard he blocked viewCornerRadius somehow. So render effect and offscreen rendering's corner radius animation would not work right now.

So maybe no problem.

src/Tizen.NUI/src/public/ViewProperty/BackgroundExtraData.cs
jmmhappy: I think we don't need to remove Obsolete here. 
Users may set/get view properties, not BackgroundExtraData. Duplication at setter is just for the animation.

hinohie:
`View without image, shadow, and background might crash by corner animation.` <<- We already got the crash before for this case. This was know, issue, so we don't need to worry about it.

hinohie:
Or, we can add something like this code at ApplyCornerRadius

```cs
if (IsBackgroundEmpty())
{
  SetBackgroundColor(Color.Transparent);
}
```

jmmhappy:
> Or, we can add something like this code at ApplyCornerRadius
> 
> ```cs
> if (IsBackgroundEmpty())
> {
>   SetBackgroundColor(Color.Transparent);
> }
> ```

On second thoughts, I wonder when will we need this.
So like.. corner radius on RenderEffect and offscreen rendering is already blocked, so this won't make it animate it's corner anyway.
Then, why would we need to animate corners on empty view? Let's happily ignore the empty view situation.

hinohie (approved):
Looks good to me

PR #6925, opened 2025-05-21
[NUI] Add creation option to visual factory

hinohie:
We need to make some way to create `ImageVisual` forcibly even if it is animation images.

For now, let we make the property to change internal creation options.

Relative dali patch :
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/324608


tscholb (approved):
LGTM :)

PR #6943, opened 2025-06-04
[Applications.Common] Fix OnTimeZoneChanged method

hjhun:
### Description of Change ###

This patch adds to call the CultureInfo.CurrentCulture.ClearCacheData() in the OnTimeZoneChanged() method.


upple (approved):
[ClearCachedData](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.clearcacheddata?view=net-9.0)
LGTM 👍

PR #6945, opened 2025-06-04
[NUI]Fix CA2000 warnings in InnerShadow.

huayongxu:
### API Changes ###

 - ACR:


hinohie (approved):
Looks good to me

dongsug-song:
because of this, https://github.com/Samsung/TizenFX/blob/main/src/Tizen.NUI/src/public/ViewProperty/ShadowBase.cs#L48
`using` can be used.

PR #6953, opened 2025-06-07
[NUI][API12] Fix crash when we use NUI.Layout if IsUsingXaml = false

hinohie:
Seperate API call by xaml case, and fix crash issue.


dongsug-song (approved):
approved

jaehyun0cho (approved):
LGTM

PR #6955, opened 2025-06-09
[NUI] Fix text label layout

wonrst:
Bug fix when the width and height specification of the text label is wrap, wrap. 
I think even if the width is wrap, it should not exceed the size of the parent. 
But I can not predict what compatibility this will break.


jaehyun0cho:
This PR requires the following PR for scrollable content case.
https://github.com/Samsung/TizenFX/pull/6956

PR #6989, opened 2025-06-27
[NUI] Add TargetMaskOnce, SourceMaskOnce Property to MaskEffect

ANZ1217:
### Description of Change ###
Add TargetMaskOnce, SourceMaskOcne Property to MaskEffect
and Add Mask Effect Sample.


src/Tizen.NUI/src/public/RenderEffects/MaskEffect.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.ComponentModel;
+using System;
+
+namespace Tizen.NUI
+{
+    /// <summary>
+    /// mask effect of a View.
+    /// Applications can apply MaskEffect as the example below :
+    /// <code>
+    /// MaskEffect effect = new MaskEffect();
```
bshsqa, line 32: Can you explain more how to use MaskEffect?
For example, alignment of target and source when the size is different.
As I know, currently, it uses same camera for both target and source.
So, if the target and source have different size, something cropped.
To fit the source and target, user should control position, size, parentorigin, and pivotpoint. (and position uses pivot ~~)

or, you can change the mask's operation, for example, if you use different camera for each source and target.
You might make two textures of target and source contains all of content of each.
And then the mask result would be always same(actually very similar).

How do you think which is better?
If you select better things, you can fix the dali code and document here more.

ANZ1217: Done!

hinohie (approved):
Please add relative dali patch at description

https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/326155

PR #6990, opened 2025-06-30
Public open for Visual's ContainerRange as public

hinohie:
Open enum to choose the container of visual to user.
Also, Make API to get which container was we add this visuals.


dongsug-song (approved):
approved

rabbitfor (approved):

PR #7000, opened 2025-07-01
[NUI] Add UpdateLayout in Window

jaehyun0cho:
UpdateLayout calculates sizes and positions of views in all layers if the view's layout is required to be updated.
e.g. RequestLayout() is called for the view.

Since UpdateLayout calculates sizes and positions of views manually right now, do not abuse calling it.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window/Window.cs
+++ b/src/Tizen.NUI/src/public/Window/Window.cs
@@ -2668,6 +2668,31 @@ namespace Tizen.NUI
             }            
         }
 
+        /// <summary>
+        /// Calculates sizes and positions of views in all layers if the view's layout is required to be updated.
+        /// e.g. RequestLayout() is called for the view.
+        /// Since UpdateLayout calculates sizes and positions of views manually right now, do not abuse calling it.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


src/Tizen.NUI/src/public/Window/Window.cs
```diff
@@ -2668,6 +2668,32 @@ public WindowBlurInfo BlurInfo
...
+        {
+            LayersChildren?.ForEach(layer =>
+            {
+                if (layer?.LayoutCount > 0)
+                {
+                    layer?.Children?.ForEach(view =>
+                    {
+                        if (view != null)
+                        {
+                            using (var size = GetSize())
```
rabbitfor: I think you may reuse size by storing it to a local variable to avoid duplicate DALi call and heap allocation for every iteration.

jaehyun0cho: 리뷰 감사드립니다.
말씀해주신대로 size를 iteration 전에 얻어온 후 사용하도록 수정하였습니다.

rabbitfor (approved):

PR #7016, opened 2025-07-04
[Markdown] Add margin to ui paragraph

wonrst:
This only applies to single paragraph blocks.

If we should support margins for another cases, such as list, quote, table, etc., it is better to add another style.

### DIFF
--- a/src/Tizen.NUI.MarkdownRenderer/src/internal/MarkdownUI/UIParagraph.cs
+++ b/src/Tizen.NUI.MarkdownRenderer/src/internal/MarkdownUI/UIParagraph.cs
@@ -27,16 +27,25 @@ namespace Tizen.NUI.MarkdownRenderer
     /// </summary>
     internal class UIParagraph : UIText
     {
+        private readonly ParagraphStyle paragraph;
+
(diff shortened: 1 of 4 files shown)


src/Tizen.NUI.MarkdownRenderer/src/internal/MarkdownUI/UIParagraph.cs
```diff
@@ -27,16 +27,25 @@ namespace Tizen.NUI.MarkdownRenderer
     /// </summary>
     internal class UIParagraph : UIText
     {
+        private readonly ParagraphStyle paragraph;
```
rabbitfor, line 30: https://confluence.sec.samsung.net/spaces/GFX/pages/191380639/NUI+Ground+Rule#NUIGroundRule-NUICodingguide

Please check the NUI coding convention. -> `_paragraph`

> We use `_camelCase` for internal and private fields and use readonly where possible. Prefix internal and private instance fields with `_`, static fields with `s_` and thread static fields with `t_`. When used on static fields, readonly should come after static (e.g. static readonly not readonly static). Public fields should be used sparingly and should use PascalCasing with no prefix when used.

PR #7020, opened 2025-07-07
[NUI] Add AccessibilityHighlightChanged event and optimize data fields

rabbitfor:
### Description of Change ###
* Add AccessibilityHighlightChanged event
* Move accessibility fields of View to extra objects that are created on demands.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

PR #7027, opened 2025-07-08
[NUI] GetRenderEffect() does not attempt to convert null result

rabbitfor:
### Description of Change ###
Avoid converting null pointer of GetRenderEffect result 

### API Changes ###

 - ACR:


jmmhappy (approved):
Looks good to me either 👍

PR #7032, opened 2025-07-08
[NUI] Fix the crash that occurs when excluding the tv profile

dongsug-song:
### Description of Change ###
[NUI] Fix the crash that occurs when excluding the tv profile

### API Changes ###
nothing


Seoyeon2Kim (approved):
LGTM

PR #7033, opened 2025-07-08
[NUI] Recover NUI.Components in tv profile

dongsug-song:
### Description of Change ###
[NUI] Recover NUI.Components in tv profile
- By TCSACR-606, NUI.Components has been deprecated in tv profile only.
- Because, it is deprecated ACR, so the package should not be removed.
- The remove ACR will be processed after the future 2 versions of tizen release.

### API Changes ###
nothing


Seoyeon2Kim (approved):
LGTM

PR #7041, opened 2025-07-09
[NUI] Disable default theme manager

rabbitfor:
### Description of Change ###
For the other frameworks using NUI as a foundation and uses its own theme manager other than NUI's, the option that disable NUI ThemeManager is needed.

### API Changes ###

 - ACR:


hinohie (approved):
Looks good to me

PR #7043, opened 2025-07-09
[NUI] Reinforce view's ownership on its render effect.

jmmhappy:
- An effect is applied only to a single View.
- Clear before disposing an effect.

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/RenderEffects/RenderEffect.cs
```diff
@@ -47,10 +47,41 @@ public enum MaskEffectMode
     [EditorBrowsable(EditorBrowsableState.Never)]
     public class RenderEffect : BaseHandle
     {
+        private WeakReference<View> ownerView;
+
+        public WeakReference<View> OwnerView
```
hinohie:
```suggestion
        internal View OwnerView
```

Please make it as `View`, and return `ownerView.Target`, instead of WeakReference itself.

Also, we'd better make it as internal, to avoid application set this value directly without `SetRenderEffect` API.

src/Tizen.NUI/src/public/RenderEffects/RenderEffect.cs
```diff
@@ -47,10 +47,41 @@ public enum MaskEffectMode
...
+        {
+            get
+            {
+                return ownerView;
+            }
+            set
+            {
+                if((ownerView != null) && ownerView.TryGetTarget(out View previousTarget))
```
hinohie:
```suggestion
                if((ownerView != null) && ownerView.TryGetTarget(out View previousTarget) && previousTarget != value)
                {
                        ownerView = null;
```

Please do not call `ClearRenderEffect()` If we try to add same view here.

src/Tizen.NUI/src/public/RenderEffects/RenderEffect.cs
```diff
@@ -47,10 +47,41 @@ public enum MaskEffectMode
...
         {
         }
 
+        protected override void Dispose(DisposeTypes type)
+        {
+            if(!Disposed)
+            {
+                if((ownerView != null) && ownerView.TryGetTarget(out View target))
+                {
+                    target.ClearRenderEffect();
```
hinohie, line 74:
```suggestion
                    target.ClearRenderEffect();
                    ownerView = null;
```

src/Tizen.NUI/src/public/RenderEffects/RenderEffect.cs
rabbitfor: OwnerView를 얻어서 사용하려는 사용성이 없기 때문에 method가 나을것 같은데 어떠신가요?
```C#
void SetOwnerView(View)
```
```C#
void ClearOwnerView()
```

src/Tizen.NUI/src/public/RenderEffects/RenderEffect.cs
```diff
@@ -47,10 +47,41 @@ public enum MaskEffectMode
     [EditorBrowsable(EditorBrowsableState.Never)]
     public class RenderEffect : BaseHandle
     {
+        private WeakReference<View> ownerView;
```
hinohie:
```suggestion
        private WeakReference<View> _ownerView;
```

Code style guide

src/Tizen.NUI/src/public/RenderEffects/RenderEffect.cs
jmmhappy: `SetOwnerView(null)`로 수정해보았는데 역시 `ClearOwnerView()`를 따로 떼는 것이 나을까요?

rabbitfor: 지금 형태도 좋아 보입니다

PR #7047, opened 2025-07-10
[NUI][Test] Use NUI Interop APIs directly in PropertyMapSample.

huayongxu:
- Use NUI Interop APIs firstly in PropertyMapSample.
- Remove Tizen.NUI.Samples Interop APIs whose referece count is 0.

### API Changes ###

 - ACR:


zg2nets (approved):
lgtm

PR #7049, opened 2025-07-11
[NUI] Get `AliveCount` for more items

hinohie:
Let we count some more class's AliveCount values, so Application easy to detect leak of objects more easy


dongsug-song (approved):
approved

PR #7050, opened 2025-07-11
[NUI] Accessibility event handlers are invoked with valid view handle

rabbitfor:
### Description of Change ###
Accessibility event handlers are invoked with valid view handle.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/View/ViewAccessibilityData.cs
+++ b/src/Tizen.NUI/src/internal/View/ViewAccessibilityData.cs
@@ -16,9 +16,7 @@
  */
 
 using System;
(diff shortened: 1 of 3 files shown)


dongsug-song (approved):
approved

hinohie (approved):
Looks good to me but minor..

Dont we need to check `_owner` Disposed, or disposequeued?

rabbitfor:
> Looks good to me but minor..
> 
> Dont we need to check `_owner` Disposed, or disposequeued?

Thank you. I agree that it'd better to have aliveness check on accessiblity event.

PR #7053, opened 2025-07-11
[Tizen.System.Session] Fix native struct marshaling bug

adaszkiewicza:
### Description of Change ###
On 32-bit system past implementation of `SubsessionEventInfoNative` caused a crash due to misalignment with the native struct. 
The struct now uses fixed size buffers. Some `unsafe` blocks are added, but we can be certain the buffer and struct size will be constant on native side.

Some typos in docs were fixed along the way.

### API Changes ###
 - ACR:
No API changes.


chanwoochoi (approved):
LGTM

PR #7055, opened 2025-07-14
[NUI] Allow to dispose some Disposable class at workerthread if possible

hinohie:
Let we call Dispose() for some Disposable classes if we define them possible to call it on worker thread.

For example, let we don't wait main thread wake to release `Vector4` or etc.


dongsug-song (approved):
approved

PR #7063, opened 2025-07-16
[NUI][VisualObject] Revive VisualObject API which was public before

hinohie:
Since some application might build app based API12, we should keep API which was exist at API12.


dongsug-song (approved):
approved

PR #7066, opened 2025-07-17
[NUI] Add IncrementalDispose at Applicatoin + DisposeQueue more thread safe

hinohie:
Let we make static value for `Application.IsUsingIncrementalDispose` whether we dispose GC queue incrementally or not.

`true` will dispose GC items maximum 20% per each loop, `false` will dispose every items immediately.
Default is false.

The number of maximum GC item is experimental values. Incremental logic could be changed in future, like `StopWatch` used.


dongsug-song (approved):
approved

PR #7068, opened 2025-07-17
[NUI] More Disposable Dispose at worker thread (TouchPoint / WindowData / Builtin AlphaFunction)

hinohie:
Add more classes that we could Dispose at worker thread.

+

Make some demo dispose items well.


everLEEst (approved):
Looks good to me.

PR #7070, opened 2025-07-18
[NUI] Check Animation's body before get properties from native

hinohie:
Fix crash issue when nobody animation disposed by GC


dongsug-song (approved):
approved

PR #7085, opened 2025-07-24
[Telephony] Deprecate Telephony API

HONGCHAEHEE:
### Description of Change ###
Deprecate Telephony API

### API Changes ###

 - ACR: https://jira.sec.samsung.net/browse/TCSACR-615


src/Tizen.Telephony/Interop/Interop.Call.cs
```diff
@@ -30,31 +30,31 @@ internal static partial class Interop
     internal static partial class Call
```
wiertel, line 31: those Interops are internal classes, any comments here are just for internal use, could have added just one comment per such file

src/Tizen.Telephony/Tizen.Telephony/Call.cs
```diff
@@ -26,6 +26,7 @@ namespace Tizen.Telephony
     /// It provides the list of CallHandle which can be used to get the information about call related actions.
     /// </summary>
     /// <since_tizen> 3 </since_tizen>
+    /// [Obsolete("Deprecated since API13, will be removed in API15.")]
```
wiertel: [Obsolete Attribute](https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute?view=net-9.0) is a C# language construct rather than a part of documentation/description. Please remove triple slash from those lines.
```suggestion
    [Obsolete("Deprecated since API13, will be removed in API15.")]
```

wiertel (changes requested):

src/Tizen.Telephony/Interop/Interop.Call.cs
HONGCHAEHEE: Modified.

src/Tizen.Telephony/Tizen.Telephony/Call.cs
HONGCHAEHEE: Modified.

wiertel (approved):

dh79pyun (approved):

PR #7086, opened 2025-07-24
[Tizen.Network] Implementation of Tizen.Network.Tethering APIs

r-bhalotia:
### Description of Change ###
Implementation of Tizen.Network.Tethering APIs

### API Changes ###

 - ACR: None (Internal CSAPI)


src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtEvents.cs
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+
+        private void RegisterEnabledEvent()
+        {
+            Log.Info(Globals.LogTag, "RegisterEnabledEvent");
+            _enabledCallback = (int result, bool isRequested, IntPtr userData) =>
+            {
+                TetheringExtEnabledEventArgs e = new TetheringExtEnabledEventArgs(result, isRequested);
+                _tetheringExtEnabled.SafeInvoke(null, e);
```
jiung-yu: How about using ?.Invoke

public event EventHandler Started;
private void SendStartedEvent() {
    Started?.Invoke(this, EventArgs.Empty);

}

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtEvents.cs
```diff
@@ -0,0 +1,223 @@
...
+            }
+        }
+
+        private void RegisterDisabledEvent()
+        {
+            Log.Info(Globals.LogTag, "RegisterDisabledEvent");
+            _disabledCallback = (int result, TetheringDisabledCause cause, IntPtr userData) =>
+            {
+                TetheringExtDisabledEventArgs e = new TetheringExtDisabledEventArgs(result, cause);
+                _tetheringExtDisabled.SafeInvoke(null, e);
```
jiung-yu: How about using ?.Invoke

public event EventHandler Started;
private void SendStartedEvent() {
    Started?.Invoke(this, EventArgs.Empty);

}

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtEvents.cs
```diff
@@ -0,0 +1,223 @@
...
+            }
+        }
+
+        private void RegisterConnectionStateChangedEvent()
+        {
+            _connectionStateChangedCallback = (IntPtr client, bool opened, IntPtr userData) =>
+            {
+                TetheringExtClient tetheringExtClient = new TetheringExtClient(client);
+                ConnectionStateChangedEventArgs e = new ConnectionStateChangedEventArgs(tetheringExtClient, opened);
+                _connectionStateChanged.SafeInvoke(null, e);
```
jiung-yu: How about using ?.Invoke

public event EventHandler Started;
private void SendStartedEvent() {
    Started?.Invoke(this, EventArgs.Empty);

}

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringErrorFactory.cs
```diff
@@ -0,0 +1,92 @@
...
+ * limitations under the License.
+ */
+
+using System;
+using Tizen.Internals.Errors;
+
+namespace Tizen.Network.Tethering
+{
+
+    internal enum TetheringError
```
jiung-yu: This can go to TetheringEnumerations as public.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringErrorFactory.cs
```diff
@@ -0,0 +1,92 @@
...
+            ThrowException(e, (handle == IntPtr.Zero), false, message);
+        }
+
+        static internal void ThrowTetheringException(int e, IntPtr handle1, IntPtr handle2, string message)
+        {
+            ThrowException(e, (handle1 == IntPtr.Zero), (handle2 == IntPtr.Zero), message);
+        }
+
+        // Used for return value of native API
+        static private void ThrowException(int e, bool isHandle1Null, bool isHandle2Null, string message)
```
jiung-yu: isHandle1Null and isHandle2Null are not used anywhere

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringErrorFactory.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            ThrowException(e, (handle1 == IntPtr.Zero), (handle2 == IntPtr.Zero), "");
+        }
+
+        static internal void ThrowTetheringException(int e, string message)
+        {
+            ThrowException(e, false, false, message);
+        }
+
+        static internal void ThrowTetheringException(int e, IntPtr handle, string message)
```
jiung-yu: Why these overloaded methods are needed?

ThrowTetheringException(int e, IntPtr handle, string message)
ThrowTetheringException(int e, IntPtr handle1, IntPtr handle2, string message)

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// </summary>
+        protected override bool ReleaseHandle()
+        {
+            Interop.TetheringExt.Deinitialize(this.handle);
+            this.SetHandle(IntPtr.Zero);
+            return true;
+        }
+
+        // TODO: need not to implement thread safe, can remove this
+        internal void SetTID(int id)
```
jiung-yu: Is this method needed?

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
```diff
@@ -0,0 +1,258 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Gets the tethering safe handle.
+        /// </summary>
+        /// <since_tizen> 13 </since_tizen>
+        /// <returns>The instance of the SafeTetheringExtManagerHandle.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static SafeTetheringExtManagerHandle GetTetheringExtHandle()
```
jiung-yu: Who'll use this method?

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtEnumerations.cs
```diff
@@ -0,0 +1,136 @@
+using System;
+using Tizen.Internals.Errors;
+
+
+namespace Tizen.Network.Tethering
+{
+    /// <summary>
+    /// Enumeration for the Tethering type.
+    /// </summary>
+    /// <since_tizen> 3 </since_tizen>
```
wiertel: 3 -> 13 through this file

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtEnumerations.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// <summary>
+        /// Tethering All type.
+        /// </summary>
+        /// <since_tizen> 3 </since_tizen>
+        All = 0,
+        /// <summary>
+        /// Tethering USB type.
+        /// </summary>
+        /// <since_tizen> 3 </since_tizen>
+        USB = 1,
```
wiertel:
```suggestion
        Usb = 1,
```

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtEnumerations.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// <summary>
+        /// Tethering WiFi type.
+        /// </summary>
+        /// <since_tizen> 3 </since_tizen>
+        WiFi = 2,
+        /// <summary>
+        /// Tethering Bluetooth type.
+        /// </summary>
+        /// <since_tizen> 3 </since_tizen>
+        BT = 3,
```
wiertel:
```suggestion
        Bluetooth = 3,
```

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtEnumerations.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// </summary>
+        /// <since_tizen> 3 </since_tizen>
+        Others = 5,
+        /// <summary>
+        /// Disabled due to user request.
+        /// </summary>
+        /// <since_tizen> 3 </since_tizen>
+        Request = 6,
+        /// <summary>
+        /// Disabled due to WiFi beign on .
```
wiertel:
```suggestion
        /// Disabled due to WiFi being on .
```

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringInfo.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Tizen.Network.Tethering
+{
+
+    /// <summary>
+    /// This class contains tethering info.
+    /// </summary>
+    /// <since_tizen> 13 </since_tizen>
+    public class TetheringInfo
+    {
+        private IntPtr _info;
```
wiertel: So this class purpose is to hold a pointer. If it is a pointer that should be released automatically then consider either subclassing SafeHandle.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
```diff
@@ -0,0 +1,215 @@
...
+
+using System;
+using System.ComponentModel;
+
+namespace Tizen.Network.Tethering
+{
+
+    /// <summary>
+    /// A manager class that enables applications to create and manage a Wi-Fi hotspot, allowing devices to share an internet connection over a Wireless Local Area Network (WLAN).
+    /// The TetheringExt Manager provides functionality to activate and deactivate a Wi-Fi hotspot, configure hotspot settings, and manage connected devices.
```
wiertel: That second sentence could go into `<remarks>` to keep the summary short.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
```diff
@@ -0,0 +1,215 @@
...
+    /// </summary>
+    /// <since_tizen> 13 </since_tizen>
+    static public class TetheringExtManager
```
wiertel: What is the purpose of "Ext" infix? In C# names it's preferred to keep full words for readability. Why not TetheringManager?

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
```diff
@@ -0,0 +1,215 @@
...
+                TetheringExtManagerImpl.Instance.ConnectionStateChanged += value;
+            }
+            remove
+            {
+                TetheringExtManagerImpl.Instance.ConnectionStateChanged -= value;
+            }
+        }
+
+        /// <summary>
+        /// Initializes the TetheringExt handle.
```
wiertel: please don't mention "handle" in API description, it's an implementation detail, from user point of view handles are not visible, not accessible
there is no way to undo Initialize, which makes me thing initializing could be done automatically when Activate() is called or State changed event is registered. If there are many more "entry" points for the API then it gets a little ridiculous to auto init everywhere, but still can be considered.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
```diff
@@ -0,0 +1,215 @@
...
+        /// </summary>
+        /// <since_tizen> 13 </since_tizen>
+        static public void Initialize()
```
wiertel: This is the only publicly visible method of this class, it does not seem very useful that way. Is this really your intention?

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
```diff
@@ -0,0 +1,215 @@
...
+        {
+            return TetheringExtManagerImpl.Instance.GetTetheringInfo();
+        }
+
+        /// <summary>
+        /// Returns the Security type of the Wi-Fi hotspot managed by the TetheringExt Manager.
+        /// </summary>
+        /// <since_tizen> 13 </since_tizen>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        static public int Security
```
wiertel: why is "Security type" represented as int? If it is not an enum/flags then the meaning of the number should be described.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        /// <summary>
+        /// Retrieves and Set the Service Set Identifier (SSID) of the Wi-Fi hotspot managed by the TetheringExt Manager.
+        /// </summary>
+        /// <since_tizen> 13 </since_tizen>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        static public string Ssid
+        {
+            set
```
wiertel: set only properties should be used rarely, in special cases - like the Passphrase below, can a getter be implemented for this property?
Description of both properties says "Retrieves and Set" which not correct. Should be "Retrieves and sets ..." if there is getter, or just "Sets ..." - otherwise.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
```diff
@@ -0,0 +1,215 @@
...
+                return TetheringExtManagerImpl.Instance.Security;
+            }
+        }
+
+        /// <summary>
+        /// Returns the Visibility of the Wi-Fi hotspot managed by the TetheringExt Manager.
+        /// </summary>
+        /// <since_tizen> 13 </since_tizen>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        static public int Visibility
```
wiertel: what is the meaning of that number? shouldn't it be `bool`?

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
```diff
@@ -0,0 +1,215 @@
...
+
+    /// <summary>
+    /// A manager class that enables applications to create and manage a Wi-Fi hotspot, allowing devices to share an internet connection over a Wireless Local Area Network (WLAN).
+    /// The TetheringExt Manager provides functionality to activate and deactivate a Wi-Fi hotspot, configure hotspot settings, and manage connected devices.
+    /// </summary>
+    /// <since_tizen> 13 </since_tizen>
+    static public class TetheringExtManager
+    {
+        /// <summary>
+        /// TetheringExtEnabled is raised when the tethering is successfully enabled.
```
wiertel: Those 2 events sound like not be best usage of events in API. You can do it like this, but for informing about successful performing of something it would be better to have a method returning Task which would report success or failure. So `Task Enable()`, `Task Disable()`, and optionally StateChanged event if that state is not covered by ConnectionStateChanged.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/ConnectionStateChangedEventArgs.cs
```diff
@@ -0,0 +1,42 @@
+using System;
+
+namespace Tizen.Network.Tethering
+{
+    /// <summary>
+    /// This class contains information related to Connection State.
```
wiertel: If an event args is designed for a single event then I like to have this event referenced in summary, but for this to work the event would need to be visible, now it is hidden. Then maybe this class should also be hidden.
```suggestion
    /// This class contains information related to <see cref="TetheringExtManager.ConnectionStateChanged"/> event.
```

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtClient.cs
```diff
@@ -0,0 +1,130 @@
+using System;
+
+namespace Tizen.Network.Tethering
+{
+    /// <summary>
+    /// This class contains information related to the connected/disconnected client.
+    /// </summary>
+    /// <since_tizen> 13 </since_tizen>
+    public class TetheringExtClient : IDisposable
```
wiertel: we have namespaces, prefixing is not needed, consider name like simple "Client" or "ClientInformation" (if the class is not expected to offer any operations)

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtClient.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// Destructor function for the class.
+        /// </summary>
+        /// <since_tizen> 13 </since_tizen>
+        ~TetheringExtClient()
+        {
+            Dispose(false);
+        }
+
+        /// <summary>
+        /// This function destroys the current instance of the class.
```
wiertel: this is not precise, it rather releases (unmanaged) resources, performs kind of clean up

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
```diff
@@ -0,0 +1,215 @@
...
+        /// </summary>
+        /// <since_tizen> 13 </since_tizen>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        static public event EventHandler<TetheringExtEnabledEventArgs> TetheringExtEnabled
```
wiertel: no need for that prefix in name, just "Enabled" / "Disabled" would suffice

wiertel (changes requested):

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
r-bhalotia: This implementation is specific to Tethering Extension. There is a different implementation for Tethering in Tizen CAPI. This is the reason why main class and other related classes are prefixed with TetheringExt.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
r-bhalotia: In the CAPI, it is returning int

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
r-bhalotia: Shall I make it TetheringExtensionManager?

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
r-bhalotia: ConnectionStateChanged event comes when other device connect/disconnect to AP. It does not give any information regarding activation/deactivation of tethering. So if tethering gets enabled/disabled for some other reason (e.g. low power), there will be no way for the application to know if we use Task.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/ConnectionStateChangedEventArgs.cs
```diff
@@ -0,0 +1,42 @@
...
+        private bool _opened = false;
+
+        internal ConnectionStateChangedEventArgs(TetheringExtClient client, bool opened)
+        {
+            _client = client;
+            _opened = opened;
+        }
+
+        /// <summary>
+        /// This method returs TetheringExtClient which contains information of the connected or disconected client.
```
wiertel: not a method, same with the next one

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtClient.cs
r-bhalotia: Namespace is Tethering, but this is Tethering Extension client

src/Tizen.Network.Tethering/Interop/Interop.TetheringExtension.cs
```diff
@@ -0,0 +1,109 @@
...
+using System;
+using System.Runtime.InteropServices;
+using Tizen.Network.Tethering;
+
+internal static partial class Interop
+{
+    internal static partial class TetheringExt
+    {
+        // Callback for event
+        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
```
wiertel, line 27: x

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
wiertel: Well "TetheringExtensionManager" would be better than "TetheringExtManager", but there are more with that prefix, a sub namespace could be considered, like "Tizen.Network.Tethering.Extension.Client" etc. or "Tizen.Network.Tethering.TetheringExtension.Client".
Because the word "extension" is not used in the description of this class, then "Ext / Extension" part in the class name seems unjustified. If you mention "Tethering Extension" (and maybe say what it is) then it would look better.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
wiertel: Ok, if the state is changed externally then events make sense. Having both  async (Task returing) methods and an event informing about state change would also make sense, but the situation now is not so clear and current design can stay.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
```diff
@@ -0,0 +1,202 @@
...
+                TetheringExtManagerImpl.Instance.TetheringExtEnabled += value;
+            }
+            remove
+            {
+                TetheringExtManagerImpl.Instance.TetheringExtEnabled -= value;
+            }
+        }
+
+        /// <summary>
+        /// TetheringExtDisabled is raised when the tethering is successfully disabled.
```
wiertel: the name of the event here and in line 33 are wrong

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtManager.cs
wiertel: It's a hidden API so it can stay, but please add information about meaning of the returned number.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtClient.cs
wiertel: I keep being distracted by the word Extension and thinking about [Extension methods](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods). I've added a suggestion for namespace by the Manager class.

jiung-yu:
Based on the discussion, I believe there is no need to create a separate Tethering.Extension namespace. The functionality in question is quite limited, and it does not seem to warrant the level of scalability that would justify a separate namespace. However, it is important to clarify the naming of the classes by ensuring that the "Ext" in their names is explicitly spelled out as "Extension" for better readability and understanding.

Dear @r-bhalotia Please check addtional @wiertel 's comments and change TetheringExtManager and TetehringExtClient class

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringData.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Runtime.InteropServices;
+using Tizen.Internals;
+
+
+
+namespace Tizen.Network.Tethering
+{
+    [NativeStruct("_tethering_ext_tethering_info", Include="tethering_ext_internal.h", PkgConfig="capi-network-tethering")]
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct TetheringInfoStruct
```
SeonahMoon, line 11: Class가 아닌 struct로 정의한 이유가 있나요?
만약, 이 파일에 TetheringInfo만 정의할 예정이라면 파일명은 TetheringInfo.cs가 더 적합할 것 같습니다.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringInfo.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+        /// <summary>
+        /// Gets the IP Address of the connected/disconnected client.
+        /// </summary>
+        /// <since_tizen> 13 </since_tizen>
+        public string Passphrase
+        {
+            get
+            {
+                return _passphrase;
```
SeonahMoon, line 43: ```passphrase```를 저장하고, getter를 제공하는 것에 대해 보안 문제가 없을지 한번 검토 부탁드립니다.

src/Tizen.Network.Tethering/Tizen.Network.Tethering/TetheringExtensionClient.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// Gets the IP Address of the connected/disconnected client.
+        /// </summary>
+        /// <since_tizen> 13 </since_tizen>
+        public string IPAddress
+        {
+            get
+            {
+                if (_disposed)
+                {
+                    throw new ObjectDisposedException("Invalid TetheringExtension client instance(Object may have been disposed or released)");
```
SeonahMoon: property getter에서는 exception 발생을 지양하는게 좋습니다.
예외상황에서 빈값으로 반환하던가, 꼭 exception이 필요하다면 method도 고려해보시는게 좋을 것 같아요 😉 

```
❌ AVOID throwing exceptions from property getters.

Property getters should be simple operations and should not have any preconditions. If a getter can throw an exception, it should probably be redesigned to be a method. 
```
참고: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/property?redirectedfrom=MSDN

PR #7092, opened 2025-07-28
[Network.Nsd] Deprecate SSDP API

joan-juyeon:
### Description of Change ###
Deprecate Tizen.Network.Nsd APIs regarding ssdp

### API Changes ###
 - ACR: https://jira.sec.samsung.net/browse/TCSACR-617


wiertel (approved):
The change is ok, the word "regarding" in the PR title does not sound natural, something like "[Network.Nsd] Deprecate SSDP API" would be nicer

PR #7096, opened 2025-07-29
[NUI] Fix a crash When WebView.EvaluateJavaScript is called frequently.

huayongxu:
The callstack:
Process terminated. A callback was made on a garbage collected delegate of type 'Tizen.NUI!Tizen.NUI.BaseComponents.WebView+JavaScriptMessageHandler::Invoke'.
   at Tizen.NUI.NDalicPINVOKE.ApplicationMainLoop(System.Runtime.InteropServices.HandleRef)
   at Tizen.NUI.Application.MainLoop()
   at Tizen.NUI.NUICoreBackend.Run(System.String[])
   at Tizen.Applications.CoreApplication.Run(System.String[])
   at Tizen.NUI.NUIApplication.Run(System.String[])
   at Tizen.TV.FLUX.FluxApplication.Run(System.String[])
   at Tizen.TV.Service.CNCapsule.Program.Main(System.String[])
DotNET onSigabrt called on com.samsung.tv.cnvoice-assistant / DN_ce-assistant(20535) onSigabrt called

The reason might be that a local variable is gc-ed, but it is still used in unmanaged code, according to [1] and [2].

[1] https://stackoverflow.com/questions/6193711/call-has-been-made-on-garbage-collected-delegate-in-c
[2] https://learn.microsoft.com/en-us/dotnet/framework/debug-trace-profile/callbackoncollecteddelegate-mda

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/WebView/WebView.cs
+++ b/src/Tizen.NUI/src/public/WebView/WebView.cs
@@ -2729,7 +2729,7 @@ namespace Tizen.NUI.BaseComponents
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
         }
 
-        private Dictionary<int, JavaScriptMessageHandler> _evaluateJavaScriptHandlerMap = new Dictionary<int, JavaScriptMessageHandler>();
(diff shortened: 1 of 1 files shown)


dongsug-song (approved):
approved

PR #7105, opened 2025-08-01
[NUI] Add Window properties to react ime show/hide

rabbitfor:
### Description of Change ###
To control window content size when ime shown/hidden, Window (and it's LayoutController) should provide several APIs.

* LayoutController.ExtraWidth
* LayoutController.ExtraHeight
* LayoutController.OffsetX
* LayoutController.OffsetY

Also attachable data list added to Window to extend functionality from outside.

### API Changes ###

 - ACR:


hinohie (approved):
Looks good to me. But maybe [NUI] tag required for the title?

PR #7106, opened 2025-08-03
[NUI] Print log when IsUsingXaml property changed

hinohie:
Since it is hard to determine whether current application set `IsUsingXaml` value as false or not, let we print log if it is changed.
Default is true so, if the log not printed, mean they using xaml now..


JoogabYun (approved):

Seoyeon2Kim (approved):
LGTM!

PR #7107, opened 2025-08-04
[NUI] Call ProcessorController.Initialize at application OnInit time

hinohie:
To make ensure we can call Awake(), we should initialize this instance at mainthread at least once.


dongsug-song (approved):
approved

PR #7109, opened 2025-08-04
[NUI] Do not call `ThemeManager` constructor at Preload case

hinohie:
Since it is relative with Appfw callback, we must not touch ThemeManager at preload time.


dongsug-song (approved):
approved

PR #7128, opened 2025-08-11
[NUI] Add PropertyBridge

wonrst:
Register delegate in NUI first..
PropertyBridge.RegisterStringGetter();

Currently, only StringGetter exists, but additional types can be added as needed. Classes sending properties to DALi must inherit from IPropertyProvider and implement GetStringProperty.

This patch implements TranslatableText (which does not exist in DALi) for Text. Once this is completed, NUI-side preparation is done.

In DALi, it can be used as follows:
std::string sid = Dali::Toolkit::PropertyBridge::Get().GetStringProperty(label, "TranslatableText");

https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/328739
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/328740


hinohie (approved):
Looks good to me

PR #7135, opened 2025-08-12
[NUI] Fix SVACE issue

dongsug-song:
### Description of Change ###
[NUI] Fix SVACE issue
- WID:510174 [STATISTICAL] Value wheelEvent.Point, which is result of method invocation with possible null return value, is dereferenced in member access expression wheelEvent.Point.X

### API Changes ###
nothing


hinohie (approved):
Oh... I forgot this Svace issue! looks good to ne

PR #7130, opened 2025-08-12
[NUI][API12] Support NUIApplication.SetMaximumRenderFrameRate

hinohie:
Let we support to change maximum fps per each frames.

Relative dali patches :
- https://review.tizen.org/gerrit/c/platform/core/uifw/dali-adaptor/+/329009
- https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/329010

PR #7143, opened 2025-08-18
[NUI] Support NUIApplication.SetMaximumRenderFrameRate

hinohie:
Let we support to change maximum fps per each frames.

Relative dali patches :
- https://review.tizen.org/gerrit/c/platform/core/uifw/dali-adaptor/+/329278
- https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/329279

(Same PR with #7130 )


dongsug-song (approved):
approved

PR #7153, opened 2025-08-20
[TCSACR-618][NUI] Deprecate XAML related APIs

dongsug-song:
### Description of Change ###
[NUI] Deprecate XAML related APIs
 - XAML is not used, so proceed Deprecated ACR

### API Changes ###
https://jira.sec.samsung.net/browse/TCSACR-618


wiertel (approved):
It's unusual to hide the API along with marking it Obsolete, a little deviation from deprecation policy. Still, it can be done as it does not break the existing applications.

PR #7158, opened 2025-08-25
[Applications.Common] Add a new internal method for lightweight multi…

hjhun:
### Description of Change ###

The added method is for remounting the subsession path in runtime. The platform application can use the function to remount the subsession path.

Adds:
 - ApplicationManager.RemountSubsession()


ilho159kim (approved):
Thank you for implementing this feature Looks Good To Me

upple (approved):
Looks good to me 👍

PR #7164, opened 2025-08-26
[PackageManager] Minimize the number of event registration when adding EventHandler

ilho159kim:
### Description of Change ###
Each time PackageManagerSetEvent is called, event is unsubscribe and a new subscribe is executed
If EventHandler has already been added, PackageManagerSetEvent doesn't called

### API Changes ###
 - ACR: NONE

### DIFF
--- a/src/Tizen.Applications.PackageManager/Tizen.Applications/PackageManager.cs
+++ b/src/Tizen.Applications.PackageManager/Tizen.Applications/PackageManager.cs
@@ -36,8 +36,6 @@ namespace Tizen.Applications
         private const string LogTag = "Tizen.Applications.PackageManager";
 
         private static SafePackageManagerHandle s_handle = new SafePackageManagerHandle();
-        private static Interop.PackageManager.EventStatus s_eventStatus = Interop.PackageManager.EventStatus.All;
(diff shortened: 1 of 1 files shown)


src/Tizen.Applications.PackageManager/Tizen.Applications/PackageManager.cs
```diff
@@ -104,20 +102,21 @@ public static event EventHandler<PackageManagerEventArgs> InstallProgressChanged
         {
             add
             {
+                Log.Error(LogTag, string.Format("InstallProgressChanged add start"));
```
hjhun: Please remove these logs. I think it's not needed.
If you want to add these logs, please move these to RegisterPackageManagerEventIfNeeded() & UnregisterPackageManagerEventIfNeeded().

ilho159kim: There were logs left for debugging, but they were removed

hjhun: Yes. Thanks.

PR #7169, opened 2025-08-28
[NUI] Expose key math methods as APIs

bshsqa:
This PR enhances the Tizen.NUI mathematical library by exposing several key methods as APIs. This change allows application developers to utilize more advanced 3D mathematical functions directly in their code, which were previously internal.

- Vector3 & Vector4 Classes:
  - The `Dot` and `Cross` product methods, which were previously marked as `internal`, have been changed to `public`.

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/Common/Vector4.cs
```diff
@@ -753,56 +753,82 @@ public void Clamp(Vector4 min, Vector4 max)
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
         }
 
-        /// <inheritdoc/>
+        /// <summary>
+        /// Calculates the dot product of this vector and a Vector3.
+        /// The W component of this vector is used in the calculation.
```
hinohie:
```suggestion
        /// The W component of this vector is not used in the calculation.
```

Not used!

PR #7188, opened 2025-09-03
[NUI.Devel.Tests] Remove build warnings about WebRequest

hinohie:
We don't use network request, but just for remove build warnings.


dongsug-song (approved):
approved

PR #7198, opened 2025-09-06
[NUI.Component] Check more -1 case consider for RecyclerView

hinohie:
Make sure -1 index could not be 0
+
Avoid 0 length list access


dongsug-song (approved):
approved

PR #7199, opened 2025-09-08
[NUI] Fix LayoutItem not to dispose padding and margin

jaehyun0cho:
The initial values of padding and margin of LayoutItem is static value Extents.Zero not to allocate memory for zero extents.

If so, those padding and margin should not be disposed.

Like View, LayoutItem is fixed not to dispose padding and margin.

### API Changes ###

 - ACR:


huayongxu (approved):
lgtm

PR #7200, opened 2025-09-09
[NUI.Gadget] Fix NUIGadgetLifecycleState enumeration

hjhun:
### Description of Change ###

The PreCreated should be '5'. This patch adjusts the order of the enumeration.


upple (approved):

dongsug-song (approved):
approved

PR #7214, opened 2025-09-15
[Tizen.NUI.Gadget] Fix the OneShotService's lifecycle handling

Hyeon-Uk:
### Description of Change ###

Fix the state and if condition

Changed:
- change Running state after OnCreate()
- Wrapping OnCreate, OnDestroy with Create(), Destroy()


upple (approved):
Thank you for good patch 😊

PR #7235, opened 2025-09-22
Add missing _task null checking

upple:
### Description of Change ###

Add missing `_task` null checking

### API Changes ###

 - ACR: N/A


src/Tizen.Applications.UnitedService/Tizen.Applications/Service.cs
```diff
@@ -113,6 +113,12 @@ public void Run(AppControlReceivedEventArgs args)
             }
 
             _task = ServiceInfo.UseThread ? TizenCore.Spawn(Name) : TizenCore.Find("main");
+            if (_task == null)
+            {
+                Log.Error("Not found task(" + Name + ")");
+                return;
```
hjhun: I think we should consider throwing an exception to notify that the error is occurred.

upple: I've added exception throwing.
Thank you for the review 👍

PR #7245, opened 2025-09-24
[NUI] If it is already maximized, you should not resize the window.

JoogabYun:
### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/public/Window/BorderWindow.cs
+++ b/src/Tizen.NUI/src/public/Window/BorderWindow.cs
@@ -173,7 +173,17 @@ namespace Tizen.NUI
 
                 if (isNeedResizeByLine == true || isNeedResizeByBorder == true)
                 {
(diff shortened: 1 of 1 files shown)


src/Tizen.NUI/src/public/Window/BorderWindow.cs
```diff
@@ -173,7 +173,17 @@ internal void UpdateProperty()
 
                 if (isNeedResizeByLine == true || isNeedResizeByBorder == true)
                 {
-                    Interop.Window.SetSize(SwigCPtr, Uint16Pair.getCPtr(val));
+                    if (IsMaximized())
+                    {
+                        // When maximized, only the border can be updated without changing the window size.
+                        ResizedEventArgs e = new ResizedEventArgs();
+                        e.WindowSize = WindowSize;
```
hinohie, line 180: Is this WindowSize value confirmed the `borderInterface.MinSize` and etc?

hinohie (approved):
Minor question. and just approve

PR #7260, opened 2025-10-01
[NUI] When maximizing, OnMaximize is called after both resizing and screen rendering are complete.

JoogabYun:
### Description of Change ###

When maximizing, OnMaximize is called after both resizing and screen rendering are complete.

### API Changes ###

 - ACR:


src/Tizen.NUI/src/public/Window/BorderWindow.cs
```diff
@@ -521,12 +521,24 @@ private void DoOverlayMode(bool enable)
...
+
         private void DoMaximize(bool isMaximized)
         {
             if (this.isMaximized != isMaximized)
             {
                 borderView?.OnMaximize(isMaximized);
-                borderInterface.OnMaximize(isMaximized);
+                if (isMaximized)
+                {
+                    AddFramePresentedCallback(OnFramePresented, 0);
```
hinohie:
```suggestion
                    if (!isFramePresentedCallbackRequested)
                    {
                        isFramePresentedCallbackRequested = true;
                        AddFramePresentedCallback(OnFramePresented, 0);
                    }
```

Can't we call this presnted callback only 1 time?

If we call `DoMaximize()` multiple times, the callback enqueue very large values, and it will reduce next frame's framerate.

hinohie (approved):
Looks good to me

PR #7262, opened 2025-10-02
Fix Tizen.System.Session GetUserList

mbloch1:
We receive a buffer of 20 chars from C, but it's guaranteed to be null terminated (with potential garbage behind the null) so we shouldn't copy the whole buffer.


metiulekm:
Seems to help.

PR #7534, opened 2026-03-24
Add API for Tethering Mode changed and Update Tethering Info

Naveen1-Saini:
### Description of Change ###

1. Tethering Mode State Callback
Event is triggered whenever the mode changed callback is received from the native Tethering C API

2. Dynamic SSID and Passphrase Update API
Updates the tethering information with new SSID (service set identifier) and passphrase.


JoonghyunCho:
## Code Review

**1. Missing license header in new file**
TetheringExtensionModeChangedEventArgs.cs is missing the Apache 2.0 license header that every other source file in this module carries. Please add the standard license block at the top of the file.

**2. Incorrect exception cref types in XML docs**
In TetheringExtensionManager.cs, both ModeChanged and UpdateTetheringInfo use exception cref pointing to TetheringError.InvalidParam and TetheringError.InvalidOperation. TetheringError is an enum, not an exception type. The exception cref tag must reference actual CLR exception types (e.g., System.ArgumentException, System.InvalidOperationException).

**3. Empty-string not validated in UpdateTetheringInfo**
UpdateTetheringInfo guards against null SSID/passphrase but does not reject empty strings. Passing an empty string to the native tethering_ext_update_tethering_info entry point may result in undefined native behavior. Consider using string.IsNullOrEmpty checks instead of null-only checks.

JoonghyunCho (commented):
🤖 [AI Review]

**PR #7534 — Add API for Tethering Mode changed and Update Tethering Info**

---

**🔴 Issue: `exception cref` tags reference enum values, not exception types**

In `TetheringExtensionManager.cs`, both the `ModeChanged` event and `UpdateTetheringInfo` method use:

```xml
<exception cref="TetheringError.InvalidParam">...</exception>
<exception cref="TetheringError.InvalidOperation">...</exception>
```

`TetheringError` is an enum, not an exception class. The `cref` attribute in `<exception>` tags must reference CLR exception types. Use the actual thrown types instead, for example:

```xml
<exception cref="System.ArgumentException">Thrown when SSID or passphrase is null or empty.</exception>
<exception cref="System.InvalidOperationException">Thrown when tethering is not active or the update fails.</exception>
```

---

**🟡 Suggestion: `UpdateTetheringInfo` should also guard against empty strings**

The null checks in `TetheringExtensionManagerImpl.UpdateTetheringInfo` are good, but empty strings (`""`) passed to the native `tethering_ext_update_tethering_info` entry point may cause undefined or erroneous native behavior. Replace the null-only checks with `string.IsNullOrEmpty`:

```csharp
if (string.IsNullOrEmpty(ssid))
    throw new ArgumentException("SSID cannot be null or empty");
if (string.IsNullOrEmpty(passphrase))
    throw new ArgumentException("Passphrase cannot be null or empty");
```

---

**🟡 Suggestion: Missing Apache 2.0 license header in `TetheringExtensionModeChangedEventArgs.cs`**

Every other source file in the `Tizen.Network.Tethering` module begins with the standard Apache 2.0 license block, but the new `TetheringExtensionModeChangedEventArgs.cs` file starts directly with `using System;`. Please add the license header for consistency and legal compliance.

JoonghyunCho:
🤖 [AI Review] — Following up on @JoonghyunCho's review above, which has not yet received a response from the author.

All three points raised are valid and should be addressed before merge:

1. **Missing license header** — `TetheringExtensionModeChangedEventArgs.cs` must include the Apache 2.0 header, as is standard for every file in this module.

2. **Incorrect `exception cref` types** — `TetheringError.InvalidParam` and `TetheringError.InvalidOperation` are enum members, not exception classes. The `<exception cref="..."/>` tags in XML doc must reference actual CLR exception types (e.g., `System.ArgumentException`, `System.InvalidOperationException`). Without this, the generated API docs will have broken references.

3. **Empty-string validation** — `string.IsNullOrEmpty` is the appropriate guard here instead of null-only checks. Passing an empty string to the underlying native API is likely to silently fail or produce unexpected behavior.

@Naveen1-Saini Could you please address these items? Happy to clarify any of the points if needed.

PR #7553, opened 2026-04-06
Deprecated ComponentBased.Common api

kdk3776:
### API Changes ###
 - ACR:  TCSACR-647

### DIFF
--- a/src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
+++ b/src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
@@ -28,6 +28,7 @@ namespace Tizen.Applications.ComponentBased.Common
     /// It serves as a base class to be inherited by other components.
     /// </remarks>
     /// <since_tizen> 6 </since_tizen>
+    [Obsolete("This has been deprecated in API14")]
     public abstract class BaseComponent
     {
         internal IntPtr Handle;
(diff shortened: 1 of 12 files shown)


src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
```diff
@@ -106,6 +118,7 @@ public void Finish()
         /// A Context Handle
         /// </summary>
         /// <since_tizen> 6 </since_tizen>
+        [Obsolete("This has been deprecated in API14")]
```
wiertel: It's hidden, so there it is not necessary. Still, it's not a problem.

wiertel (commented):
Could you change "api" -> "API" in the PR title?

JoonghyunCho:
Code review posted by automated pipeline.

JoonghyunCho:
## Code Review (continued)

**1. Inconsistent Obsolete message format**
This PR uses a different deprecation message than the related PR #7556. These two PRs should align on the same wording for consistency across the Tizen.Applications.ComponentBased namespace.

**2. Avoid marking Dispose() as Obsolete**
Adding [Obsolete] to Dispose() causes a compiler warning on the implicit call inside a using statement, which is a poor developer experience. The class-level [Obsolete] attribute already warns consumers at the point of instantiation.

**3. Redundant per-member Obsolete on override methods**
[Obsolete] on overrides like OnRestoreContents and OnSaveContent is not necessary because the compiler propagates the warning from the base class. Consider removing from overrides unless API tooling specifically requires it.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
JoonghyunCho: Agreed -- since ContextHandle is [EditorBrowsable(Never)], the [Obsolete] attribute is redundant there. The note is appreciated; the author may want to remove it for cleanliness even though it causes no harm.

JoonghyunCho:
Responding to @wiertel's review: seconding the request to update the PR title from 'Deprecated ComponentBased.Common api' to 'Deprecated ComponentBased.Common API' to match the naming convention used elsewhere in TizenFX PR titles.

JoonghyunCho (commented):
🤖 [AI Review]

**PR #7553 — Deprecated ComponentBased.Common API**

---

**🟡 Suggestion: `[Obsolete]` attribute is missing a deprecation message**

All `[Obsolete]` annotations in this PR use the parameterless form:

`[Obsolete("This has been deprecated in API14")]`

Wait — the attributes do carry a message string, but it contains no migration guidance. Best practice is to direct developers to a replacement API:

`[Obsolete("Deprecated in API14. Use Tizen.Applications.XXX instead.")]`

If there is a recommended replacement for each type/member, please include it. If the entire namespace is being retired with no direct successor, at minimum add a link to the relevant ACR or release notes so developers understand what to use instead.

---

**🟡 Suggestion: XML documentation should reflect deprecation**

The existing `<summary>` XML doc comments do not mention the deprecation or point to alternatives. Consumers using IntelliSense will only see the original description, with no hint that the API is retired. Consider adding a `<remarks>` element to key types (e.g., `BaseComponent`, `ComponentBasedApplication`) such as:

```xml
<remarks>This class is deprecated since API14. Migrate to [replacement] before the API is removed.</remarks>
```

JoonghyunCho:
🤖 [AI Review]

**PR #7553 — Deprecated ComponentBased.Common api**

---

**Inconsistent deprecation message with companion PR #7556**

This PR annotates all members with `"This has been deprecated in API14"`, while PR #7556 (same deprecation effort, `ComponentBased.ComponentManager`) uses `"Deprecated since API level 14"`. Developers consuming both packages will see two different warning messages for the same deprecation wave. Please coordinate with @Hyeon-Uk and agree on a single consistent format to use across both PRs.

---

**Regarding the `ContextHandle` inline discussion**

As noted by @wiertel and @JoonghyunCho in the inline comments: adding `[Obsolete]` to `ContextHandle` (which is already `[EditorBrowsable(EditorBrowsableState.Never)]`) is redundant but harmless. Removing it would keep the annotation scope consistent — only publicly visible members need the attribute, since hidden members are already excluded from the public API surface. The decision is left to the author, but removal is the cleaner choice.

---

@kdk3776 Please address the message consistency point before this PR merges.

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/BaseComponent.cs
kdk3776: @wiertel

I can't change title.
It keeps saying "Failed to fetch"

PR #7567, opened 2026-04-15
[APPFW] Implement Team Application C# Interface

newb1e-kim:
### [Tizen.Applications] Implement Team Application C# Interface
- Implement Interop, Wrapper Class to implement C# Team Application
- Add OnCreate Error Handling
- When the application start, TeamManager will hold the application
  instance to guarantee the object alive until exit.
- Implement TeamViewApplication

---

### [NUI] Add Tizen.Application.Team into FriendAssembly

However, it does not inherit NUIApplication. Instead of inheriting
NUIApplcation, it use DALI native in backend. For developes to utilize
C# NUI object in TeamApplication, the backend should depend on NUI
deeply, require to be friend of the project.

### DIFF
--- a/src/Tizen.Applications.Team/CultureInfoHelper.cs
+++ b/src/Tizen.Applications.Team/CultureInfoHelper.cs
@@ -20,6 +20,9 @@ using System.Runtime.InteropServices;
 
 namespace Tizen.Applications
 {
+    /// <summary>
+    /// Resolves culture names from the platform-provided CultureInfo INI file.
+    /// </summary>
     internal static class CultureInfoHelper
     {
         private const string _pathCultureInfoIni = "/usr/share/dotnet.tizen/framework/i18n/CultureInfo.ini";
--- a/src/Tizen.Applications.Team/CultureInfoHelper.cs
+++ b/src/Tizen.Applications.Team/CultureInfoHelper.cs
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2026 Samsung Electronics Co., Ltd All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Tizen.Applications
+{
+    internal static class CultureInfoHelper
+    {
+        private const string _pathCultureInfoIni = "/usr/share/dotnet.tizen/framework/i18n/CultureInfo.ini";
+        private static bool _fileExists = File.Exists(_pathCultureInfoIni);
+
+        public static string GetCultureName(string locale)
+        {
+            if (!_fileExists)
+            {
+                return string.Empty;
+            }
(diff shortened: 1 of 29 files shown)


JoonghyunCho (commented):
🤖 [AI Review]

**PR #7567 — [APPFW] Implement Team Application C# Interface**

---

**🔴 Bug: `GSourceManager._tizenUIContext` is dead code (logic error in `Post()`)**

In `GSourceManager.cs`, the `Post()` method always uses `_tizenContext` because `context` is unconditionally set to `IntPtr.Zero` and never reassigned:

```csharp
public static void Post(Action action)
{
    IntPtr context = IntPtr.Zero;  // always Zero
    var sourceContext = context == IntPtr.Zero ? _tizenContext : _tizenUIContext;  // always _tizenContext
    sourceContext.Post(action, context, _wrapperHandler);
}
```

`_tizenUIContext` can never be selected. The same dead branch exists in `Handler()`. If the intent is to dispatch to either a Tizen main-loop context or a UI context depending on a parameter or current thread, the selection logic needs to be implemented. If only one context is needed, `_tizenUIContext` and the branching should be removed.

---

**🟡 Suggestion: `CultureInfoHelper._fileExists` is computed once at class load time**

`private static bool _fileExists = File.Exists(_pathCultureInfoIni);`

This static field is evaluated once when the class is first accessed. If the `.ini` file is created after the application starts, `_fileExists` will remain `false` for the process lifetime. Consider calling `File.Exists` inside `GetCultureName` on each call (or using lazy initialization) so that late-arriving files are handled correctly.

---

**🟡 Suggestion: Missing newline at end of file**

`CultureInfoHelper.cs` and `GSourceManager.cs` both lack a trailing newline. This causes noisy diffs and is inconsistent with the rest of the codebase.

---

**🟡 Note: Duplicate PR**

PR #7568 appears to contain identical content to this PR (same author, same date, same diff). One of the two should be closed to avoid duplicated review effort.

JoonghyunCho:
🤖 [AI Review] — Tizen.Applications.Team: New Package

**Bug: `_tizenUIContext` is unreachable dead code in `GSourceManager.cs`**

In `GSourceManager.Post()`, `context` is always hardcoded to `IntPtr.Zero`:

```csharp
IntPtr context = IntPtr.Zero;
var sourceContext = context == IntPtr.Zero ? _tizenContext : _tizenUIContext;
```

The ternary condition is always `true`, so `_tizenUIContext` is never selected and is dead code. The dispatch logic in `Handler()` that checks `userData == IntPtr.Zero` also always resolves to `_tizenContext`. If the intent was to support posting to both the main context and the UI context, `context` must come from a parameter. Please either expose it (e.g., `Post(Action action, IntPtr context = default)`) or remove the unused `_tizenUIContext` field.

---

**Bug: Missing `[UnmanagedFunctionPointer(CallingConvention.Cdecl)]` on `AppInfoResControlCallback` in `Interop.ApplicationManager.cs`**

Every other unmanaged delegate in this file has `[UnmanagedFunctionPointer(CallingConvention.Cdecl)]`, but `AppInfoResControlCallback` does not. Without this attribute, the runtime may use the wrong calling convention when the native library invokes the callback, potentially causing stack corruption or crashes. Please add `[UnmanagedFunctionPointer(CallingConvention.Cdecl)]` to match all sibling delegates.

---

**Suggestion: Cache the ini dictionary in `CultureInfoHelper.cs`**

`GetCultureName` calls `Interop.LibIniParser.Load` and `FreeDict` on every invocation. Since `CultureInfo.ini` is a read-only system file, the parsed dictionary should be loaded once (lazily at first use) and reused across calls. The current implementation re-parses the entire ini file on every locale query.

---

**Nit: Missing newline at end of file**

Both `CultureInfoHelper.cs` and `GSourceManager.cs` end without a trailing newline. Please add one.

hjhun (approved):

dongsug-song (approved):
approved

PR #7572, opened 2026-04-17
[NUI][API13_MR](Application) Implement NUIApplicationInitializer

hinohie:
Let we make some API to support non-NUIApplication could use NUI features.

TODO : For now, we don't support to get defautl window for this case. We'd better make some unified logic to support it.

For example, bind ApplicationController + UiContext at native, and make current NUIApplication use it.

It is quite complex job. So need to seperate PR.


dongsug-song (approved):
approved

PR #7586, opened 2026-04-23
[MediaVision] Prevent GC of object in GenerateImage (#7325)

KwangSon:
* [MediaVision] Prevent GC of object in GenerateImage

* Fix typo and build error

* Remove unnecessary try-catch

---------

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.Multimedia.Vision/MediaVision/BarcodeGenerator.cs
+++ b/src/Tizen.Multimedia.Vision/MediaVision/BarcodeGenerator.cs
@@ -237,6 +237,7 @@ namespace Tizen.Multimedia.Vision
                 imageConfig.Width, imageConfig.Height, type, qrMode, qrEcc, qrVersion,
                 imageConfig.Path, imageConfig.Format).
                 Validate("Failed to generate image");
+            GC.KeepAlive(config);
         }
 
         /// <summary>
@@ -308,6 +309,8 @@ namespace Tizen.Multimedia.Vision
         public static void GenerateImage(string message, QrConfiguration qrConfig,
             BarcodeImageConfiguration imageConfig, BarcodeGenerationConfiguration config)
         {
+            BarcodeGenerationConfiguration config_ = null;
+
             if (qrConfig == null)
             {
                 throw new ArgumentNullException(nameof(qrConfig));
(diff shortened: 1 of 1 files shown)


src/Tizen.Multimedia.Vision/MediaVision/BarcodeGenerator.cs
```diff
@@ -325,14 +328,16 @@ public static void GenerateImage(string message, QrConfiguration qrConfig,
...
                 }
             }
 
-            SetDesignQrOptions(qrConfig, config);
+            SetDesignQrOptions(qrConfig, config ?? config_);
 
-            GenerateImage(config, message, BarcodeType.QR, imageConfig, (int)qrConfig.Mode,
+            GenerateImage(config ?? config_, message, BarcodeType.QR, imageConfig, (int)qrConfig.Mode,
                 (int)qrConfig.ErrorCorrectionLevel, qrConfig.Version);
+            config_?.Dispose();
```
JoonghyunCho, line 340: 🤖 [AI Review]
🔴 **Critical**: If `SetDesignQrOptions` or the inner `GenerateImage` throws (e.g. native side returns an error), the locally-created `config_` is never disposed and its native handle leaks. Wrap the usage in `try/finally` so disposal happens on every path.

```suggestion
            try
            {
                SetDesignQrOptions(qrConfig, config ?? config_);

                GenerateImage(config ?? config_, message, BarcodeType.QR, imageConfig, (int)qrConfig.Mode,
                    (int)qrConfig.ErrorCorrectionLevel, qrConfig.Version);
            }
            finally
            {
                config_?.Dispose();
            }
```

KwangSon: Thanks for the code review.
However, this change is a cherry-pick from API14.
Would it be possible to merge it first and then make modifications afterward (to keep the code in sync)?

JoonghyunCho: 🤖 [AI Review]
Cherry-picking as-is to keep API14 in sync is a reasonable trade-off — the leak only triggers on the exception path of `SetDesignQrOptions`/`GenerateImage` (an uncommon failure case), so merging this first and applying the `try/finally` cleanup as a small follow-up that is also back-ported to API14 keeps both branches consistent without blocking this fix.

JoonghyunCho: @KwangSon Hello, the code owner of `Tizen.Multimedia.Vision` can decide what to do with the AI reviewed contents.

hsgwon (approved):

PR #7626, opened 2026-05-12
[NUI][API14] (RenderTask) Make API to setup internal camera setup without `Tizen.NUI.Camera`

hinohie:
Let we make some API to use CameraActor without use Tizen.NUI.Camera class. Setup automatcally attach / detached camera by the API call.

Their are 3 kind of API support now.

- Camera attach at RootLayer
- Camera attach at SourceView
- Camera attach at StopperView

It will show similar result what RenderTask doing by camera actor, for standard cases.

### API Changes ###

 - ACR:


dongsug-song (approved):
approved

Issue #7634, opened 2026-05-14
Clarify API13/API14 net8.0 Tizen.NET dependency and Tizen 10.1 platform support


dh329-kim:
In the public `Samsung/TizenFX` `API13` and `API14` branches, `pkg/Tizen.NET/Tizen.NET.nuspec` declares:

```xml
<dependency id="Microsoft.NETCore.Platforms" version="8.0.0-preview.7.23375.6" />
```

for the `net8.0*` target groups.

However, `packaging/depends` does not contain the matching package:

```text
microsoft.netcore.platforms.8.0.0-preview.7.23375.6.nupkg
```

Only older versions such as `3.1.0` and `6.0.9` are present.

## Affected Targets

`API13`:

```text
net8.0-tizen10.0
net8.0
```

`API14`:

```text
net8.0-tizen10.0
net8.0-tizen10.1
net8.0
```

## Impact

Offline/product builds that restore packages only from the local TizenFX NuGet source can fail because the exact dependency is not available.

## Additional Finding

There may be another related issue around the `API14` / `net8.0-tizen10.1` support model.

The `API14` branch declares support for:

```text
net8.0-tizen10.1
```

However, the currently checked `Samsung.Tizen.Sdk` from `dotnet-build-tools` appears to list supported Tizen target platform versions only up to `10.0`, and supported API levels only up to API13 / Tizen 10.0.

For example, `Samsung.Tizen.Sdk/10.0.113` contains entries such as:

```xml
<TizenSdkSupportedTargetPlatformVersion Include="10.0" />
<SupportedAPILevel Include="v13.0" MappedAPIVersion="10.0"/>
```

but I could not find equivalent built-in support for `10.1` or API14 in that build tools package.

Some consumer projects appear to work around this by adding:

```xml
<SdkSupportedTargetPlatformVersion Include="10.1" />
```

directly in their project/build props.

## Questions

Could maintainers clarify the intended support model for API13/API14 `net8.0*` packages?

Specifically:

1. Should `Microsoft.NETCore.Platforms` be removed from the `net8.0*` Tizen.NET package groups, as currently seen in `main`?
2. Or should the dependency be kept, updated to stable `8.0.0`, and the matching nupkg added to `packaging/depends`?
3. Should `dotnet-build-tools` / `Samsung.Tizen.Sdk` be updated to officially support Tizen 10.1 / API14?
[...]

PR #7636, opened 2026-05-14
Remove net8 Microsoft.NETCore.Platforms dependency from API13

dh329-kim:
## Summary
- Remove the missing `Microsoft.NETCore.Platforms 8.0.0-preview.7.23375.6` dependency from the `net8.0*` Tizen.NET package groups.
- Keep the existing legacy and net6.0 `Microsoft.NETCore.Platforms` dependencies unchanged.

## Verification
- Parsed `pkg/Tizen.NET/Tizen.NET.nuspec` as XML.
- Ran `git diff --check`.

Related to #7634.


JoonghyunCho (approved):
Removing unnecessary `Microsoft.NETCore.Platforms 8.0.0-preview.7.23375.6` dependency on API13.

PR #7556, opened 2026-04-09
[TCSACR-642][Tizen.Applications.ComponentBased.ComponentManager] Deprecate Component Manager APIs

Hyeon-Uk:
https://jira.sec.samsung.net/browse/TCSACR-642

PR #7637, opened 2026-05-14
Deprecated ComponentBased.Common API

kdk3776:
### API Changes ###

 - TCSACR-647

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/FriendAssembly.cs
+++ b/src/Tizen.NUI/src/internal/Common/FriendAssembly.cs
@@ -53,6 +53,8 @@ using System.Runtime.CompilerServices;
 [assembly: InternalsVisibleTo("Tizen.NUI.FLUX.Components, " + Tizen.NUI.PublicKey.TizenFLUX)]
 [assembly: InternalsVisibleTo("Tizen.DA.FLUX.Components, " + Tizen.NUI.PublicKey.TizenFLUX)]
 
+[assembly: InternalsVisibleTo("Tizen.Multimedia.OnePlayer, " + Tizen.NUI.PublicKey.MMOnePlayer)]
+
 namespace Tizen.NUI
 {
     internal static class PublicKey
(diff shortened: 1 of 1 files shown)


src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
```diff
@@ -131,6 +134,7 @@ public void RegisterComponent(Type compType, string compId)
         /// </code>
         /// </example>
         /// <since_tizen> 6 </since_tizen>
+        [Obsolete("This has been deprecated in API14")]
         public override void Run(string[] args)
```
JoonghyunCho, line 134: 🤖 [AI Review]
🔴 **Critical**: `Application.Run(string[])` in the base class is not `[Obsolete]`, so marking the override here triggers **CS0809 (Obsolete member overrides non-obsolete member)**. The deprecation belongs on the type itself; drop `[Obsolete]` from the override — this matches the fix in the reference deprecation PR #7556, where `[Obsolete]` was removed from `Dispose()` overrides for the same reason.

```suggestion
        /// <since_tizen> 6 </since_tizen>
        public override void Run(string[] args)
```

src/Tizen.Applications.ComponentBased/Tizen.Applications.ComponentBased.Common/ComponentBasedApplication.cs
```diff
@@ -154,6 +158,7 @@ public override void Run(string[] args)
         /// Exits the application's main loop.
         /// </summary>
         /// <since_tizen> 6 </since_tizen>
+        [Obsolete("This has been deprecated in API14")]
         public override void Exit()
```
JoonghyunCho, line 157: 🤖 [AI Review]
🔴 **Critical**: Same as the `Run` override above — `Application.Exit()` is abstract on the base and not marked `[Obsolete]`, so applying `[Obsolete]` to this override produces **CS0809**. Drop the attribute here; the class-level `[Obsolete]` already covers consumers.

```suggestion
        /// <since_tizen> 6 </since_tizen>
        public override void Exit()
```

JoonghyunCho:
🤖 [AI Review]
Left 2 inline comments (🔴 2, 🟡 0). See inline for details.

**Scope checked:**
- `[Obsolete("This has been deprecated in API14")]` correctly applied to types, events, properties, constructors, and newly-introduced-on-this-type members (consistent with the convention used in merged PR #7556).
- Identified CS0809 risk on the two `override` members (`Run`, `Exit`) — base members in `Tizen.Applications.Application` are not obsolete.
- Verified the deprecation message string matches the project convention established in #7556 (`This has been deprecated in API14`).
- Confirmed `<since_tizen>` tags are preserved on all changed members; no XML doc regressions.

---
*Automated review by AI assistant*

PR #7653, opened 2026-05-20
Fix NullReferenceException in NUIGadgetManager

upple:
### Description of Change ###

### API Changes ###

 - ACR: N/A

### DIFF
--- a/src/Tizen.Applications.UIGadget/Tizen.Applications/UIGadgetInfo.cs
+++ b/src/Tizen.Applications.UIGadget/Tizen.Applications/UIGadgetInfo.cs
@@ -222,9 +222,10 @@ namespace Tizen.Applications
 
         internal static UIGadgetInfo CreateUIGadgetInfo(string packageId)
         {
+            IntPtr handle = IntPtr.Zero;
             try
             {
-                Interop.PackageManagerInfo.ErrorCode errorCode = Interop.PackageManagerInfo.PackageInfoGet(packageId, out IntPtr handle);
+                Interop.PackageManagerInfo.ErrorCode errorCode = Interop.PackageManagerInfo.PackageInfoGet(packageId, out handle);
                 if (errorCode != Interop.PackageManagerInfo.ErrorCode.None)
                 {
                     Log.Error("Failed to get package info. packageId: " + packageId + ", error = " + errorCode);
@@ -232,21 +233,10 @@ namespace Tizen.Applications
                 }
 
                 UIGadgetInfo info = new UIGadgetInfo(packageId);
-                if (info == null)
-                {
-                    Log.Error("Failed to create UIGadgetInfo. packageId: " + packageId);
-                    return null;
-                }
 
                 SetResourceType(info, handle);
                 SetResourceVersion(info, handle);
                 SetMetadata(info, handle);
-
-                errorCode = Interop.PackageManagerInfo.PackageInfoDestroy(handle);
-                if (errorCode != Interop.PackageManagerInfo.ErrorCode.None)
-                {
-                    Log.Warn("Failed to destroy package info. error = " + errorCode);
-                }
 
                 SetExecutableFile(info);
                 SetResourceFile(info);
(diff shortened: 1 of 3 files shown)


src/Tizen.Applications.UIGadget/Tizen.Applications/UIGadgetInfo.cs
```diff
@@ -222,31 +222,45 @@ private static void SetResourceType(UIGadgetInfo info, IntPtr handle)
...
+                SetResourcePath(info);
+                SetUIGadgetResourcePath(info);
+                return info;
+            }
+            catch (Exception e)
             {
-                Log.Warn("Failed to destroy package info. error = " + errorCode);
+                Log.Error("Exception occurred while creating UIGadgetInfo. packageId: " + packageId + ", exception: " + e);
+                return null;
+            }
```
hjhun: I think it's not needed Because, the UIGadgetManager handles exceptions.

src/Tizen.Applications.UIGadget/Tizen.Applications/UIGadgetManager.cs
```diff
@@ -39,43 +39,53 @@ public static class UIGadgetManager
...
-                        var info = UIGadgetInfo.CreateUIGadgetInfo(pkg);
-                        if (info != null)
+                        foreach (var pkg in packages.Split(':'))
                         {
-                            try
-                            {
-                                _gadgetInfos.TryAdd(info.ResourceType, info);
-                            }
-                            catch (Exception e) when (e is ArgumentNullException || e is OverflowException)
+                            var info = UIGadgetInfo.CreateUIGadgetInfo(pkg);
```
hjhun, line 52: This block should be handled by try/catch block.

src/Tizen.Applications.UIGadget/Tizen.Applications/UIGadgetManager.cs
```diff
@@ -39,43 +39,53 @@ public static class UIGadgetManager
 
         static UIGadgetManager()
         {
-            var ptr = Interop.Libc.GetEnvironmentVariable("GADGET_PKGIDS");
-            if (ptr != IntPtr.Zero)
+            try
```
hjhun: Please remove this.

src/Tizen.Applications.UIGadget/Tizen.Applications/UIGadgetManager.cs
upple: I added an exception handling.

src/Tizen.Applications.UIGadget/Tizen.Applications/UIGadgetManager.cs
upple: Done

src/Tizen.Applications.UIGadget/Tizen.Applications/UIGadgetInfo.cs
upple: Done

upple (commented):
@hjhun 
Thank you for the reveiw.
Please review again.

JoonghyunCho:
🤖 [AI Review]

**Reviewed — no findings.**

**Scope checked:**
- `UIGadgetInfo.CreateUIGadgetInfo` refactored to `try/finally` — fixes the native-handle leak when any of the `SetResourceType` / `SetResourceVersion` / `SetMetadata` calls throw. `PackageInfoDestroy(handle)` now runs unconditionally; handle remains valid until `finally` because the unmanaged calls in the try body don't depend on it being already destroyed.
- Moving `SetExecutableFile` / `SetResourceFile` / `SetResourceClassName` / `SetResourcePath` / `SetUIGadgetResourcePath` inside the try block is benign — they take only `info`, not `handle`, so timing relative to `PackageInfoDestroy` is irrelevant.
- `UIGadgetManager` widens `catch (ArgumentNullException || OverflowException)` to `catch (Exception)` around per-package `CreateUIGadgetInfo` calls — defensible in a foreach-over-packages loop where one bad package should not stop the others; each failure is logged with `packageId` for diagnosis. No callers above rely on specific exception types here.
- `NUIGadgetManager.LoadGadgetInfos` drops the redundant `if (gadgetInfo != null)` check — `new NUIGadgetInfo(info)` either throws or returns non-null, so the guard was unreachable.
- No public API changes; all touched types/members are `internal` or non-public-surface — no XML-doc obligations.

---

PR #7654, opened 2026-05-20
Fix uigadget api14

upple:
### Description of Change ###

### API Changes ###

 - ACR: N/A

### DIFF
--- a/src/Tizen.Applications.UIGadget/Tizen.Applications/UIGadgetInfo.cs
+++ b/src/Tizen.Applications.UIGadget/Tizen.Applications/UIGadgetInfo.cs
@@ -222,9 +222,10 @@ namespace Tizen.Applications
 
         internal static UIGadgetInfo CreateUIGadgetInfo(string packageId)
         {
+            IntPtr handle = IntPtr.Zero;
             try
             {
-                Interop.PackageManagerInfo.ErrorCode errorCode = Interop.PackageManagerInfo.PackageInfoGet(packageId, out IntPtr handle);
+                Interop.PackageManagerInfo.ErrorCode errorCode = Interop.PackageManagerInfo.PackageInfoGet(packageId, out handle);
                 if (errorCode != Interop.PackageManagerInfo.ErrorCode.None)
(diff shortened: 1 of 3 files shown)


JoonghyunCho:
🤖 [AI Review]

**Reviewed — no findings.**

**Scope checked:**
- `UIGadgetInfo.CreateUIGadgetInfo`: try/catch/finally with `IntPtr.Zero` guard around `PackageInfoDestroy` correctly closes the handle-leak path; catch block returns `null` so the contract ("null means failed") is preserved for callers.
- `UIGadgetManager` static constructor: hard cast `(CoreApplication)CoreApplication.Current` replaced by `as`-null-check pattern — this is the actual NRE/InvalidCastException fix called out in the PR title; lifecycle handlers are simply skipped when `Current` is unavailable, which is the right call inside a static ctor where throwing would yield `TypeInitializationException`.
- Outer `catch (Exception)` blocks added to: `UIGadgetInfo.CreateUIGadgetInfo`, `UIGadgetManager` static ctor, `NUIGadgetManager` static ctor, `LoadGadgetInfos`, and public `Refresh()` — broad catches are defensible at the boundary of error-recovery code (per-package iteration, type init) and each logs a packageId or call-site context. `Refresh()` now swallows-and-logs instead of propagating; this is an intentional behavior change called out in the PR description (no explicit `<exception>` doc to break).
- Inner inner `TryAdd` catch in `UIGadgetManager` retained its narrower `when (e is ArgumentNullException || e is OverflowException)` filter — only the message was enriched with `packageId`. Specific filter preserved where applicable; not over-widened.
- `NUIGadgetManager.LoadGadgetInfos` drops the redundant `if (gadgetInfo != null)` check — `new NUIGadgetInfo(info)` either throws or returns non-null. Same cleanup as sibling PR #7653.
- Diff vs. #7653: this API14 backport additionally wraps the static ctors and `Refresh()` and broadens the per-item catch — consistent with API14 being a stabilization branch where "don't crash app init" outweighs "fail loudly".

---

PR #7655, opened 2026-05-20
Fix uigadget api13

upple:
### Description of Change ###

### API Changes ###

 - ACR: N/A

### DIFF
--- a/src/Tizen.Applications.UIGadget/Tizen.Applications/UIGadgetInfo.cs
+++ b/src/Tizen.Applications.UIGadget/Tizen.Applications/UIGadgetInfo.cs
@@ -222,9 +222,10 @@ namespace Tizen.Applications
 
         internal static UIGadgetInfo CreateUIGadgetInfo(string packageId)
         {
+            IntPtr handle = IntPtr.Zero;
             try
             {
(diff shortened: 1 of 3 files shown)


JoonghyunCho:
🤖 [AI Review]

**Reviewed — no findings.**

**Scope checked:**
- `UIGadgetInfo.CreateUIGadgetInfo`: `PackageInfoDestroy` moved into a `finally` — the native package handle is now released on every path including exceptions (fixes the prior leak), and the resource setters still run while the handle is valid.
- Static-constructor hardening: `(CoreApplication)CoreApplication.Current` hard cast → `as CoreApplication` + null guard avoids `InvalidCastException`/`TypeInitializationException`; lifecycle event wiring remains unconditional.
- `NUIGadgetManager.LoadGadgetInfos`: removal of the dead `if (gadgetInfo != null)` after `new NUIGadgetInfo(...)` is safe (a constructor cannot return null) and per-item failures are now isolated in the inner catch.
- Public API surface: no signature changes; only `Refresh()` body changed and it retains its `<since_tizen> 13` documentation — no missing-doc surface across the 3 files (all internal/private/static-ctor or body-only).

---

PR #7656, opened 2026-05-20
[NUI][API14][TCSACR-652] Make deprecated class `Tizen.NUI.Adaptor` as internal

hinohie:
### Description of Change ###

[NUI] Make deprecated class `Tizen.NUI.Adaptor` as internal

Note : Some friend assembled project use `Adaptor.Instance` so we cannot remove it. Just keep it as internal.

### API Changes ###

 - ACR: [TCSACR-652](https://jira.sec.samsung.net/browse/TCSACR-652)

Removed:
 - class Tizen.NUI.Adaptor
 - static Adaptor Instance { get; }
 - void FeedWheelEvent(Wheel);
 - void FeedKeyEvent(Key);

### DIFF
--- a/src/Tizen.NUI/src/internal/Common/Adaptor.cs
+++ b/src/Tizen.NUI/src/internal/Common/Adaptor.cs
@@ -36,9 +36,10 @@ namespace Tizen.NUI
     /// need to wait for an initialize signal as per the Tizen.NUI.Application class).
     ///
     /// </summary>
-    /// <since_tizen> 4 </since_tizen>
-    [Obsolete("This has been deprecated in API8 and will be removed in API10. This is not used anymore, do not use.")]
-    public class Adaptor : Disposable
+    /// <remarks>
+    /// Deprecated at API8; made internal in API14.
+    /// </remarks>
+    internal class Adaptor : Disposable
     {
         private static readonly Adaptor instance = Adaptor.Get();
 
@@ -50,9 +51,7 @@ namespace Tizen.NUI
         /// Returns a reference to the instance of the adaptor used by the current thread.
         /// </summary>
         /// <remarks>The adaptor has been initialized. This is only valid in the main thread.</remarks>
-        /// <since_tizen> 4 </since_tizen>
-        [Obsolete("This has been deprecated in API8 and will be removed in API10. This is not used anymore, do not use.")]
-        public static Adaptor Instance
+        internal static Adaptor Instance
         {
             get
             {
@@ -64,9 +63,7 @@ namespace Tizen.NUI
         /// Feeds a wheel event to the adaptor.
         /// </summary>
         /// <param name="wheelEvent">The wheel event.</param>
-        /// <since_tizen> 4 </since_tizen>
-        [Obsolete("This has been deprecated in API8 and will be removed in API10. This is not used anymore, do not use.")]
-        public void FeedWheelEvent(Wheel wheelEvent)
+        internal void FeedWheelEvent(Wheel wheelEvent)
         {
             Interop.Adaptor.FeedWheelEvent(SwigCPtr, Wheel.getCPtr(wheelEvent));
             if (NDalicPINVOKE.SWIGPendingException.Pending) throw NDalicPINVOKE.SWIGPendingException.Retrieve();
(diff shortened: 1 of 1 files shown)


JoonghyunCho:
🤖 [AI Review]

**Reviewed — no findings.**

**Scope checked:**
- Public → internal transition of `Tizen.NUI.Adaptor` and its 4 members (`Instance`, `FeedWheelEvent`, `FeedKeyEvent`, `ReleaseSwigCPtr`) — accessibility downgrade is consistently applied; `[Obsolete]` and `<since_tizen>` correctly dropped since the surface is no longer public.
- Static field initialization `private static readonly Adaptor instance = Adaptor.Get();` is unaffected — type-init still runs for friend-assembly consumers of `Adaptor.Instance`.
- `protected override ReleaseSwigCPtr` keeps `[EditorBrowsable(Never)]` — harmless on an internal class; left in place to preserve override correctness.
- Breaking-change motivation matches PR description: friend assemblies still call `Adaptor.Instance`, so deletion is not possible and `internal` is the minimal-surface choice.
- No new public API introduced, no XML-doc obligations triggered.

---

src/Tizen.NUI/src/internal/Common/Adaptor.cs
```diff
@@ -36,9 +36,10 @@ namespace Tizen.NUI
     /// need to wait for an initialize signal as per the Tizen.NUI.Application class).
     ///
     /// </summary>
-    /// <since_tizen> 4 </since_tizen>
-    [Obsolete("This has been deprecated in API8 and will be removed in API10. This is not used anymore, do not use.")]
-    public class Adaptor : Disposable
+    /// <remarks>
+    /// Deprecated at API8, Become internal at API14.
+    /// </remarks>
+    internal class Adaptor : Disposable
```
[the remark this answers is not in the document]
hinohie: No. We don't need to add obsolete tag for change public->internal.

JoonghyunCho:
🤖 [AI Review]

**Reviewed — no findings.**

**Scope checked (delta since previous review):**
- File relocated from `src/Tizen.NUI/src/public/Common/Adaptor.cs` to `src/Tizen.NUI/src/internal/Common/Adaptor.cs` — relocation matches the accessibility downgrade; no other source moves.
- Class and 4 members (`Adaptor`, `Instance`, `FeedWheelEvent`, `FeedKeyEvent`, `ReleaseSwigCPtr`) are now `internal`; `[Obsolete]` and `<since_tizen>` correctly dropped — symbol surface is no longer public, so neither attribute is meaningful.
- `protected override ReleaseSwigCPtr` retains `[EditorBrowsable(Never)]` — harmless on an internal class; left intact to preserve override semantics.
- Static field initialization `private static readonly Adaptor instance = Adaptor.Get();` unchanged — friend-assembly consumers of `Adaptor.Instance` still resolve correctly.
- `<remarks>Deprecated at API8; made internal in API14.</remarks>` is a stylistic preposition choice ("at" vs "in"); not a correctness issue.

---

src/Tizen.NUI/src/internal/Common/Adaptor.cs
```diff
@@ -36,9 +36,10 @@ namespace Tizen.NUI
     /// need to wait for an initialize signal as per the Tizen.NUI.Application class).
     ///
     /// </summary>
-    /// <since_tizen> 4 </since_tizen>
-    [Obsolete("This has been deprecated in API8 and will be removed in API10. This is not used anymore, do not use.")]
-    public class Adaptor : Disposable
+    /// <remarks>
```
wiertel, line 39: This `<remarks>` look like a use in the native/Doxygen style where remarks adds important information, in C# xmldoc `<remarks>` works more like `@details` from Doxygen - should contain additional description which expands the short `<summary>`. IMO the best structure would be keep only first 2 paragraphs in summary, add this information about deprecation, and move rest of the summary into remarks. But this description will not be seen by public so improving it slightly is not a big deal.

wiertel (approved):

src/Tizen.NUI/src/internal/Common/Adaptor.cs
JoonghyunCho: 🤖 [AI Review]
Agreed on the semantics — in C# XMLdoc `<remarks>` extends `<summary>` (closer to Doxygen `@details`), so the deprecation/internal-status note is the misfit, not the surrounding description. Since the type is now `internal` and these docs are not part of the published surface, the cleanest minimal change is to keep only the deprecation/migration note inside `<remarks>` and leave the rest of the prose as-is; a full split into summary/remarks paragraphs would be over-investment for an IntelliSense-hidden, internal class.

hinohie: Thank you for your infomations. I used to using `remarks` tag somthing like "notes".  As you said, this is remove ACR so documents are not the important points. So I'll just merge this PR now, and will consider remarks tags in future :)

PR #7671, opened 2026-06-01
[NUI][TCSACR-652] Make deprecated class Tizen.NUI.Adaptor as internal

hinohie:
### Description of Change ###

[NUI] Make deprecated class Tizen.NUI.Adaptor as internal

Note : Some friend assembled project use Adaptor.Instance so we cannot remove it. Just keep it as internal.

### API Changes ###

ACR: [TCSACR-652](https://jira.sec.samsung.net/browse/TCSACR-652)
Removed:

class Tizen.NUI.Adaptor
static Adaptor Instance { get; }
void FeedWheelEvent(Wheel);
void FeedKeyEvent(Key);


dongsug-song (approved):
looks good to me

PR #7752, opened 2026-07-15
[NUI] Add LottieAnimationView.RefreshDynamicProperty()

JoogabYun:
### Description of Change ###

Forces dynamic properties registered via DoActionExtension() to be re-evaluated at the current frame without changing which frame is displayed. Needed while the animation is paused or stopped, where re-setting CurrentFrame to the same value is a no-op for the underlying renderer, so a dynamic property callback's updated return value would otherwise never be picked up.

Marked EditorBrowsable(Never) as Inhouse API pending ACR, same as the existing DoActionExtension()/FlushLottieMessages() pattern.

refer
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-csharp-binder/+/348329
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-toolkit/+/348327
https://review.tizen.org/gerrit/c/platform/core/uifw/dali-adaptor/+/348326

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/internal/Interop/Interop.LottieAnimationView.cs
+++ b/src/Tizen.NUI/src/internal/Interop/Interop.LottieAnimationView.cs
@@ -29,6 +29,9 @@ namespace Tizen.NUI
 
             [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_AnimatedVectorImageVisual_Actions_FLUSH_get")]
             public static extern int AnimatedVectorImageVisualActionFlushGet();
+
+            [global::System.Runtime.InteropServices.DllImport(NDalicPINVOKE.Lib, EntryPoint = "CSharp_AnimatedVectorImageVisual_Actions_REFRESH_DYNAMIC_PROPERTY_get")]
+            public static extern int AnimatedVectorImageVisualActionRefreshDynamicPropertyGet();
         }
     }
 }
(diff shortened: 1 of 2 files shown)


JoonghyunCho:
🤖 [AI Review]

**Reviewed — no findings.**

**Scope checked:**
- New public `RefreshDynamicProperty()` carries XML docs (`<summary>`, `<remarks>`), and `[EditorBrowsable(EditorBrowsableState.Never)]` correctly exempts it from the `<since_tizen>` requirement, matching the pre-ACR in-house pattern of `DoActionExtension()`/`FlushLottieMessages()`.
- `Disposed` guard is in place before the interop call, preventing native access through a disposed handle.
- Verified the interop call against all 6 existing `DoActionWithEmptyAttributes` call sites (`ImageView.Play/Pause/Stop/Reload`, `FlushLottieMessages`, `AnimatedImageVisual`): none performs a `SWIGPendingException` check on this action path — the omission here follows the established convention for this specific interop (only `DoActionExtension`, a different interop entry point, uses that check).
- New P/Invoke `AnimatedVectorImageVisualActionRefreshDynamicPropertyGet` and the `ActionRefreshDynamicProperty` static field mirror the existing FLUSH action wiring exactly; the native entry point depends on the linked dali-csharp-binder change, consistent with the **NO MERGE** label until the native side lands.
- Change is purely additive (1 hidden public method, 1 internal interop method, 1 internal field) — no breaking change to the API14 surface; identical to the DevelNUI counterpart #7751.

---

dongsug-song (approved):
approved

PR #7769, opened 2026-07-22
[VoiceControlManager] Fix svace null-dereference issue in recognized …

syeonholic:
…commands callback

### Description of Change ###
Marshal.PtrToStringAnsi() returns null when the native pointer is IntPtr.Zero. The resulting null msgString (and recognizedString) was passed into SelectRecognizedCommandsDelegate.Invoke(), which svace reports as a null-value dereference and which could cause a NullReferenceException inside consumer delegate implementations.

Coalesce both marshaled strings to string.Empty at the source so the delegate and AllRecognitionResultEventArgs always receive non-null values.


stom-hwang (approved):
LGTM 👍

JoonghyunCho:
🤖 [AI Review]

**Reviewed — no findings.**

**Scope checked:**
- Root cause matches the reported svace null-dereference: `Marshal.PtrToStringAnsi` returns `null` when the native pointer is null, and both call sites now coalesce to `string.Empty`.
- Both changed lines live inside the `private static` `_recognizedCommandsSelectionCallback` delegate — no public API surface or signature change, so no XML-doc impact.
- Downstream usage stays null-safe with the empty-string substitution; no behavioral regression versus the previous nullable value.

---

JoonghyunCho (approved):

wonnamy (approved):
LGTM

urmain (approved):
LGTM

PR #7702, opened 2026-06-15
Removed deprecated Tizen.Applications.AttachPanel

Md-OmarFaruque:
### Description of Change ###

Removed deprecated Tizen.Applications.AttachPanel

### API Changes ###

 - TCSACR-654

PR #7787, opened 2026-08-03
Removed deprecated Tizen.Applications.AttachPanel

Md-OmarFaruque:
### API Changes ###

 - TCSACR-654


wiertel (approved):
It's ok. This is exactly the same change as #7702 but for the API14 branch. The change is intended for API 14 so it's good.

PR #7762, opened 2026-07-19
[AI Task] [Tizen.Pims.Contacts] Remove Convert.ChangeType boxing from ContactsRecord Get/Set

JoonghyunCho:
## Summary
Removes the `Convert.ChangeType` / `Convert.To*` conversion paths from `ContactsRecord.Get<T>(uint)` and `Set<T>(uint, T)`. Each branch already verifies the exact type via `typeof(T) == typeof(X)`, so the value is converted with a direct cast (`(T)(object)val`) instead of going through the culture-aware `IConvertible` dispatch. `Get<T>` also drops the intermediate `object parsedValue` local and returns directly from each branch.

Per-call effect (e.g. `Get<int>`): no `Convert.ChangeType` call, no `IConvertible` vtable dispatch, no `CultureInfo` lookup. For `Get<string>`/`Set<string>` the boxing allocation disappears entirely (reference-type cast). `ContactsRecord` property access is the per-field unit of contact serialization, so bulk sync/import workloads (10k records × ~10 properties) save hundreds of thousands of dispatch/boxing operations.

## Changes
- `src/Tizen.Pims.Contacts/Tizen.Pims.Contacts/ContactsRecord.cs`
  - `Get<T>`: each type branch now ends with `return (T)(object)val;`; removed `object parsedValue` local, `Convert.ChangeType` calls (5), and the trailing unbox `return (T)parsedValue;`. Unsupported-type branch still throws `NotSupported` as before.
  - `Set<T>`: `Convert.ToString/ToInt32/ToBoolean/ToInt64/ToDouble` replaced with direct casts. The string branch uses `(string)(object)value ?? string.Empty` to preserve the exact prior behavior of `Convert.ToString(object)`, which maps `null` to `string.Empty` (so native `SetStr` still receives an empty string, not `null`).

## Mode
Refactoring

## Verification
- [x] Build: passed (0 errors, 0 warnings)
- [ ] Tests: N/A
- [ ] Benchmark: skipped (sdb error: no device connected)

Public API signatures (`T Get<T>(uint)`, `void Set<T>(uint, T)`) are unchanged. Every direct cast is guarded by the exact `typeof(T)` check, so no `InvalidCastException` is reachable on paths that previously succeeded; normal-path results are bit-identical, including `Get<string>` returning `null` when the native value is `NULL`.

Fixes #7667

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7800, opened 2026-08-09
[AI Task] [Tizen.Pims.Calendar] Remove Convert.ChangeType boxing from CalendarRecord Get/Set

### LINKED ISSUE #7710: [AI Refactoring] CalendarRecord.Get<T>/Set<T> 의 Convert.ChangeType 박싱 + IConvertible 디스패치 제거 → 직접 캐스트 [Scope: src/Tizen.Pims.Calendar] (2026-06-21)
[Type: Refactoring]
[Scope: src/Tizen.Pims.Calendar]
[Priority: 🟡 Improvement]
[Lens: Performance, Modernization]

## Observation
`CalendarRecord.Get<T>(uint)` and `Set<T>(uint, T)` are the primary record field accessors. Each branch already resolves the concrete type via a `typeof(T) == typeof(...)` guard, yet still routes the already-typed value through `Convert.ChangeType` / `Convert.ToXxx`.

`src/Tizen.Pims.Calendar/Tizen.Pims.Calendar/CalendarRecord.cs:252-317` (Get) and `:328-388` (Set):
```csharp
public T Get<T>(uint propertyId)
{
    object parsedValue = null;
    if (typeof(T) == typeof(string))
    {
        string val;
        int error = Interop.Record.GetString(_recordHandle, propertyId, out val);
        ...
        parsedValue = Convert.ChangeType(val, typeof(T));   // IConvertible round-trip
    }
    else if (typeof(T) == typeof(int))
    {
        int val;
        ...
        parsedValue = Convert.ChangeType(val, typeof(T));   // box + IConvertible.ToType
    }
    ... // long, double, CalendarTime — all the same shape
    return (T)parsedValue;                                   // unbox
}
```
And in `Set<T>` at L372:
```csharp
CalendarTime time = (CalendarTime)Convert.ChangeType(value, typeof(CalendarTime));
```
`Convert.ChangeType` appears **6 times** in this file.

This is the exact pattern already approved for the sibling module in `ContactsRecord.Get<T>/Set<T>` (#7667).

## Problem
**Performance.** Inside each `typeof(T)` branch the value is already the requested type, so `Convert.ChangeType(val, typeof(T))` is pure overhead: it boxes the value into `object`, runs the `IConvertible.ToType` machinery (type-code switch + virtual dispatch), and the result is then unboxed by `return (T)parsedValue;`. For the `string` branch it is a no-op round-trip. For `CalendarTime` (L372) it redundantly boxes an already-typed reference. `Get`/`Set` are called once per field, so iterating a `CalendarList` multiplies this by (records x fields).

## Proposed Improvement
Drop the `parsedValue` indirection and cast directly through `object` per branch (identical to the approved #7667 fix). For value types this keeps the single unavoidable box but removes the `Convert`/`IConvertible` dispatch; for `string`/`CalendarTime` it removes the redundant conversion entirely.

Before:
```csharp
parsedValue = Convert.ChangeType(val, typeof(T));
...
return (T)parsedValue;
```
After:
```csharp
return (T)(object)val;   // per branch; val is already the concrete T
```
And L372:
```csharp
CalendarTime time = (CalendarTime)(object)value;
```

### Target Files
- `src/Tizen.Pims.Calendar/Tizen.Pims.Calendar/CalendarRecord.cs` (Get: L252-317, Set: L328-388)

## Expected Impact (Quantitative Metrics)
- `Convert.ChangeType` / `IConvertible.ToType` dispatch per `Get`/`Set` call: **eliminated** (6 sites)
- `string` branch: redundant conversion round-trip removed (allocations unchanged, dispatch removed)
- `CalendarTime` Set branch: 1 redundant box removed
- Net allocations for value-type branches: unchanged (single box is inherent to a generic returning `T` from a value type), but reflection/dispatch overhead per access removed on the warm record-access path

## API Compatibility Check
- Public API signature change: **none** (`Get<T>`/`Set<T>` signatures untouched; only the method bodies change)
- Behavior change: **none** — the `typeof(T)` guards already constrain `T` to the supported set; outputs are bit-identical
- Tizen API Level floor: **maintained** — plain cast; no new API used. Assembly targets `net8.0`

## Impact Scope
- Internal call sites within the assembly: `Get<int>` at `CalendarDatabase.cs:403` (inside a per-record `Delete(CalendarList)` loop). `Get<T>`/`Set<T>` are public API, so dominant call volume is in consumer apps — but the change is confined to the method bodies and is behavior-preserving.
- Within the same assembly for internal callers; no cross-assembly signature exposure
- Well under 100 sites — no scope split needed

JoonghyunCho:
## Summary
Removes the `Convert.ChangeType` / `Convert.To*` conversion paths from `CalendarRecord.Get<T>(uint)` and `Set<T>(uint, T)`. Each branch already verifies the exact type via `typeof(T) == typeof(X)`, so the value is now converted with a direct cast (`(T)(object)val`) instead of going through the culture-aware `IConvertible` dispatch. `Get<T>` also drops the intermediate `object parsedValue` local and returns directly from each branch.

Per-call effect (e.g. `Get<int>`): no `Convert.ChangeType` call, no `IConvertible` vtable dispatch, no `CultureInfo` lookup. For `Get<string>`/`Set<string>` and the `CalendarTime` branches the redundant conversion round-trip disappears entirely (reference-type cast). `Get`/`Set` run once per record field, so iterating a `CalendarList` multiplies the savings by (records x fields).

This applies the identical pattern already approved for the sibling module in #7762 (`ContactsRecord`, issue #7667).

## Changes
- `src/Tizen.Pims.Calendar/Tizen.Pims.Calendar/CalendarRecord.cs`
  - `Get<T>`: each type branch now ends with `return (T)(object)val;`; removed the `object parsedValue` local, the `Convert.ChangeType` calls (5), and the trailing unbox `return (T)parsedValue;`. The unsupported-type branch still throws `NotSupported` as before.
  - `Set<T>`: `Convert.ToString/ToInt32/ToInt64/ToDouble` replaced with direct casts; `Convert.ChangeType` in the `CalendarTime` branch replaced with `(CalendarTime)(object)value`. The string branch uses `value as string ?? string.Empty` to preserve the exact prior behavior of `Convert.ToString(object)`, which maps `null` to `string.Empty` (so native `SetString` still receives an empty string, not `null`).

## Mode
Refactoring

## Verification
- [x] Build: passed (0 errors, 0 warnings)
- [ ] Tests: N/A
- [ ] Benchmark: skipped (sdb error: no device connected)

Public API signatures (`T Get<T>(uint)`, `void Set<T>(uint, T)`) are unchanged. Every direct cast is guarded by the exact `typeof(T)` check, so no `InvalidCastException` is reachable on paths that previously succeeded; normal-path results are bit-identical, including `Get<string>` returning `null` when the native value is `NULL` and `Convert.ChangeType`'s same-type shortcut for `CalendarTime` (not `IConvertible`).

Fixes #7710

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Tizen.Pims.Calendar/Tizen.Pims.Calendar/CalendarRecord.cs
+++ b/src/Tizen.Pims.Calendar/Tizen.Pims.Calendar/CalendarRecord.cs
@@ -251,7 +251,6 @@ namespace Tizen.Pims.Calendar
         /// <exception cref="ArgumentException">Thrown when one of the arguments provided to a method is not valid</exception>
         public T Get<T>(uint propertyId)
         {
-            object parsedValue = null;
             if (typeof(T) == typeof(string))
             {
                 string val;
@@ -261,7 +260,7 @@ namespace Tizen.Pims.Calendar
                     Log.Error(Globals.LogTag, $"Get String Failed [{error}]{propertyId:X}");
                     throw CalendarErrorFactory.GetException(error);
                 }
-                parsedValue = Convert.ChangeType(val, typeof(T));
+                return (T)(object)val;
             }
             else if (typeof(T) == typeof(int))
             {
@@ -272,7 +271,7 @@ namespace Tizen.Pims.Calendar
                     Log.Error(Globals.LogTag, $"Get Intger Failed [{error}]{propertyId:X}");
                     throw CalendarErrorFactory.GetException(error);
                 }
-                parsedValue = Convert.ChangeType(val, typeof(T));
+                return (T)(object)val;
             }
             else if (typeof(T) == typeof(long))
             {
@@ -283,7 +282,7 @@ namespace Tizen.Pims.Calendar
                     Log.Error(Globals.LogTag, $"Get Long Failed [{error}]{propertyId:X}");
                     throw CalendarErrorFactory.GetException(error);
                 }
-                parsedValue = Convert.ChangeType(val, typeof(T));
+                return (T)(object)val;
             }
             else if (typeof(T) == typeof(double))
             {
@@ -294,7 +293,7 @@ namespace Tizen.Pims.Calendar
                     Log.Error(Globals.LogTag, $"Get Double Failed [{error}]{propertyId:X}");
                     throw CalendarErrorFactory.GetException(error);
                 }
-                parsedValue = Convert.ChangeType(val, typeof(T));
+                return (T)(object)val;
             }
             else if (typeof(T) == typeof(CalendarTime))
             {
@@ -306,14 +305,13 @@ namespace Tizen.Pims.Calendar
                     throw CalendarErrorFactory.GetException(error);
                 }
                 CalendarTime val = ConvertIntPtrToCalendarTime(time);
-                parsedValue = Convert.ChangeType(val, typeof(T));
+                return (T)(object)val;
             }
             else
             {
                 Log.Error(Globals.LogTag, "Not supported Data T/ype");
                 throw CalendarErrorFactory.GetException((int)CalendarError.NotSupported);
             }
-            return (T)parsedValue;
         }
 
         /// <summary>
@@ -329,7 +327,7 @@ namespace Tizen.Pims.Calendar
         {
             if (typeof(T) == typeof(string))
             {
-                string val = Convert.ToString(value);
+                string val = value as string ?? string.Empty;
                 int error = Interop.Record.SetString(_recordHandle, propertyId, val);
                 if (CalendarError.None != (CalendarError)error)
                 {
@@ -339,7 +337,7 @@ namespace Tizen.Pims.Calendar
             }
             else if (typeof(T) == typeof(int))
             {
-                int val = Convert.ToInt32(value);
+                int val = (int)(object)value;
                 int error = Interop.Record.SetInteger(_recordHandle, propertyId, val);
                 if (CalendarError.None != (CalendarError)error)
                 {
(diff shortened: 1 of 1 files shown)


JoonghyunCho:
⚠️ **Benchmark skipped**: Encountered an issue with sdb device connection/deployment/execution, so the benchmark could not be run. Manual benchmark verification is required. Error: `sdb devices` reports no attached devices (empty device list), so the comparative benchmark could not be deployed or executed.

JoonghyunCho:
🤖 [AI Review]

**Reviewed — no findings.**

**Scope checked:**
- All 10 direct casts (5 `Get<T>` branches, 5 `Set<T>` branches) are dominated by an exact `typeof(T) == typeof(X)` guard, so no path that previously succeeded can now throw `InvalidCastException`.
- `value as string ?? string.Empty` in the `Set<T>` string branch preserves the `Convert.ToString(object)` null-to-empty mapping, so native `SetString` still never receives `null`.
- `Get<string>` still surfaces a native NULL as a `null` return — `(T)(object)val` matches the prior `Convert.ChangeType` null behavior.
- After removing the `parsedValue` local, every `Get<T>` branch returns or throws (the unsupported-type branch still throws `NotSupported`), so control flow remains complete.
- The `CalendarTime` branches previously relied on the `Convert.ChangeType` same-type shortcut (the type is not `IConvertible`); the direct cast is behavior-identical, including `null` pass-through.
- Public API signatures `T Get<T>(uint)` and `void Set<T>(uint, T)` are unchanged.

---

PR #7813, opened 2026-08-21
[NUI] Supports View Annotation

sparrow74:
Supports View Annotation for AIOS
It has methods and sample.


dongsug-song (approved):
approved

PR #7831, opened 2026-09-11
[NUI] Use dedicated interop for Renderable.UpdateArea

heeyongsong:
### Description of Change ###

DALi now stores the renderer update area margin as Insets, so setting it through an Extents PropertyValue no longer works. Renderable.UpdateArea and the deprecated Renderer.UpdateArea call
CSharp_Dali_Renderer_Set/GetUpdateAreaMargin with four floats instead, and the unused UpdateAreaExtents property index is removed. The values are no longer limited to integers.

Requires the matching dali-csharp-binder change.

### API Changes ###

 - ACR:

### DIFF
--- a/src/Tizen.NUI/src/deprecated/Rendering/Renderer_Deprecated.cs
+++ b/src/Tizen.NUI/src/deprecated/Rendering/Renderer_Deprecated.cs
@@ -728,32 +728,23 @@ namespace Tizen.NUI
         /// Gets and Sets extents of partial update area.
         /// </summary>
         /// <remarks>
-        /// Extents the area - the position and the size - used for the attached View's partial update area calculation.
+        /// Extends the area - the position and the size - used for the attached View's partial update area calculation.
         /// This value be appended after calculate all update area, like visual offset.
         /// Change  <see cref="Tizen.NUI.BaseComponents.View.UpdateAreaHint"/> value if you want to change View's partial update area.
-        /// Warning : Only 0u ~ 65535u integer values are allowed for each parameters.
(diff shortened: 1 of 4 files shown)


dongsug-song (approved):
looks good to me

JoonghyunCho:
🤖 [AI Review]

**Reviewed — no findings.**

**Scope checked:**
- Searched the full tree at the PR head for remaining references to the removed `UpdateAreaExtents` property index — none left (`Renderer.Property`, `RendererProperty`, and the `UpdateAreaExtentsGet` P/Invoke are removed together, so no dangling compile references).
- Compared the two new P/Invoke signatures (`SetUpdateAreaMargin` taking 4 floats, `GetUpdateAreaMargin` with 4 out floats) against both call sites in `Renderable.cs` and `Renderer_Deprecated.cs` — parameter order (start, end, top, bottom) matches `UIExtents` construction on both sides.
- Both rewritten getter/setter bodies call `NDalicPINVOKE.ThrowExceptionIfExists()` after the interop call, consistent with NUI error propagation; the dead `temp == null` check and the disposable `Extents`/`PropertyValue` round-trip are gone.
- `UpdateArea` is `[EditorBrowsable(EditorBrowsableState.Never)]` on both classes, so no doc-tag requirements apply; the remark correctly drops the obsolete 0–65535 integer limitation now that floats are passed directly.
- Cross-repo dependency on the matching dali-csharp-binder entry points (`CSharp_Dali_Renderer_Set/GetUpdateAreaMargin`) is explicitly declared in the PR description.

---

PR #7840, opened 2026-09-13
[NUI] Rebase main to DevelNUI

dongsug-song:
[NUI] Rebase main to DevelNUI


dongsug-song:
these patches are already reviewed, so just merge them without review now.

PR #7845, opened 2026-09-15
[NUI] Wake event processing after TextField TextChanged

wonrst:
### Description of Change ###
Wake the processor after TextChanged handlers return so pending relayout requests from controls updated during IME input can be processed without waiting for another input or cursor blink event.

### API Changes ###

 - ACR:


jaehyun0cho (approved):
LGTM

PR #7849, opened 2026-09-16
[Tizen.System.Session] Ensure no-op (un)registration works correctly

metiulekm:
### Description of Change ###

The C# bindings coalesce multiple C# callbacks into one C callback, which is registered with the first C# callback and unregistered with the last C# callback. These operations should only happen on non-null<>null edges, otherwise there might be multiple registrations, or we can attempt unregistering when there is no registration.

However C# delegate removal and combination operators is actually more flexible than expected, in particular C# accepts nulls in both cases. Additionally, removal allows unregistering callbacks that are not registered, which is a no-op. These cases mean that some operations can start from a null and end in a null. In this case, we will call C callback (un)registration incorrectly.

This commit adds null checks:
- registrations check if the callback being registered is non-null,
  - if it is a non-null, the result will always be non-null,
  - if it is a null, it is safe to return because adding a null does not do anything,
- unregistrations check if the existing callback list is non-null,
  - it being non-null is enough to eliminate null<>null edges,
  - if it is a null, it is safe to return because there is nothing to unregister anyway.

### API Changes ###

None.


mbloch1 (approved):

JoonghyunCho:
🤖 [AI Review]

**Reviewed — no findings.**

**Scope checked:**
- All 4 custom event accessors in `Session.cs` (`AddUserWait`, `RemoveUserWait`, `SwitchUserWait`, `SwitchUserCompleted`) receive the same `add`/`remove` guard pair; fetched the full file at `68532f9` and confirmed no other event in the class calls `RegisterCallbackForEvent`/`UnregisterCallbackForEvent`, so nothing was missed.
- `add` path: the `value == null` early-return sits before the `_handler == null` registration check, so `+= null` can no longer reach the native register on a null→null edge. A non-null `value` always yields a non-null combined delegate, so the register-on-first-subscriber invariant is preserved.
- `remove` path: the `_handler == null` early-return stops `-= x` on an empty list from reaching `UnregisterCallbackForEvent`. Removing a null or unsubscribed delegate from a non-null list leaves it non-null, so the post-removal null check still fires only on a true non-null→null edge.
- Both guards are inside the existing `lock`, so edge detection stays atomic with the native (un)register call.
- `RegisterCallbackForEvent` throws via `CheckError` before `_handler += value` executes, so a failed native registration leaves the handler null and a later subscribe retries; unchanged by this PR and consistent with the new guard.
- Resulting semantics match the compiler-generated event accessor (`+= null`, `-= null`, and removing an absent delegate are all no-ops), so no observable public API behavior change.

---

chanwoochoi (approved):

